PR #14, opened 2022-06-20
Testing different way of installing Vulkan SDK

drsnuggles8:


drsnuggles8:
New builds are working, googletest / ctest currently not yet implemented though.

PR #45, opened 2023-07-01
Fonts

drsnuggles8:
First pull request, just adding dependencies


drsnuggles8:
Seems like build order is not right in the github actions, it works locally for me (zlib needs to be built first, so that libpng can find it, so that freetype can find it, and so on). To be fixed..

PR #376, opened 2026-06-22
feat(mcp): implement structured tool output with redaction and output…

drsnuggles8:
… schema


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `8bd662bd5c1aad4a69b202935b190157639d091c`
Reviewed the PR diff at **high** effort (7 files, +423/-8) after merging `origin/master` in to resolve the conflict.

- **Findings requiring a fix:** 0.
- **Verified:** every handler unconditionally populates each field its new `OutputSchema` marks `required` (error paths return early via `ToolResult::Error` before the `Structured()` return) — no schema/handler mismatch across `olo_scene_summary`, `olo_scene_get_entity`, `olo_memory_report`, `olo_perf_snapshot`, `olo_physics_raycast`. Text mirror stays byte-identical (`dump(2)`), and structuredContent is redacted consistently with the `content` text.
- **Dismissed:** `RedactStructuredContent` does not redact object *keys* (nlohmann range-for over an object yields values only) — **by design**: keys are always hardcoded schema field names; filesystem paths only ever appear in *values* (e.g. `componentsYaml`). Redacting keys would risk corrupting document structure for no benefit.

**Build/test after merge:** `OloEngine-Tests` + `OloEditor` build clean; `McpStructuredOutput` 10/10, broader MCP 135/135, Cinematic suite 198/198 — all green locally.

PR #377, opened 2026-06-22
feat(inventory): implement shop trading system with buy/sell function…

drsnuggles8:
…ality and events

### DIFF
--- a/OloEngine/src/OloEngine/Gameplay/Inventory/InventoryEvents.h
+++ b/OloEngine/src/OloEngine/Gameplay/Inventory/InventoryEvents.h
@@ -40,4 +40,26 @@ namespace OloEngine
         std::string SlotName;
     };
 
+    // Published when a buyer purchases items from a shop container. TotalPrice
+    // is the currency the buyer paid (unit BuyPrice × Quantity).
+    struct ItemBoughtEvent
+    {
+        UUID BuyerEntityID;
+        UUID ShopEntityID;
+        std::string ItemDefinitionID;
+        i32 Quantity = 0;
+        i32 TotalPrice = 0;
+    };
+
+    // Published when a seller offloads items to a shop container. TotalPrice is
+    // the currency the seller received (unit SellPrice × Quantity).
+    struct ItemSoldEvent
+    {
+        UUID SellerEntityID;
+        UUID ShopEntityID;
+        std::string ItemDefinitionID;
+        i32 Quantity = 0;
+        i32 TotalPrice = 0;
+    };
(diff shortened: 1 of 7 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `434cddd8665d9b97a515cbf4b9ac44e76e278ac6`
Reviewed the full PR diff against `master` at **high** effort (correctness + cleanup/altitude/conventions finder passes, recall-biased). Scope: `InventorySystem` shop trading (`BuyItem`/`SellItem`/`GetBuyPrice`/`GetSellPrice` + file-local `TransferByDefinition`), the two new gameplay events, and the 10-case functional test.

- **Findings (real, requiring a fix):** 0
- **Verified clean:**
  - **Atomicity** — both trades validate every precondition (shop flag, stock/ownership, affordability, destination room) before any currency or item moves; `TransferByDefinition` stages adds against a `destCopy` and only commits after the source debit, so a full destination / insufficient stock rolls back to a true no-op (the 4 rejection tests pin this).
  - **Overflow** — buy affordability and sell gain both compute in `i64`; buy debit fits `i32` (guarded `totalCost <= Currency`), sell credit is rejected (not wrapped) when `Currency + gain` would exceed `i32` max.
  - **InstanceID hygiene** — whole-stack pulls carry the original instance; partial pulls mint a fresh `InstanceID` so source/dest can't share an ID. Multi-stack re-stacking in the destination is the intended/tested behavior.
  - **Pricing** — half-`BuyPrice` buyback truncation (incl. the `BuyPrice 1 → 0` edge), explicit `SellPrice` precedence, and the saturating UI helpers all match the tests.
- **Dismissed (no change):**
  - `SellItem` `if (unitPrice < 0)` (line 256) is technically unreachable today (`UnitSellPrice` clamps to ≥0), but it's deliberate defensive symmetry with `BuyItem` and guards a future `UnitSellPrice` change — negligible-value churn to remove.
  - `BuyItem` `if (unitPrice < 0)` (line 208) is **not** dead: `UnitBuyPrice` returns the unvalidated `def.BuyPrice`, so a negative authored price is genuinely reachable and correctly rejected.
- **Behavior check:** non-visual gameplay-logic change (no renderer/shader/visual surface) — CLAUDE.md's visual-evidence mandate doesn't apply. Covered by 10 functional tests; CI build + test jobs ran them green.
- **Note:** CodeRabbit hit its review rate limit and never actually reviewed this PR (only a 'limit reached' notice), so this self-review is the substantive review on record.

PR #384, opened 2026-06-24
Implement persistent cache metadata functionality in SoundGraphCache

drsnuggles8:
- Added SaveCacheMetadata and LoadCacheMetadata methods for serializing and restoring cache metadata using YAML.
- Introduced GetAudioDataSizeBytes method in WavePlayer to introspect audio data size.
- Updated memory estimation logic to accurately reflect WavePlayer audio data footprint.
- Added unit tests for cache metadata round-trip, handling of missing/corrupt files, and memory estimation.

## Summary by CodeRabbit

* **New Features**
  * Improved audio graph memory tracking for more accurate cache usage estimates.
  * Added support for reporting per-node heap usage in audio processing components.

* **Bug Fixes**
  * Removed an outdated fixed memory estimate so cache size calculations better reflect actual audio data.
  * Clarified that the in-memory cache does not use its own persistent metadata format.

* **Tests**
  * Added coverage for memory reporting and cache size calculations with audio nodes.

### DIFF
--- a/OloEngine/src/OloEngine/Audio/SoundGraph/NodeProcessor.h
+++ b/OloEngine/src/OloEngine/Audio/SoundGraph/NodeProcessor.h
@@ -485,6 +485,16 @@ namespace OloEngine::Audio::SoundGraph
             (void)numFrames;
         }
 
+        /// Large heap buffers this node owns beyond its own sizeof — e.g. a
+        /// WavePlayer's decoded audio samples. Base nodes own none; a node type
+        /// with a significant buffer overrides this so SoundGraphCache memory
+        /// accounting stays type-agnostic (no per-type dynamic_cast) and counts
+        /// any future buffer-owning node automatically.
+        [[nodiscard]] virtual sizet GetHeapBytes() const
+        {
+            return 0;
+        }
+
         //==============================================================================
         /// Endpoint access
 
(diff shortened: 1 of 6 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `d7a42668f60b839a7659fe251d98692277476c31`
Reviewed the full PR diff against `master` at **high** effort (correctness + cleanup + altitude + conventions angles).

**Verdict: no correctness bugs.** The change replaces a flat 2 MB-per-node memory guess with a type-agnostic `NodeProcessor::GetHeapBytes()` virtual that `WavePlayer` overrides to report its real decoded-sample buffer. Verified:
- Removed `Save/LoadCacheMetadata` placeholders have **zero callers** anywhere in `src`/`OloEditor`/`tests` — safe deletion; the in-memory-cache vs. `CompilerCache` persistence split is correct.
- `m_AudioData.m_Samples` is `std::vector<f32>`, so `capacity() * sizeof(f32)` is the correct byte count; `capacity()` (not `size()`) is a documented, deliberate choice.
- Virtual hook dispatches over the existing `m_Nodes` loop; `Put`/`GetMemoryUsage`/`AddNode` signatures all match the new test's usage.
- New `SoundGraphCacheTest` passes locally (3-WavePlayer graph now estimates 533 KB vs. the retired 6 MB); full suite is green in CI incl. ASan/TSan/UBSan.

- **Findings:** 1 (SonarCloud `S5817`)  ·  **Fixed:** made the test helper `MakeSourceFile` `const` (doesn't mutate the fixture).
- **Dismissed:** 3× SonarCloud `S6166` ("add message to [[nodiscard]]") — the codebase uses bare `[[nodiscard]]` 2437:17, so the new code matches the overwhelming convention; adopting the message form here would be inconsistent.

_The `const` fix is in the working tree, not yet pushed — pushing re-triggers the multi-hour sanitizer suite for a cosmetic nit, so holding for an explicit go-ahead vs. merging as-is._

PR #386, opened 2026-06-24
Refactor ECS component management: automate AllComponents tuple gener…

drsnuggles8:
…ation and update related documentation

## Summary by CodeRabbit

* **New Features**
  * Component registration is now generated automatically, reducing manual upkeep when adding new components.
  * The build process now produces an additional generated output for component coverage and related bindings.

* **Bug Fixes**
  * Improved validation to keep generated component coverage in sync and prevent stale component lists from being used.

* **Documentation**
  * Updated contributor guidance for component lifecycle, serialization, save/load, and scripting registration requirements.


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `40718358`

Reviewed the full PR diff (`origin/master...HEAD`) at **high** effort — 3 correctness + 5 quality finder angles, then verification.

**Correctness findings: 0.** The change is behaviour-preserving by construction, and I verified it empirically:
- The generated `AllComponents.Generated.inl` tuple is **set-identical** to the prior hand-written tuple in `Components.h` — 96 components ⟷ 96, zero drift (extracted both, `comm`/`diff` clean).
- `ComponentTupleCoverageTest` guards the generated file in **both** directions (forward: every declared component is in the tuple; inverse: every tuple entry resolves to a declared struct — which also catches drift between the generator's recursive scan and the test's curated header-root list).
- The generator's `kComponentsNotInTuple` and the test's `kNotInTuple` exclusion sets match exactly (6 entries each).
- `main()`'s `errors` flag is correctly hoisted so the empty-tuple guard and the no-OLO_PROPERTY early-out both propagate failure; the empty-tuple guard refuses to overwrite a good file with `ComponentGroup<>` and fails the build loudly.

**Quality candidates (5): 3 dismissed, 2 refuted — no code change warranted.**
- *Dismissed — duplicate recursive dir walk:* `CollectComponentStructs` walks `OloEngine/src` separately from `ParseHeaders`. Immaterial for a build-time tool that runs in <1s; merging risks regressing a set-identity-verified generator for no measurable gain.
- *Dismissed — struct-detection overlap with `ParseHeaders`:* shared helpers (`Trim`/`StripBlockComments`) are already reused; the two passes detect different things (OLO_PROPERTY fields vs `struct *Component` defs) and are intentionally kept congruent with the guard test.
- *Dismissed — two hand-synced exclusion lists:* inherent to the two-binary design (generator can't import the test's list); documented in both files + CLAUDE.md, and generator-newer drift IS caught by the forward coverage test. Known design tension, tracked.
- *Refuted — "unreachable `return false`" in `WriteAllComponentsTuple`:* required to silence MSVC C4715 — a `switch` over the `WriteResult` enum is not treated as exhaustive in a `bool`-returning function. Not dead code.
- *Refuted — "`constexpr kSuffix` redeclared in loop":* `constexpr` is compile-time, zero runtime cost; the loop-hoisting rule targets runtime invariants.

**CI:** the only red check is a **pre-existing `FrameExportTest` flake unrelated to this diff** — `FrameExportTest.*` share a fixed temp dir (`olo_frame_export_test`) raced across `ctest --parallel 4` separate processes. It also flakes on `master` (run 28098055255, `MarkdownSortAnalysisPresent`). Re-triggered the Windows workflow to clear it. This diff touches only OloHeaderTool / the generated tuple / the coverage test and cannot affect frame-export.

**Behaviour/visual check:** N/A — build-tooling + generated code + test only; no runtime or visual surface. The coverage test is the behavioural guard and runs in CI.

PR #387, opened 2026-06-24
feat: Add VehicleComponent with serialization, scripting, and runtime…

drsnuggles8:
… support

- Implemented VehicleComponent with properties for vehicle dynamics including half track width, axle offsets, wheel dimensions, suspension parameters, and driver inputs.
- Added serialization support for VehicleComponent to ensure proper saving and loading of component state.
- Integrated VehicleComponent into the Lua and C# scripting systems, allowing for manipulation of vehicle properties from scripts.
- Created tests to validate the functionality of VehicleComponent, including physics interactions, suspension behavior, and save-game round-trip integrity.
- Updated Scene class to handle vehicle addition and removal during runtime, ensuring proper management of vehicle constraints in the physics engine.


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `29ca083d2803449539069609bae1eb1cbc50569c`
Reviewed the PR diff against `master` at **high** effort (8 finder angles + verification), after merging `origin/master` in to resolve the `Components.h` `AllComponents` conflict (master's #386 moved that tuple into OloHeaderTool codegen — took the generated-include side; `VehicleComponent` is auto-registered, verified by `ComponentTupleCoverage`).

**Findings: 5 fixed + 2 regression tests** (commit `29ca083d`). The review surfaced a reachable use-after-free class in the vehicle runtime lifecycle — a `VehicleConstraint` is a live Jolt step listener that dereferences its chassis body every tick, so it must be torn down before that body is freed.
- **Fixed:** `Scene::DestroyEntity` now destroys the vehicle before `registry.destroy()` (entt doesn't fire `OnComponentRemoved` there) — was a UAF/assert when destroying a car at runtime.
- **Fixed:** `OnComponentRemoved<Rigidbody3DComponent>` destroys the vehicle before the chassis body it drives.
- **Fixed:** `UpdateVehicleControllers` uses non-asserting `TryGetEntityWithUUID` instead of the asserting `GetEntityByUUID`, so a transient `m_Vehicles`/entity desync skips cleanly instead of crashing the tick.
- **Fixed:** `CreateVehicle` preserves an authored `SuspensionMinLength == 0` (serializers/Jolt allow `0 <= min`; `SanitizeVehiclePositive` had rewritten 0 to the default).
- **Fixed:** `CreateVehicle` steer-angle falls back to 0 (steering disabled) on a non-finite value, matching the documented intent.
- **Tests:** added `DestroyingVehicleEntityReleasesTheConstraint` + `RemovingChassisRigidbodyReleasesTheVehicle` (both UAF without the fixes).

**Dismissed (4)** — all faithfully mirror the existing `PhysicsJoint3DComponent` pattern, so out of scope for this PR: per-site duplicated sanitization tables; `Scene.cpp` re-looping vs `JoltScene::CreateVehicles()` (joints do the same); `m_RuntimeVehicleToken` mirroring `m_Vehicles` membership; per-step UUID hashing in `UpdateVehicleControllers`.

**Verification:** full suite **3485 passed / 4 skipped (expected) / 0 failed**; all component cross-binding coverage guards pass; `VehicleTest` (7 cases incl. the 2 new) green. Runtime-only change (no visual surface) — behavior verified through the Functional tests' real `Scene::OnUpdateRuntime` + Jolt.

PR #394, opened 2026-06-25
SoundGraph: sample-accurate external (game-code) triggers

drsnuggles8:
Thread a per-event sample offset through SoundGraphSource's input-event queue and SoundGraph::SendInputEvent so externally scheduled triggers land on the exact sample instead of the block boundary.

### DIFF
--- a/OloEngine/src/OloEngine/Audio/SoundGraph/SoundGraph.h
+++ b/OloEngine/src/OloEngine/Audio/SoundGraph/SoundGraph.h
@@ -500,9 +500,18 @@ namespace OloEngine::Audio::SoundGraph
         /// Connect Input Event to Input Event
         void AddRoute(InputEvent& source, InputEvent& destination) const noexcept
         {
+            // Phase 4 (docs/soundgraph-metasounds-refactor.md): forward the sample
+            // offset so a graph-input trigger fired at frame N reaches the destination
+            // node's trigger input at frame N instead of being quantised to the block
+            // boundary. This is the hop an *external* (game-code) trigger takes —
+            // SoundGraphSource enqueues the event with an offset, the audio thread
+            // drains it into SoundGraph::SendInputEvent, and this route carries the
+            // offset the last hop to the consuming node. Binding m_OffsetEvent makes
+            // InputEvent::operator() prefer the offset-aware path; a value-only fire
+            // still arrives with the default offset 0 (block start).
             InputEvent* dest = &destination;
-            source.m_Event = [dest](float v)
-            { (*dest)(v); };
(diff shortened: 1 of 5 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `117ae001`
Reviewed the full PR diff (`origin/master...HEAD`, 5 files / +351) at **high** effort: line-by-line correctness scan, removed-behavior audit, cross-file caller tracing, plus cleanup/altitude/convention angles. Cross-checked the 8 SonarCloud new-code findings.

**Correctness: no bugs found.** Verified the load-bearing paths:
- `Identifier(string)`→`GenerateFNVHash` and `Identifier(u32)` round-trip to the same `m_Hash` that `InEvents` keys on, so the external string overload resolves to the right endpoint.
- SPSC hand-off (`m_InputEventQueue`) mirrors the existing `m_ParameterUpdateQueue` contract (main-thread producer / audio-thread consumer); `DrainInputEvents` runs before `Process`; `ReplaceGraph` clears the queue under the same suspend protocol as the param queue.
- `ClampInputEventOffset` correct for negative (→0), in-range, over-range (→blockSize-1) and blockSize==0 (→0); `SendInputEvent` sanitises NaN/Inf payload to 1.0f per the float-validation rule.
- `InputEvent::operator()` prefers `m_OffsetEvent`, so the route rebind in `AddRoute` is correct and value-only fires still default to offset 0.

CI is fully green and `ctest` discovers + **runs** the new `SoundGraphSampleAccurateTriggerTest.*` cases (not just builds), so green covers them.

- **Findings:** 10 (0 correctness)  ·  **Fixed:** none required
- **Dismissed:**
  - SonarCloud S1238 (pass `choc::value::ValueView` by const-ref) — false positive: `ValueView` is a lightweight Type+pointer view (pass-by-value like `string_view`); signature already passed by value pre-PR.
  - SonarCloud S1712 ×5 (replace default args with overloads) — declined: default args are the established pattern across the SoundGraph event system and the defaults intentionally preserve existing callers' block-boundary timing.
  - SonarCloud S5350 (`capturePtr` → pointer-to-const) — trivial test-only cosmetic, no functional effect; not worth resetting a fully-green CI run.
  - SonarCloud S1917 (mixed `//` and `/* */`) — false positive: `/*sampleOffset=*/` inline param label alongside a `//` banner is idiomatic.
  - code-review: `std::clamp` in `ClampInputEventOffset` — declined: `std::clamp(x,0,maxOffset)` is UB when blockSize==0 (maxOffset=-1, lo>hi); the explicit form keeps the necessary guard and documents the legacy-negative intent.
  - code-review: duplicated 6-line envelope setup across two tests — declined: self-contained test bodies are preferred for independence; duplication is minor.

PR #397, opened 2026-06-25
Dedup StreamWriter/StreamReader container helpers via shared element …

drsnuggles8:
…helper

### DIFF
--- a/OloEngine/src/OloEngine/Serialization/StreamReader.h
+++ b/OloEngine/src/OloEngine/Serialization/StreamReader.h
@@ -8,6 +8,7 @@
 #include <vector>
 #include <string>
 #include <span>
+#include <type_traits>
 
 namespace OloEngine
 {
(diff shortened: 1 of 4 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `70713f59443cd5344631d31792b022d38922f623`
Reviewed the full PR diff at **high** effort (traced every caller + pinned the wire format + ran the new tests locally).

- **Findings:** 0 real.
- **Verified:**
  - Behavior-preserving for every previously-supported instantiation. The only production caller, `FontSerializer::WriteArray(std::vector<FontCodepointRange>)`, is trivially-copyable and dispatches identically through the new `WriteElement`.
  - `WriteElement`/`ReadElement` correctly made `private`; the public API (`WriteMap`/`ReadMap`/`WriteArray`/`ReadArray`) is unchanged.
  - Wire format unchanged & pinned by new byte-level tests: `u32` count prefix for arrays/maps, `u64`-length-prefixed strings (matches `StreamWriter::WriteString`), raw bytes for trivially-copyable, static `Serialize`/`Deserialize` for class elements.
  - Bonus capability (not a regression): now also handles `std::map<std::string, …>` keys and string map *values* — cases the old code couldn't even compile.
  - 13/13 new `StreamHelpers.*` tests pass locally; CI build + ASan/LSan/TSan/UBSan all green on this SHA.
- **Dismissed:** the unified `WriteMap(const Map&)`/`ReadMap(Map&)` is now an unconstrained template — but a non-pair container (e.g. `std::set`) fails to compile on the structured binding rather than silently misbehaving, so a `concept` constraint would be gold-plating, not a correctness fix.
- **CI note:** the only red check is the **non-required** `SonarCloud Scan`, which timed out at the 6h cap. Successful SonarCloud runs take 272–357 min (cap is 360) — a chronic infra flake unrelated to this 4-file dedup; does not block merge.

Issue #380, opened 2026-06-23
Codegen: collapse the six-touchpoint ECS component registration via OloHeaderTool


drsnuggles8:
## Problem

Adding a single ECS component currently requires hand-editing **~6 touch-points** across multiple files (documented in `CLAUDE.md` → "Cross-binding check"). Miss one and scripting / scene saves / save-games / runtime copies silently drop the component, or the build fails to link:

1. `Scene/Components.h` — the struct **and** the `AllComponents` tuple. *(guarded by `ComponentTupleCoverageTest`)*
2. `Scene/Scene.cpp` — `OnComponentAdded<T>` specialization **and** `OLO_ON_COMPONENT_REMOVED_NOOP(T)`. *(link error if missing — not silent)*
3. `Scene/SceneSerializer.cpp` — `SerializeEntity` (write) + `DeserializeEntities` (read). *(not guarded)*
4. **SaveGame — 3 edits across 2 files:** `SaveGameComponentSerializer.{h,cpp}` (decl + def + `RegisterAll()`), **and** the `SAVE_COMPONENT(...)` + `TRY_LOAD_COMPONENT(...)` enumeration lists in `SaveGameSerializer.cpp`. *(silently dropped on save/load if missed)*
5. `Scripting/C#/Generated/` — driven by `OLO_PROPERTY` annotations via OloHeaderTool. *(generated)*
6. `Scripting/Lua/LuaScriptGlue.cpp::RegisterAllTypes()` — Sol2 usertype. *(intentionally not completeness-guarded)*

The coverage tests (`ComponentTupleCoverageTest`, `ComponentSerializerCoverageTest`) *catch* some misses after the fact, but several touch-points (SceneSerializer, SaveGame enumeration lists) are unguarded, and the whole thing is tedious and error-prone.

## Proposal

Extend OloHeaderTool (or add a sibling generator) to **emit the mechanical, list-shaped touch-points from the `OLO_PROPERTY`-annotated component struct** rather than requiring hand edits — collapsing ~6 manual edits toward one annotated struct:

- The SaveGame `SAVE_COMPONENT` / `TRY_LOAD_COMPONENT` enumeration lists + the `SaveGameComponentSerializer` decl/def/`RegisterAll()` registration.
- The `SceneSerializer` `SerializeEntity` / `DeserializeEntities` field read/writes (where field types are serializable per their annotation).
[...]

PR #405, opened 2026-06-26
Codegen: generate the SaveGame component capture/restore lists (#380)

drsnuggles8:
## What & why

Second slice of #380 (collapse the six-touchpoint ECS component registration via OloHeaderTool). The first slice generated the `AllComponents` tuple (#386). This one generates the **`SAVE_COMPONENT(...)` capture list and `TRY_LOAD_COMPONENT(...)` restore list** in `SaveGame/SaveGameSerializer.cpp`.

Those two ~97-entry hand-maintained enumerations were the **most dangerous unguarded** touch-point in the component checklist: a component missing from **either** was *silently dropped* on save/load — no compile error, no warning. It round-tripped fine through scene YAML but vanished through save-games (inventory/quest/ability progress lost). Generating them removes the hand-labour and the silent-drop footgun.

## How

- `tools/OloHeaderTool/main.cpp` emits `SaveGame/Generated/SaveGameComponent{Capture,Restore}.Generated.inl` from the **same `struct *Component` scan** as the tuple (`CollectComponentStructs`), minus a save-game-specific exclusion set `kComponentsNotInSaveGame`.
- That exclusion set **deliberately differs** from `kComponentsNotInTuple`: save-games *keep* `IDComponent`/`TagComponent` (identity is persisted) but *drop* every component with no `Serialize()` overload — the 4 runtime `*StateComponent`/`UIResolvedRectComponent` **plus** `AudioSoundGraphComponent`/`LocalizedTextComponent` (tuple members with no save serializer yet).
- `SaveGameSerializer.cpp` `#include`s the generated lists in place of the hand lists. `IDComponent` restore is now a uniform `TRY_LOAD_COMPONENT` rather than the bare-first `LOAD_COMPONENT` — functionally identical since restore matches each stream block by `typeHash`, so list order is irrelevant (the entity is created with its UUID before any block is read).
- `SaveGameComponentSerializerCoverageTest` now parses the generated `.inl` files (the documented "touch-point moved into a generated file → repoint the grep" gotcha), with `RegisterAll()` staying hand-written as the live guard.
- New 5th CLI arg + `OHT_SAVEGAME_OUT` cache var wire it into `GenerateBindings`.
- `CLAUDE.md` updated (SaveGame cross-binding bullet, OloHeaderTool section, pitfalls).

## Verification

- Generated lists are **byte-for-byte identical in coverage** to the prior hand lists (97 components; capture == restore == registered), confirmed by set-diff.
- `OloEngine-Tests` builds clean. Passing: `SaveGameComponentSerializerCoverage` (×2, reading generated files), `SaveGameIntegration` round-trip (`CaptureAndRestoreWithEntities`, `FullFileRoundTrip`), `ComponentTupleCoverage`, `ComponentSerializerCoverage`, `SaveableRegistry`.
- Generated `.inl` files and all edited C++ are clang-format-idempotent.

## Scope

Deliberately limited to the two enumeration lists + repointing the coverage test. The `SceneSerializer` per-field blocks and the per-component `Serialize()` overloads (need field-type introspection) remain later slices — #380 stays open.

## Summary by CodeRabbit

* **New Features**
  * Save-game component handling is now generated automatically, improving coverage and reducing the chance of missing save/load support for supported components.
  * Save/load lists are now kept in sync with component definitions through the build process.

* **Bug Fixes**
  * Improved save-game reliability by removing manual component list maintenance during save and restore.
  * Updated coverage checks to verify the generated save-game lists are present and complete.


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `2a79bb57ce38bb9cf2883ce69bcbb53b80311d27`
Reviewed the PR diff at **high** effort (3 finder angles: correctness, conventions/test-guard, cleanup → verify).

- **Correctness:** clean. Verified the one behavior-change risk — IDComponent moved from a bare unguarded `LOAD_COMPONENT` (old first-entry) to a generated guarded `TRY_LOAD_COMPONENT` — is behavior-preserving: restore matches each stream block by `typeHash`, so list order is irrelevant. Macros in scope, `#undef TRY_LOAD_COMPONENT` intact, empty-list guard present.
- **Conventions / test-guard:** clean. `SaveGameComponentSerializerCoverageTest` correctly repoints its regexes at the generated `.inl`; no vacuous-pass (existence + non-empty asserts); IDComponent not silently missed.
- **Findings:** 5 (all cleanup-tier on the codegen tool). **Fixed: 1** — `ReportWrite` signaled failure two ways (return value + `bool&` out-param, caller used only the param); dropped the out-param, returns bool, caller ANDs the writes (pure refactor, regenerated `.inl` byte-identical).
- **Dismissed: 4** — emitter copy-paste (the two output shapes genuinely differ; one-fn-per-output matches the file's pattern); double iteration (trivial); `kComponentsNotInSaveGame` mixing two exclusion reasons (well-documented **and** guarded by this very coverage test); `WriteAllComponentsTuple` inconsistency (pre-existing untouched code, out of scope).

**Merge note:** the `CONFLICTING` state was `RagdollComponent` landing on master after the branch forked; resolved by merging master in and regenerating the lists (regenerated capture/restore match master's old hand-list exactly, 99 components). The earlier Windows CI "build" failure was the **unrelated** pre-existing `QuestDatabaseFloatValidationTest` parallel-temp-dir race (shared hardcoded `olo_quest_floatval_test` dir under `ctest --parallel 4`), not a build break — passes locally in isolation.

Local verification: `OloEngine-Tests` builds clean; SaveGame coverage + round-trip + Ragdoll suites all green (6/6 + 31/31).

PR #407, opened 2026-06-26
chore(audio): clear SonarQube S1271/S6166/S127 in Audio/ (no behaviour change)

drsnuggles8:
## SonarQube maintainability cleanup — Audio subsystem (no behaviour change)

Mechanical "batched signal hygiene" pass confined to `OloEngine/src/OloEngine/Audio/**` (28 files, +335/-325, nothing outside Audio). Three SonarQube CODE_SMELL rule-classes cleared; zero behaviour change.

### Rules cleared
| Rule | What | Flagged (approx) | How |
|---|---|---|---|
| `cpp:S1271` | `::` to access global functions | ~197 | Qualified global C-API calls — SIMD intrinsics (`_mm*`), miniaudio (`ma_*`), libc math/`mem*` — with `::`. Applied uniformly across all `#if` SIMD paths (SonarQube only sees the active one), so 274 call sites touched. A comment/string-aware allowlist regex did the rewrite. |
| `cpp:S6166` | `[[nodiscard]]` needs an explanation | 42 | Added a short message to each existing bare `[[nodiscard]]`. No new attributes → no new warnings. |
| `cpp:S127` | hoist loop-invariant stop condition | 11 | Cached `.size()` into a `const` before each loop whose container is not modified in the body. |

### Deliberately out of scope (left for follow-up)
- **`cpp:S953`** ("unions should not be used", 56 in `SampleBufferOperations.h`) — deliberate type-punning for SIMD/sample layout; "fixing" it would change behaviour.
- **`cpp:S5536`** (dead code, ~114) — risky: Audio is reachable via Lua/Sol2 bindings, DSP node registration and tests; needs per-symbol project-wide proof before deletion.
- **`cpp:S8417`** (memory-order style, ~107) — known noise.
- **`cpp:S5981` / `S1712` / `S4136` / `S6177`** — real but require judgement (downcast semantics, API-surface changes, reordering, `using enum` collisions); deferred to keep this diff purely mechanical.
- A few `S1271` hits on namespace-local helpers (e.g. `AsSplitter`) were left — `::` would resolve to global scope and is wrong for them. The `_MM_SHUFFLE` macro was likewise excluded.

### Verification
- `cmake --build build --target OloEngine --config Debug` — green (the `::` prefixes compile; no macro breakage).
- `cmake --build build --target OloEngine-Tests --config Debug` — green.
- 183/183 audio-related tests pass (`*Audio*` / `*SoundGraph*` / `*Spatial*` / `*Reverb*` / `*VBAP*` / `*DSP*` / `*Filter*` …).

SonarQube auto-resolves these on the next scan.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio processing stability and consistency across playback, filtering, spatialization, reverb, and sound graph paths.
  * Fixed several audio routines to avoid unnecessary repeated work, which can help performance during real-time processing.

* **Style**
  * Improved developer-facing warnings when important return values are ignored, making audio-related code safer to use.
  * Standardized internal audio API usage for clearer, more consistent behavior.

### DIFF
--- a/OloEngine/src/OloEngine/Audio/AudioCallback.h
+++ b/OloEngine/src/OloEngine/Audio/AudioCallback.h
@@ -116,14 +116,16 @@ namespace OloEngine::Audio
         bool InitBase(u32 sampleRate, u32 maxBlockSize, const BusConfig& busConfig) final
         {
             m_InDeinterleavedBuses.resize(busConfig.m_InputBuses.size());
-            for (sizet i = 0; i < m_InDeinterleavedBuses.size(); ++i)
+            const sizet inputBusCount = m_InDeinterleavedBuses.size();
+            for (sizet i = 0; i < inputBusCount; ++i)
             {
                 m_InDeinterleavedBuses[i].resize({ busConfig.m_InputBuses[i], maxBlockSize });
                 m_InDeinterleavedBuses[i].clear();
             }
 
             m_OutDeinterleavedBuses.resize(busConfig.m_OutputBuses.size());
-            for (sizet i = 0; i < m_OutDeinterleavedBuses.size(); ++i)
+            const sizet outputBusCount = m_OutDeinterleavedBuses.size();
+            for (sizet i = 0; i < outputBusCount; ++i)
             {
                 m_OutDeinterleavedBuses[i].resize({ busConfig.m_OutputBuses[i], maxBlockSize });
                 m_OutDeinterleavedBuses[i].clear();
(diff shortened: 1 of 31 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `9ad55e51`
Reviewed the full PR diff (28 files) at **high** effort.

**Root cause of the red CI** — not a sanitizer regression: all 3 Linux/Clang sanitizer jobs (ASan+LSan, UBSan, TSan) failed at the **build** step (`error: expected unqualified-id`, 12×). The S1271 `::`-qualification pass prefixed lowercase SIMD intrinsics in `SampleBufferOperations.h` (`_mm256_shuffle_ps`, `_mm256_permute4x64_pd`, `_mm256_extractf128_ps`, …). Those `imm8`-taking intrinsics are **function-like macros** in Clang/GCC `<immintrin.h>`, so `::_mm256_extractf128_ps(...)` is illegal. MSVC implements them as real functions → **Windows compiled clean while Linux/Clang went red**.

**Findings:** 1 · **Fixed:** stripped `::` from every `_mm*` intrinsic in `SampleBufferOperations.h` (commit `9ad55e51`). `::` added nothing (intrinsics are global; unqualified lookup already resolved them) and only broke the macro ones — the file is now byte-identical to `master` for those lines, so it compiles on all three toolchains. `_MM_SHUFFLE` (uppercase macro) was correctly left unprefixed. No behaviour change.

**Reviewed, no findings:** the rest of the diff is three behaviour-preserving SonarQube transforms — S127 `.size()` hoists (verified no iterated container is mutated mid-loop), `[[nodiscard("reason")]]` (diagnostic-only), and `::`-qualification of **real** free functions (`ma_*`, `::cosf/sinf/sqrtf/atanf`, `::memcpy` — all exist in the global namespace, safe).

**Dismissed:** none.

_Note: the Linux-Clang macro path can't be reproduced on this Windows box (MSVC/clang-cl treat the intrinsics as functions); the fix is a mechanical revert to master-identical code and is confirmed by this push re-triggering the sanitizer jobs._

PR #409, opened 2026-06-26
Async compute-shader creation in GPUResourceQueue + FTask fire-and-forget teardown fix

drsnuggles8:
## Summary

Two related changes (split into two commits):

1. **`feat(renderer)`: async compute-shader creation in `GPUResourceQueue`** — the original task. The async GPU-resource queue's `CreateShaderCommand` had a stubbed compute branch that logged *"compute shader creation not yet implemented"* and silently produced a **null** shader. Because `ComputeShader` and `Shader` are **unrelated sibling types** under `RendererResource`, a `Ref<ComputeShader>` can't be delivered through `CreateShaderCommand`'s `Ref<Shader>` callback at all — so the stub could never have worked.

2. **`fix(task)`: clean up fire-and-forget `FTask` with `TDeleter`, not a body-side delete** — a latent task-system bug surfaced by wiring up the producer above.

## (1) Async compute shaders

- Added a dedicated **`CreateComputeShaderCommand`** with its own `std::function<void(Ref<ComputeShader>)>` callback, building the program via the production `ComputeShader::CreateFromSource` (mirrors the existing one-command-per-resource-type pattern). The old `CreateShaderCommand` compute branch now fails cleanly with a clear error instead of a misleading "not implemented" warning.
- Added the **producer**: `ComputeShader::CreateFromFileAsync(path, onReady)`. The `.comp` file is read and its `#include`s resolved on a task-system worker (`ComputeShader::LoadSourceFromFile` — CPU/disk only, thread-safe), which enqueues the GPU-creation command. The next `GPUResourceQueue::ProcessAll()` on the main thread (already pumped each frame from `RenderPipeline`) compiles it and fires `onReady` with the shader, or `nullptr` on failure.
- The 15+ existing compute systems (Particle/Snow/Ocean/GTAO/…) still load synchronously at init and hard-bail on a null shader, so they're **not** converted here — async is opt-in for new callers.

## (2) FTask teardown fix

`AsyncTask()` and the `Async()`/`AsyncThread()` helpers deleted their heap `LowLevelTasks::FTask` from **inside the runnable body** (`delete Task;` / `delete CleanupTask;`). That runs while the task is still in the `Running` state, tripping `~FTask()`'s `IsCompleted()` assertion and handing the scheduler a freed task — which then **crashed `FScheduler::StopWorkers()` / scheduler teardown** the moment any such task had actually run. `AsyncTask` had no callers before, so the bug was latent until `CreateFromFileAsync` used it.

**This was a porting mistake, not an upstream UE bug.** UE's `~FTask()` has the identical assertion, and UE's `Async()`/`AsyncTask` use the TaskGraph (`TGraphTask`/`FFunctionGraphTask`) and `IQueuedWork` — never a raw-`FTask` body-delete. UE cleans up a low-level task with a `TDeleter` captured in the runnable (`Tasks/TaskPrivate.h`: `TDeleter<FTaskBase, &FTaskBase::Release>{ this }`), which fires *after* the scheduler flags the task `Completed`.

The fix ports that idiom: a `Private::FFireAndForgetTask` owns the `FTask` as a member (like UE's `FTaskBase`) and is released via a `TDeleter` captured by value in the runnable. Applied to **all five** fire-and-forget sites (`AsyncTask`, `Async()` TaskGraph + ThreadPool, and the two `Async(Thread)`/`AsyncThread` cleanup tasks).

## Tests

- `TaskSystemTest.AsyncTaskFireAndForgetCompletesAndTearsDownCleanly` — drives a real `AsyncTask` to completion **and** the suite's `StopWorkers` teardown (the exact crash scenario).
- `GPUResourceQueueComputeShader.*` (5) + `GPUResourceQueueAsyncComputeShader.*` (1, scheduler-driven) — off-thread load + `#include` resolution, missing-file failure, command null-callback plumbing, producer hand-off, and the real worker-thread dispatch end-to-end.

## Verification

- New task + compute tests: pass, exit 0.
- Full `TaskSystemTest.cpp` suites (exercise `Async`/`AsyncTask`/`Async(Thread)` + `StopWorkers`): 24 pass, exit 0.
- Full suite: **3573 passed**. The only failure is the pre-existing flaky `FogVisualEvidenceTest` GPU golden test (passes in isolation, zero relation to tasks/compute).

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous compute shader loading and creation, including background file reading and include handling.
  * Added support for creating compute-shader GPU resources through the resource queue.
  * Improved async task handling with safer fire-and-forget execution.

* **Bug Fixes**
  * Compute shader loading now fails cleanly when files are missing or source preprocessing is invalid.
  * Resource queue now correctly rejects compute-shader work sent through the wrong path.
  * Added coverage for async task shutdown behavior to prevent teardown issues.

* **Tests**
  * Expanded automated coverage for compute shader loading, async creation, and task lifecycle behavior.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -244,6 +244,11 @@ jobs:
 
     - name: Install dependencies
       run: |
+        # The ubuntu-24.04 runner image ships Microsoft apt repos (azure-cli, prod)
+        # that periodically return 403 Forbidden; we never use them (Clang comes from
+        # apt.llvm.org), and `apt-get update` fails hard on any single repo error.
+        # Drop any Microsoft source so a Microsoft-side outage can't break our update.
+        sudo grep -rlZ packages.microsoft.com /etc/apt/sources.list.d/ 2>/dev/null | sudo xargs -0 -r rm -f || true
         sudo apt-get update
         sudo apt-get install -y \
           clang-19 lld-19 \
@@ -376,6 +381,11 @@ jobs:
 
     - name: Install dependencies
       run: |
+        # The ubuntu-24.04 runner image ships Microsoft apt repos (azure-cli, prod)
+        # that periodically return 403 Forbidden; we never use them (Clang comes from
+        # apt.llvm.org), and `apt-get update` fails hard on any single repo error.
+        # Drop any Microsoft source so a Microsoft-side outage can't break our update.
+        sudo grep -rlZ packages.microsoft.com /etc/apt/sources.list.d/ 2>/dev/null | sudo xargs -0 -r rm -f || true
         sudo apt-get update
         sudo apt-get install -y \
           clang-19 lld-19 \
@@ -489,6 +499,11 @@ jobs:
 
     - name: Install dependencies
       run: |
+        # The ubuntu-24.04 runner image ships Microsoft apt repos (azure-cli, prod)
+        # that periodically return 403 Forbidden; we never use them (Clang comes from
+        # apt.llvm.org), and `apt-get update` fails hard on any single repo error.
+        # Drop any Microsoft source so a Microsoft-side outage can't break our update.
+        sudo grep -rlZ packages.microsoft.com /etc/apt/sources.list.d/ 2>/dev/null | sudo xargs -0 -r rm -f || true
         sudo apt-get update
         sudo apt-get install -y \
           clang-19 lld-19 \
--- a/.github/workflows/video-ffmpeg.yml
+++ b/.github/workflows/video-ffmpeg.yml
@@ -56,6 +56,11 @@ jobs:
       - name: Install Linux deps
         if: matrix.os == 'Linux'
         run: |
+          # The ubuntu runner image ships Microsoft apt repos (azure-cli, prod) that
+          # periodically return 403 Forbidden; we never use them, and `apt-get update`
+          # fails hard on any single repo error. Drop any Microsoft source so a
+          # Microsoft-side outage can't break our update.
+          sudo grep -rlZ packages.microsoft.com /etc/apt/sources.list.d/ 2>/dev/null | sudo xargs -0 -r rm -f || true
           sudo apt-get update
           sudo apt-get install -y \
             nasm \
--- a/OloEngine/src/OloEngine/Async/Async.h
+++ b/OloEngine/src/OloEngine/Async/Async.h
@@ -128,6 +128,28 @@ namespace OloEngine
         };
 
         /**
+         * @brief Heap owner for a fire-and-forget LowLevelTasks::FTask.
+         *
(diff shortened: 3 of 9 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `8ea0751be5a33087b130ab366d3d1a1c3e5d3632`
Reviewed the PR diff at **high** effort (8 finder angles → verify). Most candidates were pre-existing/latent and confirmed *not* regressions; two genuine improvements to this PR's own code were folded in.

**Fixed (2):**
- **Altitude:** extracted `Private::LaunchFireAndForget(name, priority, runnable)` and routed all five fire-and-forget launch sites (`AsyncTask`, `Async()` TaskGraph + ThreadPool, the two `Async(Thread)`/`AsyncThread` cleanup tasks) through it. The `FFireAndForgetTask`+`TDeleter` ownership boilerplate was copy-pasted 5×; centralizing it means the load-bearing lifetime contract lives in one place and a 6th caller can't reintroduce the body-side `delete` this PR fixes.
- **Efficiency:** `ComputeShader::LoadSourceFromFile` scanned the path for the last separator twice; now a single `find_last_of`.

**Dismissed (with reasons):**
- Use-after-move of `Function` in the `Async(Thread)`/`AsyncThread` **thread-creation-failure** fallback — real but **pre-existing**, unrelated to this PR's intent (separate code path, only on OS thread-creation failure, untestable here). Worth a separate fix; out of scope.
- `CreateComputeShaderCommand`→`CreateFromSource` asserts on `RendererAPI::None` — consistent with the existing `CreateShaderCommand`→`Shader::Create`; `ProcessAll` runs on the main thread when the (OpenGL) renderer is up, so unreachable in practice.
- Leak if a launched task is never run (scheduler stopped) / `TryLaunch` returns false / cancellation — not regressions; the old `delete Task` had the same-or-worse behavior, and the new model actually **fixes** the cancellation-path leak. Latent/unreachable for fresh tasks.
- `SourceLoadResult` could return `RawShaderData` directly — intentional decoupling; returning it would pull `GPUResourceQueue.h` into the widely-included `ComputeShader.h`.
- Sync + async tests assert the same end state — kept both; they exercise different paths (deterministic queue plumbing vs. real worker-thread dispatch), and the scheduler-driven test was the point.
- `OpenGLComputeShader` ctor/Reload duplicate the name+read+ProcessIncludes logic — real duplication, **deferred**: refactoring the hot synchronous compute path (used by 15+ systems, GL-context-gated coverage that skips headless) can't be safely validated in CI without a manual editor run.

**Verification:** local full suite **3573 passed**; the only failure is the pre-existing flaky `FogVisualEvidenceTest.DistantGeometryFogsNearStaysClear` (GPU golden test, passes in isolation, unrelated to tasks/compute). Task + compute async tests pass with clean scheduler teardown (exit 0).

PR #413, opened 2026-06-27
Codegen: generate the Scene OnComponentAdded/Removed no-op specializations (#380)

drsnuggles8:
## What & why

Third slice of #380 (collapse the six-touchpoint ECS component registration via OloHeaderTool). Prior slices generated the `AllComponents` tuple (#386) and the SaveGame capture/restore lists (#405). This one generates the **`Scene::OnComponentAdded<T>` / `OnComponentRemoved<T>` no-op specializations** in `Scene/Scene.cpp` — touch-point #2 of the six.

`Scene.cpp` carried **~93 hand-written empty `OnComponentAdded` + ~93 empty `OnComponentRemoved`** specializations, one per component, interleaved with the handful that do real work. Every new component meant remembering to add two more boilerplate lines; the primary templates are declaration-only, so a miss is an *engine* link error (add, via `AddComponent<T>`) or an *OloEditor* link error (remove, via the remove-component button) — loud, but pure busywork. Generating the no-ops removes the labour and leaves only the genuinely-custom handlers hand-written.

## How

- `tools/OloHeaderTool/main.cpp` emits `Scene/Generated/OnComponent{Added,Removed}.Generated.inl` from the **same `struct *Component` scan** as the tuple (`CollectComponentStructs`), minus **two distinct** custom-handler exclusion sets `kComponentsCustomOnAdd` / `kComponentsCustomOnRemove`. They live in the existing `Scene/Generated` dir, so **no new CLI arg / CMake wiring** was needed (reuses `scene_out_dir`).
- The two sets **deliberately differ** — the add/remove callbacks are independent: `CameraComponent` / `CinematicComponent` / `LocalizedTextComponent` / `AudioSoundGraphComponent` do real work on **add** but no-op on remove, while `Rigidbody2DComponent` / `SpringBoneComponent` / `NoiseAnimationComponent` are the reverse (no-op add, real teardown). The 3D-physics group (`Rigidbody3D`, `PhysicsJoint3D`, `Vehicle`, `Ragdoll`, `CharacterController3D`) is custom in both.
- `Scene.cpp` `#include`s the generated lists inside `OLO_ON_COMPONENT_{ADDED,REMOVED}_NOOP` macros, replacing the hand-written no-op blocks. The custom-body specializations stay exactly where they were. `Skeleton` (a bare component type, not a `struct *Component`) is the one the scan can't see, so its no-ops stay hand-written.
- New `ComponentHandlerCoverageTest` reconstructs the full component universe (`AllComponents` tuple ∪ `kComponentsNotInTuple`) and asserts each generated list `== universe − custom`, in both directions, plus a sanity check that the two custom sets differ. Mirrors both exclusion sets — same sync contract `ComponentTupleCoverageTest` enforces for `kComponentsNotInTuple`.
- `CLAUDE.md` updated (Scene.cpp cross-binding bullet, OloHeaderTool section, pitfalls).

## Verification

- Generated lists are **coverage-identical to the prior hand-written set**: 104 `struct *Component` scanned, 93 Added no-ops (= 104 − 11 custom-add) and 93 Removed no-ops (= 104 − 11 custom-remove), confirmed by set-diff against the stripped specializations (the removal was script-driven off the generated lists; counts matched exactly: 37 namespace-form + 56 qualified-form add, 93 macro remove).
- `OloEngine.lib` compiles clean — **no duplicate-definition errors** (every custom body is correctly excluded). `OloEngine-Tests.exe` links — **no missing `OnComponentAdded`**. `OloEditor.exe` + `OloRuntime.exe` link — **no missing `OnComponentRemoved`**.
- Passing: `ComponentHandlerCoverage` (×3), `ComponentTupleCoverage` (×2), `ComponentSerializerCoverage` (×2), `SaveGameComponentSerializerCoverage` (×2) — 9/9.
- Generated `.inl` files re-run idempotent ("unchanged"); edited C++ is clang-format-clean.

## Scope

Deliberately limited to the two no-op enumeration lists + the coverage test. The `SceneSerializer` per-field blocks and per-component `Serialize()` overloads (need field-type introspection) remain later slices — **#380 stays open**.


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `f005765fb32f7477607804f3ce6996c1640330dc`
Reviewed the full PR diff against `master` at **high** effort (8 finder angles → verify). **No correctness bugs found** — the refactor is coverage-preserving (set-diff of old hand-written specializations vs generated `.inl` + surviving custom bodies: zero dropped, zero duplicate, zero missing on both Add and Remove sides; `Skeleton` correctly stays hand-written; the 6 `kComponentsNotInTuple` runtime-only components correctly appear in the generated lists).

- **Findings:** 5, all minor/optional — **0 fixed (none are defects), 5 dismissed.**
- **Dismissed:** Added/Removed emitter near-duplication — deliberately mirrors the merged `EmitSaveGameCapture/RestoreList` pair; collapsing one alone would be inconsistent.
- **Dismissed:** test `RepoRoot`/`ReadFile`/`CollectTupleMembers` copy — no shared meta-test util header exists; every coverage test redefines them in its own anon namespace (established precedent).
- **Dismissed:** mirror exclusion sets duplicated in the test — **by design**; the test must hold its own copy to *detect* drift from the generator (importing them would defeat the cross-check, same contract as `ComponentTupleCoverageTest`).
- **Dismissed:** `VideoOverlay`/`VideoSurface`/`Ragdoll` are empty-but-excluded — the authoritative `kComponentsCustomOnAdd` comment already states "either real init work, *or an intentionally-empty exception whose inline comment documents why*"; self-correcting via duplicate-definition compile error.
- **Verified locally:** `OloEngine.lib` (no dup-def), `OloEngine-Tests.exe` (no missing OnComponentAdded), `OloEditor.exe`+`OloRuntime.exe` (no missing OnComponentRemoved) all link; coverage suite 9/9 pass.

PR #415, opened 2026-06-27
chore(gameplay): clear SonarQube maintainability smells in Gameplay subsystem

drsnuggles8:
## What

Batched SonarQube **CODE_SMELL / MAINTAINABILITY** cleanup of the **`Gameplay`** subsystem (`OloEngine/src/OloEngine/Gameplay/`), in the same spirit as the merged Audio (#407) and Animation (#406/#388) sweeps. **No behaviour change** — these are signal-hygiene fixes, not bug fixes.

There is no GitHub issue for this; the source of truth is the SonarQube project `drsnuggles8_OloEngineBase`.

**Before:** 331 OPEN MAINTAINABILITY issues across the 49 `Gameplay/` files. The next SonarQube scan after merge will reflect the reduction (issues stay OPEN until a rescan).

## Rules cleared (mechanical, behaviour-preserving)

| Rule | Count | Fix |
|---|---|---|
| `S6166` | 71 | Added explanation text to every `[[nodiscard]]`. |
| `S6045` | 32 | Transparent `StringHash`/`StringEqual` on `std::string`-keyed unordered containers. |
| `S6177` | 18 | `using enum` in ToString/FromString/switch scopes. |
| `S6007` | 17 | `[[nodiscard("…")]]` on pure getters/helpers. |
| `S131` | 6 | `default:` on exhaustive switches. |
| `S127`/`S6165`/`S886` | 11 | `std::erase_if` for decrement-and-erase loops; hoisted `.size()`; pseudo-counter out of the for-condition. |
| `S1066` | 4 | Merged nested `if` with short-circuit `&&` (side-effect order preserved). |
| `S5566` | 3 | range-for → `std::ranges::all_of` / `any_of`. |
| `S6009`/`S945` | ~13 | `const std::string&` → `std::string_view` where the param is only compared / looked up (lookups now heterogeneous via the transparent maps). |
| `S881`/`S912` | 3 | Lifted increment/decrement-with-side-effect out of `if` conditions. |
| `S6003` | 2 | `push_back` → `emplace_back`. |
| `S6004`,`S5523`,`S4136`,`S6186` | 4 | init-statement; initialize-immediately; group overloads; drop redundant `operator!=` (C++20 synthesizes it). |

### Shared / cross-binding touch-points
- **NEW `Core/TransparentStringHash.h`** — shared `StringHash`/`StringEqual` transparent functors (extracted from the existing `InputActionManager` idiom; that file now includes the header instead of defining its own copies — same names, no behaviour change).
- **`Serialization/ArchiveExtensions.h`** — `operator<<(std::unordered_map<K,V>)` generalized to be variadic over the hash/equal policy (`Rest...`) so the transparent gameplay maps serialize; default maps are unaffected.
- **`SaveGame/SaveGameComponentSerializer.cpp`** — `SerializeStringSet` made a template over the set's hash/equal, for the same reason.

These two serializer generalizations are the necessary cross-binding updates for the `S6045` container-type change (the SaveGame save path consumes `QuestJournal` / `ItemInstance` containers).

## Deliberately skipped (with rationale)

Non-mechanical / behaviour-changing clusters, left in place:

- **`S1142`** (too many returns) — control-flow refactor, risks behaviour.
- **`S1712`** (default params → overloads) — high churn, doubles method count; **already slated for project-wide deactivation** in `docs/sonarqube-rule-suggestions.md`.
- **`S1771`/`S5018`** (struct→class / noexcept move) — every hit is a POD struct whose only member fn is a defaulted `operator==` (the sanctioned undo-tracking idiom, CLAUDE.md §7) or an ECS `*Component` (aggregate-init + the OloHeaderTool `struct *Component` scan). Documented as a false positive in this PR's doc update.
- **`S5536`** (unused function) — all are public API methods; removal is an API change, **also slated for deactivation**.
- **`S925`/`S3776`/`S1448`** — large structural refactors out of scope.
- **`S1706`** (no-exceptions) — the YAML asset-loading `try/catch` is intentional per-file error handling.
- **`S1181`** (catch more specific) — narrowing `catch(std::exception)` would change which exceptions propagate (e.g. `bad_alloc`).
- **`S6010`** (filesystem::path param) — would change fmt log formatting of the path.
- False positives left alone: **`S5008`** (type-erasure `void*` in `GameplayEventBus`), **`S3230`** (move-init from a mutable static), **`M23_279`** (`std::forward` on a loop-invoked callable would be a bug), **`M23_280`** (`optional::reset()` after move is well-defined).

A new `## 9. cpp:S1771` section in `docs/sonarqube-rule-suggestions.md` records the recurring defaulted-`operator==` FP plus the S1706/S5008/M23_280 one-offs.

## Verification
- `OloEngine` and `OloEngine-Tests` build clean (Debug, MSVC).
- All **228** Gameplay/Inventory/Quest/Ability/SaveGame tests pass — including the YAML and save-game round-trip tests that exercise the modified serializer paths.
- Full suite **3607/3612 pass**; the one failure (`FogVisualEvidenceTest`, a GPU visual test untouched by this PR) passes in isolation — a known flaky visual test, not a regression.

### DIFF
--- a/OloEngine/src/OloEngine/Core/InputActionManager.h
+++ b/OloEngine/src/OloEngine/Core/InputActionManager.h
@@ -2,6 +2,7 @@
 
 #include "OloEngine/Core/IInputProvider.h"
 #include "OloEngine/Core/InputAction.h"
+#include "OloEngine/Core/TransparentStringHash.h"
 
 #include <string>
 #include <string_view>
@@ -9,24 +10,8 @@
 
 namespace OloEngine
 {
-    // Transparent hash/equal for heterogeneous lookup with std::string_view
-    struct StringHash
-    {
-        using is_transparent = void;
-        [[nodiscard]] std::size_t operator()(std::string_view sv) const
-        {
-            return std::hash<std::string_view>{}(sv);
-        }
-    };
-
-    struct StringEqual
-    {
-        using is_transparent = void;
-        [[nodiscard]] bool operator()(std::string_view lhs, std::string_view rhs) const
-        {
-            return lhs == rhs;
-        }
-    };
+    // StringHash / StringEqual (transparent heterogeneous lookup) now live in
+    // Core/TransparentStringHash.h so other subsystems can reuse them.
 
     class InputActionManager
     {
--- a/OloEngine/src/OloEngine/Core/TransparentStringHash.h
+++ b/OloEngine/src/OloEngine/Core/TransparentStringHash.h
@@ -8,10 +8,11 @@
 namespace OloEngine
 {
     // Transparent hash/equal functors for heterogeneous lookup on associative
-    // containers keyed by std::string. Pairing these (plus std::equal_to<>) with
-    // an unordered container lets find/count/contains/erase accept std::string_view
-    // and const char* without materialising a temporary std::string (SonarQube
-    // cpp:S6045). The is_transparent marker is what enables the heterogeneous
+    // containers keyed by std::string. Pairing StringHash + StringEqual with an
+    // unordered container (std::unordered_map<std::string, V, StringHash, StringEqual>)
+    // lets find/count/contains/erase accept std::string_view and const char*
+    // without materialising a temporary std::string (SonarQube cpp:S6045). The
+    // is_transparent marker on both functors is what enables the heterogeneous
     // overloads.
     struct StringHash
     {
--- a/OloEngine/src/OloEngine/Core/TransparentStringHash.h
+++ b/OloEngine/src/OloEngine/Core/TransparentStringHash.h
@@ -1,0 +1,33 @@
+#pragma once
+
+#include <cstddef>
+#include <functional>
+#include <string>
+#include <string_view>
+
+namespace OloEngine
+{
+    // Transparent hash/equal functors for heterogeneous lookup on associative
+    // containers keyed by std::string. Pairing these (plus std::equal_to<>) with
+    // an unordered container lets find/count/contains/erase accept std::string_view
+    // and const char* without materialising a temporary std::string (SonarQube
+    // cpp:S6045). The is_transparent marker is what enables the heterogeneous
+    // overloads.
+    struct StringHash
+    {
+        using is_transparent = void;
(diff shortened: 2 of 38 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `4f23983088038685b465f2f89b14d1180af7d761`

Reviewed the full PR diff against `master` at **high** effort (8 finder angles → verify). Correctness finders (line-by-line, removed-behaviour, cross-file caller tracer) and the CLAUDE.md-conventions finder all returned **zero** findings — the refactor is verified behaviour-preserving and source-compatible (every QuestJournal getter consumer uses `auto`/iterators/range-for; `string_view` params are caller-compatible; no new float `==`; IWYU intact). A cleanup finder surfaced 5 consistency items (no correctness):

- **Fixed:** `Item.h` `ItemCategoryFromString`/`ItemRarityFromString` → `std::string_view` (matching the other `*FromString` helpers in this sweep) with unqualified `using enum` fallback returns; corrected the `TransparentStringHash.h` doc comment to reference the shipped `StringEqual` (was `std::equal_to<>`). Pushed as `4f239830`; `OloEngine-Tests` rebuilt clean, 155 Inventory/Item/Quest tests green.
- **Dismissed:** (1) `QuestJournal`'s local `StringSet`/`StringMap<V>` aliases vs the inline `StringHash, StringEqual` spelling in the 1–2-map files — deliberate; the alias pays off in the 9-container file, the explicit spelling is self-documenting elsewhere, and rewiring 6 verified files is churn disproportionate to the style gain. (2) `[[nodiscard]]` on the private static map accessors — added to clear `cpp:S6007`, which the analyzer raised on exactly those accessors; harmless, and clearing flagged issues is this PR's purpose.

**Verification:** `OloEngine` + `OloEngine-Tests` build clean (Debug/MSVC); all 228 Gameplay/Inventory/Quest/Ability/SaveGame tests pass (incl. the YAML + save-game round-trip tests exercising the modified serializer paths); full suite 3607/3612 (the lone `FogVisualEvidenceTest` failure passes in isolation — a known flaky GPU visual test untouched by this PR).

PR #418, opened 2026-06-27
chore(physics3d): SonarQube CODE_SMELL cleanup (no behaviour change)

drsnuggles8:
Mechanical SonarQube MAINTAINABILITY cleanup confined to
`OloEngine/src/OloEngine/Physics3D/**`. Same kind of work as the merged
per-subsystem sweeps #406 (Animation) and #407 (Audio): query SonarQube, fix a
coherent batch of smells, change no behaviour, keep physics tests green.

35 issues cleared across 8 files, 11 distinct rules:

- **S6166** (6): added explanation strings to existing `[[nodiscard]]`
  attributes (BuoyancySystem, JoltContactListener.{h,cpp}, JoltScene.h). No new
  bare attributes, so no new S6166 warnings.
- **S6007** (4): declared non-mutating const getters / pure mapping function
  `[[nodiscard(...)]]` (JoltScene.h count getters, JoltScene.cpp
  `ToJoltPathRotation`) — with a message to pre-empt S6166.
- **S6012** (2): CTAD on `std::unique_lock` / `std::lock_guard`
  (JoltJobSystemAdapter).
- **S6936** (1): renamed the local `clock` type alias → `Clock` (collides with
  libc `clock`) in JoltJobSystemAdapter.
- **S1271** (3): `::`-qualified global C-API / global constants — `::strlen`,
  `::_open` (JoltShapes), `::i32_max` (MeshCookingFactory).
- **S3574** (2): removed redundant lambda return types (`makeWheel`,
  `findExtreme`).
- **S1481** (1): removed a dead empty range-for over `m_Bodies` whose only effect
  was the unused binding; `m_Bodies.clear()` already runs the destructors.
- **S6004** (1): folded two `Ref<JoltBody>` locals into the `if` init-statement.
- **S6171** (2): `map.find(k) != end()` → `map.contains(k)` (constraints,
  vehicles).
- **S1121** (9): split chained `a1 = a2 = v;` constraint-axis assignments into
  independent statements.
- **S5350** (4): made vehicle/constraint range-for bindings `const auto&` and a
  `JPH::Body*` a `const JPH::Body*`.

No behaviour change. OloEngine + OloEngine-Tests build green; physics tests pass.
SonarQube auto-resolves the cleared rules on its next scan; no GitHub issue.

### Deliberately left (judgement / behaviour-risk / precedent)
- **S8417** (atomics memory-order, ~7) — changing the carefully-chosen
  acquire/release/relaxed orders to seq_cst is a behaviour change; deferred as
  "noise" exactly as in #407.
- **S5981** (`dynamic_cast` downcasts, ~13) — Jolt disables `std` RTTI and uses
  `static_cast` for constraint/controller downcasts by design; `dynamic_cast`
  would not compile / is wrong. Judgement rule, deferred as in #407.
- **S5025 / M23_329** (raw `new` / advanced allocation, ~12) — Jolt's
  `Ref<T>` / settings APIs take ownership of `new`-allocated objects; not a
  free-standing `make_unique` swap.
- **S1142** (too-many-returns, 11), **S3776** (cognitive complexity, 3),
  **S1188** (long lambda, 2), **S1151** (long switch case, 2) — branchy
  validation / converter flow; structural refactors out of scope for a
  behaviour-preserving sweep (matches #406's "clean guard merges only").
- **S131** (1) — the `JointType3D` switch already covers every enumerator with a
  post-switch fallback `return nullptr`; adding a `default` would trip clang's
  `-Wcovered-switch-default`.
- **S5945** (1) — a bounded `const JPH::BodyID[2]` feeding a pointer+count Jolt
  API; the C array is the idiomatic form and a swap would pull in `<array>` for
  one site.
- **S799** (2), **S1712** (1), **S5419** (3), **S1706 / S2486** (try/catch),
  **M23_235** (explicit lambda capture) — public-API renames / overload &
  parameter restructuring / try-catch reshaping; left as in #406/#407.
- **S6022** (3) — false positive: `ECollisionFlags` bitwise-OR via
  `std::to_underlying`; `std::byte` would be semantically wrong for a typed enum.
- **S5421** (1) — false positive: `s_JoltGlobalRefCount` is a mutable atomic
  ref-count and cannot be `const`.

`sonar-project.properties` is owned by the sibling `chore/sonarqube-analysis-setup`
branch and was **not** touched — all findings were addressed in code.

### DIFF
--- a/OloEngine/src/OloEngine/Physics3D/BuoyancySystem.cpp
+++ b/OloEngine/src/OloEngine/Physics3D/BuoyancySystem.cpp
@@ -38,7 +38,7 @@ namespace OloEngine
             f32 m_FFTHeightScale = 1.0f; ///< artist multiplier (WaterComponent::m_FFTHeightScale, u_FFTParams.z)
         };
 
-        [[nodiscard]] bool IsOverFootprint(const WaterVolume& w, const glm::vec3& worldPos)
+        [[nodiscard("footprint test result must be used")]] bool IsOverFootprint(const WaterVolume& w, const glm::vec3& worldPos)
         {
             const glm::vec4 local = w.m_InvModel * glm::vec4(worldPos, 1.0f);
             if (!std::isfinite(local.x) || !std::isfinite(local.z))
@@ -46,7 +46,7 @@ namespace OloEngine
             return local.x >= -w.m_HalfX && local.x <= w.m_HalfX && local.z >= -w.m_HalfZ && local.z <= w.m_HalfZ;
         }
 
-        [[nodiscard]] WaterSurface::Params MakeParams(const WaterComponent& wc, f32 planeHeight)
+        [[nodiscard("constructed water-surface params must be used")]] WaterSurface::Params MakeParams(const WaterComponent& wc, f32 planeHeight)
         {
             WaterSurface::Params p;
             p.m_WaveDir0 = wc.PackWaveDir0();
--- a/OloEngine/src/OloEngine/Physics3D/JoltContactListener.cpp
+++ b/OloEngine/src/OloEngine/Physics3D/JoltContactListener.cpp
@@ -171,13 +171,13 @@ namespace OloEngine
         }
     }
 
-    [[nodiscard]] UUID JoltContactListener::GetEntityIDFromBody(const JPH::Body& body) const noexcept
+    [[nodiscard("entity UUID lookup result must be used")]] UUID JoltContactListener::GetEntityIDFromBody(const JPH::Body& body) const noexcept
     {
         // The entity ID is stored in the body's user data
         return static_cast<UUID>(body.GetUserData());
     }
 
-    [[nodiscard]] u32 JoltContactListener::GetPhysicsLayerFromBody(const JPH::Body& body) const noexcept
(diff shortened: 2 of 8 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `9e49b8a294b1d985a6329149780aceb45f19be95`
Reviewed the full PR diff at **medium** effort (line-by-line correctness, removed-behavior audit, cross-file caller trace, conventions).

- **Findings:** 0 actionable.
- Verified behavior-preserving: the 9 split chained-assignments keep identical values; the `if`-init two-declarator fold doesn't change `bodyA`/`bodyB` scope; the deleted empty range-for had no side effects (`m_Bodies.clear()` still runs the destructors); `const auto&` / `const JPH::Body*` bindings are never mutated through; `::strlen`/`::_open`/`::i32_max` resolve to the same global symbols (`i32_max` lives at global scope in `Core/Base.h`); `std::unique_lock`/`std::lock_guard` CTAD deduces the identical type; `[[nodiscard("...")]]` messages match between decl and def and no call site discards a now-nodiscard return.
- Build: `OloEngine` + `OloEngine-Tests` green (Debug). Tests: 186 physics/joint/vehicle tests pass locally.

Note: CodeRabbit's automated review is currently rate-limited (retry ~1h); its pass will land later and can be picked up by the next `/pr-status`.

Issue #357, opened 2026-06-20
MCP server: protocol-conformance & tool-authoring roadmap (benchmarked vs UE 5.8)


drsnuggles8:
Follow-up to #306 and #316. Where those track the MCP **feature surface** (explain

tools, eventing, writes, rendering/visual-inspection harness), this issue tracks the

**protocol-conformance & tool-authoring** axis — surfaced by a line-level comparison of

OloEngine's `OloEditor/src/MCP` against UE 5.8's

`Engine/Plugins/Experimental/ModelContextProtocol`.

The original baseline (verified against `master` 2026-06-20): **36 `olo_*` tools** in a

flat `std::vector<ToolDef>`; protocol negotiation tops at `2025-06-18`; `ToolResult` was

`content` + `isError` only; `ToolDef` had no `title`/annotations/output-schema/toolset.

> **Overlap note:** the headline "live push" bet (Tier 3) is the *same work* as

> **#306-B** (`resources/subscribe` / `GET /mcp`). It shipped under #306-B (#414).

---

## Status — updated 2026-06-28

**Tier 1 and Tier 2 are fully delivered. The Tier-3 live-push bet shipped under #306-B

(#414). The only remaining work is gated/blocked: script-defined tools (needs a sandboxing

design) and the deferred-by-design protocol items (P3 phase 2, P5b, P6).**

| Item | Tier | State | Delivered by |

|---|---|---|---|

| P1 — tool annotations + `title` | 1 | ✅ done | annotations on `ToolDef`, emitted in `tools/list` |

| P5a — tool-name validation at `RegisterTool` | 1 | ✅ done | |

| P2 — `structuredContent` + `outputSchema` | 2 | ✅ done | Json-producing tools migrated |

| P4 — schema-builder DSL | 2 | ✅ done | **#421** (`McpSchemaBuilder.h`); enforced on dispatch by **#423** |

| P3 phase 1 — toolset tagging | 2 | ✅ done | `Toolset` tag + `tools/list` filter |

| Live `resources/subscribe` push | 3 | ✅ done | **#414** (== #306-B; `GET /mcp` SSE) — incidentally the progress/cancel substrate |

| Script-defined tools (C#/Lua) | 3 | ⬜ **blocked** | needs a sandboxing design (read-only-project posture) |

| P3 phase 2 — toolset meta-tools / tool-search | 2 | ⬜ deferred | until the `tools/list` payload is a *measured* context cost |

[...]

PR #421, opened 2026-06-28
feat(mcp): schema-builder DSL + migrate tool schemas (#357 P4)

drsnuggles8:
## What

Adds a small header-only C++ **schema-builder DSL** for the MCP diagnostics-server tools and migrates every hand-written JSON `inputSchema` / `outputSchema` literal to it. This is **item P4** of #357 ("MCP server: protocol-conformance & tool-authoring roadmap", Tier 2).

## Why

Today the 41 built-in MCP tools hand-write raw `nlohmann::json` schema literals inside `RegisterBuiltinTools` (~48 literals) — verbose and easy to get subtly wrong (a dropped `additionalProperties`, an `"integer"` typo'd as `"string"`, a forgotten bound). A fluent builder is a force-multiplier for every future tool and removes that whole class of mistake, with none of the reflection machinery a full reflection system would need.

## Changes

- **New `OloEditor/src/MCP/McpSchemaBuilder.h`** — header-only DSL, `namespace OloEngine::MCP::Schema`:
  - Factories: `Object()`, `EmptyObject()`, `Int()/Number()/Bool()/String()/Array()/Array(elem)`.
  - Modifiers: `Desc/Min/Max/ExclusiveMin/MinItems/MaxItems/Enum/Items/Prop/Required/NoAdditional`.
  - Recurring-shape helpers: `Vec3(desc)`, `EntityId(desc=…)`, `.Pagination(pageSizeDesc)`, plus a `Raw(json)` escape hatch (e.g. the `sinceId` multi-type `{"type":["integer","string"]}`).
  - A `Node` converts implicitly to `Json`, so it drops straight into `ToolDef::InputSchema` / nested `Prop()` / `Array()`.
- **`OloEditor/src/MCP/McpTools.cpp`** — migrated all 48 schema literals across the 41 tools (41 input + 5 output). Net **−138 lines**; zero raw `Json{…}` schema literals remain.
- **New `OloEngine/tests/MCP/McpSchemaBuilderTest.cpp`** (+ `tests/CMakeLists.txt`) — header-only (does not link `McpTools.cpp`, which pulls the whole renderer/physics stack); 20 tests pinning each DSL shape and proving a representative cross-section is byte-identical to the old literals.

No behaviour / protocol / dispatch change — schemas are read-only metadata. Scope is confined to `OloEditor/src/MCP/` plus the one test file.

## Equivalence guarantee (the wire schema clients see is unchanged)

The repo uses the default `nlohmann::json` (a sorted `std::map`), so object-key insertion order is irrelevant to `==`/`.dump()`; only array order matters (enum / required / items), which the builder preserves. Numeric bounds are stored as JSON **integers** (`Min/Max` take `i64`) so `.dump()` stays `"1"`, never `"1.0"` (and `json(1)==json(1.0)` is true, so the tests assert `.dump()` too, not just `==`).

Proven three independent ways:
1. A standalone harness compiling the real header against real nlohmann — 10/10 representative schemas `==` **and** `.dump()`-identical.
2. A full string-literal diff of the original vs migrated `RegisterBuiltinTools`: zero strings introduced (no typos), zero content dropped (the only relocated strings moved into the `EntityId`/`Pagination` helpers).
3. The 20 `McpSchemaBuilder*` gtest cases.

## Testing

- `OloEditor` (Debug) and `OloEngine-Tests` (Debug) both build clean under MSVC — no warnings on the new/changed files.
- `OloEngine-Tests.exe --gtest_filter=Mcp*` → **191/191 passed**, including the new `McpSchemaBuilder` (9) + `McpSchemaBuilderEquivalence` (11) and the existing `McpDispatch` / `McpToolAnnotations` / `McpStructuredOutput` `tools/list` tests (no regression).

Part of #357 (Tier 2, P4).

### DIFF
--- a/OloEditor/src/MCP/McpSchemaBuilder.h
+++ b/OloEditor/src/MCP/McpSchemaBuilder.h
@@ -103,13 +103,6 @@ namespace OloEngine::MCP::Schema
             return *this;
         }
 
-        // Array element schema (alternative to the Array(elem) factory).
-        Node& Items(const Node& elem)
-        {
-            m_Json["items"] = elem.ToJson();
-            return *this;
-        }
-
         // ---- object modifiers --------------------------------------------------
 
         // Add one named property, lazily creating the `properties` map. An object
@@ -147,14 +140,6 @@ namespace OloEngine::MCP::Schema
         Node& Pagination(std::string_view pageSizeDesc)
         {
             return Prop("page", IntMin0Page()).Prop("pageSize", PageSize(pageSizeDesc));
-        }
-
-        // Escape hatch: set an arbitrary keyword (e.g. a multi-type `type`). Kept so
-        // a one-off shape never forces a literal back into McpTools.cpp.
-        Node& Set(std::string_view keyword, Json value)
-        {
-            m_Json[std::string(keyword)] = std::move(value);
-            return *this;
         }
 
         [[nodiscard]] const Json& ToJson() const
--- a/OloEditor/src/MCP/McpSchemaBuilder.h
+++ b/OloEditor/src/MCP/McpSchemaBuilder.h
@@ -1,0 +1,255 @@
+#pragma once
+
+// A tiny, header-only fluent DSL for authoring the JSON Schemas the MCP
+// diagnostics tools declare (issue #357 P4). Every built-in tool used to
+// hand-write a raw `nlohmann::json` literal for its `inputSchema` /
+// `outputSchema` (≈48 literals across McpTools.cpp), which is verbose and easy to
+// get subtly wrong — a dropped `additionalProperties`, an `"integer"` typo'd as
+// `"string"`, a `minimum` bound forgotten. This builder captures the handful of
+// shapes those schemas actually use behind a small chainable API so a schema
+// reads like the constraint it expresses:
+//
+//     tool.InputSchema = Schema::Object()
+//         .Prop("count", Schema::Int().Min(1).Max(200).Desc("How many lines (default 50)."))
+//         .Prop("tag", Schema::String().Desc("Only lines whose [Tag] matches."))
+//         .NoAdditional();
+//
+// Design / equivalence guarantee
+//   * A `Node` is a thin wrapper around one `nlohmann::json` object. Every
+//     factory/modifier just sets keys, so the emitted JSON is exactly what the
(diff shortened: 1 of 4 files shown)


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `20aeac4794b5472eeebe830693b4f49e4658c407`
Reviewed the full PR diff vs `master` at **high** effort (correctness + cleanup + altitude + conventions).

- **Correctness findings:** 0. An independent reviewer rebuilt the actual branch header against vendored nlohmann and compared every migrated schema's `.dump()` to its verbatim pre-P4 literal — all 41 tools (23 not in the test, the 14 in the test, and the 11 `EmptyObject()` cases) matched byte-for-byte. The repo's default `nlohmann::json` is a sorted `std::map`, so object-key order is irrelevant; array order (enum/required/items) and integer-vs-float bounds are preserved.
- **Fixed (1):** removed two dead public methods from the new DSL — `Node::Items(elem)` (zero call sites; fully redundant with the `Array(elem)` factory that every site uses) and `Node::Set(key, value)` (zero call sites; the one multi-type node uses `Raw()`). Trims untested/unused surface from a brand-new public DSL. Commit `20aeac4794b5472eeebe830693b4f49e4658c407`; `Mcp*` tests remain 191/191.
- **Dismissed (3):**
  - `MinItems`/`MaxItems` "no direct callers" — kept: they're exercised through the tested `Vec3()` helper and are natural JSON-Schema array keywords worth exposing.
  - `EmptyObject()` vs `Object().NoAdditional()` "wire-identity bandaid" — kept: the lazy-`properties` design is *required* so an opaque sub-object (`Object().Desc()`, e.g. screenshot `camera`/`orbit`) carries **no** `properties` key; `EmptyObject()` is the correct explicit helper for the distinct closed-empty shape, not a bandaid.
  - `operator Json()` copy vs `ToJson()` ref — kept: schemas build once at editor startup; the value copy is intrinsic to the immutable-node value semantics.

**Conventions (CLAUDE.md):** no violations — 4-space indent matches `McpServer.h`, includes split `<...>`/`"..."`, `#pragma once`, header self-contained, test carries `// OLO_TEST_LAYER: unit`. The test's `EXPECT_EQ` compares `nlohmann::json`/`std::string`, not raw `float`, so it doesn't break the float-equality rule.

Issue #231, opened 2026-05-22
Sphere area lights (component + shader + editor gizmo)


drsnuggles8:
## Summary

Today's lights are point, spot, and directional — all of them treated as zero-area emitters with a `dot(N,L)` per fragment. That's fine for game-style lighting but produces hard, plastic-looking highlights on PBR surfaces. Add **sphere area lights**: lights with a physical radius that produce soft shadows and a representative-point specular highlight matching the analytic shape. This is the lowest-cost area-light primitive and the most natural fit for the existing `s_Data.SphereMesh` editor visualization.

## Current state

- [`Components.h`](OloEngine/src/OloEngine/Scene/Components.h#L888-L965) defines `DirectionalLightComponent`, `PointLightComponent`, `SpotLightComponent`. No area lights.
- [`DrawPointLightGizmo`](OloEngine/src/OloEngine/Renderer/Renderer3DUtilityDraws.cpp#L861) draws a 3-circle wireframe sphere as the bulb icon — same wireframe convention as collider/audio gizmos.
- PBR shaders (`PBR.glsl`, `PBR_MultiLight.glsl`) compute `L = normalize(lightPos - fragPos)` and treat lights as points.
- The light UBO layout would need to grow (or a separate area-light UBO).

### 1. `SphereAreaLightComponent`

```cpp
struct SphereAreaLightComponent
{
    glm::vec3 Color = glm::vec3(1.0f);
    f32 Intensity = 1.0f;
    f32 Radius = 0.5f;       // Physical emitter radius
    f32 Range = 10.0f;       // Falloff distance
    bool CastsShadow = false; // Soft shadows are a separate follow-up
};
```

Five-touchpoint integration per `CLAUDE.md` (Components tuple, SceneSerializer, SaveGameComponentSerializer, OLO_PROPERTY, LuaScriptGlue).

### 2. Shader math

Use the **representative-point** technique (Karis 2013, *Real Shading in Unreal Engine 4*) for the specular term — cheap, well-known, looks right:

```glsl
vec3 r = reflect(-V, N);
vec3 L = lightPos - fragPos;
vec3 centerToRay = dot(L, r) * r - L;
vec3 closestPoint = L + centerToRay * clamp(radius / length(centerToRay), 0.0, 1.0);
vec3 Lspec = normalize(closestPoint);
```

Use this `Lspec` for the GGX specular evaluation. The diffuse term keeps using `normalize(L)` plus a solid-angle correction:

```glsl
[...]

PR #422, opened 2026-06-28
feat(renderer): sphere area-light shadows via point-cubemap pool

drsnuggles8:
## Summary

Gives the existing **sphere area lights** the ability to **cast shadows**. The PBR *shading* for sphere area lights already shipped (#231); the `SphereAreaLightComponent::m_CastShadows` field was a hardcoded no-op with a "follow-up" comment. This implements that follow-up: a first slice of area-light shadowing (hard shadow from the representative point).

## Design

A sphere area light treated as a point at its centre (the representative point) is exactly a point light for shadowing — so casting area lights **reuse the existing point-light cubemap shadow pool** (max 4, shared with point lights). The win: the entire shadow *generation* path (`ShadowRenderPass`, `ShadowMap`, render graph) needs **zero changes** — the existing point-cubemap loop renders the area light's shadow automatically once it's registered.

- **`Scene.cpp`** — casting area lights share the point-shadow index counter: tagged into `direction.w` in the light-upload loop, registered via `SetPointLightShadow(centre, range)` in the shadow-setup loop (both continuing after point lights, capped at `MAX_POINT_SHADOWS`).
- **5 shaders** (`DeferredLightingShared`, `PBR_MultiLight`, `PBR_MultiLight_Skinned`, `Terrain_PBR`, `Terrain_Voxel`) — extended the existing `POINT_LIGHT` shadow branch to `|| SPHERE_AREA_LIGHT`; the cubemap lookup is identical.
- **`Components.h`** — `m_CastShadows` is now opt-in working; comment updated.

No new serialized fields (`m_CastShadows` was already serialized) — shadow tunables (bias, resolution) stay renderer-internal constants for this slice.

## Verification (all three layers per CLAUDE.md)

1. **CPU contracts** (16 tests) — representative-point cubemap depth-compare (occluded vs lit, range-not-radius as far plane) + pool-sharing invariants (no point/area collision, cap, casters-only).
2. **Differential PNG evidence** — `SphereAreaLightShadowEvidenceTest` renders the real pipeline from 2 angles; CastShadows off→on darkens the cube's lee floor while the lit side is unchanged (golden-free, GPU-independent). Committed On/Off PNGs show a clean hard shadow.
3. **Live editor (MCP)** — confirmed the shadow from multiple angles; `ShadowMapPoint0` capture showed the occluder rendered into the cubemap (min-depth 0.16).

**Regression:** 95 tests across 27 shadow/light/rendering suites pass; `AssetSceneLoad` loads the new demo scene cleanly.

## Known limitation

Like point/spot shadows, this applies on the non-Forward+ shading path (< ~8 lights). Under Forward+, area lights — same as point/spot today — are shaded without shadows. Documented in code.

## Follow-up (Phase 2)

Soft penumbra sized from the emitter radius (PCSS), already wired for directional/spot/point via `SoftShadowMode` — left as an explicit follow-up.


drsnuggles8:
## 🤖 Self-review (finish-pr) @ `77ae4c6cad71adeb80117df2c9f9ed349d364b7c`
Reviewed the full PR diff against `master` at **high** effort (8 finder angles → verify): correctness on `Scene.cpp` + the 5 shader edits, plus a dedicated test-quality pass and a cleanup/altitude/conventions pass.

- **Correctness:** none. Both upload and shadow-setup loops iterate the same EnTT views in the same order with the same `m_CastShadows && idx < MAX_POINT_SHADOWS` predicate, so the `direction.w` slot a light is tagged with always matches the cubemap registered for it. The removed unconditional `direction = (0,-1,0,-1)` is safe — `calculateSphereAreaLightContribution` never reads `direction`. The shader branch guards `pointShadowIdx >= 0`, so the −1 (no-shadow) case is handled.
- **Fixed (2, test-quality):**
  - `ClearLineOfSightLeavesReceiverLit` was tautological (`storedDepth := currentDepth` → held for any bias, pinned nothing) → rewrote as `BlockerBehindReceiverLeavesReceiverLit` with an independently-computed stored depth from a blocker behind the receiver.
  - Documented `AssignPool`'s scope (models the MAX_POINT_SHADOWS slot cap, not the separate MAX_LIGHTS UBO cap) to prevent future misuse.
- **Dismissed:**
  - Visual-test scan bands overlapping the cube — refuted by directly inspecting the committed On/Off PNGs: the lit/lee bands sample floor, the cube sits between them.
  - `MaxDarkening` max-cell noise sensitivity — mirrors the shipped `ContactShadowVisualEvidenceTest` pattern; the real signal is ~100 luma vs the 8.0 floor, and the `litDrop < leeDrop*0.5` control band guards against global noise.
  - Upload-loop MAX_LIGHTS `continue` not mirrored in setup — pre-existing point-light behavior, needs >256 lights, out of scope.

Issue #424, opened 2026-06-28
Audio: route SoundGraph voices through the 3D spatializer (SetLocation/SetVelocity/SetOrientation)


drsnuggles8:
## Background

`SoundGraphSound::SetLocation` / `SetVelocity` / `SetOrientation` (`OloEngine/src/OloEngine/Audio/SoundGraph/SoundGraphSound.cpp`) currently only store their values (backing `GetLocation()`/`GetVelocity()`); nothing reaches a spatializer, so a SoundGraph voice is **not 3D-positioned**.

This was split off from the parameter-routing work (branch `feature/audio-soundgraph-param-routing`), which wired the cleanly-routable controls (Volume/Pitch/Looping/LowPass/HighPass) into the graph as conventional input parameters. 3D was deferred because it needs new infrastructure, not a parameter route.

## Why it needs infra (not a param route)

- `Audio::DSP::Spatializer` (`AudioEngine::GetSpatializer()`) does VBAP panning, but its `InitSource(u32 sourceID, ma_engine_node* nodeToInsertAfter, const AudioSourceConfig&)` is built around an `ma_engine_node` and is wired only into the `ma_sound`-based `AudioSource` path.
- `SoundGraphSource::Initialize` attaches a **bare custom `ma_node`** (`SoundGraphMiniaudioNode`) straight to `ma_node_graph_get_endpoint`. There is no `ma_engine_node` to insert a spatializer node after, and the output currently bypasses any panning stage.

## Scope of the follow-up

1. Give `SoundGraphSource` an output path that can host a spatializer node (insert a `Spatializer` source node between the SoundGraph node and the endpoint), with a per-voice `sourceID` and lifecycle (register on init / `ReleaseSource` on shutdown/ReplaceGraph).
2. Route `SoundGraphSound::SetLocation/SetVelocity/SetOrientation` to `Spatializer::UpdateSourcePosition(sourceID, transform, velocity)`; ensure listener updates happen (`Spatializer::UpdateListener`).
3. Decide ownership of the spatialization-enabled flag (cf. `SoundConfig::m_SpatializationEnabled`) and wire it from `AudioSoundGraphComponent`/Scene.
4. Tests: position -> gain/pan effect (the spatializer already exposes `GetCurrentDistanceAttenuation`/`GetCurrentConeAngleAttenuation`/`GetCurrentDistance` for assertions).

## References
[...]

PR #426, opened 2026-06-28
feat(audio): route SoundGraphSound volume/pitch/looping/filter setter…

drsnuggles8:
…s into the live graph

SoundGraphSound's high-level controls (SetVolume/SetPitch/SetLooping/ SetLowPassFilter/SetHighPassFilter) only stored a member and never reached the running SoundGraph, so a scene's authored VolumeMultiplier/PitchMultiplier/Looping (forwarded by Scene::InitializeAudioSoundGraph) was silently dropped.

They now route the finite-checked, clamped value into the graph as a conventionally- named input parameter (Volume/Pitch/Loop/LowPass/HighPass) via the proven SoundGraphSource::SetParameter path. Routing is opt-in/best-effort: a graph reacts only if it exposes a matching input stream, else the write no-ops. SyncControlParameters- ToGraph re-applies stored controls on graph install so a set-before-init value lands.

Also: InitializeDetachedSource() seam for headless/device-free use; clamp + std::isfinite guards on the float setters; honest comments replacing the stale "stub"/"dummy voice ID" markers (StopNow, SetParameter int/bool catch blocks, Update).

3D spatialization (SetLocation/SetVelocity/SetOrientation) is deferred — the Spatializer is ma_engine_node-based while SoundGraphSource uses a bare ma_node straight to the endpoint, so it needs new infra, not a param route. Tracked in #424.

Tests: SoundGraphControlRoutingTest.cpp (7 unit tests) drives the real ProcessSamples path device-free, proving each control reaches its graph cell (and per-frame downstream), clamping-before-routing, non-finite rejection, set-before-init ordering, and graceful no-op when the endpoint is absent.

## Summary by CodeRabbit

* **New Features**
  * Added live routing for sound controls into the audio graph, including volume, pitch, looping, and filter settings.
  * Preset control values now carry over when a graph is installed.

* **Bug Fixes**
  * Added safer handling for invalid values and runtime errors during sound updates.
  * Improved behavior when controls are used without an active graph connection.

* **Tests**
  * Added end-to-end coverage for control routing, value validation, and graph synchronization.

### DIFF
--- a/OloEngine/src/OloEngine/Audio/SoundGraph/SoundGraphSound.cpp
+++ b/OloEngine/src/OloEngine/Audio/SoundGraph/SoundGraphSound.cpp
@@ -7,6 +7,8 @@
 
 #include <choc/containers/choc_Value.h>
 #include <algorithm>
+#include <cmath>
+#include <string_view>
 // AudioEngine.h forward-declares ma_engine; we need the full miniaudio API surface here
 // for ma_engine_get_sample_rate (and the implicit pointer arithmetic on ma_engine*).
 #include <miniaudio.h>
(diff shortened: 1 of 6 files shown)


drsnuggles8:
@coderabbit review

PR #450, opened 2026-06-28
feat(codegen): generate SceneSerializer blocks for trivial components (#380)

drsnuggles8:
## Summary

Extends **OloHeaderTool** to generate the per-component scene-YAML **serialize/deserialize** blocks, collapsing the last big *unguarded* hand-maintained ECS touch-point (`Scene/SceneSerializer.cpp`). A component whose field you forgot to hand-serialize used to round-trip through nothing — **silent scene-data loss** that the old name-only `ComponentSerializerCoverageTest` couldn't catch. Codegen now *prevents* the miss.

Fourth slice of the #380 epic (after AllComponents tuple #386, SaveGame lists #405, OnComponentAdded/Removed #413).

## Scope — the safe 13 of 99

The generator drives off a **new full data-member scan** (`CollectComponentFields`), **not** the `OLO_PROPERTY` scan — the serializer persists *every* field, not just script-exposed ones (e.g. `DirectionalLightComponent` serializes 9 fields but annotates only 3).

A component is auto-generated **iff** every data member is a primitive / `glm::vec*` / `std::string` **and** it isn't in the new `kComponentsCustomSerialize` exclusion set. Anything with a non-trivial field (enum, `AssetHandle`, `Ref<T>`, `std::vector`, nested struct, `glm::quat`/`mat`/`ivec`) is classified non-trivial and **skipped automatically** — no exclusion entry needed.

The 13 migrated components: CharacterController3D, CircleRenderer, DirectionalLight, PointLight, SpotLight, LocalizedText, LuaScript, Nameplate, NetworkInterest, Perceptible, UICheckbox, UIRectTransform, UIToggle.

Each was verified to be a **plain round-trip** (parsed-field set == hand-written YAML-key set) before migration. That check also correctly *rejected* `VehicleComponent` (a runtime-only `RuntimeVehicleToken` field the serializer deliberately omits) and `ScriptComponent` (writes the external C# ScriptField map, not just its `ClassName` member) — both stay hand-written.

## What changed

- **`tools/OloHeaderTool/main.cpp`** — field scanner + classifier (`SceneSerType`), two emitters, `kComponentsCustomSerialize` exclusion set, and an empty-output guard (refuses to overwrite with 0 components).
- **`Scene/Generated/Scene{Serialize,Deserialize}Components.Generated.inl`** (new, tracked) — `#include`d at the end of `SerializeEntity` / `DeserializeEntityComponents`. Floats validated via `TryReadFiniteF32` / glm `Decode`; a missing key keeps the constructor default.
- **`SceneSerializer.cpp`** — 26 hand-written blocks removed, 2 includes added.
- **`ComponentSerializerCoverageTest.cpp`** — existence check now spans `SceneSerializer.cpp` ∪ the generated `.inl`; new **`NoComponentIsBothHandWrittenAndGenerated`** disjointness test fails loudly on a double-emit.
- **`ComponentRoundTripTest.cpp`** — 3 new tests (LuaScript/string, Perceptible/i32, Nameplate/mixed-vec); the other 10 migrated components already had round-trip tests.
- **`CLAUDE.md`** — touch-point + OloHeaderTool docs updated.

## Contract / safety

Every component is handled by **exactly one** of the hand-written serializer or the generated `.inl`:
- **drop** → existence check fails (name in neither corpus)
- **double-emit** → disjointness check fails (duplicate YAML key on write / `AddComponent<T>` twice on read)

A new *trivial* component is auto-serialized with zero edits; a new *complex* one fails the coverage test until hand-written.

## Testing

> ⚠️ This worktree's environment cannot link the full `OloEngine-Tests` binary (pre-existing: ffmpeg source-build + assimp `.rc`-under-clang-cl both fail), so the GoogleTest run is left to CI.

Validated locally with the real clang-cl C++23 toolchain:
- All four changed/new translation units **compile cleanly**.
- OloHeaderTool builds, runs, and emits exactly the intended 13-component output (other generated files unchanged).
- Both generated `.inl` are clang-format-stable.

## Deferred (follow-up)

Complex-field serialization (asset handles, `Ref<T>`, enums, `std::vector`, nested structs, setter-only `SceneCamera`, and trivial components needing custom range-clamp/`Sanitize` validation) remains hand-written. A follow-up issue will track the rest of #380's SceneSerializer work.


drsnuggles8:
Deferred complex-field SceneSerializer codegen is now tracked in #451. Source issue #380 commented (umbrella, not closed).

Issue #428, opened 2026-06-28
Physics: terrain has no collision — wire Jolt HeightFieldShape into TerrainComponent


drsnuggles8:
## Problem
Terrain renders fully (quadtree LOD, tessellation, splatmaps, streaming) but has **no physics collision**. Jolt's heightfield shape is explicitly rejected today:

- `OloEngine/src/OloEngine/Physics3D/JoltShapes.cpp:542` — `GetShapeType: HeightField shape type {0} not supported, defaulting to Box`
- `OloEngine/src/OloEngine/Physics3D/JoltBinaryStream.cpp:394` reads a `HeightField` case, but no shape is ever constructed for terrain.
- No terrain-collider component exists under `OloEngine/src/OloEngine/Terrain/`.

Consequence: the player falls through terrain, NPCs can't walk on it, vehicles can't drive on it.

## Why it matters
**Genre-blocking** for an FPS — collision against the ground is table stakes. This is the single most urgent Tier 1 gap.

## Proposed approach
- Build `JPH::HeightFieldShapeSettings` from the terrain heightmap (CPU heights already available via `TerrainData::GetHeightAt`), matching world scale and tile origin.
- Add a `TerrainColliderComponent` (or extend `TerrainComponent`) that creates a static Jolt body from the heightfield on scene start / terrain edit. (Remember the cross-binding touch-points if adding a component.)
- For streamed terrain (`TerrainStreamer`), create/destroy heightfield bodies per loaded tile; update the dirty region on sculpt/erosion edits (partial heightfield update).
- Resolve `JoltShapes.cpp::GetShapeType` to handle `HeightField` for serialization round-trip.

## Acceptance criteria
- A character controller and a vehicle rest on / drive over terrain without falling through.
- Raycasts (AI perception, weapons) hit the terrain surface.
- Streamed tiles add/remove collision in step with their visual load/unload.
- Sculpt/erosion edits update collision for the affected region.

---
**Tier 1 — architectural.** From the "production-ready open-world FPS" gap analysis.

## Score
```olo-score
capability: 8
craft: 3
stability: 3
decay: 1
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```
[...]

PR #468, opened 2026-06-29
feat(physics): wire Jolt HeightFieldShape into terrain collision (#428)

drsnuggles8:
## What & why

Closes the **#428** Tier-1 gap: terrain had **no collision** — players fell through it, NPCs/vehicles couldn't traverse it, and AI/weapon raycasts missed it. This builds a static `JPH::HeightFieldShape` body from the terrain's CPU height field so characters, vehicles, and raycasts interact with the surface, matching the rendered mesh exactly.

## How

- **`JoltShapes::CreateTerrainHeightFieldShape`** — maps the terrain's `worldSize` / `heightScale` / entity scale onto Jolt's `offset + scale·(x, sample, z)` surface formula. Jolt's sample layout (`index = z·N + x`, normalized heights) matches `TerrainData::GetHeightData()` directly. Odd resolutions are padded up to the block-size multiple (Jolt requires `sampleCount % blockSize == 0`); bit-depth is chosen adaptively for ~2 cm precision; min/max is auto-derived.
- **`ShapeType::HeightField`** + stringifier + `GetShapeType` round-trip fix (it was silently defaulting to `Box`).
- **`JoltScene`** static terrain-body lifecycle (`CreateTerrainBody` / `DestroyTerrainBody` / `DestroyAllTerrainBodies`) on the `NON_MOVING` layer; body `UserData` = entity UUID so raycasts/contacts resolve back to the terrain entity; torn down in `Shutdown` / `OnRuntimeStop` / `OnPhysics3DStop`.
- **`TerrainComponent.m_CollisionEnabled`** (serialized, default **on**) + runtime body token. Round-trips through **scene YAML** and **save-game** (format `v5 → v6`).
- **Scene wiring** — collision is built **GPU-free** from CPU heights at `OnPhysics3DStart` (reads `TerrainData` if already built, else `TerrainGenerator::GenerateHeightField` — the identical field the render path uploads), so it works headless and on the dedicated server. Re-synced on runtime `Regenerate`.

## Tests (104/104 green, incl. all cross-binding coverage guards)

New `TerrainCollisionTest.cpp` (Functional, headless — real `Scene::OnPhysics3DStart` + Jolt step):
- a dynamic body **rests on flat** terrain (translated up via TransformComponent — proves transform application),
- a dynamic body **rests on procedural** terrain and the rest height **tracks the local surface** (proves the shape follows the height field; uses an **odd** resolution to exercise the padding path),
- a **raycast** down hits the terrain surface and resolves to the terrain entity.

Coverage guards (ComponentTuple/Handler/Serializer + SaveGame), save-game round-trips, `SceneSerializer`, `LuaDrivesTerrainRegeneration`, and `JoltShapes` all still pass.

## Scoped to a follow-up (filed separately)

Streamed per-tile collision (`TerrainStreamer`) and sculpt/erosion dirty-region partial updates (`HeightFieldShape::SetHeights`).

## Summary by CodeRabbit

* **New Features**
  * Added terrain collision support so static terrain can now participate in physics and raycasts.
  * Terrain collision can be enabled or disabled per terrain, and the setting is saved with scenes and game data.

* **Bug Fixes**
  * Improved terrain physics setup and shutdown so collision bodies are created and removed reliably.
  * Fixed terrain shape handling so height-based terrain is recognized correctly instead of falling back to a generic shape.


drsnuggles8:
@coderabbit review

Issue #480, opened 2026-06-30
Renderer: FSR1 EASU/RCAS spatial upscale (render-at-lower-res + upscale) — follow-up to #432 CAS slice


drsnuggles8:
## Problem
The CAS sharpening slice of #432 has shipped (PR #479): `UpscalerRenderPass` + `PostProcess_CAS.glsl`, a contrast-adaptive sharpen running post-tonemap on the native-res LDR image. What remains is the actual **spatial upscale** — the part that buys the GPU-time win: render the scene below display resolution, then upscale to display res with **FSR1 EASU** (edge-adaptive spatial upsampling) followed by **RCAS** (robust contrast-adaptive sharpening) before the UI composite.

## Why it matters (Tier 2 — performance ceiling)
A 50 km² world at 60 fps on mid-range PC GPUs realistically needs render-at-lower-res + upscale; TAA + CAS at native res can't recover that. EASU/RCAS are pure pixel/compute shaders (AMD FidelityFX, license-permissive) and port to GLSL.

## Scope (builds on the shipped CAS scaffold)
- Port **EASU** and **RCAS** (FSR1). RCAS replaces / extends the existing CAS kernel; EASU is the new upscale stage.
- **Render-scale plumbing**: render the scene at a Quality/Balanced/Performance scale factor into a smaller target; EASU upscales to display res. This is the missing piece — the current `UpscalerRenderPass` is resolution-preserving (sharpen only).
- EASU must run **early** (on the HDR/scene colour, before display-res post like Bloom/DOF), with RCAS/CAS sharpening **late** (post-tonemap, where the current CAS pass already sits) — so the single pass likely splits into an EASU stage + the existing RCAS/CAS stage. See the post-tonemap vs HDR placement note in `docs/agent-rules/glsl-shaders.md §9`.
- Hook into the existing DRS bounds UBO (binding 33, `DRSUBOData`) so screen-space passes clamp UV to the rendered region.
- Quality / Balanced / Performance scale factors selectable in `RendererSettingsPanel`.

## Acceptance criteria
- Scale factors selectable and applied (scene renders below display res, upscales cleanly).
- Measurable GPU-time win at a fixed display resolution with acceptable image quality (multi-angle screenshot evidence + a contract test, per the CLAUDE.md rendering rule).

## Relationship
[...]

Issue #306, opened 2026-06-07
MCP diagnostics server: follow-ups (explain tools, eventing, writes, hardening)


drsnuggles8:
Follow-ups to the read-only MCP diagnostics server (#285). That PR landed Phase 0–3:
the localhost Streamable-HTTP transport, the main-thread marshal primitive, and a
read-only surface of tools + resources + prompts (see `docs/mcp-diagnostics-server.md`).
These are the deliberately-deferred next steps, grouped so each bullet is a clean,
self-contained PR.

The guiding principle stays **"expose, don't embed"** and **read-only until the
advisory version earns trust** — except item C below, which is the considered move
across that line.

---

## Status — updated 2026-07-02

**All four items (A, B, C, D) are now fully delivered.** Item C (consented, undoable
writes) — its own epic — is complete: the session write gate + four write tools + the
per-action consent **dialog** (#512) have all landed. This issue can be closed.

| Item | State | Delivered by |
|---|---|---|
| A — Explain tools | ✅ done | `olo_render_why_not_visible` + full `olo_physics_*` family |
| B — Eventing | ✅ done | `olo_events_tail` (#343) **and** SSE push over `GET /mcp` (#414) |
| C — Consented, undoable writes | ✅ **done — epic complete (5 slices)** | gate + `olo_set_collision_layer` (#427), generic `olo_entity_set_field` + `olo_entity_list_fields` (#474), `olo_reload_script` (#489), `olo_renderer_settings_set` (#503), per-action consent **dialog** (#512) |
| D — Hardening | ✅ done | dispatch tests (#315) + inputSchema enforcement & payload cap (#423); per-pass capture hooks |

---

## A. "Explain" tools for the top user pains
Promote the worked examples from manual multi-tool flows into first-class tools that
gather all signals in one call:
- [x] `olo_render_why_not_visible({entity})` — checks mesh/material presence, scale≠0,
      frustum/HZB/LOD culling, behind-camera, render layer, and material shader
      compile status, and reports the likely reason.
- [x] `olo_physics_*` family — collision-layer matrix, list colliders + layers,
      active contacts, point/ray overlap queries, and
      `olo_physics_why_no_collision({a,b})` (the "player falls through the floor" case).
      Shipped: `olo_physics_layer_matrix`, `olo_physics_list_colliders`,
      `olo_physics_contacts`, `olo_physics_overlap`, `olo_physics_raycast`,
      `olo_physics_why_no_collision`.

[...]

PR #503, opened 2026-07-01
feat(mcp): olo_renderer_settings_set — consented renderer/post-process settings write (#306)

drsnuggles8:
## What

Adds **`olo_renderer_settings_set`** — a consented MCP write tool (issue **#306** item C) that sets multi-valued, session-global renderer settings over MCP, so a rendering feature can be **verified live at each value** from an agent session. It is the enum-valued sibling of the boolean `olo_render_toggle_pass`.

Settings covered (the ones the issue names / that made #480 hard to verify):
- **`upscale`** — FSR1 spatial-upscale mode (`off`/`quality`/`balanced`/`performance`/`ultraperformance`) — the #480 motivating case.
- **`tonemap`** — tone-map operator (`none`/`reinhard`/`aces`/`uncharted2`).
- **`renderpath`** — rendering path (`forward`/`forwardplus`/`deferred`; switching rebuilds the render-graph topology).

Call with **no arguments** to list every setting + current value + allowed values.

## Design

- **Gated** behind the existing session write gate (`ToolDef::ProjectWrite` / "Allow writes", default OFF, not persisted). `readOnlyHint:false`, `idempotentHint:true`. Schema-enforced (#423) + payload cap.
- **Restore-prior-value, NOT `CommandHistory`.** These are global renderer settings, not scene/ECS data, so an undo-stack entry would be wrong (unlike `olo_set_collision_layer` / `olo_entity_set_field`). The response reports `previousValue`/`restoreWith`; the agent reverts by setting it back. The settings are never written to disk (a scene reload restores them) — the same ephemeral boundary the render-override / sun tools respect.
- **Testable seam split:** the httplib/editor-free apply core lives in `OloEditor/src/MCP/McpRendererSettings.h` (mutates the POD `PostProcessSettings`/`RendererSettings` by reference), mirroring `McpSetCollisionLayer.h` / `McpRenderOverrides.h`. The enum value tables reference the engine enums directly, so the token↔int mapping can't drift. A render-path switch signals `ApplyResult::PathChanged` so the handler calls `Renderer3D::ApplyRendererSettings()` (render-graph rebuild) — the only renderer-bound bit, kept in `McpTools.cpp`.

## Tests

- **17 new seam tests** (`OloEngine/tests/MCP/McpRendererSettingsTest.cpp`): gate off/on, schema enforcement, `Apply`, parse + token round-trip. **300 total MCP tests green**, clean compile.

## Live verification (attached editor over MCP)

- Tool registered live with correct schema/annotations (render toolset 10→11).
- **Gate OFF** refuses `set` + introspection + malformed call (gate runs before schema check).
- **Gate ON** applies: tonemap `reinhard→aces` showed a clear visible change (more saturated greens / higher contrast), and **restore was byte-identical to the baseline frame**.

## Note — a pre-existing #480 bug this surfaces

Setting `upscale` non-off **at runtime in the editor viewport** renders a **black frame** + a flood of `GL_INVALID_VALUE: y values exceeds the boundaries of the corresponding image object`. The editor's own "Spatial Upscale (FSR1)" dropdown (`PostProcessSettingsPanel::DrawUpscaleSection`) applies `Upscale` identically, so this is a **pre-existing #480/#491 EASU render-scale bug**, not introduced here — the new tool just makes it observable live, which is exactly #306-C's purpose. Tracked separately.

## Docs

`docs/guides/mcp-diagnostics-server.md` — tool table, `render` toolset, tool count (41→42), and a new "Multi-valued renderer settings" section.

## Summary by CodeRabbit

* **New Features**
  * Added a new MCP tool for viewing and updating renderer session settings.
  * Supports introspection of current renderer/post-process values and allowed options.
  * Lets users change settings like upscale, tonemap, and render path, with restore-friendly responses.

* **Bug Fixes**
  * Improved validation so invalid or incomplete tool inputs are rejected clearly.
  * Changes that affect render path now trigger the needed renderer refresh.

* **Documentation**
  * Updated MCP diagnostics documentation to cover the new renderer settings tool and its usage.

### DIFF
--- a/OloEditor/src/MCP/McpRendererSettings.h
+++ b/OloEditor/src/MCP/McpRendererSettings.h
@@ -43,6 +43,7 @@
 
 #include <nlohmann/json.hpp>
 
+#include <algorithm>
 #include <array>
 #include <cctype>
 #include <optional>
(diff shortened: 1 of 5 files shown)


drsnuggles8:
The FSR1 runtime black-frame issue surfaced while live-verifying the `upscale` write is now tracked as **#504** (pre-existing #480/#491 EASU bug — this PR just makes it observable live). The `tonemap` and `renderpath` writes verify clean.

PR #552, opened 2026-07-04
fix(renderer): GTAO-lit scenes render too dark

### LINKED ISSUE #533: Renderer: GTAO-lit scenes render too dark — AOApplyPass reuses the SSAO half-res shader for GTAO's full-res AOBuffer
## Summary

Discovered while investigating #504 (fixed in #532). Separate, pre-existing bug — not caused by and not related to the FSR1 fix in #532.

`AOApplyPass` reuses `PostProcess_SSAOApply.glsl` for **both** SSAO and GTAO output, but the shader is written and commented specifically for SSAO's semantics:

```glsl
// Blurred SSAO result (R channel = AO value, 0 = full occlusion, 1 = no occlusion)
// This texture is at HALF resolution.
layout(binding = 20) uniform sampler2D u_SSAOTexture;
```

GTAO's `AOBuffer` (declared in `RenderPipeline.cpp::PopulateBlackboard`, GTAO branch) is **full** scene-band resolution — SSAO's is half. `bilateralUpsampleAO()` in the shader is written around the half-res assumption. The practical effect: with `ActiveAOTechnique == AOTechnique::GTAO` and `GTAOEnabled == true`, the composited scene renders far too dark — in a minimal synthetic test scene (a few directional lights, cubes, a plane) it comes out essentially fully black.

This reproduces regardless of `PostProcessSettings::Upscale` (confirmed at `Upscale == Off` too), so it's unrelated to the FSR1 render-scale mechanics in #504/#532.

## Repro

1. In a scene with normal lighting, set `PostProcessSettings::ActiveAOTechnique = AOTechnique::GTAO` and `GTAOEnabled = true` (no FSR1 upscale needed).
2. Render a frame and read back `SceneColor` / the composited output.
3. Observe the frame is much darker than with AO off or with SSAO active at equivalent settings — in simple scenes, fully black.

## Suspected fix direction

Either:
- Give GTAO its own apply shader (or a resolution-aware branch in the existing one) instead of reusing `PostProcess_SSAOApply.glsl`'s half-res-specific `bilateralUpsampleAO()`, or
- Confirm whether the GTAO compute pass itself (`GTAORenderPass::DispatchGTAO`/`DispatchDenoise`) is producing near-zero AO values independent of the apply-side resolution mismatch — worth ruling out before assuming the apply shader is the sole cause.

## Found by

Investigation for #504, surfaced while adding `EASUVisualEvidenceTest.GTAOSurvivesRuntimeUpscaleSwitch` in #532 — that test's assertion was deliberately scoped to a GL-error scan + relative brightness check rather than an absolute one, specifically to avoid coupling to this bug.

## Score
```olo-score
capability: 3
craft: 5
stability: 5
decay: 1
effort: 3
confidence: 0.8
learning: 3
fun: 5
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary

Fixes #533. Two independent bugs combined to produce the reported near-black GTAO frames:

- **`GTAO.comp`** built its per-slice tangent-elevation basis as `cross(directionVS, viewNormal)`. Crossing anything with `viewNormal` is trivially perpendicular to `viewNormal` (a pure cross-product identity), so the tangent-elevation angle `n` collapsed to exactly 0 for every slice on every pixel, regardless of the surface's real tilt relative to the camera — any surface not exactly face-on then measured its horizon against the wrong, untilted baseline and self-occluded. Fixed by building the basis from the camera's view axis instead.
- **`McpTools.cpp`'s `olo_render_toggle_pass`** mutated `PostProcessSettings::ActiveAOTechnique` directly without calling `Renderer3D::ApplyRendererSettings()` afterward, so the render graph stayed wired for the previously-active AO pass. `GTAOPass` never executed, its `AOBuffer` was never written (stayed zero), and `AOApplyRenderPass` still multiplied the entire composited frame (sky included) by that all-zero buffer at intensity 1.0 — an all-black frame. This is very likely how the original bug was reproduced, via this repo's own MCP diagnostics tooling.

Root-cause writeup with full derivation posted on #533.

## Test plan

- [x] `GTAOMathTest.cpp` (new, CPU-only) — pins the tangent-elevation math and guards against reintroducing the degenerate pre-fix formula.
- [x] `GTAOVisualEvidenceTest.cpp` (new, GPU) — mirrors `SSAOVisualEvidenceTest`; PNG evidence in `OloEditor/assets/tests/visual/GTAO_*.png` shows sky/floor render normally post-fix (previously all-black).
- [x] Full SSAO/GTAO/AOApply/RenderPipeline suite (40 tests, 10 suites) passes with no regressions.
- [x] One known follow-up gap documented in the new test, not fixed here: GTAO's horizon search doesn't produce a visually distinct localized contact-occlusion crease the way SSAO does — a separate, lesser quality issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GTAO rendering stability so ambient occlusion no longer collapses on flat or grazing surfaces, keeping shading consistent across viewing angles.
  * Switching between SSAO and GTAO now applies immediately, avoiding stale ambient occlusion behavior after render-pass toggles.
* **Tests**
  * Added automated headless GTAO math and visual regression checks to detect the issue and related regressions.
* **Documentation**
  * Updated shader guidance with a common GLSL vector-basis pitfall and the corrected, consistent approach.


drsnuggles8:
Addressed in c1a6b19e: the axisVS/cosN basis now consistently uses the per-pixel view vector (`normalize(-pixCenterPos)`) instead of the surface normal or a fixed screen axis, for `orthoDirectionVS`, `axisVS`, and `cosN`'s dot target.

Turned up a second, subtler bug while verifying: even after the first fix, `cosN` was dotting `projectedNormal` against `viewNormal`, which is a linear-algebra tautology (`dot(P, V) == |P|^2` for any orthogonal projection `P` of `V`) — non-degenerate, but not the intended quantity. Verified numerically before/after. `GTAOVisualEvidenceTest`'s floor/crease AO-buffer assertions now pass with real headroom instead of just the loosest composite-brightness check. Full regression suite (43 tests) green.

Issue #505, opened 2026-07-01
Renderer: full-scene render binds a stale texture (benign GL_INVALID_OPERATION) after cross-test render-graph/asset churn


drsnuggles8:
## Summary

The #485 GL-error-state listener surfaced this: after a run of GPU render tests churns `Renderer3D`'s process-wide render-graph / asset state, a subsequent **full-scene render** emits `GL_INVALID_OPERATION` — *"`<texture>` is not a valid texture name"* — via the GL debug callback, every frame. The render output is **correct** (all victim tests pass their pixel assertions); only the GL error queue is dirtied. So it is *benign today* but indicates a **stale texture handle** being bound somewhere in the full render path.

## Evidence (from #485)

- Victim tests (`SceneRenders3D`, `SphereAreaLightShadowEvidenceTest`, `MetallicSphereUnderNebula`, `EASUVisualEvidenceTest.*`, `AutoExposureBrightScene`, `OccludedInstanceFieldScene`) **all pass in isolation**, and fail only in the full single-process suite → the bad state is inherited from *earlier* tests, not self-generated.
- It is **diffuse / accumulative with multiple sources**: fixing one confirmed source (`GPUOcclusionCullGPUTest`, which left an HZB bound to a texture unit then deleted it) removed it for some victims, but others still tripped and the failing set shifted → more than one upstream test leaves shared renderer state that makes a later render bind an invalid texture name.
- The debug message is `id 1282 (GL_INVALID_OPERATION) "<texture> is not a valid texture name"`, i.e. `glBindTextureUnit(unit, name)` with a non-zero name that is no longer a valid texture.

## Current containment (not a fix)

`RendererAttachedTest::RunFrames` / `RunEditorFrames` now drain the GL error queue after each guarded render tick (mirroring the #485 production fix 69aa9357 that drains stray errors in the readback helpers), so a render self-contains its error queue and cannot poison a later GPU test. The #485 listener stays strict everywhere else. This masks the symptom in tests — the underlying stale-handle bind is unfixed.

## What to do

Root-cause which shared `Renderer3D` resource (render-graph target / material / IBL / HZB / bindless handle) is left referencing a deleted texture across a render-graph resize / path switch / asset churn, and fix the lifetime so the full pipeline binds no stale texture name. Then the per-tick drain in `RendererAttachedTest` can be removed and the listener will guard render-attached tests too.

[...]

PR #557, opened 2026-07-04
fix(renderer): invalidate IBL disk cache when source env-map edited in place (#542)

### LINKED ISSUE #542: Renderer: IBL disk cache never invalidated when a source environment map is edited in place -> stale lighting persists across restarts
## Summary

`IBLCache` keys its disk cache on **source path + IBL config only** — never the source file's modification time or a content hash — and its `Invalidate(...)` method is **dead code** (no callers). So editing an environment-map `.hdr`/`.png` *in place* (same path, same config) yields a **cache hit that returns the IBL derived from the OLD image**. The diffuse ambient + specular reflections stay locked to the stale environment **permanently, across editor restarts**, until `assets/cache/ibl/` is hand-deleted. This is the same "valid-until-invalidated but never invalidated by the event that changes the underlying state" species as #505/#530.

## Evidence

- [IBLCache.cpp:169-197](../blob/master/OloEngine/src/OloEngine/Renderer/IBLCache.cpp#L169) (`ComputeHash`) folds only the source-path string and config fields into the FNV key — no `last_write_time`, no content hash.
- `IBLCache::TryLoad` (IBLCache.cpp:214+) validates only file **existence** of the cache entries, not freshness vs the source.
- `IBLCache::Invalidate(...)` is defined (IBLCache.cpp:355, declared IBLCache.h:119) but has **no caller** anywhere in the tree — nothing ever clears a stale entry.
- Contrast the sibling `MeshCache`, which embeds and compares the source timestamp (`MeshCache.cpp:130-147`) and is correct.
- `ProceduralSky` sidesteps this by setting `UseDiskCache=false` (ProceduralSky.cpp:348); file-backed maps do not.
- The developers' own comment states the faulty assumption verbatim (`EnvironmentMap.h:48-56`: "a given .hdr always produces the same IBL") — true only if the file never changes in place, which hot-reload violates.

## Reproduction

1. Scene has an `EnvironmentMapComponent` → `assets/env/sky.hdr` (equirect). First run generates + caches the IBL under `hash(path + config)`.
2. Repaint `sky.hdr` in place (same path, same config).
3. Reopen the scene (or restart the editor): the runtime `Ref` isn't serialized, so the IBL is lazily rebuilt → `IBLCache::TryLoad` → same hash → **cache HIT → old IBL**. The skybox background (equirect→cubemap, not IBL-cached) shows the new sky; diffuse/specular lighting stays on the old sky.

## Fix direction

Add the source `last_write_time` (as `MeshCache` does) to `ComputeHash`, or add a timestamp/staleness comparison to `TryLoad`, and/or wire `IBLCache::Invalidate` into the `Texture2D`/env-map branch of the asset-reload handler. Per CLAUDE.md, verify visually (edit an HDR, reload, confirm ambient/reflections update).

## Score
```olo-score
capability: 3
craft: 3
stability: 5
decay: 1
effort: 2
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
## Summary

`IBLCache` keyed its disk cache on **source-path + config only** — no source `last_write_time`/content hash — and `IBLCache::Invalidate` was **dead code**. Editing an environment-map `.hdr`/`.png` **in place** (same path, same config) yielded a cache hit returning the IBL baked from the **old** image, so diffuse ambient + specular reflections stayed locked to the stale environment **across editor restarts** until `assets/cache/ibl/` was hand-deleted. This is the same "valid-until-invalidated but never invalidated by the event that changes the underlying state" species as #505/#530.

## Fix

Mirror the correct `MeshCache` pattern — store the source's timestamp in the cache header and reject a stale entry on load:

- **`IBLCacheHeader`**: add `SourceTimestamp` (u64); bump `kIBLCacheVersion` 5→6 so older files are cleanly rejected; `static_assert` 36→44 bytes.
- **`Save`** stamps every cache file with the source file's `last_write_time`.
- **`TryLoad`** compares the source's current mtime against the stored value; on mismatch it wires the previously-**dead** `Invalidate` (removing the stale files) and misses, so the caller re-bakes. Synthetic keys (in-memory / procedural cubemaps with no backing file) stay `0 == 0` on compare and **still hit**, preserving `UseDiskCache=false` behavior (ProceduralSky unaffected).
- **`HasCache`** also checks freshness for a consistent contract.
- Fix the stale "a given `.hdr` always produces the same IBL" comment in `EnvironmentMap.h` that the issue flagged.

## Verification

- **New GPU-gated CPU/contract test** `IblCacheInvalidatesWhenSourceFileEditedInPlace`: real temp source file + real in-place mtime bump → asserts stale-detect, cache invalidation, and re-bake-hit; the existing `IblCacheCubemapRoundTripPreservesAllMips` still passes bit-exact through the new 44-byte header.
- **Live editor, real in-place edit + restart** (MaterialSpheres → `newport_loft.hdr`): the log shows the exact production path — first launch `Cache miss … Generating IBL … Saved`, then after overwriting the HDR in place and restarting: `IBLCache: Stale cache for assets/textures/hdr/newport_loft.hdr (source modified) — invalidating` → `Invalidated cache … (3 files)` → `Generating IBL textures`. The rebaked on-disk cache is confirmed version 6 with a populated `SourceTimestamp`.

**Note on the viewport pixel-delta:** this change adds **no render/shader math** — it only changes *which* cached textures load — so the appropriate evidence is the live-editor invalidation log + the content-verified unit test rather than a viewport color-delta screenshot (a synthetic uniform-bright test HDR bakes to a degenerate all-white IBL, which can't produce a clean before/after color delta; that's a bake property, not a cache-fix issue).

Fixes #542

## Summary by CodeRabbit

* **Bug Fixes**
  * Environment map caching now refreshes correctly when a source file is edited in place, helping prevent outdated lighting results.
  * Cached lighting data is now more reliably invalidated and rebuilt when source content changes.
  * Procedurally generated environment maps are handled more safely to avoid reusing stale cached results after rebakes.

### DIFF
--- a/OloEngine/src/OloEngine/Renderer/EnvironmentMap.h
+++ b/OloEngine/src/OloEngine/Renderer/EnvironmentMap.h
@@ -46,12 +46,14 @@ namespace OloEngine
         // single-context bake rationale.
 
         // Disk-cache control. The IBL disk cache is keyed on the source
-        // cubemap's path + this config. That's correct for file-backed
-        // environment maps (a given .hdr always produces the same IBL), but
-        // WRONG for procedurally generated cubemaps whose pixels change at
-        // runtime while their debug path stays constant ("Generated Cubemap").
-        // Such sources must set this false so a stale cache entry can't be
-        // served for freshly-baked content (e.g. ProceduralSky changing sun
+        // cubemap's path + this config, and the cache header additionally
+        // stores the source file's last-write-time, so an in-place edit of a
+        // file-backed .hdr/.png is detected and re-baked (IBLCache::TryLoad
+        // rejects a stale entry). That timestamp check only works for real
+        // files, though: procedurally generated cubemaps whose pixels change at
+        // runtime keep a constant debug path ("Generated Cubemap") with no
+        // mtime, so they must still set this false so a stale cache entry can't
+        // be served for freshly-baked content (e.g. ProceduralSky changing sun
         // direction / turbidity / exposure).
         bool UseDiskCache = true;
     };
--- a/OloEngine/src/OloEngine/Renderer/IBLCache.cpp
+++ b/OloEngine/src/OloEngine/Renderer/IBLCache.cpp
@@ -32,7 +32,15 @@ namespace OloEngine
     // shaders. The advanced shaders now exist (IBLPrefilterImportance,
     // IrradianceConvolutionAdvanced, BRDFIntegrationAdvanced) with
     // mip-biased importance sampling, so the baked output differs.
-    constexpr u32 kIBLCacheVersion = 5;
+    // Version 6 adds SourceTimestamp to the header (staleness tracking):
+    // the cache is keyed on source-path + config only, so editing an
+    // environment-map file in place (same path, same config) previously
+    // served the IBL baked from the OLD image forever. The header now
+    // stores the source file's last-write-time; TryLoad rejects (and
+    // invalidates) an entry whose stored timestamp no longer matches the
+    // source, so an in-place edit re-bakes on next load. The layout change
+    // also makes older v1–v5 files unreadable, which is the intended reset.
+    constexpr u32 kIBLCacheVersion = 6;
 
     // Cache file header for version tracking
     // Use pragma pack to ensure consistent binary layout across compilers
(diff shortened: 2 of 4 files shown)


drsnuggles8:
@coderabbit review

Issue #462, opened 2026-06-28
Networking: extend snapshot interpolation/reconciliation beyond TransformComponent


drsnuggles8:
## Problem
The netcode stack is strong, but `SnapshotInterpolator` interpolates **only `TransformComponent`** (`Networking/Replication/SnapshotInterpolator.cpp`), and client reconciliation blends only transforms. Animation state, rigidbody velocity, and other replicated component state aren't interpolated/reconciled — causing jitter on rotating / animated / physics-driven networked entities.

## Proposed approach
- Generalize interpolation/reconciliation to a registered set of interpolatable components (transform, velocity, animation state, key gameplay state) via the existing `ComponentReplicator` path.
- Per-component interpolation policy (lerp / slerp / step) declared alongside replication.

## Acceptance criteria
- A networked animated / rotating / physics entity moves smoothly on remote clients (no per-snapshot popping).

---
**Refinement of the existing (strong) netcode stack.**

## Score
```olo-score
capability: 3
craft: 3
stability: 2
decay: 1
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
@
Implemented in #483.

**Acceptance criterion met** — *a networked animated / rotating / physics entity moves smoothly on remote clients (no per-snapshot popping)*:

- Interpolation + reconciliation are now generalized to a data-driven `ComponentInterpolationRegistry` with per-component policies (lerp / slerp / step), built on the existing `ComponentReplicator` path.
- Registered set: **Transform** (translation/scale lerp + rotation slerp), **Rigidbody3D** (velocity lerp, type/mass step), **AnimationState** (step). Adding more components later is a single `Register()` call.
- The Functional test `InterpolatedComponentsSmoothOnRemoteClientTest` drives a real physics body (moving + spinning) through the actual `EntitySnapshot` → `SnapshotInterpolator` path on a separate client scene and asserts the pose blends smoothly between server snapshots (strictly interior, monotonic, no popping).

The issue will auto-close when #483 merges.
@

PR #561, opened 2026-07-05
fix(networking): correct prediction/reconciliation and state-lifecycle bugs

### LINKED ISSUE #546: Networking: client-prediction/reconciliation correctness + state-lifecycle bugs to fix before wiring the snapshot pipeline into a live loop
## Summary

Auditing the snapshot interpolation / client-prediction / reconciliation code (the #462/#483 policy-registry work) surfaced several genuine correctness + state-lifecycle defects. **Scope caveat:** the snapshot pipeline is **built and unit-tested but not yet wired into any running loop** — `NetworkManager::TickSnapshots()` has no caller, `SnapshotInterpolator::Interpolate()` is called only from tests, and the server never sends `InputAck` so `ClientPrediction::Reconcile` never fires in production. None crashes a live app **today**; all will bite the moment the pipeline is wired up. Filing so the base is hardened first (they are cheap fixes).

## Findings

**N1 — reconciliation smoothing eases toward the SERVER state, not the prediction (`glm::mix` args inverted). CONFIRMED.**
In `Reconcile`, the server snapshot is applied to the scene *before* resim ([ClientPrediction.cpp:39](../blob/master/OloEngine/src/OloEngine/Networking/Prediction/ClientPrediction.cpp#L39)), and `entry.Capture` records the entity's **current = server** pose as `preReconcile` (:93). Then `TransformSmooth`/`Rigidbody3DSmooth` do `mix(pre, t, rate)` (`ComponentInterpolationRegistry.cpp:73-75, 116-117`) = `server*(1-rate) + resim*rate`. But the `Smooth` contract (`ComponentInterpolationRegistry.h:55-56`, "ease the *current* (resimulated) component back toward a pre-reconcile server snapshot by rate") and `SetSmoothingRate`'s doc ("Lower = smoother but slower correction") both imply `mix(t, pre, rate)` — the **arguments are swapped**, inverting the rate. With the default `rate=0.1` the entity is yanked ~90% back to the latency-delayed server pose every correction, reintroducing most of the input latency prediction exists to hide. No test asserts the post-`Smooth` position.

**N2 — reconnect/disconnect leaves all client interpolation + prediction state stale. CONFIRMED (state-pollution species).**
`NetworkManager::Disconnect()` ([NetworkManager.cpp:274-285](../blob/master/OloEngine/src/OloEngine/Networking/Core/NetworkManager.cpp#L274)) resets only `s_Client`. It does **not** clear the persistent statics `s_ClientInterpolator` (its `SnapshotBuffer`, `m_LatestReceivedTick`), `s_ClientReceivedTick`, or `s_ClientPrediction` (`m_LastConfirmedTick`, `m_InputBuffer`); `Connect()` doesn't either. Reconnecting (same server after restart, or a UUID-reusing persistence world) mixes stale session-A snapshots with new session-B ones in the 64-slot ring → reused-UUID entities rubber-band; the carried `m_LastConfirmedTick` suppresses session-B reconciles until ticks climb past it.

**N3 — `ServerInputHandler::m_LastProcessedTicks` grows unbounded (LOW).**
One `clientID → tick` entry per client, never pruned; `NetworkServer::OnConnectionStatusChanged` (NetworkServer.cpp:361-373) erases the connection but nothing erases the handler entry (ServerInputHandler.h:37). Client IDs are monotonic so it is not a cross-client bleed, but it is a slow unbounded leak on the long-lived headless `OloServer`.

**N4 — `ComponentInterpolationRegistry` lazy-init not thread-safe (SPECULATIVE).**
`s_Entries`/`s_Initialized` are plain statics mutated under no lock in `EnsureInitialized` (ComponentInterpolationRegistry.cpp:156-200), and (unlike `ComponentReplicator::RegisterDefaults`, eagerly called under `s_Mutex`) it's never eagerly initialized. On a listen-server the registry is first-touched from both the network and game threads → possible double-register / read-during-`push_back`-realloc.

## Fix direction

N1: swap the `mix` argument order (and `slerp`) so it eases current→server by `rate`; add a post-`Smooth` position assertion. N2: clear `s_ClientInterpolator`/`s_ClientReceivedTick`/`s_ClientPrediction` on `Connect` (or `Disconnect`). N3: add a `RemoveClient` hook wired from `OnConnectionStatusChanged`. N4: eagerly call the interpolation `RegisterDefaults()` under a lock in `NetworkManager::Init` (mirror `ComponentReplicator`).

## Score
```olo-score
capability: 1
craft: 1
stability: 3
decay: 3
effort: 2
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
## Summary
Fixes four latent correctness/state-lifecycle bugs in the client-prediction / snapshot-replication pipeline (#462/#483), found while auditing before the pipeline gets wired into a live game loop.

- **N1** — `ComponentInterpolationRegistry::TransformSmooth`/`Rigidbody3DSmooth` mixed toward the *pre-reconcile server snapshot* instead of the *resimulated* state, contradicting `ClientPrediction::SetSmoothingRate`'s documented "lower = smoother but slower correction" contract (a low rate would instead snap hard to the server value every frame). Swapped the `mix()`/`slerp()` argument order; added a regression test (`ReconcileSmoothsResimulatedStateTowardPreReconcileSnapshot`) that pins the corrected direction.
- **N2** — `NetworkManager::Disconnect()` only reset `s_Client`, leaving `s_ClientInterpolator`, `s_ClientPrediction`, and `s_ClientReceivedTick` stale across a reconnect, so a new session could start by blending against the previous session's buffered snapshots/inputs. Added `SnapshotInterpolator::Reset()` and `ClientPrediction::ResetSession()`, called from `Disconnect()`.
- **N3** — `ServerInputHandler::m_LastProcessedTicks` grew without bound over the lifetime of a long-running headless `OloServer`, since nothing removed a client's entry on disconnect. Added `ServerInputHandler::RemoveClient()`, wired through a new `NetworkServer::ClientDisconnectedCallback` invoked from `OnConnectionStatusChanged`'s peer-close/problem-detected branch, registered in `NetworkManager::StartServer`.
- **N4** — `ComponentInterpolationRegistry`'s lazy `EnsureInitialized()` mutates plain statics with no lock, racy under concurrent first access. `NetworkManager::Init()` now eagerly calls `RegisterDefaults()` under `s_Mutex`, mirroring the existing `ComponentReplicator::RegisterDefaults()` call.

**Scope note:** the snapshot/prediction pipeline is built but not yet wired into a live loop — `TickSnapshots` has no caller, `SnapshotInterpolator::Interpolate` is exercised only by tests, and the server never sends `InputAck`. These bugs are therefore latent; this PR hardens the base ahead of that wiring, which remains a follow-up.

Fixes #546

## Test plan
- [x] `OloEngine-Tests` built clean (Debug, MSVC)
- [x] Full networking test sweep (113 tests across 27 suites: `*Network*`, `*Prediction*`, `*Snapshot*`, `*Rollback*`, `*Interpolation*`, `*ServerInputHandler*`/`AuthorityRejectionTest`, `*LagCompensat*`, `*Interest*`) — all pass
- [x] New regression test for the N1 mix-direction fix passes

## Summary by CodeRabbit

* **New Features**
  * Added a per-client disconnect callback to support session teardown.
  * Added explicit session reset for client prediction and snapshot interpolation to prevent cross-session carryover.
* **Bug Fixes**
  * Fixed reconciliation smoothing direction so translation/rotation/scale (and rigidbody velocity lerps) blend toward the pre-reconcile snapshot during resimulation.
  * Improved server disconnect cleanup by removing per-client “last processed tick” tracking to avoid stale data.
* **Tests**
  * Added a reconciliation smoothing regression test covering resimulated-state blending.


drsnuggles8:
@coderabbit review

PR #581, opened 2026-07-07
perf(animation): reuse persistent scratch buffers in AnimationGraph/AnimationStateMachine

### LINKED ISSUE #445: Perf/DOD: stop heap-allocating per-frame skinning scratch — write into the renderer frame allocator (offset+count)
## Problem
The renderer already has strong per-frame linear allocators + per-worker scratch buffers (`Renderer/Commands/FrameResourceManager.h`, `Commands/FrameDataBuffer.h`) and a POD command queue. But gameplay/animation hot paths bypass it:
- Animation/skinning matrices are passed as `std::vector<glm::mat4>` **copies** (e.g. `DrawAnimatedMesh(const std::vector<glm::mat4>&)` in `Renderer/Renderer3D.h`) rather than an offset+count into the existing frame buffer.
- ECS hot-path systems use default-allocator containers, not a per-frame scratch arena (the `FMemStack` / linear allocators in `OloEngine/src/OloEngine/Memory/` are not wired into these loops).

Result: per-frame heap allocation + copies for skinning data that could be written once into the contiguous frame buffer the renderer already maintains.

## Why it matters (performance)
Heap churn and copies every frame for every skinned character; defeats the contiguous-buffer / frame-arena pattern the renderer is built around.

## Proposed approach
- Change skinning/animation submission to write bone matrices directly into `FrameDataBuffer` (per-worker scratch) and pass **offset+count** to the draw call instead of a `std::vector` copy.
- Where gameplay systems need per-frame scratch, expose/reuse a frame arena (`FMemStack` / `CommandAllocator`-style) rather than default-allocator `std::vector`.

## Acceptance criteria
- `DrawAnimatedMesh` (and similar) take offset+count into the frame buffer; no `std::vector<glm::mat4>` copy on the hot path.
- Reduced per-frame allocations, verified via `RendererMemoryTracker` / allocation counts.

---
**Data-oriented scene layer.** From the DOD audit — closes the gap between the (strong) renderer and (conventional) gameplay memory model.

## Score
```olo-score
capability: 2
craft: 1
stability: 2
decay: 1
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary

Closes #445.

- `Renderer3D::DrawAnimatedMesh`/`DrawAnimatedMeshParallel` (the issue's headline example) already write bone matrices into `FrameDataBuffer` via offset+count (`AllocateBoneMatrices`/`WriteBoneMatrices`) — in place since commit `789c48f2`, predating this issue. No changes needed there.
- The remaining per-frame heap churn was one layer up: `AnimationGraph::Update()` built a fresh `std::vector<BoneTransform>` **inside its per-layer loop** every tick, and `AnimationStateMachine::Update()`'s cross-fade path did the same for every ticked frame of a transition. Both are now persistent per-instance scratch members, `resize()`d in place instead of reallocated.
- Verified safe: every downstream writer (`BlendTree::FillBindPose`/`Evaluate*`/`BlendBoneTransforms`, `AnimationStateMachine::Update`) fully overwrites `[0, boneCount)`; each entity owns its own `AnimationGraph` instance (`Scene.cpp` `Clone()`s one per entity); the `AnimationGraph`/`AnimationGraphSystem` gameplay step is not `.Parallelizable()`.
- Added a stale-scratch guard (`targetTransforms.clear()`) for the one case where a reused buffer could otherwise carry over a previous transition's target pose.
- Documented the reusable pattern in `docs/agent-rules/per-frame-scratch-reuse.md`.

## Test plan

- [x] Added 2 regression tests in `AnimationStateMachineTest.cpp` targeting the reuse-specific hazard (stale data bleeding across transitions/layers/ticks): `SequentialCrossFadesDoNotLeakScratchBetweenTransitions`, `MultiLayerGraphUpdateReusesScratchAcrossLayersAndTicksWithoutBleed`.
- [x] Full `OloEngine-Tests` suite: 4184 passed, 5 skipped (env-gated), 0 failed.
- [x] Visually verified against the animated `fox.olo` scene in the editor — clean frame, no shader/GL errors, `DrawAnimatedMesh` still logs correct bone-matrix offsets.

## Summary by CodeRabbit

* **New Features**
  * Added guidance for reusing per-frame scratch buffers to reduce repeated allocations in animation updates.
  * Improved animation evaluation to keep temporary pose data around between frames for faster steady-state performance.
* **Bug Fixes**
  * Fixed stale pose data from leaking between animation transitions and layered updates.
  * Added regression coverage to verify cross-fades and multi-layer graph updates remain correct across ticks.


drsnuggles8:
@coderabbit review

Issue #525, opened 2026-07-03
Scene: YAML load takes 24-59s at 50k-100k entities — bulk/binary scene load or runtime spawner


drsnuggles8:
**Is your feature request related to a problem? Please describe.**
Scene YAML load time grows to **tens of seconds to a minute** at stress-scene entity counts (Dist build, NVMe, editor launch→MCP-ready wall time minus ~2 s baseline):

| scene | entities | file size | startup+load |
|---|---|---|---|
| baseline (PinkCube) | 2 | — | 1.9 s |
| physics_pile_10000 | 10 007 | 6.0 MB | 12–42 s (run variance) |
| ecs_static_100000 (transform-only) | 100 002 | 18.2 MB | 24 s |
| physics_pile_50000 | 50 007 | 30.1 MB | 53–59 s |

Per-entity YAML deserialize (yaml-cpp node traversal + per-component `AddComponent`) is the floor; a 100k-entity scene of *nothing but transforms* takes ~24 s. For editor iteration on large levels — and for the perf battery itself — this is the dominant cost, and it rules out "just author a big scene" for runtime content at scale. Related: `anim_crowd_200` does a full `Ref<AnimatedModel>::Create` **per entity** on deserialize (no model sharing/caching across entities with the same `SourceFilePath`).

**Describe the solution you'd like**
Any of (in rough preference order):
1. A **bulk/binary scene format** (or cached binary sidecar) for entity-heavy scenes — parse once, memcpy-ish restore.
2. A **runtime spawner path** (spawn N entities from a template at scene start) so stress/procedural content doesn't round-trip YAML at all.
3. Cheap wins inside the current path: share `AnimatedModel` instances by source path, reserve registry storage up front, batch component construction.

**Describe alternatives you've considered**
Keeping stress scenes generator-only (current approach — they're git-ignored YAML) works for testing but doesn't help real large levels or the editor's own iteration loop.

**Additional context**
Measured by `scripts/perf/run-perf-battery.ps1` (launch→MCP-discovery wall time; the MCP server starts after project+scene load). Battery + generator on `feature/perf-stress-scenes`; runbook `docs/guides/perf-stress-scenes.md`.

## Score

```olo-score
capability: 5
craft: 3
stability: 2
decay: 2
effort: 5
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
[...]

PR #584, opened 2026-07-07
perf(scene): reserve EnTT storage + dedup AnimatedModel loads on deserialize

drsnuggles8:
## Summary
Cheap-wins slice of #525 ("Scene: YAML load takes 24-59s at 50k-100k entities"), scoped to the two structural fixes identified without needing a binary scene format or runtime spawner rewrite:

- **Registry/component reserve**: `SceneSerializer::DeserializeFromYAML` and `DeserializeAdditive` now reserve the EnTT entity pool plus the `IDComponent`/`TransformComponent`/`TagComponent` storages up front, using the known entity count from the YAML `Entities` sequence, instead of letting every pool reallocate repeatedly as entities stream in.
- **AnimatedModel dedup cache**: `SceneSerializer` now caches `Ref<AnimatedModel>` per resolved absolute source path for the lifetime of one `Deserialize()`/`DeserializeFromYAML()` call. A crowd of entities sharing one rigged-character source file (the common case for NPCs/enemies) now pays Assimp import + GPU-buffer allocation once instead of once per entity. Per-entity playback state (`AnimationStateComponent`) stays independent — verified by a new regression test.

## Results
Measured with the perf-stress battery (`docs/guides/perf-stress-scenes.md`), scene-open time:

| scene | entities | before | after | speedup |
|---|---|---|---|---|
| `anim_crowd_5000` | 5,003 | 56.9 s | 2.2 s | ~26x |
| `anim_crowd_10000` | 10,003 | >120s (perf-battery tool timeout) | 4.3 s | >28x |
| `ecs_static_100000` | 100,002 | >120s (timeout) | 20.0 s | >6x |

`ecs_static_100000` has no `AnimationStateComponent` at all, isolating the reserve-only win; the `anim_crowd_*` scenes get both fixes stacked.

## Test plan
- [x] New `AnimatedModelDedupTest.cpp` — two entities sharing one glTF source path share the same `MeshSource`/`Skeleton` Ref after deserialize, while independently-set `AnimationStateComponent` playback scalars (`m_CurrentTime`, `m_IsPlaying`) survive un-aliased. Log evidence confirms only one Assimp import fires for the shared path.
- [x] Full `OloEngine-Tests` suite (4198 tests) passes, 0 failures, 0 regressions.
- [x] Perf-stress battery before/after comparison (above).

## Summary by CodeRabbit

* **Bug Fixes**
  * Scene loading now avoids reloading the same animated asset multiple times, improving load times and reducing memory use for scenes with repeated characters.
  * Animated entities keep their individual playback state after reloading, so one entity’s animation changes no longer affect another’s.

* **Tests**
  * Added coverage for animated scene reloads to verify shared asset reuse and independent per-entity animation state.

### DIFF
--- a/OloEngine/src/OloEngine/Scene/SceneSerializer.cpp
+++ b/OloEngine/src/OloEngine/Scene/SceneSerializer.cpp
@@ -1601,7 +1601,8 @@ namespace OloEngine
         return Static;
     }
 
-    static void DeserializeEntityComponents(Entity& deserializedEntity, const YAML::Node& entity)
+    static void DeserializeEntityComponents(Entity& deserializedEntity, const YAML::Node& entity,
+                                            std::unordered_map<std::string, Ref<AnimatedModel>>& animatedModelCache)
     {
         if (auto transformComponent = entity["TransformComponent"]; transformComponent)
         {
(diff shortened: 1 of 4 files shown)


drsnuggles8:
@coderabbit review

Issue #519, opened 2026-07-03
MCP perf tooling: CPU-scope profiling, GPU-timing accuracy under load, overdraw debug view


drsnuggles8:
Follow-up to #306 and #316, surfaced by two live MCP perf-diagnosis sessions (2026-07-02,
Sponza CSM/atrium, Dist build) run against the real-GPU-timings tooling from #517. Tracks
the remaining perf-diagnostics gaps that are neither scriptable-repro-setup (#316 Part 5)
nor part of #306's now-complete write-consent epic.

## New tool wanted

- [x] **`olo_perf_cpu_scopes`** — expose `PerformanceProfiler` per-scope CPU timings (every
      system in `Scene.cpp` is wrapped in `OLO_PERF_SCOPE`) over MCP, read-only. Today this
      data only reaches the editor Statistics panel, and `OLO_PERF_SCOPE` is compiled out
      entirely in Dist, so a Dist session has zero CPU-scope visibility (frame totals only).
      **Shipped in #537.**

## Accuracy bugs in the #517 GPU-timing tooling (diagnosed, not fixed)

Root-caused during a heavy-GPU-load session (frame >250 ms, Sponza atrium, Dist, maximized
viewport) — surfaced in the #316 comment thread (2026-07-02/07-03 live perf sessions):

- [x] **`cpuMs` exceeds `frameTimeMs` at saturation** (e.g. cpu 905 ms vs frame 519 ms).
      Root cause: `RendererProfiler` measures `m_FrameTime` at `BeginFrame` (previous
      frame-boundary delta) but `m_CPUTime` at `EndFrame` (current bracket), and the MCP
      snapshot reads `m_CurrentFrame` mid-frame — the two numbers describe *different
      frames* whenever frame times swing.
- [x] **`gpuWaitMs` reads 0.00 even when the GPU is the bottleneck by ~100x.** Only the
      `FrameResourceManager::BeginFrame` fence wait is attributed; under deep GPU backlogs
      the real stall lands in `SwapBuffers`, outside the profiler bracket entirely (and the
      editor's over-budget render throttle further suppresses fence waits).
- [x] **Per-pass GPU numbers get erratic on very long frames** (ShadowPass read 13.9 ms at
      native vs 0.59 ms at 1080p vs 4.3 ms at 4K with a fixed camera — shadow cost cannot
      depend on viewport resolution). `GPUPassTimerPool` has a 4-slot timestamp ring; when
[...]

PR #585, opened 2026-07-07
fix(mcp): settle render-throttle transient before perf-lever writes return

drsnuggles8:
## Summary
- `olo_renderer_settings_set` and `olo_scene_open` could return before the editor's render-budget throttle actually cleared, so a caller reading `olo_perf_snapshot`/`olo_perf_pass_timings` right afterward could see stale pre-change data — `RendererProfiler::BeginFrame`/`EndFrame` only run inside `Renderer3D::BeginScene`, which the throttle skips for a beat after any main-thread stall (a scene load being the common trigger).
- Both handlers now wait out that transient via the existing `AwaitRenderedFrames`/`IsCaptureUnready` settle discipline (the same mechanism the camera-pose screenshot tools already use), so a couple of real frames have rendered with the new state by the time either call returns.
- Closes the last open item in #519 ("First perf-lever write right after scene load doesn't take effect on the GPU").

## Test plan
- [x] Added `OloEngine/tests/Rendering/PropertyTests/McpPerfSettleTest.cpp` (headless MCP host, GPU-gated) asserting `olo_renderer_settings_set` doesn't return until the frame counter advances past the settle window.
- [x] Live-verified against a running editor over MCP: `olo_scene_open` → immediate `olo_renderer_settings_set(depthprepass=on)` → immediate `olo_perf_pass_timings` shows the `DepthPrepass` GPU sub-pass with fresh (`gpuResultsAgeFrames: 1`) data; flipping back to `off` and re-checking immediately correctly shows it gone.
- [ ] `OloEngine-Tests.exe --gtest_filter=McpPerfSettleTest.*` — build was still compiling in this environment when I finished; please run before merging if CI doesn't cover it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scene loading and renderer setting updates now wait briefly for rendering to settle before completing, so changes are visible immediately in subsequent reads.
  * This reduces stale results after opening a scene or changing performance/rendering options.

* **Tests**
  * Added coverage to verify renderer setting changes do not return until the updated state has been rendered.

### DIFF
--- a/OloEditor/src/MCP/McpTools.cpp
+++ b/OloEditor/src/MCP/McpTools.cpp
@@ -3689,6 +3689,32 @@ namespace OloEngine::MCP
 
             if (result.is_object() && result.contains("__error"))
                 return ToolResult::Error(result["__error"].get<std::string>());
+
+            // Settle before returning (#519 "first perf-lever write right after
+            // scene load doesn't take effect on the GPU"). A lever flip right
+            // after a heavy scene load lands while the editor's render-budget
+            // throttle is still skipping frames — the just-blocked main thread
+            // reports a huge timestep for the next OnUpdate, which trips
+            // skipRender for a beat. While throttled, Renderer3D::BeginScene
+            // never runs, so RendererProfiler::BeginFrame/EndFrame don't either:
+            // GetLastCompletedFrameData() (what olo_perf_snapshot reads) stays
+            // frozen on whatever rendered before the write, making the change
+            // invisible until an unrelated later frame finally renders. Waiting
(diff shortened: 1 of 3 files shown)


drsnuggles8:
Ran the full MCP test suite locally: \`OloEngine-Tests.exe --gtest_filter="Mcp*.*"\` → **344 passed, 1 skipped** (\`McpHeadlessAttachTest.HostUntilDetached\`, the interactive-only opt-in test — expected). \`McpPerfSettleTest.RendererSettingsSetSettlesBeforeReturning\` passes.

Issue #616, opened 2026-07-10
Navigation: CrowdManager is initialized+ticked but no agent is ever registered → all NavAgents avoidance-blind


drsnuggles8:
## Summary
`CrowdManager` (DetourCrowd) is **initialized and ticked every frame but no code ever registers an agent
with it** — so it is entirely inert, and every `NavAgentComponent` silently falls back to a naive manual
path-follower with **no inter-agent separation, no local/dynamic obstacle avoidance, and no steering**.
Agents walk straight through each other and through anything not baked into the static navmesh.

Discovered during a `/start-work` code-review sweep of the Navigation subsystem (2026-07-10).

## Evidence
- `CrowdManager::AddAgent` / `SetAgentTarget` have **zero production callers** — only `NavMeshTest.cpp`
  calls them. (`OloEngine/src/OloEngine/Navigation/CrowdManager.cpp:59`.)
- `CrowdManager` is nonetheless constructed + `Initialize`d on every `SetNavMesh`
  (`Scene.cpp:1301-1302`) and its `update()` is ticked every frame (`NavigationSystem.cpp:22-23`) —
  running DetourCrowd against **0 agents** (wasted per-frame cost).
- `NavAgentComponent::m_CrowdAgentId` is only ever assigned `-1` (`Components.h:3689`, `Scene.cpp:1315`),
  so the crowd-sync branch (`NavigationSystem.cpp:34`, `m_CrowdAgentId >= 0`) is **unreachable dead code**;
  every agent takes the naive fallback at `NavigationSystem.cpp:46-81`.

## Impact
Any scene with multiple `NavAgentComponent`s: NPCs overlap, clip through each other, and ignore dynamic
obstacles — the entire reason DetourCrowd exists. Also a small constant per-frame cost updating an empty
crowd.

## Proposed fix
- Register/unregister agents with the crowd on `NavAgentComponent` add/remove (or lazily on first update
  when `m_CrowdAgentId < 0`); store the returned id into `m_CrowdAgentId`.
- Route target changes through `CrowdManager::SetAgentTarget` instead of the one-shot `FindPath`.
- Remove agents on entity destroy / `SetNavMesh` reset (the reset loop at `Scene.cpp:1311` already zeroes
  the ids).
- **Test:** spawn two agents converging on one point and assert they do not overlap (impossible to pass
  today).

## Notes / relationship
Behavior-changing; needs multi-agent functional (and ideally visual) verification. Overlaps
[...]

PR #618, opened 2026-07-10
fix(nav): report partial paths as terminal + validate bake bounds

drsnuggles8:
## Summary

Fixes two verified Navigation correctness/robustness bugs found in a code-review sweep (not a pre-existing issue).

### 1. Partial paths reported as full success → NPC infinite hang (primary)

Detour's `dtNavMeshQuery::findPath` returns a path to the **nearest reachable** poly and sets `DT_PARTIAL_RESULT` when the target is off-navmesh or in a disconnected region. `NavMeshQuery::FindPath` checked only `dtStatusFailed || npolys == 0` and reported this as success. The manual follower walked the agent to the closest point and cleared its target; `BTMoveTo` — checking arrival against the **raw** never-reached target — re-issued the same target and returned `Running` **forever**. The NPC was physically stuck while the behavior-tree node hung silently.

- `FindPath` now returns `FindPathResult { Failed, Partial, Complete }`, detecting `DT_PARTIAL_RESULT`.
- `NavAgentComponent` gains a runtime `m_TargetUnreachable` flag (`OLO_SERIALIZE(Skip)`).
- `NavigationSystem` latches the flag on Failed/Partial, keeps `m_HasTarget` **observable** so consumers see a stable terminal state, and stops recomputing every frame.
- `BTMoveTo` returns **Failure** on an unreachable target instead of spinning.
- Exposed to scripts: C# `NavAgentComponent.IsTargetUnreachable`, Lua `targetUnreachable`.

### 2. Unvalidated nav-bake bounds (hardening)

`NavMeshGenerator::Generate` (public API) fed `boundsMin/boundsMax` straight into `rcCalcGridSize` with no check. A NaN/Inf or inverted (`max <= min`) bound would yield a garbage/huge grid allocation. Added an `isfinite` + `max > min` guard alongside the existing settings validation. The only production caller passes sane bounds, so this is latent — but the API is public.

## Tests

- **New** `NavAgentUnreachableTargetTerminatesTest` — raw follower latches `m_TargetUnreachable` (doesn't cross the gap); `BTMoveTo` terminates with Failure in <10 ticks instead of spinning.
- **New** `NavMeshGeneratorBoundsValidationTest` — NaN/Inf/inverted/degenerate bounds rejected cleanly; valid control passes.
- Existing `FindPath` call-sites migrated to the enum API; `NavMeshOffMeshLinkBridgesGapTest` strengthened to assert `Partial` directly.
- Full build green; 138 behavior-tree / component-coverage / serializer tests pass (no regressions from the new component field). All nav tests pass.

## Related follow-up (not in this PR)

#616 — `CrowdManager` is initialized+ticked but `AddAgent` has zero production callers, so every `NavAgent` falls back to the naive follower with no inter-agent avoidance. That fix is behavior-changing and overlaps `NavigationSystem`/`Scene`; kept separate so this PR stays a shippable correctness fix.

### DIFF
--- a/OloEngine-ScriptCore/src/OloEngine/InternalCalls.cs
+++ b/OloEngine-ScriptCore/src/OloEngine/InternalCalls.cs
@@ -326,6 +326,8 @@ namespace OloEngine
 		[MethodImpl(MethodImplOptions.InternalCall)]
 		internal static extern bool NavAgentComponent_HasPath(ulong entityID);
 		[MethodImpl(MethodImplOptions.InternalCall)]
+		internal static extern bool NavAgentComponent_IsTargetUnreachable(ulong entityID);
+		[MethodImpl(MethodImplOptions.InternalCall)]
 		internal static extern void NavAgentComponent_ClearTarget(ulong entityID);
 		#endregion
 
--- a/OloEngine-ScriptCore/src/OloEngine/Scene/Components.cs
+++ b/OloEngine-ScriptCore/src/OloEngine/Scene/Components.cs
@@ -384,6 +384,11 @@ namespace OloEngine
 	{
 		public bool HasPath => InternalCalls.NavAgentComponent_HasPath(Entity.ID);
 
+		// True when the current target cannot be reached (no path, or only a partial
+		// path to the nearest reachable point). Poll this alongside HasPath: a partial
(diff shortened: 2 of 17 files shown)


drsnuggles8:
@coderabbit review

PR #637, opened 2026-07-11
fix(mcp): honour notifications/cancelled during write-consent wait (#610)

### LINKED ISSUE #610: MCP: write-consent wait ignores notifications/cancelled (and a minor screenshot pose-strand edge)
Two cancellation/timeout robustness gaps surfaced while reviewing the #357 MCP tools. Both were deliberately left out of the review-fix commit (b8ba3d35) because they need more than a minimal edit; filing here so they aren't lost.

### 1. Consent-gate wait is not cancellable (primary)

`McpServer::DispatchToolCall` registers the in-flight call (and its `CancelFlag`) in `m_InFlightCalls` **after** the `ProjectWrite` consent gate, and `RequestConsent` blocks on `m_ConsentCv.wait_for` without consulting any cancel flag. So while a write tool is parked on the `Prompt`-mode consent modal, an arriving `notifications/cancelled` cannot reach it — the call runs to the human's decision or the consent timeout (`m_ConsentTimeoutMs`) regardless.

**Impact:** bounded (the wait is already capped by the consent timeout), so this is a responsiveness/correctness-of-cancellation gap, not a hang.

**Fix sketch:**
- Move the in-flight registration + `CancelFlag` init *before* the consent gate in `DispatchToolCall` (McpServer.cpp ~1599-1667). Keep the existing `InFlightGuard` so every exit path still erases the entry.
- Thread the `CancelFlag` into `RequestConsent`; add it to the `wait_for` predicate and wake `m_ConsentCv` from the cancellation path so the wait returns promptly with a `Cancel` decision.
- Handle that decision as a cancelled response.

Care needed around the `m_ConsentMutex` / `m_ConsentCv` interaction to avoid a deadlock — this is why it wasn't a drive-by edit.

### 2. Screenshot pose-apply timeout can strand the camera (minor)

In `Handle_Screenshot` (McpToolsCamera.cpp), step 1 (apply pose) is a `MarshalRead` **outside** the try/restore block, and `posed` is only set to `true` after it returns. If that first `MarshalRead` times out (throws), the enqueued game-thread job may still run later and apply the requested pose — but with `posed == false` nothing restores it, leaving the user's camera at the screenshot pose.

Only reachable when the editor main thread is unresponsive for the marshal timeout, so rare. On the normal path (including cancellation during the settle wait) the pose is already restored inside the capture job, so there is no strand there.

**Fix sketch:** make the apply/restore a single atomic transaction, or detect the abandoned-job case and restore. Lower priority than #1.

drsnuggles8:
Closes #610.

Fixes the two cancellation/timeout robustness gaps in the #357 MCP tools, plus a pre-existing test port-bind flake surfaced while verifying them.

## 1. Consent-gate wait is now cancellable (primary)

`DispatchToolCall`/`HandleToolsCall` registered the in-flight `CancelFlag` **after** the `ProjectWrite` consent gate, and `RequestConsent` blocked on `m_ConsentCv.wait_for` without consulting any cancel flag — so a `notifications/cancelled` arriving while a write was parked on the Prompt-mode consent modal could not reach it; the call ran to the human's decision or the consent timeout regardless.

- In-flight registration (`CancelFlag` + `m_InFlightCalls` + RAII `InFlightGuard`) now happens **before** the consent gate, so every early-return path still erases the entry.
- `RequestConsent` takes the `CancelFlag`, adds it to the `wait_for` predicate, and returns a new `ConsentDecision::Cancel` (cancellation wins over a racing human Approve). `HandleToolsCall` maps it to a `kRequestCancelledCode` response and **runs no write**.
- The `notifications/cancelled` path wakes `m_ConsentCv` after setting the flag.

**Deadlock warning from the issue — addressed:** `m_ConsentMutex` and `m_InFlightMutex` are **never nested** (no ABBA cycle). The cancel path sets the atomic (under/after `m_InFlightMutex`, released), then a bare `{ lock(m_ConsentMutex); }` barrier serializes with the waiter's predicate check *before* `notify_all()`, closing the lost-wakeup window.

## 2. Screenshot pose-strand edge (minor)

In `Handle_Screenshot`, step-1 (apply pose) was a `MarshalRead` **outside** the try/restore block, so a marshal timeout could leave the enqueued apply job to move the camera later with nothing restoring it. The prior pose + an `Applied` flag now live in a shared `PoseGuardState` captured **by value** into every marshaled job; step-1 apply moved inside the try, and the error-path restore is enqueued **after** any still-pending apply (game-thread FIFO), so a late apply is always undone.

## Also: test port-bind flake

`McpHeadlessHost::Start` derived a base port in `[20000, 59999]` and tried only **24 consecutive** ports. Windows reserves *contiguous* excluded TCP blocks there (Hyper-V/WSL/Docker; up to ~700 wide, e.g. `50000–50480` — confirmed those refuse binding with `AccessDenied`), so ASLR landing the base inside one block failed every attempt (`McpHeadlessAttachTest.ScreenshotAndShaderErrorsOverMcp` / `McpPerfSettleTest.RendererSettingsSetSettlesBeforeReturning` flaky-failing with `port==0`). The derived path now strides by a prime (769) wider than any block; the explicit-`OLO_MCP_PORT` path is unchanged.

## Tests

- 3 new `McpConsentedWriteTest` cases: cancel-while-parked → cancelled response + no write + prompt drained; cancel-wins-over-late-approve; unrelated-cancel-doesn't-abort.
- The two previously-flaky port-bind tests now pass.
- Full suite green: **4348 passed, 0 failed**.

Note: verified through the transport-agnostic dispatch seam (real `HandleToolsCall`/`RequestConsent`/`DispatchRpc` via `HandleMessage`); the HTTP transport layer is untouched.

## Summary by CodeRabbit

* **New Features**
  * MCP tool requests can now be cancelled while awaiting write consent.
  * Cancelled requests return a clear cancellation response and do not apply changes.
  * Consent prompts are safely cleared after cancellation.
  * Screenshot capture now reliably restores the previous camera position, including after errors.

* **Bug Fixes**
  * Prevented cancelled requests from being approved by a late consent decision.
  * Improved test host port selection and connection retry behavior.

* **Documentation**
  * Clarified consent prompt behavior when the MCP panel is closed or a request is cancelled.

### DIFF
--- a/OloEditor/src/MCP/McpServer.cpp
+++ b/OloEditor/src/MCP/McpServer.cpp
@@ -1733,6 +1733,20 @@ namespace OloEngine::MCP
             }
         }
 
+        // Linearization point for the consent→handler transition (issue #610 review):
+        // a notifications/cancelled observed by now — after consent was granted but
+        // before the write starts — prevents the handler from running at all, instead
+        // of executing the (undoable) write and discarding its result at the
+        // post-handler check below. The CancelFlag is a single atomic, so this load is
+        // ordered strictly before or after the cancel store in the flag's modification
+        // order: load == true ⇒ the cancel is "before" write initiation ⇒ don't run;
+        // load == false ⇒ execution has started and any later cancel is handled by the
+        // handler's cooperative IsCurrentCallCancelled() polling + the post-handler
+        // discard. (A CAS state machine would add no further guarantee here — a
+        // mid-run cancel can always reach an opaque handler.)
+        if (scope.CancelFlag->load(std::memory_order_acquire))
+            return MakeError(id, kRequestCancelledCode, "Request cancelled");
+
         ToolResult result;
         try
         {
--- a/OloEditor/src/MCP/McpServer.cpp
+++ b/OloEditor/src/MCP/McpServer.cpp
@@ -984,15 +984,25 @@ namespace OloEngine::MCP
         m_ConsentCv.notify_all();
     }
 
-    ConsentDecision McpServer::RequestConsent(const ToolDef& tool, const Json& arguments)
+    ConsentDecision McpServer::RequestConsent(const ToolDef& tool, const Json& arguments,
+                                              const std::shared_ptr<std::atomic<bool>>& cancelFlag)
     {
         const auto timeout = std::chrono::milliseconds(m_ConsentTimeoutMs.load());
+
+        // Predicate helper: a call cancelled via notifications/cancelled while parked
+        // here must return promptly (issue #610). The cancellation path stores the
+        // flag then synchronizes on m_ConsentMutex before notifying, so a read taken
+        // under this lock cannot miss a store the notifier has already published.
+        const auto cancelled = [&cancelFlag]
+        { return cancelFlag && cancelFlag->load(std::memory_order_acquire); };
 
         u64 myId = 0;
         {
             std::unique_lock lock(m_ConsentMutex);
             if (m_ConsentAborting)
                 return ConsentDecision::Deny;
+            if (cancelled())
+                return ConsentDecision::Cancel;
             // A concurrent mode change may have already settled the outcome before we
             // enqueue anything — honour it without prompting.
             const WriteConsentMode mode = m_ConsentMode.load();
@@ -1009,9 +1019,9 @@ namespace OloEngine::MCP
             entry.Summary = BuildConsentSummary(arguments);
             m_ConsentQueue.push_back(std::move(entry));
 
-            const bool resolved = m_ConsentCv.wait_for(lock, timeout, [this, myId]
+            const bool resolved = m_ConsentCv.wait_for(lock, timeout, [this, myId, &cancelled]
                                                        {
-                                                           if (m_ConsentAborting)
+                                                           if (m_ConsentAborting || cancelled())
                                                                return true;
                                                            for (const ConsentEntry& e : m_ConsentQueue)
                                                            {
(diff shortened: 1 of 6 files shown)


drsnuggles8:
Addressed the two review findings in `3a2870f4`:

**1. Consent→handler race.** Added a `CancelFlag` check immediately before the handler runs (`HandleToolsCall`): a `notifications/cancelled` observed after consent is granted but before write initiation now prevents the (undoable) write entirely, instead of running it and discarding the result at the existing post-handler check. The flag is a single atomic, so that load is the linearization point (`load==true` ⇒ cancel precedes write initiation ⇒ skip; `load==false` ⇒ execution started, later cancels fall to the handler's cooperative `IsCurrentCallCancelled()` + post-handler discard). I did **not** add a CAS state machine — it adds no further guarantee for an opaque handler a mid-run cancel can always reach. New regression test `CancelBeforeHandlerStartNeverInitiatesWrite` uses a handler-entered probe + the consent modal as a barrier to prove a cancellation before the handler never initiates the write.

**2. Port constants.** Hoisted the ephemeral-port range/stride into shared class-scope constants (`kPortRangeLo` / `kPortRangeSpan` / `kPortStride`) used by both `Start()` and `DerivePidPort()`, replacing the duplicated locals/literals. No behavioural change; range bounds and modulo preserved.

Validated: all 26 consent + progress-cancel + port-bind tests pass, full `Mcp*` filter green (367 passed, 1 interactive skip), OloEditor + tests compile clean.

Issue #635, opened 2026-07-11
Gameplay: RPG character progression — XP/leveling, stats, skill trees, classes


drsnuggles8:
## Problem

OloEngine already ships a comprehensive GAS-style ability system, an RPG inventory/equipment/loot layer, and a full quest system — all wired through ECS, scene serialization, save-games, and both scripting VMs (see *Why it matters* for the exact file inventory). What it does **not** have is the pillar that makes those systems into an actual RPG: **character progression**. The loop is visibly dangling in the shipped code:

- `QuestRewards::ExperiencePoints` (`Quest.h:84`) is defined and loaded, but `QuestJournal.cpp:141` grabs the reward struct and **never grants XP** to any character — there is no XP recipient.
- `QuestJournal::SetPlayerLevel` + a `Level` quest-requirement gate (`QuestRequirement.h:21`, `QuestJournal.cpp:594`) exist, but a repo-wide grep confirms **nothing ever raises a character's level through play** — `SetPlayerLevel` is called only from deserialize / save-restore / scripting.
- `LootTable::Roll(itemLevel)` scales drops by an item level that **no character level feeds**.
- `AbilityComponent::InitializeDefaultRPGAttributes` **hard-codes** four attributes — there is no class/archetype system, no growth, no leveling.

Grep for `ProgressionComponent` / `CharacterClass` / `SkillTree` / `TalentTree` / `ExperienceComponent` / `LevelUpEvent` across `OloEngine/src` and `OloEditor/src` returns **empty**. The progression subsystem is genuinely absent.

## Why it matters

Per the #440 discipline, the GAS + inventory + quest theme is **already shipped and must not be re-proposed**. What exists today:

- **Abilities/attributes/effects/tags** — `OloEngine/src/OloEngine/Gameplay/Abilities/`: `AttributeSet.{h,cpp}`, `Effects/GameplayEffect.h` + `ActiveEffectsContainer.{h,cpp}` (Instant/HasDuration/Infinite, periodic, stacking, granted/required/blocked tags), `GameplayAbility.h` (cost/cooldown/tag-gating, channeled/toggled), `CooldownManager.{h,cpp}`, `Tags/*`, `Damage/*`, `GameplayAbilitySystem.{h,cpp}`, `AbilityComponents.h`.
[...]

Issue #632, opened 2026-07-11
Renderer: realtime dynamic global illumination (DDGI probe volumes)


drsnuggles8:
## Problem

OloEngine's light-probe GI is **fully baked and static**. `LightProbeBaker` captures a cubemap at each probe offline (editor-driven), projects it to L2 SH, and stores it in a serialized `LightProbeVolumeAsset`. At runtime `Scene::OnUpdateRender` (`Scene.cpp:5521-5580`) uploads that frozen SH data to the GPU every frame and `LightProbeSampling.glsl` trilinearly blends it. Nothing relights when a light moves or geometry changes, and the trilinear sampler has **no visibility term — it light-leaks through walls**. The only dynamic indirect light today is `SSGIRenderPass`, a short-range screen-space one-bounce effect that vanishes off-screen and off-contact.

This epic adds a **realtime Dynamic Diffuse Global Illumination (DDGI)** subsystem: probe irradiance that updates every frame in response to moving lights and changed geometry, with proper wall occlusion, multi-bounce, and bounded cost. It reuses the existing probe-grid placement, component, and trilinear scaffolding, and replaces the static SH payload with GPU-relit atlases.

## Why it matters

- **Realtime, not baked.** #439 tracks the *baked* lightmap/GI path; this is the dynamic counterpart — day/night cycles, destructible/moving geometry, and moving lights all bounce correctly with no re-bake.
- **Fixes a real shipping bug.** The current trilinear sampler leaks light through walls (no visibility term). The Chebyshev visibility atlas is the standard DDGI fix and lands here as part of the sampler rewrite.
- **Flagship scope.** A new subsystem with two GPU atlas representations, per-frame ray gather, three compute stages, an amortized scheduler, and a two-path lighting integration — the multi-day single-shot runway the froxel anchor (#435) targets.

## What exists today (confirmed by grep + read — the #440 discipline)

- `Renderer/LightProbeBaker.{h,cpp}` — offline editor cubemap capture -> L2 SH. **Baked only.**
- `Renderer/LightProbeVolumeAsset.h` — stored, serialized SH coefficient array.
- `Scene/Components.h:2163` `LightProbeVolumeComponent` — grid/bounds/spacing/intensity + `m_BakedDataAsset`; `m_Dirty` triggers an **editor re-bake only**.
[...]

Issue #497, opened 2026-07-01
clang-cl + ASan miscompiles C++ exception handling (9 EH tests excluded from Windows ASan job)


drsnuggles8:
## Summary

The `ASan (Windows/clang-cl)` job (introduced by #490's switch from MSVC cl.exe to clang-cl) **cannot run C++ exception-handling paths**: clang-cl + `/fsanitize=address` miscompiles MSVC-style C++ throw/catch, so any test that triggers a real `throw` unwinding into a `catch` crashes.

## Root cause

ASan puts redzones on the compiler-emitted C++ EH metadata globals (`_ThrowInfo`, `_CatchableTypeArray`, RTTI type descriptors). The MSVC C++ EH runtime reads across these structures during a throw, tripping the redzone → `global-buffer-overflow` / `0xC0000005`.

Reproduces with a minimal program under `clang-cl /fsanitize=address /EHsc /MD /O2`:

```cpp
int main() {
    try { throw std::runtime_error("boom"); }
    catch (const std::exception&) { return 0; }
}
```

→ `AddressSanitizer: global-buffer-overflow ... READ of size 8` (shadow byte `f9` = global redzone), aborts.

This is an **LLVM toolchain bug, not an engine defect** — the previous MSVC cl.exe ASan job passed all these tests, and they pass on the Linux ASan job. No compile/link/runtime workaround fixes it (verified: `-mllvm -asan-globals=0`, `-fno-strict-aliasing`, `-fno-delete-null-pointer-checks`, `/Od`, `ASAN_OPTIONS=detect_stack_use_after_return=0` / `replace_intrin=0`, and clang-cl-driver vs direct lld-link linking all still crash).

## Impact

9/3895 tests fail — exactly the ones that throw+catch (malformed-input robustness + MCP error handling). Excluded from the clang-cl ASan ctest run in `.github/workflows/asan.yml` so the job is green; any *future* throw/catch test will also fail this job until the underlying bug is resolved.

Excluded tests:
- `SceneSerializerFuzzRegression.UnterminatedFlowMap`
- `SceneSerializerFuzzRegression.SinglePrintableChar`
- `EngineSubsystemSmoke.ProjectLoadMissingPathFailsCleanly`
- `StreamingRegionSerializer.ParseInvalidYAML`
- `InputActionSerializerTest.MalformedYAML`
- `ShaderGraphSerializationTest.DeserializeInvalidYAMLReturnsFalse`
[...]

PR #659, opened 2026-07-16
feat(atmosphere): weather director, astronomical time-of-day, volumetric cloudscape (#633)

### LINKED ISSUE #633: World: atmosphere & sky — weather director, astronomical time-of-day, volumetric clouds
## Problem  The engine has the *ingredients* of an atmosphere but no atmosphere. Five sky/weather pieces exist and none of them talk to each other:  - **`ProceduralSky`** (`OloEngine/src/OloEngine/Renderer/ProceduralSky.{h,cpp}`) — Preetham analytic daylight, baked to a cubemap only when its parameter hash changes. The sun is a **static authored `SunDirection` vector**. - **`StarNestSky`** (`Renderer/StarNestSky.{h,cpp}`) — a separate raymarched nebula, not linked to time. - **`ProceduralSkyComponent`** (`Scene/Components.h` ~l.1986) has an `m_LinkSunToDirectionalLight` flag and an **ephemeral MCP-only** sun override (`OloEditor/src/MCP/McpToolsRender.cpp` ~l.759 `olo_scene_set_time_of_day` / `set_sun_angle` → `Renderer3D::SetSunDirectionOverride`). Its own comment says *"useful when authoring a time-of-day controller"* — **the controller was never built.** No clock, no ephemeris, no dynamic light curve, no automatic day/night, no night-sky blend. - **Four independent weather structs** — `m_WindSettings` / `m_SnowSettings` / `m_PrecipitationSettings` / `m_FogSettings` (`Scene.h` ~l.889-894; structs in `Renderer/PostProcessSettings.h`), driven by `WindSystem` / `SnowAccumulationSystem` / `PrecipitationSystem` + `PostProcess_Fog.glsl`. **Nothing couples them** — you can get rain particles under a cloudless Preetham sky with no wind, all contradicting each other. - **No volumetric clouds at all.** `OloEditor/assets/shaders/` contains only `ProceduralSky.glsl`, `StarNestSky.glsl`, `Skybox*.glsl`, `PostProcess_Fog.glsl`. Fog's volumetric ray-march is a *screen-space post pass*, not a cloudscape.  ## Why it matters  A dynamic sky is the single highest-leverage visual multiplier in an engine: it elevates *every* scene simultaneously, and it is the thing the current static baked sky most conspicuously lacks. Coupling the five weather systems under one director also fixes real correctness gaps (contradictory weather state) and gives gameplay/scripting/save-games a single authoritative "what is the weather / what time is it" source instead of five drifting knobs. This is a portfolio-grade flagship on par with the #435 froxel-lighting epic.  ## Proposed approach  Three fused, interlocking pillars — the value is in wiring them **to each other**, not just shipping three features:  **Pillar A — Weather director.** A scene-level `WeatherSystem` + serialized `WeatherStateComponent` owning named states (Clear / Overcast / Rain / Storm / Snow / FogBank), per-state presets spanning all 5 subsystems, and a transition engine that cross-blends toward a target state over a configurable duration with per-parameter easing. Registered in `Scene::GetGameplayScheduler()` with correct Reads/Writes channels. Publishes a derived global **wetness** signal (rises while raining, slow dry-out) into a UBO so PBR materials darken/smooth up-facing surfaces.  **Pillar B — Astronomical time-of-day driver.** A `TimeOfDaySystem` + serialized `TimeOfDayComponent` (time-of-day, day length, date, latitude, time-scale/pause) computing **sun AND moon** world direction from a simplified ephemeris (pure-math core, headless-testable). Drives the directional light color+intensity and sky exposure/turbidity along a dawn → golden-hour → noon → dusk → blue-hour → night curve; cross-fades the Preetham day sky to a moon-lit star night sky through twilight with a moon disk + moonlight secondary light. Becomes the **single authoritative sun source** — retiring `m_LinkSunToDirectionalLight` and the ephemeral MCP override in favor of a real clock-advancing `olo_scene_set_time_of_day`.  **Pillar C — Raymarched volumetric cloudscape.** A `CloudscapeComponent` + a weather-map texture (coverage / type / wetness) and Perlin-Worley 3D-noise assets; a two-layer stratus+cumulus vertical density profile with coverage/erosion and wind-driven animation (reads the existing `WindSystem` field). A new render pass raymarches the field with cone-sampled in-scatter lighting (Henyey-Greenstein phase + powder term + cheap multiscatter approx) lit by the **time-of-day sun/moon**, ambient from sky IBL, Beer-Lambert transmittance — at half-res with a blue-noise jittered start, temporally reprojected (history buffer + neighborhood clamp) and depth-composited (clouds behind geometry, aerial perspective). Casts **cloud shadows** onto the world (attenuate the directional light), and **reads** the fog/aerial-perspective output (scope this coupling to consume froxel #435's fog rather than rewrite it — sequence after/alongside #435), and contributes an IBL term so PBR surfaces + SSR reflect the sky.  **The cross-wiring (what makes it a flagship, not three features):** weather state sets cloud coverage + fog + wind + precipitation *together*; time-of-day lights the clouds and colors the sky; clouds shadow the ground and dim the sun.  ## Acceptance criteria  - [x] Weather director core: `WeatherSystem` + serialized `WeatherStateComponent`, named states, per-state presets across all 5 subsystems, cross-blend transition engine, scheduler registration with correct Reads/Writes channels. - [x] Derived global wetness signal feeding PBR material darkening/roughness on up-facing surfaces; ties into precipitation-accumulation feedback. - [x] Astronomical clock + sun/moon ephemeris: serialized `TimeOfDayComponent`, deterministic pure-math core pinned by a unit test (mirroring `ProceduralSkyMathTest`). - [x] Dynamic sun lighting curve drives directional light color/intensity + sky exposure/turbidity across the day; TimeOfDaySystem is the single authoritative sun source; `m_LinkSunToDirectionalLight` + ephemeral MCP override retired in favor of a real clock-advancing tool. - [x] Night sky + twilight cross-fade (day sky ↔ moon-lit star sky) with moon disk + moonlight; cubemap/IBL hash-gated on quantized time (no per-frame rebake). - [x] Cloud density field: `CloudscapeComponent` + weather-map texture + Perlin-Worley noise, two-layer stratus+cumulus profile, coverage/erosion, wind-driven animation from `WindSystem`. - [x] Cloud raymarch + lighting pass: cone-sampled in-scatter (HG + powder + multiscatter approx) lit by time-of-day sun/moon, sky-IBL ambient, Beer-Lambert transmittance, half-res + blue-noise jitter. - [x] Temporal reprojection + upsample + depth/aerial-perspective composite (clouds occluded by geometry, correct against the sky). - [x] Cloud shadows attenuate the directional light on the ground; cloud coverage feeds fog/aerial perspective (reading froxel #435 output, not rewriting it); cloud IBL contribution reflected by PBR surfaces + SSR. - [x] Full plumbing: three new components serialized through the codegen touch-points; editor inspector panels (weather-state dropdown + transition, time-of-day scrubber, cloud/weather-map controls); Lua bindings; MCP tools (set weather, advance/set time-of-day, capture cloud + cloud-shadow buffers); save-game round-trip. - [x] Testing + visual verification: CPU/contract tests for ephemeris, weather-blend, and cloud density/phase math (L1-L5); headless `ShaderHarness` compile guard for the new shaders; multi-angle screenshot evidence across a dawn/noon/dusk/night × clear/overcast/storm matrix, PNGs read back and confirmed (per the repo's "rendering changes MUST be visually verified" rule).  ```olo-score capability: 9 craft: 9 stability: 7 decay: 3 effort: 13 confidence: 0.9 learning: 9 fun: 10 kano: delighter blocked_by: [] blocks: [] ```

drsnuggles8:
Implements all three pillars of #633 — the weather director, the astronomical
time-of-day driver, and the raymarched volumetric cloudscape — wired to each
other, on one branch.

Closes #633

## Pillar A — Weather director

- `WeatherStateComponent` with six named states (Clear, Overcast, Rain, Storm,
  Snow, Fog Bank), each a full `WeatherPreset` spanning clouds / fog / wind /
  precipitation / snow accumulation / sun dimming / wetness target.
- Snapshot-based cross-blend transition engine (`WeatherSystem`): changing the
  target starts a timed transition; a mid-flight retarget eases from the
  currently applied blend, never snaps. Disabled endpoints contribute zero;
  precipitation kind switches at the transition midpoint.
- Registered in the gameplay scheduler as `"Weather"` with an explicit
  `Before("PhysicsKick")` edge (cloth samples wind inside the kick).
- Derived global wetness: rises while rain/sleet falls, dries out slowly after
  (asymmetric rates), uploaded per-frame in the new `AtmosphereShadingUBO`
  (binding 53) and consumed by all five PBR surface shaders
  (`atmosphereApplyWetness`: up-facing albedo darkening + roughness drop).

## Pillar B — Astronomical time-of-day

- `TimeOfDayComponent` + pure-math `Atmosphere/Ephemeris.{h,cpp}` (Cooper
  declination, hour-angle horizontal coordinates, synodic moon approximation,
  rebake quantization, the day-lighting curves) — headless-tested.
- `TimeOfDaySystem::Apply` runs on the render path in BOTH editor and runtime
  and drives the scene's first directional light: warm→white color curve,
  intensity ramp, and a sun↔moon swap across the horizon — one authoritative
  light, so every single-sun consumer stays photometrically consistent by
  construction.
- `ProceduralSkyComponent::m_LinkSunToDirectionalLight` retired (scene YAML
  drops the key; save-games v9 read-and-discard the legacy bool), and the
  ephemeral `Renderer3D::SetSunDirectionOverride` plumbing removed — the MCP
  time/sun tools now write the real component (`olo_scene_set_sun_angle`
  solves for the closest clock time), plus new `olo_scene_set_weather` /
  `olo_scene_get_atmosphere` tools (writes consent-gated).
- New `AtmosphereSky` combined bake: Preetham day + procedural starfield with
  sidereal rotation + phase-lit moon disk/halo + twilight cross-fade + damped
  cloud tint, into ONE cubemap + IBL set, hash-gated on time quantized to
  `m_RebakeQuantumGameMinutes` (no per-frame rebake), hash-recorded-on-failure.

## Pillar C — Volumetric cloudscape

- `CloudscapeComponent` + GPU-generated Perlin-Worley base (128³ RGBA8) and
  Worley detail (32³) noise, a weather map (authored `AssetHandle` or a
  procedural default with measured-min/max-normalized coverage), two-layer
  stratus↔cumulus height gradients, wind-driven advection from the
  `WindSystem` settings on a mockable clock.
- Half-res raymarch (`PostProcess_Cloudscape.glsl`): IGN-jittered start,
  dual-lobe HG phase + powder + one reduced-extinction multi-scatter octave,
  Beer-Lambert with a shared `kCloudExtinction`, cone light-march toward the
  active sun/moon, sky-ambient term.
- TAA-pattern temporal reprojection (graph-tracked history, 3×3 neighborhood
  clamp), depth-aware bilateral upsample, `scene·T + inscatter` composite
  BEFORE the fog pass (depth-clipped against geometry; froxel fog + analytic
  tail supply aerial perspective over the clouds). `CloudsColor` inserted into
  every downstream candidate ladder.
- Top-down cloud-shadow transmittance map (`CloudShadow_Generate.comp`, same
  density field + extinction) multiplied into the directional light in all
  five surface shaders OUTSIDE the CSM gate and mirrored into the froxel fog
  scatter's sun term (photometric-parity discipline); weather-blended coverage
  also feeds the (damped) sky-bake tint so IBL/SSR see the cloudscape.

## Plumbing

Three components fully codegen-serialized (scene YAML + binary; nested
`WeatherPreset` sub-maps; `OLO_SERIALIZE(Skip)` runtime fields); save-game
format v9 with capture/restore registration; editor inspector panels (live
time scrubber, weather-state dropdown + per-preset editors, cloudscape
controls + weather-map drag-drop); Lua usertypes + round-trip tests;
`OLO_PROPERTY` C# bindings; MCP field registry. All coverage tests green.

## Testing & visual verification

- ~200 targeted tests: `EphemerisMathTest` / `WeatherBlendTest` /
  `CloudDensityMathTest` / `AtmosphereSkyMathTest` (L1),
  `WeatherTimeOfDayDirectorTest` (Functional, incl. save-game round-trip),
  scheduler seams, MCP sun-solve.
- `AtmosphereVisualEvidenceTest`: a 12-cell {dawn,noon,dusk,night} ×
  {Clear,Overcast,Storm} matrix rendered through the FULL pipeline with 4
  driver-independent contracts + 12 committed golden PNGs, each read back and
  inspected. The fixture snapshots/restores the process-global Renderer3D
  weather settings AND pins the settings-applied live renderer state
  (`ApplyRendererSettings` after `EnableRendering`) so captures are
  suite-order-independent — both traps are written up in the new
  `docs/agent-rules/volumetric-cloud-debugging.md`.
- Live editor verification over MCP (multi-angle screenshots, buffer captures,
  the new weather/time tools): broken-cumulus noon, storm snap, moonlit
  starfield night, golden-hour dusk.
- **Full suite: 4737 ran, 4734 passed, 0 failed** (3 environment-gated skips).

## Notes for review

- **Deliberate behavior change:** StarNest bake failures now record the hash on
  BOTH paths (editor + runtime), so a failed bake no longer retries every
  frame.
- Known v1 simplifications (documented in code): uniform moon disk (no
  crescent terminator), cloud-shadow map at the y=0 reference plane (no
  per-pixel height parallax), precipitation kind switches at the transition
  midpoint, clouds receive last-slice froxel fog + analytic tail rather than
  per-slice samples.
- Hygiene: removes `FoliageGenerationTest.olo.auto`, an editor autosave
  artifact accidentally committed in ff0c4024.
- `OloEditor/SandboxProject/Assets/Scenes/AtmosphereTest.olo` is a tuned demo
  scene — open it and scrub the Time Of Day slider / switch the weather target
  to explore.

## Summary by CodeRabbit

* **New Features**
  * Added dynamic time-of-day lighting (sun/moon, sky baking, day/night blending) driven by components and persistent scene state.
  * Added full volumetric cloudscape rendering with temporally-resolved clouds, procedural cloud noise, and cloud-shadowing.
  * Added weather director support with smooth transitions, precipitation/fog/snow/wetness effects, and wind/scene “twin” syncing.
  * Added editor inspectors and component authorship (time-of-day, weather, cloudscape), plus MCP/Lua controls and save-game support.

* **Documentation**
  * Added volumetric cloud debugging guide and updated MCP atmosphere/weather tool guidance.


drsnuggles8:
Addressed the CI failure, both SonarQube reliability issues, and the CodeRabbit review. Summary:

## CI: Windows clang-cl Release+ASan OOM — fixed (structural)

`LLVM ERROR: out of memory` compiling `McpToolsScene.cpp.obj` in the Sanitizers workflow. Root cause: the generated MCP writable-field registry (`McpFieldRegistry.Generated.inl`, ~100 components × two type-erased lambdas per field) was built by an **inline** `BuildRegistry()` in `McpGenericFieldWrite.h`, so the full instantiation storm recompiled in every includer — `McpToolsScene.cpp` plus two MCP test TUs (their Debug objects were 470–560 MB each). The #635 + #633 component additions pushed the biggest TU past the CI runner's memory. Note **master is red too** (its Sanitizers run fails since the #658 merge, same cause); this PR carries the fix.

Fix: `Registry()` is now defined in a dedicated `OloEditor/src/MCP/McpFieldRegistry.cpp` TU (linked by both OloEditor and OloEngine-Tests) that contains nothing but the registry build; the header keeps the declaration. The instantiation storm compiles exactly once, in a TU with no other content.

## SonarQube (reliability E)

- **`CloudscapeComponent` memcmp equality (cpp:S5000)** — its `operator==` was whole-object `Math::BitwiseEqual` (memcmp), which compares indeterminate padding bytes (bool → f32 alignment gaps). Replaced with a member-wise comparison (bitwise per float member, `static_cast<u64>` for the AssetHandle per the documented UUID C2666 footgun) + a `PreferValueComparison<CloudscapeComponent>` opt-in in `SceneHierarchyPanel.cpp` so editor undo uses the value path — the same pattern as `PerceptibleComponent`.
- **Scene.cpp single-iteration loops (S1751)** — five "take the first entity of a view" `for`-loops with unconditional `break` rewritten as explicit first-entity lookups (`view.begin() != view.end()`), with comments stating the one-cloudscape/one-clock-drives-the-scene contract.

## CodeRabbit findings — 4 fixed, 7 skipped (each verified against current code)

**Fixed:**
1. `AtmosphereTest.olo` — `XFogSettingsDisabledX` renamed to the recognized `FogSettings` key **with `Enabled: false`** (the suggestion's literal wording would have kept `Enabled: true` and turned fog on; the intent — an explicit disabled fog block instead of an unknown-key hack — is implemented).
2. `Scene::OnUpdateEditor` now calls `WeatherSystem::ApplyImmediate` after the time-of-day edit-mode tick, so WeatherStateComponent edits preview live without Play (the zero-dt apply recomputes the blend only — no transition/wetness drift from an open scene). This was the header's own documented contract with no call site.
3. `TimeOfDayComponent::operator==` excludes the clock fields while `m_Enabled && m_AdvanceInEditMode && !m_Paused` — the editor undo snapshot persists across frames, so the ticking clock generated a spurious undo entry every frame the inspector was open. Comparison stays exact when advancement is off (time-slider scrubbing must remain undoable); symmetry holds because the gating bools are compared first.
4. `WeatherTimeOfDayDirectorTest` golden-hour block — moved to a deterministic daylight hour (17:00, ~9.6° elevation; 18:00 was actually −0.4° and only lit via the intensity ramp's below-horizon tail) and the conditional guard replaced with unconditional assertions (`intensity > 0`, `!m_IsNight`, warmth checks), so the block can never silently skip.

**Skipped (verified, with reasons):**
1. *McpRenderOverrides.h declination duplication* — deliberate mirror-with-named-comment ("the same model `Ephemeris::SolarDeclination` pins"), the repo's sanctioned pattern (light-path-photometric-parity.md); drift is pinned by the `RoundTripsThroughEngineEphemeris` differential test (0.5° tolerance). Reusing the engine function would break the header's documented minimal-dependency contract and inject f32 into the solver's f64 pipeline.
2. *AtmosphereSky::EvaluateAtDirection normalize guard* — test-only CPU mirror kept structurally identical to the shader (same convention as StarNestSky/ProceduralSky evaluators); sanitization lives at the authored-data boundary (`ComputeUBO`, which already falls back for a degenerate moon vector).
3. *CloudShadowMap ShadowWorldSize NaN sanitize* — every ingress path is already finite-validated + range-clamped ([500, 50000]): YAML `TryReadFiniteF32`+clamp, scenebin `isfinite` read, save-game `isfinite ? clamp : default`, MCP clamp, bounded editor widget. Consumer-side re-validation would be redundant layering inconsistent with the repo's validate-at-ingress convention.
4. *CloudscapeRenderState tolerance-based operator==* — the defaulted `==` is change-detection, not value comparison; all seven sibling settings structs in the file use the identical defaulted form, and the one explicit sibling (`UnderwaterFogState`) is explicit precisely to keep **bitwise** semantics. A tolerance compare would miss real small changes and break transitivity.
5. *CloudscapeRenderState / Renderer3D.h / CommandDispatch m_PascalCase renames* — the flagged members match their immediate surroundings exactly (WindSettings/UnderwaterFogState, the Renderer3DData fields, `CSMShadowTextureID`/`AtlasShadowTextureID` siblings); renaming only the new members would make those structs internally inconsistent.
6. *EphemerisMathTest EXPECT_NE(a, c)* — the requested tolerance assertion already exists on the next line (`EXPECT_NEAR(c, 10.0f + 5.0f/60.0f, 1e-4f)`); the `EXPECT_NE` additionally pins the bucket-crossing change-detection contract (the quantized value must move bitwise so the sky-bake hash re-bakes), and a/c differ by a full bucket (0.083 h) — not flake-prone.
7. *Test-local struct member renames (Case, BandStats)* — plain PascalCase members are the dominant convention for local test-helper aggregates across `OloEngine/tests/` (Pose, LaunchResult, Failure, …); `m_` is reserved for fixture-class members.

## Drive-by test-robustness fix

`McpProgressHttpTest`'s port sweep (ASLR-derived base, 32 consecutive ports) drowned entirely when the base landed inside one of Windows' Hyper-V/WSL excluded-port blocks (netsh reserves ~100-port chunks; nothing listens there but every bind fails, and the base is sticky per binary image so every rerun failed identically). The sweep now strides by a prime (137) larger than a reservation block, so no single block can swallow all attempts.

All 4,826 tests green locally after the changes (MSVC Debug full suite).

drsnuggles8:
CI follow-up (`b4830c75`) — the previous registry-TU split fixed the original OOM mechanism but surfaced two new failures, both in the new `McpFieldRegistry.cpp`:

1. **Windows (MSVC Ninja Release)** — the TU was missing the `OloEnginePCH.h` first-include every sibling MCP TU has, so nothing defined `NOMINMAX` before windows.h and `std::numeric_limits<T>::max()` in `ContainerAllocationPolicies.h` got macro-clobbered (`C2589 '(': illegal token on right side of '::'`). The local VS-generator build force-includes the CMake PCH, which masked the omission. Fixed by adding the include.

2. **Sanitizers (ASan Windows/clang-cl Release)** — the dedicated TU still exhausted the runner ("The paging file is too small", 0x5AF): compiler peak memory scales with the largest *function*, and `BuildRegistry()` was one flat body of 983 closure-heavy `push_back`s. OloHeaderTool now emits the registry as `BuildRegistryChunkN` functions (component-granular, ~48-entry target → 15 chunks today) plus a `BuildRegistryChunks` driver, and the TU includes the generated `.inl` at anonymous-namespace scope. The chunking is in the generator (`McpChunkWriter`), so it scales automatically as future components grow the registry.

`McpFieldRegistryTest` / `McpGenericFieldWriteTest` exercise `Registry()` at runtime (no text parsing of the `.inl`), all green locally along with the compile of the changed TU. Letting CI validate the two failing configs now.

drsnuggles8:
Third CI round (\9ee934b\): the previous fixes worked — **Windows is green** and the ASan job now **builds and runs the full suite** (the chunked registry compiled fine). The remaining failure was 2 tests out of 4,807: \ExperienceCurveTest.SerializerRejectsMalformedYAML\ and \CharacterClassDatabaseTest.SerializerRejectsMalformedYAML\, both inherited from master's #635 merge (master's own Sanitizers run is red on them too). They crash with the documented clang-cl+ASan exception-handling miscompile (0xc0000005 on any real throw+catch, issue #497) — the exact category the workflow already excludes nine tests for. Added both to the exclude list with the same rationale; they stay covered by the green Linux ASan/UBSan/TSan jobs. Workflow-only change, no code touched.

drsnuggles8:
Rebased onto master (now includes the DDGI epic #632) and addressed the second CodeRabbit round — **all six findings verified valid and fixed**.

## Rebase notes (the conflicts were semantic this time)

- **Save-game version**: both #632 and this branch claimed v9. DDGI's v9 (probe fields) stands; the atmosphere slice moved to **v10** and the legacy `m_LinkSunToDirectionalLight` read-and-discard gate now fires for v9-and-older saves.
- **Binding slots**: DDGI took UBO 51 + TEX 56–58, so the cloud slots shifted — UBO_ATMOSPHERE_SKY/CLOUDSCAPE/SHADING → 52/53/54, TEX_CLOUD_* → 59–62, TEX_SHADER_GRAPH_0 → 63 — in `ShaderBindingLayout.h` and the four GLSL declaration sites. The reflection-binding/UBO-size guard tests pin the pairing.
- **Dropped my `asan.yml` exclusions commit**: master fixed the same two #635 malformed-YAML tests better (in-source `OLO_ASAN_ENABLED && _WIN32` guards that keep the non-throwing assertions running everywhere). The workflow exclusions would have skipped those tests entirely on Windows ASan — redundant and strictly worse, so they're gone.

## CodeRabbit findings (6/6 valid, all fixed)

1. **`WeatherSystem::AdvanceWetness` rate sanitize** — real gap: `std::max(NaN, 0.0f)` returns NaN (the comparison is false), and a Lua script writes the rate fields directly with no ingress validation, so a NaN rate would poison the wetness signal every surface shader consumes. The chosen rate is now `isfinite ? max(rate, 0) : 0`.
2. **`CloudShadowMap::Shutdown` cache invalidation** — real: the shadow map is bound through the *tracked* PBR dispatch path (`CommandDispatch::SetCloudShadowTextureID`), and `InvalidateTextureBinding`'s contract requires the call before deletion so a recycled GL ID can't be skipped against stale tracking (the raw-ID teardown bypasses the `OpenGLTexture2D` destructor that normally does this). Added before `DeleteTexture`.
3. **Exact cloud sampler names in the binding validator** — fixed during the conflict resolution: each `TEX_CLOUD_*` case now matches only its own sampler name, so swapped base/detail/weather/shadow bindings are rejected instead of sliding through a shared `contains("Cloud")`.
4. **`m_WeatherMapScaleKm` save-game fallback** — stale `30.0f` from before the live-tuning changed the component default to `12.0f`; updated (clamp range unchanged).
5. **Lua read-only coverage** — each property is registered individually in Sol2, so one accidentally-writable sibling would slip past a single-property probe. The ToD test now attempts writes to `sunDirection`/`moonDirection`/`isNight` (plus the existing `sunElevationDegrees`), the weather test to `transitionProgress`/`wetness` — each asserting the script errors and the field is unchanged.
6. **Second camera angle in the visual matrix** — added a 13th capture, `NoonClearAerial`: 45 m up, ~22° down-tilt. The committed golden shows the ground plane from above with the pillar's shadow, the props depth-composited against the atmosphere, and the noon cumulus band — a substantially different vantage from the horizon pose. (First attempt at 380 m/50° produced a featureless haze — the whole 60°-FOV frustum sat below the horizon; caught by reading the PNG back, per the repo rule, and documented in the test comment.)

Full suite green locally after the rebase + fixes; the 12 existing goldens were verified unchanged (re-render drift RMSE ≤ 0.13 vs threshold 8 — pure driver noise — so they were restored rather than churned).

drsnuggles8:
Third CodeRabbit round: 2 of the 3 new findings valid and fixed; 1 invalid; the duplicates were adjudicated in round 1 and stand.

**Fixed:**

1. **MCP guide security model** — valid: the "no tool writes scenes, assets, or files" bullet predated the consent-gated write tier. It now reads "read-only **by default**" and explicitly distinguishes the three tiers: editor-only viewport state (never persisted), ephemeral session-global renderer overrides (scene reload restores them), and the **(consented write)** tools — `olo_entity_set_field`, `olo_scene_set_time_of_day`, `olo_scene_set_sun_angle`, `olo_scene_set_weather` — which mutate serialized components of the loaded scene **in memory** (undoable, discarded on reload, reaching disk only via an explicit user save; no tool writes files). The consent-gate framing is preserved and cross-linked.

2. **Stale binding numbers in comments** — valid: the DDGI rebase shifted the cloud slots (UBO 53/54, TEX 59–62) but a dozen prose comments still cited the pre-rebase numbers. Fixed all cited sites (`ShaderBindingLayout.h` CloudscapeUBO doc, `CloudShadowMap.cpp`, `CloudShadow_Generate.comp`, `CommandDispatch.h`, five spots in `RenderPipeline.cpp`) **plus five more the sweep found** (`AtmosphereSky.glsl`, `AtmosphereShading.glsl`, `CloudscapeCommon.glsl`, two in `CloudShadowMap.h`). Comment-only; the reflection-binding guard tests pin the actual numbers.

**Skipped:**

3. **Cloud history invalidation when the pass is inactive** — already guaranteed structurally. The history-sink protocol (`RenderGraph::ExtractHistoryTextures`) force-clears **every** sink's valid-flag at the top of each frame's extraction pass and re-asserts it only after a successful `glCopyImageSubData` from a source that was *actually produced this frame* — a culled/disabled Cloudscape pass fails the `IsResourceReachableForExtraction` check and the flag stays false. So `CloudsHistoryValid` is false for every frame the pass doesn't run, the `ImportHistory` call is guarded on it, and re-enabling always starts a fresh accumulation. This auto-invalidation is precisely why the TAA-pattern graph-tracked history was chosen (TAA itself relies on the same semantics when toggled).

**Duplicates** (bot re-raised round-1 findings, previously adjudicated on this PR):

- *CloudscapeRenderState tolerance-based `operator==`* — still skipped: the defaulted `==` is change-detection ("did the state change since last frame"), where bitwise-exact is the correct semantic — a tolerance compare would miss real small changes and break transitivity. All seven sibling settings structs in the file use the identical defaulted form, and the one explicit sibling (`UnderwaterFogState`) is explicit precisely to keep **bitwise** semantics (its comment says so).
- *m_PascalCase renames in CloudscapeRenderState / Renderer3DData / CommandDispatchData* — still skipped: the flagged members exactly match their immediate surroundings (bare-PascalCase struct families: WindSettings/FogSettings/UnderwaterFogState; Renderer3DData's `PostProcess`/`Fog`/`FogTime`; `CSMShadowTextureID`/`AtlasShadowTextureID`/`SnowDepthTextureID`). Renaming only the new members would make those structs internally inconsistent; per repo convention, surrounding-style consistency wins.

Issue #607, opened 2026-07-10
MCP: post-#357 follow-ups (script-tool tiers, live reload, C# tools) + capability-gap log


drsnuggles8:
Successor tracker to #357 (MCP protocol-conformance & tool-authoring roadmap, resolved by PR #608). Two standing jobs:

1. **Capability-gap log** — CLAUDE.md routes agents here: reach for an `olo_*` tool that doesn't exist yet? Log it under **§ Open capability gaps** below (tool name + what it should expose + why, one line) instead of silently working around it. For a *project-specific* diagnostic composable from existing read-only tools, prefer a Lua script tool (`RegisterMcpTool{…}` in `<project assets>/McpTools/`, see `docs/guides/mcp-diagnostics-server.md` § "Script-defined tools") — and still log it here if a native tool would serve every project.
2. **The deferred crumbs from the #357 omnibus**, each with its unblock condition.

> **📌 This post is the single source of truth — log new gaps by EDITING it, not by commenting.** The comment thread has now been cleared twice (2026-07-29, 17 comments; 2026-08-09, 4 comments) after drifting out of sync with this post. Both times every gap logged below the fold had already been superseded or needed folding up by hand, and a `/start-work` sweep wrongly concluded the tracker was dry from this post's status line alone. **A comment is not a log entry — it will be deleted.**

**Status:** all **17 capability gaps** re-verified on 2026-09-03 are implemented and under review in PR #1028; **0 of those 17 remain open**. The live tool surface is 91 built-in tools (93 in the verification project with two Lua tools), and the generated field registry now includes writable `TransformComponent.Rotation`. The deliberately deferred C# script-tool item below remains unchecked and is not part of the 17.

**Under review, 2026-09-03:** PR #1028 carries the complete 17-gap batch on `feature/mcp-capability-gaps-607`. Full Debug suite: 7,170 tests, 0 failures, 12 expected disabled; live MCP verification covered all 17 gaps. **This issue stays OPEN as a standing gap log** after the PR merges.

---
## 📋 Capability gaps — *implemented in PR #1028, pending merge*

[...]

PR #664, opened 2026-07-18
feat(mcp): map-key addressing for MorphTargetComponent + LightProbe/ReflectionProbe clamp migration (#607)

drsnuggles8:
## Summary

Closes the two remaining Section C items on #607:

- **`MorphTargetComponent::Weights` map-key addressing.** Its `std::unordered_map<std::string, f32> Weights` field had no MCP write path at all — a map's runtime key can't be a compile-time-known accessor chain. Added a new `MakeMapKeyField` mechanism (`McpGenericFieldWrite.h`): one registry entry per map field, addressed at write time by a **dotted runtime key** (`"Weights.Smile"`), resolved by a new `FindMapKeyed()` that strips the registered prefix — reusing the existing dotted sub-object-addressing convention rather than adding a second JSON argument. The write goes through `MorphTargetComponent::SetWeight`/`GetWeight` directly on the live component (never a whole-map copy+swap), so its `[0, 1]` clamp is inherited for free. Hand-registered in `McpFieldRegistry.cpp` (not codegen — it's the only map field in the engine today). `olo_entity_list_fields` expands the field into one discoverable dotted entry per *current* key.
- **`LightProbeVolumeComponent::Spacing` / `ReflectionProbeComponent::InfluenceRadius` clamp migration.** Both fields were registered as plain (unranged) MCP fields despite their hand-written `SceneSerializer.cpp` deserializers enforcing a strict `> 0` guard. Migrated both onto `OLO_SERIALIZE(Clamp, Min = 0.01f)`, so the MCP field registry inherits the range for free (no `kMcpFieldClamps` table entry needed). The hand-written `SceneSerializer.cpp` guards were updated from reject-and-reset-to-default to clamp-to-floor to match the annotation's semantics — a small-but-positive value is now preserved instead of silently replaced by the constructor default. Both components stay hand-written in the scene serializer for unrelated reasons (cross-field invariants a single-field `Clamp` can't express), so only the two named fields' range semantics changed, not the whole components' (de)serialization path.

## Test plan

- [x] `McpFieldRegistry.*` (36 tests, including 7 new map-key tests: round-trip/undo, independent keys, range clamp, bare-prefix refusal, `ListFields` expansion, empty-map handling)
- [x] `ComponentSerializerCoverage.*`, `ComponentTupleCoverage.*`, `ComponentHandlerCoverage.*`, `SaveGameComponentSerializerCoverage.*`
- [x] `*LightProbe*:*ReflectionProbe*` (17 tests, including the GPU visual-evidence test and YAML/Lua/save-game round-trips)
- [x] `GenerateBindings` rebuilt; confirmed `McpFieldRegistry.Generated.inl` emits ranged entries for both migrated fields

## Summary by CodeRabbit

* **New Features**
  * Added support for editing individual morph-target weights using dotted field names such as `Weights.Smile`.
  * Field discovery now lists available map keys and their supported ranges.
  * Morph-target weight updates support undo and report the applied value.

* **Bug Fixes**
  * Weight values are clamped to the valid range of 0–1.
  * Probe spacing and influence radius values now enforce a minimum of 0.01 while preserving safe defaults for invalid data.

* **Documentation**
  * Updated MCP guidance with map-key editing syntax, discovery behavior, and range rules.

### DIFF
--- a/OloEditor/src/MCP/Generated/McpFieldRegistry.Generated.inl
+++ b/OloEditor/src/MCP/Generated/McpFieldRegistry.Generated.inl
@@ -347,7 +347,7 @@ static void BuildRegistryChunk4(std::vector<FieldEntry>& registry)
     // LightProbeVolumeComponent
     registry.push_back(OLO_GFW_FIELD(LightProbeVolumeComponent, "BoundsMin", m_BoundsMin));
     registry.push_back(OLO_GFW_FIELD(LightProbeVolumeComponent, "BoundsMax", m_BoundsMax));
-    registry.push_back(OLO_GFW_FIELD(LightProbeVolumeComponent, "Spacing", m_Spacing));
+    registry.push_back(OLO_GFW_FIELD_RANGE(LightProbeVolumeComponent, "Spacing", m_Spacing, OLO_GFW_BOUND(0.01f), OLO_GFW_NO_BOUND));
     registry.push_back(OLO_GFW_FIELD_RANGE(LightProbeVolumeComponent, "Intensity", m_Intensity, OLO_GFW_BOUND(0.0f), OLO_GFW_NO_BOUND));
     registry.push_back(OLO_GFW_FIELD(LightProbeVolumeComponent, "Active", m_Active));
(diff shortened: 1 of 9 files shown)


drsnuggles8:
@coderabbit review

PR #667, opened 2026-07-19
feat(mcp): olo_editor_select_entity — select/clear Scene Hierarchy selection over MCP

drsnuggles8:
## Summary

Implements the highest-leverage open capability gap logged in #607: there was no way to drive editor selection programmatically, so the Properties inspector's whole `DrawComponent<T>` surface was unreachable for MCP screenshot verification. `olo_input_inject` can click a viewport entity but can't reliably land a Scene Hierarchy *panel row* click — the OS cursor reasserts over the synthetic position between injected frames.

- New consented write tool **`olo_editor_select_entity`**: select an entity by `entity` (UUID, string or number) or `clear` the selection — exactly one of the two.
  - Not routed through `CommandHistory`/undo — selection is editor UI state, not project data.
  - An unknown `entity` uuid leaves the **current** selection untouched and reports `ok:false`, rather than silently clearing it.
  - The result always reads back the **actual** post-call selection (`selected`/`entity`/`name`), and a new `changed` field distinguishes a real transition from a no-op (matches the tool's `idempotentHint:true`).
- `OloEditor/src/MCP/McpSelectEntity.h` (new): engine-free schema + arg parsing + result shaping, following the `McpSceneControl.h` split so it's unit-testable without pulling in an editor TU. Reuses the shared `McpToolsCommon::ParseUuid` instead of a third near-duplicate UUID parser.
- `McpServer.h` gains `McpSelectEntityResult` + the `SelectEntityInEditor` context hook; `EditorLayer` wires it to `SceneHierarchyPanel::SetSelectedEntity`/`ClearSelection`.
- 36 unit tests: dispatch write-gate, inputSchema enforcement, pure `ParseArgs`/`ToJson`, including a pin that a **numeric** `entity` isn't rejected by the JSON-Schema layer before the parser (which accepts one) ever sees it.
- `docs/guides/mcp-diagnostics-server.md` updated (tool table, toolset list, write-consent list, new walkthrough section). `#607`'s capability-gap checklist promoted this item to Shipped.
- `docs/agent-rules/build-trees-and-windows-asan.md`: documents a fresh-worktree vendor-bootstrap stall (GameNetworkingSockets/WebRTC clone wedged by an interrupted earlier attempt) and the robocopy-from-a-sibling-worktree workaround, hit while standing up this branch's build.

## Review

Ran `/code-review` (8-angle high-effort pass: line-by-line, removed-behavior, cross-file tracer, reuse, simplification, efficiency, altitude, CLAUDE.md conventions) against the diff before opening this PR. Findings applied:

- **UUID-parsing triplication** — `McpSelectEntity.h` had its own copy of UUID-from-JSON parsing, near-identical to `McpToolsCommon::ParseUuid`. Now delegates to the shared helper instead.
- **Schema/parser type mismatch** — the `entity` property was declared `type: "string"` only, so a numeric `entity` (which the parser explicitly supports, and a unit test asserted) would actually be **rejected by inputSchema enforcement before the handler ever ran** — a real, silent gap between documented and actual behavior. Fixed locally (widened this tool's own `entity` schema to `["string","number"]`) without touching the shared `Schema::EntityId()` helper other tools depend on (its exact JSON shape is pinned by `McpSchemaBuilderTest.cpp`).
- **Dead post-call `Available` re-check** in the handler — removed; the real hook never returns `Available:false`, and no sibling handler (`Handle_SceneOpen`) does this re-check either.
- **Branching / duplicated read-back logic** in `EditorLayer::SelectEntityInEditor` — flattened to a single source-of-truth read-back of the panel's selection at the end of the function (used for both the success and failure paths), instead of duplicating the "convert `Entity` → (id, name)" logic per branch. This also made adding the `changed` field a one-line diff (before/after uuid comparison).
- Double JSON-object lookup for `'clear'` in `ParseArgs` collapsed to a single `args.find(...)`.

A few findings were surfaced but deliberately **not** applied (out of scope for this diff — pre-existing code in unrelated files, or matching established codebase convention): `McpGenericFieldWrite.h`'s separately-shaped (stricter) UUID parser, the "exactly one of two args" validation pattern being hand-rolled at ~5 call sites codebase-wide, and the shared `Schema::EntityId()` helper itself still being `"string"`-only for every *other* consumer (`olo_scene_get_entity`, `olo_camera_frame_entity`, etc.) — a real, wider pre-existing gap, but touching that shared, already-shipped/pinned helper is a separate change.

## Test plan

- [x] `cmake --build build --target OloEngine-Tests --config Debug` — clean build.
- [x] Full `Mcp*` gtest filter: **507 tests, 506 passed, 1 skipped by design** (`McpHeadlessAttachTest.HostUntilDetached`, gated behind an env var not set in this run).
- [x] New `McpSelectEntityTest.cpp` (36 tests) covers: write-gate on/off, inputSchema enforcement (unknown property, missing both `entity`/`clear`, both given), numeric vs. string `entity`, `clear`, unknown-uuid `ok:false` shaping, and the `changed` field.
- [ ] Live-editor MCP verification (select an entity, screenshot the Properties panel) needs a human at the editor to enable "Allow writes" — the consent gate has no headless bypass by design (see `docs/agent-rules/mcp-setter-based-field-registry.md` §5). Not exercised in this PR; the pure core + dispatch seam are what's testable headlessly.

Closes one item of #607 (tracker stays open — 4 other capability gaps remain).

## Summary by CodeRabbit

* **New Features**
  * Added an MCP tool to select an entity in the editor’s Scene Hierarchy or clear the current selection.
  * Selection updates are reflected in the Properties panel and report the resulting entity, name, and whether the selection changed.
  * Unknown entities leave the current selection unchanged.
  * Tool access is controlled by the editor’s Agent writes setting.

* **Documentation**
  * Added usage, validation, response, and consent details for the new tool.

* **Tests**
  * Added coverage for selection, clearing, validation, gating, idempotency, and invalid entity handling.

### DIFF
--- a/OloEditor/src/EditorLayer.cpp
+++ b/OloEditor/src/EditorLayer.cpp
@@ -566,6 +566,12 @@ namespace OloEngine
             { return QueueMcpInput(plan); };
             mcpContext.GetInputState = [this]() -> MCP::McpInputStateSnapshot
             { return GetMcpInputState(); };
+            // olo_editor_select_entity (#607): select/clear the Scene Hierarchy
+            // panel's selection so the Properties inspector draws the requested
+            // entity — see SelectEntityInEditor for the UUID resolution + the
+            // "leave the current selection untouched on a bad uuid" contract.
+            mcpContext.SelectEntityInEditor = [this](u64 entityUuid, bool clear) -> MCP::McpSelectEntityResult
+            { return SelectEntityInEditor(entityUuid, clear); };
             mcpContext.GetFrameIndex = [this]() -> u64
             { return m_FrameIndex; };
             mcpContext.IsCaptureUnready = [this]() -> bool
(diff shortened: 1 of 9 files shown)


drsnuggles8:
@coderabbitai review

PR #700, opened 2026-07-30
refactor(rhi): sweep raw GL call sites out of OloEngine/src — Phase 2 step 2 (#691)

drsnuggles8:
Closes the headline target of **Phase 2** of #691: the raw OpenGL call sites are gone from `OloEngine/src/**`, and the translation units there can no longer name a GL symbol at all.

Step 1 (#698) converted the facade's *vocabulary*. This is step 2 of 3 — the call-site sweep. **Step 3 is deliberately not here** (reasons at the bottom), so Phase 2's checkbox stays unticked.

## Counters

| Counter | Phase 1 | After #698 | **This PR** | Target |
|---|---:|---:|---:|---|
| `sweep_gl_calls` | 313 | 313 | **0** | 0 (Phase 2) ✅ |
| `sweep_glad_includes` | 70 | 39 | **0** | 0 (Phase 2) ✅ |
| `tools_gl_calls` | 236 | 236 | 236 | 0 — Phase 8, *relocation* not exemption |
| `debug_escape_hatch` | 0 | 0 | 0 | must stay 0 |

`sweep_glad_includes` is the one that proves something: a TU that cannot see `<glad/gl.h>` provably cannot name a GL symbol. That is compiler-enforced, not a grep. `rhi_boundary_baseline.json` is updated in the same commit, as the ratchet requires.

## The finding: the facade wasn't just GL-typed, it was *incomplete*

ADR 0011 §1.7 framed Phase 2 as "strip the `GLenum`s, then sweep". That undersold it by about half.

**84 distinct GL entry points** appear across the 313 call sites, and **~60% had no `RendererAPI` equivalent at all** — so closing the sweep needed **~60 new virtuals** against an existing 74. Whole categories had never been abstracted, so every pass reached past the facade to do them:

| Category | New virtuals | Why it had no entry point |
|---|---:|---|
| Buffer binding points (`glBindBufferBase`) | 2 | Biggest single gap — 26 call sites |
| Raw buffer lifecycle | 9 | `VirtualMeshRegistry` hand-rolls an arena + persistent-mapped upload ring |
| Named-framebuffer state | 10 | `SetDrawBuffers` existed but only for the *bound* FBO; every site names one via DSA |
| Queries | 7 | `BeginConditionalRender` existed; the pools feeding it did not |
| Fences | 4 | `FrameResourceManager` used `GLsync` directly |
| Draws from bound geometry | 4 | `CommandDispatch` keeps a redundant-bind cache the `*Raw` family would defeat |
| Program / VAO / FBO binding | 5 | The POD dispatcher holds a `u32`, not a `Ref<Shader>` |
| Texture clear / offset upload / readback / dims / barrier | 8 | — |
| VAO lifecycle, debug groups, device idle, MSAA caps, separate blend, front face, clear depth, patch count | 11 | — |

**Generalisable, and worth carrying into Phase 5:** an abstraction's completeness is measured by how many distinct *operations* the layer above performs, not by how many call sites it already serves. 74 virtuals looked thorough while 60 operations went around it, because each was rare enough (1–3 sites) to read as a special case. Frequency wasn't the signal either — `glBindBufferBase` had 26 sites and was still missing.

## ADR 0011 amendments (5)–(10)

- **(6) `RHI::NoAttachment`** — a "this draw slot writes nowhere" sentinel. `DecalRenderPass` steers a decal into exactly one G-Buffer attachment using `GL_NONE` entries, which an attachment *index* cannot express. Both backends need it (`GL_NONE` / `VK_ATTACHMENT_UNUSED`); only the neutral layer lacked it.
- **(7) `glGetError` disappears** rather than being abstracted. GL's error model is a sticky global flag, Vulkan's is a per-call result — exposing either forces the other backend to fake it. The readback virtuals return `bool` instead.
- **(8) `DrawBoundIndexed` / `…Instanced` / `DrawBoundArrays`** draw from *previously bound* geometry. That is the **native Vulkan shape** (`vkCmdBindIndexBuffer` + `vkCmdDrawIndexed`), not a GL-ism — the combined bind-and-draw `*Raw` form is the less portable one. They also carry `RHI::PrimitiveTopology` / `RHI::IndexType` explicitly instead of hard-coding triangles + `u32`.
- **(9) One recorded, deliberate debt.** `SetProgramUniformFloat` is the single virtual a Vulkan backend **cannot implement faithfully** — SPIR-V has no name-queryable default uniform block. One call site (`DrawInfiniteGrid`'s `u_GridScale`). The alternatives were a shader change (forfeits golden-image parity, this branch's only safety net) or reaching for `Shader` (the POD dispatcher holds a `u32` by design). Phase 6 folds it into a UBO and deletes it. Flagged now rather than surprising Phase 7 bring-up.
- **(10) `MemoryResidency` moved** from `RHIResources.h` to `RHITypes.h` — see below.

## Two boundary leaks the ratchet cannot see, also closed

1. **A `Platform/<Backend>/` include leaks exactly as much as `glad/gl.h`.** Three passes (`FluidIntermediatesPass`, `WaterRenderPass`, `VirtualMeshRegistry`) included `Platform/OpenGL/OpenGLUtilities.h` for `GLClearProgramGuard`. Deleting their *direct* glad includes would have driven `sweep_glad_includes` to zero while all three TUs could still name every symbol in OpenGL — the exact counter-gaming the baseline's own `_comment` warns about, arrived at honestly. The fix is a layering one: the guard is an NVIDIA driver hazard, so it moved **into** the backend clear implementations, where `ClearDepthOnly()` had been carrying it correctly all along. The passes now get the behaviour for free.

2. **A loader-symbol probe is invisible by construction.** `RGCommandContext` guarded its debug markers with `if (GLAD_GL_KHR_debug)` — which does not match `gl[A-Z]` and so never appeared in any count. The capability check moved into `OpenGLRendererAPI::PushDebugGroup`. Same class as step 1's `glad_glCreateTextures != nullptr` context probe.

**A third near-miss is recorded because the catch was luck.** The sweep started inventing `RHI::BufferUsage { DynamicDraw, DynamicCopy, DynamicRead }` — a straight transcription of GL's hints. Phase 1 had *already* designed that concept, better, as `RHI::MemoryResidency { DeviceLocal, HostToDevice, DeviceToHost }`, sitting beside `BufferDesc` where nothing could reach it. What surfaced the duplication was a **name collision** with `RHIResources.h`'s bind-flag `BufferUsage` — and it only fired in the **test** build, since the engine library never includes that header. Rename either enum and a duplicate concept ships silently. Lesson written up: read an earlier phase's declaration-only header for the *vocabulary you are about to invent*, not just the types you consume.

## Testing

- **Full suite: 5195 run / 5189 passed / 0 failed / 6 skipped.** Goldens and visual-evidence tests **executed**, not skipped; `OLOENGINE_GOLDEN_REBASE` unset throughout. The 6 skips are the same environment-dependent ones as #698 (app-launch smoke ×3, video fixture, real-asset cook, MCP headless attach).
- **New tests.** `RHIEnumLoweringTest` gains 6 cases for the new enums, each with the last-ordinal `static_assert` tripwire amendment (4) established — including one asserting the `NoAttachment` sentinel cannot be folded into attachment-0 arithmetic (`GL_NONE` is not `GL_COLOR_ATTACHMENT0 + anything`, and collapsing it would leave the *previous* draw-buffer list live).
- **This PR changes no pixels.** Nine evidence PNGs regenerated; each was pixel-diffed against `HEAD` **after** measuring the run-to-run noise floor on the identical binary. Every delta sits at or below its own noise:

  | PNG | vs HEAD (mean) | noise floor | verdict |
  |---|---:|---:|---|
  | `EASU_GTAONative` | 0.4469 (max 86) | **0.4469 (max 86)** | identical — pure noise, untouched code |
  | `Fluid_Waterline` | 0.0021 | 0.0020 | same magnitude |
  | `Fluid_Side` | 0.0011 | 0.0014 | *below* noise |
  | `Fluid_ThreeQuarter` | 0.0008 | 0.0009 | below noise |
  | `VirtualGeometry_Debug_ClusterId` | 0.0006 (max 158) | 0.0009 (**max 158**) | the max-158 spike reproduces in the control |
  | `OcclusionCull_*`, `WorldOriginRebase`, `CameraRelative`, `SphereAreaLight` | ≤0.0003 | ~0 | below every threshold |

  The `max=158` on the cluster-ID debug image was the alarming one; it appears *identically* in the no-change control, so it is GPU-order-dependent cluster-ID colouring, not a regression. All nine restored.
- **Live editor.** Renders correctly; **zero shader compile/link errors and zero `GL_INVALID_*`** (exactly what a wrong enum lowering or a bad draw-buffer list would produce). `olo_render_frame_breakdown` shows the rewritten dispatch path (`DrawMeshInstanced` / `DrawSkybox`) executing with real GPU timings.

### One verification gap, stated plainly

I could not exercise **virtual geometry** or **deferred lighting** in the *live editor*. Both require the Deferred path; that path is not scene-serialized (the `VirtualGeometrySponza.olo` header documents this and says to switch it by hand after load), and the MCP `olo_renderer_settings_set` write gate is off by default and cannot be unlocked non-interactively. Both paths are covered by the suite's evidence/golden tests, which executed and matched HEAD within noise. Worth a human eyeball on `VirtualGeometrySponza.olo` + `SponzaDeferred.olo` with the path switched to Deferred.

### One behaviour delta worth knowing about

`VirtualGeometryPass` set depth state with raw `glEnable(GL_DEPTH_TEST)`, which left `OpenGLRendererAPI::m_DepthTestEnabled` stale — and `Clear()` derives its `GLbitfield` from that member. Routing the pass through the facade makes that cache truthful, so a later `Clear()` can now clear depth where it previously did not. That is a fix, but it is rendering-visible; the golden/evidence tests show no resulting pixel change.

Profiler counters also move slightly: the facade's state setters bump `RendererProfiler::StateChanges` where the raw calls did not. Draw counters deliberately do **not** move — the new `DrawBound*` family does not touch `RendererProfiler`, because the call sites it replaces never did and several keep their own `CommandDispatch::Statistics`.

## Why step 3 is not in this PR

Step 3 is `TransientPool`'s bare `RendererID` / `GetRendererID()` split three ways per ADR 0011 §1.1/§1.3/§1.4. Three reasons it is separate rather than an unjustified slice:

- **`RHI::ResourceHandle` has no producer.** `RHIResources.h` says so in its own header comment, and `GetNativeHandleForDebug` is declared but defined nowhere. Step 3 needs a generation-checked handle-minting layer built first.
- **`GetRendererID()` has ~320 call sites**, not a handful (`Scene.cpp` 56, `Renderer3DMeshSubmission.cpp` 33, …).
- **Part of it is Phase-3-entangled.** ADR 0011 §1.3 says `BindTexture` "does not get abstracted, it **disappears**" — under heap-bindless, which is Phase 3's `ARB_bindless_texture` rehearsal.

This matches how #691 is deliberately structured (phase-at-a-time PRs; step 1 shipped alone as #698).

## Summary by CodeRabbit

* **New Features**
  * Expanded renderer support for framebuffer operations, resource management, GPU queries, fences, synchronization, and debug markers.
  * Added broader multisample and rendering capability detection.
* **Bug Fixes**
  * Improved GPU readback failure handling to prevent invalid data from being used.
  * Improved synchronization, attachment handling, and state restoration for more consistent rendering.
* **Documentation**
  * Updated renderer abstraction guidance, coverage details, and migration progress documentation.


drsnuggles8:
@coderabbitai review

drsnuggles8:
Worked through all 14. Thanks — two of these were genuinely load-bearing, and one caught a bug I introduced.

**Fixed as suggested**

- **`UIRenderer.cpp` contradictory comment** — this was a real bug of mine, not just prose. A bulk `GLint`→`i32` regex rewrote the comment's own counter-example into *"i32/u32, not i32/u32"*, and a `#`→`//` substitution turned `#691` into `//691`. Rewritten.
- **Duplicated identity draw-buffer builder** (the Major one) — you were right that this wanted extracting, and it turned out the facade was simply missing the method. Added `RendererAPI::RestoreAllFramebufferDrawAttachments(framebufferID, count)`, the named-framebuffer counterpart of the `RestoreAllDrawBuffers(u32)` that already existed for the *bound* framebuffer. Nine open-coded `std::array` + fill-loop + span sites collapse to one line each, across `SceneRenderPass`, `DeferredLightingPass` (×3), `ForwardOverlayRenderPass`, `GPUDrivenOcclusionPass`, `DeferredGPUOcclusionPass`, `VirtualGeometryPass` and `GBuffer`.
- **Three readback sites logged failure but used the buffer anyway** — correct, and that was a half-fix on my part. Each now acts: DDGI returns early and retries the probe next frame (the update is amortized); `OceanFFTGpu::DebugInverseFFT2D` returns empty. `LightProbeBaker` was the one that mattered most, since those SH coefficients are *persisted* into `LightProbeVolumeAsset` — `RenderCubemapAtPosition` is now `[[nodiscard]] bool`, and `BakeProbeAtPosition` reports the probe through its existing `outValid` out-param rather than writing SH derived from undefined pixels to disk.
- **Cached patch-vertex cap left two per-call queries behind** — right, I added the cache and didn't use it at the two pre-existing sites. Both now read `m_MaxPatchVertices`.
- **Mock: record indices, not just the count** — agreed; `DecalRenderPass`'s four modes all pass 5 entries, so a count cannot distinguish them. `RecordedCall` gained a `ParamU32List`.
- **Mock: `CreateFence()` returning 0** — good catch, that made the failure branch the only one tests could reach (and made `FrameResourceManager` log an error on a healthy test). Returns an incrementing non-zero handle now.
- **`measured_on_commit` recorded a working tree** — now pinned to `11d7cea7`.
- **ADR/agent-rules/header arithmetic** — you're right that the entry-point count and the virtual count are different quantities and shouldn't be divided into each other. Corrected at all three sites to state both explicitly: **84** distinct entry points, **54** of them with no facade equivalent, **60** new virtuals — with a note that an entry point can expand into several (`glClearTexImage` → float + uint clears, mirroring `VkClearColorValue`'s union; the readbacks each gained a `bool` return). 54 is the gap, 60 is the fix.
- **Wrong internal cross-reference** — correct, the last-ordinal guard is the "One new guard" paragraph, not amendment (4) (which is `UploadTextureSubImage2D`'s source-buffer format). Fixed.
- **"the last four files standing" named only one** — fixed by naming all four and the distinction between them: `BloomRenderPass.cpp`, `OITResolveRenderPass.cpp` and `ShaderPack.cpp` named no `GL*` identifier so their includes just fell out; `UIRenderer.cpp` was the one that needed retyping first.

---

**Two I fixed differently, with reasoning**

**1. `default:` on the `RHI::FenceStatus` switch.** The concern is right — falling through to `return true` for an unrecognized status is the wrong direction for a primitive gating reuse of double-buffered GPU resources. But adding `default:` inside the switch would *suppress* the compiler's exhaustiveness warning, which is what actually catches a new enumerator at build time; it trades a build error for a silent runtime one.

So I kept the switch exhaustive with no `default:`, moved the success cases to `return true` directly, and made the post-switch fallthrough the backstop — **failing closed** with an error log. That satisfies the safety concern and keeps the compile-time check. Same idiom the lowering switches in `OpenGLRHIConversions.h` already use (fallback *after* the switch, not a `default:` label inside it).

**2. Enum tripwires don't detect appended members.** This is correct and it's the most useful finding in the review — but chasing it turned up something worse than the gap itself.

Appends *are* caught, just not by the `static_assert`: **16 of the 19 lowering switches deliberately carry no `default:` label**, so clang's `-Wswitch` errors on an unhandled enumerator. Meanwhile the header comment on `OpenGLRHIConversions.h` claimed *"Every switch carries a `default:` that logs and falls back"* — which was false for those 16, and anyone "fixing" the code to match that comment would have silently removed the only guard against appends, engine-wide.

So rather than change the assertions I documented the actual division of labour, in the test, the ADR, and that header comment:
- last-ordinal `static_assert` → catches **insert / remove / reorder**
- absence of `default:` + `-Wswitch` → catches **append**
- MSVC's C4062 is off by default even at `/W4`, so the **clang-cl CI job is what enforces it** — worth knowing before anyone drops that job
- the three switches that *do* have `default:` (`ToGLPixelFormat`, `ToGLPixelType`, `ToGLImageAccess`) are intentional: their enums have many members that are legitimately invalid for those conversions, so exhaustiveness there would be noise rather than a guard

I considered a `Count` sentinel per enum, which would make the `static_assert` catch appends directly, and rejected it: it makes an invalid value representable in the neutral vocabulary (`ToGL(Count)` becomes well-formed) and forces a dead `case` in every lowering switch. Happy to revisit if you disagree.

---

Rebuilt clean; full suite **5195 run / 0 failed / 6 skipped** (same environment-dependent skips), and the RHI ratchet still reports **0 / 0 / 236 / 0**.

drsnuggles8:
Good catch, and a fair one — that helper contradicted its own declaration comment. The comment warns that restoring a narrower list than the target actually has silently drops later fragment outputs, and then the implementation clipped to a 16-entry stack array whenever `GL_MAX_DRAW_BUFFERS` was larger.

Fixed in `2d229261`: above 16 it allocates rather than clips, so `GL_MAX_DRAW_BUFFERS` is the only ceiling — and that one already clamps *with a warning*. The stack path still covers every framebuffer in the engine (the G-Buffer is the widest at 5), so this costs nothing in practice; it just stops the cap being silent.

Rebuilt clean; RHI ratchet + enum lowering (21 tests) and the render-graph / deferred / G-Buffer / scene-pass / occlusion / framebuffer / visual-evidence selection (494 tests) all pass.

PR #732, opened 2026-07-31
feat(rhi): generation-checked RHI::ResourceHandle — the mint and the first migrations (#691 Phase 2 step 3)

drsnuggles8:
Part of #691 Phase 2 step 3 (ADR 0011 §1.1/§1.3/§1.4): replace the bare `u32 RendererID` currency with a generation-checked `RHI::ResourceHandle`.

**#691's Phase 2 checkbox stays UNTICKED and #691 stays open.** This is the mint plus the first real call-site migrations, not the completed step.

## Why the generation matters

GL recycles object names. `Texture::operator==` compares `GetRendererID()`, so two *different* objects can compare equal. That is not hypothetical here — `CommandDispatch`'s own comments record a debugged production bug of exactly this shape: `VirtualGeometryPass` binds the Hi-Z pyramid to texture unit 0 for its cull compute, unit 0 is also `u_AlbedoMap`, and any material whose albedo GL name matched the stale bind-cache entry *silently sampled HZB depth as its albedo*.

## What's here

| | |
|---|---|
| `c9b48f29` | The mint — `RHI::ResourceRegistry` (lock-free chunked slots, seqlock-style generation read), `ScopedResourceHandle`, and the `GetNativeHandleForDebug` that Phase 1 declared but never defined. 3 new ratchet counters, ADR amendments (11)–(17). |
| `153e5617` | Resource classes mint `GetRHIHandle()` *alongside* `GetRendererID()` — 9 interfaces, 10 backend classes, 26 `Sync()` sites. |
| `127a17da` | Handle-taking siblings for the 7 bind virtuals. |
| `81264f14` | Framebuffer attachments mint their own identities (an attachment is a separate GL object from its framebuffer). |
| `e8816d76` | Handle-returning raw facade creators. **L0 complete: every producer can mint.** |
| `cf3c6612` | `RenderGraph` carries both currencies as *alternatives*, never both. |
| `fc30e44c` | Renames the render graph's `ResourceHandle` → `RGResourceHandle` after a near-miss where a missing include let `RHI::ResourceHandle` silently degrade to a different in-scope type. |
| `923c5276` | **First real call-site migration** — SSAO end-to-end. |
| `f914bfcb` | Unrelated build fix (see below). |
| `046527c0` | Scoping finding for the next slice. |

## The migration is ordered by dataflow, not by size

`handle -> native` is a registry lookup; **`native -> handle` is not recoverable**. So a symbol can only take handles once every producer feeding it hands out handles. Two slice attempts were made and reverted for violating this — converting `HZBGenerator`'s output alone would have made `GetHZBTextureID()` return null whenever an external HZB was set, silently breaking GTAO while compiling green.

That is also why SSAO moves as a *whole chain* in one commit (create → configure → import → bind → delete) rather than layer by layer.

## Two defects found and fixed en route

- **`Sync()` retired a live handle on in-place reload.** A texture hot-reload transiently reports native id 0; the first implementation treated that as a release. Caught by `RHIHandleNativeIdentity.HandleSurvivesInPlaceReloadAndTracksTheNewGLName`.
- **Generation bumping was silently disabled for handle imports.** `ImportTextureHandle` delegated to `ImportTexture(name, 0u, …)` and stamped the identity on afterwards, so `AllocateTextureHandle`'s change test compared `0 != 0` forever — re-importing a name with a different texture kept the slot generation and every cached `RGTextureHandle` resolved to the new occupant. The recycled-name hazard, recreated inside the fix for it. Identity is now threaded in as a defaulted parameter so one change test covers both currencies. Pinned in both directions by two new `RenderGraph` tests.

Neither was reachable from the migrated pass (SSAO creates its noise texture once and never recreates it) — both were found by reading, which is worth knowing for the remaining slices.

## Verification

- **5220 passed / 0 failed**, three consecutive full-suite runs. 6 skips, all environmental.
- Goldens executed with `OLOENGINE_GOLDEN_REBASE` unset.
- Evidence PNGs pixel-diffed against HEAD **after** measuring the run-to-run noise floor on the identical binary. SSAO's own six are byte-identical. Six others drift; four are at or below their own floor, and the two that aren't (8/255 over 0.001% of pixels; 6/255 over 0.000%) were last regenerated 71 commits ago and belong to passes that call none of the migrated entry points. Reverted rather than committed.
- Live editor: SSAO compiles and links, 18 entities deserialized / 0 failed, no errors in `OloEngine.log`.

## Ratchet

`sweep_renderer_id` **700 → 699**, measured and locked into the baseline. `facade_native_id_params` stays **68** — correct, since this adds handle overloads without deleting any u32 form; that counter only falls as each u32 form loses its last caller. `sweep_gl_calls` and `backend_resolve_hatch` remain **0**.

Worth stating plainly: `sweep_renderer_id` matches the *accessor name* `RendererID`, not the currency, so it is a poor progress meter for this work and a regression ratchet rather than a burn-down.

## Also included: an unrelated build fix

`f914bfcb` fixes `cmake/vendor/OpenUSD.cmake` never passing `-DCMAKE_INSTALL_PREFIX`. `cmake --install --prefix` only redirects *relative* destinations, so USD baked `C:/Program Files (x86)/usd` into the absolute install rule for `pxrConfig.cmake` and the install died without elevation. It only fires on a cache miss, and USD builds once per machine, so it lay dormant until a cache wipe — then surfaced as `LNK1181` on a missing `install/Debug/lib/usd_m.lib`, because the Release failure aborted the step before Debug installed. Unrelated to #691 but it blocked all work here.

## What's next (not in this PR)

`046527c0` documents it: the `Renderer/Commands/` bind cache is **one indivisible unit** — material textures, four shadow paths, snow and cloud shadow all key the same `BoundTextureIDs` array, so `PODMaterialData` cannot migrate alone. Every producer involved already mints, so it's unblocked, just large (~180 errors from the field-type change). That migration is what turns the `InvalidateTextureSlot`/`InvalidateTextureBinding` workaround above from load-bearing correctness into a pure optimisation.

## Summary by CodeRabbit

* **New Features**
  * Added generation-checked handles for major rendering resources.
  * Added safe handle-based resource creation, binding, configuration, deletion, and render-graph integration.

* **Bug Fixes**
  * Improved stale-resource detection during reloads, resizing, and deletion.
  * Corrected resource-kind handling across rendering workflows.

* **Documentation**
  * Added guidance on resource identity, migration, and lifecycle safety.

* **Tests**
  * Expanded coverage for resource lifetime, identity, concurrency, and render-graph behavior.


drsnuggles8:
Worked all 9 findings — 8 applied, 1 applied more broadly than suggested. Pushed in `93eff464` + `a0ff3ba6`.

**Two were real defects. Thank you — both were latent rather than active, which is exactly when they're cheapest to fix.**

1. **`ResolveTextureHandle` didn't follow `m_VersionAliasTargets`** — correct, and the second-order point in the finding is the important one: the null handle already means "imported as a native id", so a versioned read returning null would have been *indistinguishable at the call site* from the documented case. Mirrored the native resolver's iterative depth-capped walk (a malformed alias map must not recurse the stack to death).

2. **Seqlock acquire fence** — correct on the memory model. The `before` acquire load stops later reads floating above it; nothing stopped the relaxed `Native`/`Owner` reads sinking *below* the `after` load, so the tearing check could validate a generation pair while returning a payload read from outside it. `std::atomic_thread_fence(acquire)` added between the payload and the second load.

**Hardening, both accepted:**

3. **`Sync` gating on `IsValid()`** — right, that's a structural test on the handle's own bits and says nothing about registry liveness. After a `Clear()` every handle still reads valid while its slot is gone. Now gates on `IsLive` and re-adopts.

4. **Kind validation** — applied wider than the suggestion. A helper alone would have been dead code, so I added `ResourceRegistry::GetKind` (same seqlock discipline as the native resolve) plus `Utils::ResolveNativeAs`, and routed **all 14** typed call sites in the GL backend through it. The reason this matters more than it first looks: GL object names are per-type, so a texture and a buffer can both legitimately be name 1 — a handle handed to the wrong family resolves to a real, valid, unrelated object, and the generation *cannot* catch it because both handles are live. Mismatch returns 0, which lands on the same benign degradation the unchecked form already documents.

**Documentation findings — all fair, and they caught prose my own commits made stale.**

Rather than just bump 700 → 699, I marked the ADR paragraph as dating from part 1 (it accurately describes what part 1 shipped) and added the caveat that matters more than the number: **`sweep_renderer_id` matches the accessor *name* `RendererID`, not the currency.** The first fully-migrated resource — SSAO's noise texture, five native-id call sites — moved it by exactly 0. It's a regression ratchet, not a burn-down; `facade_native_id_params → 0` is the completion criterion. The defect-taxonomy and qualifier-fallback corrections are applied as described.

**Verification:** 5226 tests ran, **5222 passed, 0 failed**, 4 environmental skips — two fewer than the pre-review runs, because building `OloEditor` (which `add_dependencies` on `OloRuntime`) let both `AppLaunchSmoke` launch tests execute instead of skipping.

One deliberate non-change: `a0ff3ba6` backs out 10 evidence PNGs that `git add -A` swept into the previous commit. Diffing two runs of the *identical* binary during slice 5's verification showed those files moving by up to 158/255 with no code change, so they're run-to-run noise; committing them would imply a visual change that didn't happen.

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
Second review pass: all 3 findings verified as still valid and fixed — none skipped. Pushed in `91845e50`.

**1. `Delete*` retiring another family's registry entry — a regression my own kind-check introduced.** Confirmed and the most serious of the three. `ResolveNativeAs` correctly refused to hand back the wrong family's GL name, but `Unregister` isn't kind-aware, so the delete killed the other resource's *identity* while leaving its *GL object alive*. That's strictly worse than the unchecked form it replaced — adding the check made the two halves disagree. All four overloads now skip both halves together.

Pinned by a new test that hands a **buffer** to `DeleteTexture` — GL names are per-type, so that buffer's name is very likely also a live texture's, which is exactly the confusion the check exists to stop. It asserts the buffer's entry, kind and native name all survive, *and* that a correctly-typed `DeleteBuffer` still retires it; a guard that made the wrong case safe by breaking the right case would otherwise pass.

**2. Warning fired on stale handles.** Confirmed. `KindOf` returns `Unknown` for a stale handle, which is `!= expected`, so every legitimate use-after-free degradation — the documented benign path where resolving to 0 means binding 0 unbinds — would have logged. Both this and finding 1 now share one predicate, `Utils::IsWrongKind`, true only for a **live** handle of another family.

**3. `GetKind` duplicated the existing `KindOf`.** Correct, and this one's straightforwardly my error — `KindOf` was already there with 8 test callers, and the header's own thread-safety comment names it. I added a second function doing the same job without checking. Deleted mine; callers untouched.

Taking the second option in the finding (harden the existing one) rather than migrating callers, because `KindOf` needed hardening regardless: its body was `IsLive()` followed by a relaxed read — a TOCTOU where the slot can be `Unregister`ed and `Register`ed to a new tenant between the check and the load, so a **stale handle would be told the new occupant's kind**. It now uses the same seqlock discipline as `ResolveTaggedForBackend` (generation either side of the payload, acquire fence between).

**Verified:** 5227 tests ran, **5223 passed, 0 failed**, 4 environmental skips.

PR #740, opened 2026-08-02
feat(rhi): heap-bindless rehearsal on OpenGL — Phase 3 (#691)

drsnuggles8:
Phase 3 of #691. Builds `RHI::ViewHandle` and `RHI::HeapOffset` — the pair Phase 2 amendment (11) deferred — proves the binding contract on real hardware, and converts 11 shaders / 24 bind sites onto it.

**The heap is OFF by default** (`OLO_RHI_BINDLESS=1` enables it), so no production frame changes. Goldens are green with `OLOENGINE_GOLDEN_REBASE` unset.

## What landed

| | |
| --- | --- |
| `Renderer/RHI/RHIDescriptorHeap.{h,cpp}` | Neutral view registry + heap: views own slots, two lifetime classes, generation-validated `RHI::OffsetOf`, poison-on-free, `SamplerDesc`-deduplicated sampler heap, `InvalidateResource`, reserved null descriptor |
| `Platform/OpenGL/OpenGLDescriptorHeap.{h,cpp}` | `ARB_bindless_texture` backend, extension-gated with runtime toggle |
| `OpenGLShader::CreateProgramFromRawGLSL` | The bindless compile route (see below) |
| `RGCommandContext::BindTextureOrHeapOffset` | One call per site, forks internally, degrades to the old bind |
| Converted | SSAO, AOApply, FXAA, ChromaticAberration, ColorGrading, MotionBlur, TAA, ContactShadow, DOF, EASU, Fog |

## The finding that reshaped the phase

`GL_ARB_bindless_texture` is a GLSL-only extension predating SPIR-V, and glslang rejects it whenever it generates SPIR-V — *on both target environments*:

```
error: 'GL_ARB_bindless_texture' : not allowed when using generating SPIR-V codes
```

Every production shader enters through `shaderc` targeting Vulkan, so bindless GLSL **cannot travel the normal pipeline at all**. Hence the raw-`glShaderSource` route. `BindlessShaderPipelineTest` pins this and fails in *either* direction — a toolchain that started accepting it would make a whole compile route deletable.

**Consequence for Phase 4:** the shader-side mechanism does not transfer (Vulkan uses descriptor indexing, which is SPIR-V-native), and neither does most of the backend — residency, texture immutability and sampler folding are all `ARB_bindless_texture`-specific. What *does* transfer is everything above `IDescriptorHeapBackend`, which is the part ADR 0011 designed.

## Bugs found while building it

- **Publishing the heap once per frame is wrong.** Passes mint views inside their own `Execute`, so the frame-level flush ran before this frame's transient descriptors existed → black viewport. It survived the first conversion because (a) the first converted pass wasn't executed by the active render path and (b) **the GPU test was supplying the missing call by hand** — a test that sequences a mechanism for the subject cannot detect that the subject omits it.
- **"Unbind" has no translation to an indexed model.** A pass clears an input by binding null; under the heap the offset is what the shader reads, so an untouched offset keeps sampling the old texture. Heap slot 0 is now a reserved null descriptor.
- **An unwritten heap slot is not zero.** `glNamedBufferStorage(…, nullptr, …)` leaves contents *undefined*, and slot 0 was never in a dirty range. Passed in isolation, failed only in the full suite after thousands of tests had churned GPU allocations.
- **The resource registry's stability is a liability here** — it deliberately preserves a handle across an in-place reload, leaving descriptors dangling while the view generation is unchanged. Mirror of the Phase 2 slice-6 finding; hence `InvalidateResource`.

## How far full bindless reaches (measured)

Ratchet `sweep_bind_texture_sites` 232 → **208**. The remainder is not homogeneous:

| Bucket | Sites | Shape |
| --- | ---: | --- |
| Ordinary pass binds | ~130 | Mechanical, exactly like the ones converted |
| `Renderer/Commands/` material path | ~21 | The redundant-bind cache — one indivisible unit |
| **`BindImageTexture` (storage images)** | **38** | **Not modelled — needs a second descriptor kind** |

The counter's target is a **floor, not zero**: the slot-based path is a permanent fallback for devices without the extension, and unlike `facade_native_id_params` there is no type-system equivalent available.

## Verification

- Full suite **5255/5259** (4 environment skips, 0 failures)
- 17 CPU contract tests, 8 GPU end-to-end tests (real textures reached through nothing but integer offsets; poison, transient ring, null descriptor and the real `RGCommandContext` seam all checked against read-back pixels), 3 shader-pipeline probes
- Editor with `OLO_RHI_BINDLESS=1`: 11 shaders on the bindless route, no GL or shader errors
- A/B on the water sandbox: bindless OFF vs ON differs on 14.36% of pixels while ON vs ON (identical config) differs on 14.39% — **indistinguishable within the scene's own noise floor**
- Regenerated evidence PNGs pixel-diffed against a measured run-to-run floor and reverted as noise

Docs: boundary doc §4b, ADR 0011 amendments (18)–(26), `glsl-shaders.md` §5a, `live-verification-noise-floor.md` gains the suite's measured PNG noise table. Stale bind-site counts in #691 and ADR 0011 corrected.

Leaves #691 open — Phases 4–9 remain.

## Summary by CodeRabbit

* **New Features**
  * Added OpenGL bindless texture support for more efficient rendering resource access.
  * Updated post-processing effects to use heap-aware texture binding.
  * Added automatic fallback to standard texture binding when bindless support is unavailable.
  * Added shader compilation and caching for bindless variants.

* **Bug Fixes**
  * Improved descriptor lifetime handling, resource invalidation, sampler reuse, and transient allocation safety.

* **Tests**
  * Added CPU and GPU coverage for descriptor heaps, shader compatibility, and rendering behavior.

* **Documentation**
  * Added guidance for bindless shaders, resource lifetimes, caching, and visual verification.


drsnuggles8:
@coderabbitai review

drsnuggles8:
**Phase 3's finish line is written up on the issue:** https://github.com/drsnuggles8/OloEngineBase/issues/691#issuecomment-5163071248

This PR delivers the contract, the backend, the compile route, the pass-side seam and 11 shaders / 24 sites (ratchet 232 → 208). It does **not** finish Phase 3, so the checkbox on #691 stays unticked.

The next PR closes it by working three buckets, which are not interchangeable:

1. **~128 ordinary pass binds** — mechanical, recipe proven over 11 shaders here (`glsl-shaders.md` §5a).
2. **~21 material-path sites** in `Renderer/Commands/` — the redundant-bind cache, one indivisible unit.
3. **38 `BindImageTexture` sites** — a *second descriptor kind*, not a sweep: the heap makes sampler descriptors only (ADR 0011 amendment (26)).

The issue comment carries the per-bucket detail and the four conversion rules that cost debugging time to learn.

Issue #735, opened 2026-08-01
URGENT: AMD golden baselines are unverified — cross-check against NVIDIA before they are trusted


drsnuggles8:
The AMD vendor baseline set was committed in 5d776ba2 (4 renderer goldens under `OloEditor/assets/tests/golden/amd/`, 13 Atmosphere captures under `OloEditor/assets/tests/visual/amd/`). It was baked on **AMD/Mesa only**, and until it is cross-checked on NVIDIA it is a set of *recordings*, not *references*.

### Why this is urgent

A golden makes the nightly green. It does not establish that the render is correct — and once baked, the nightly will actively **defend** whatever was captured, including a defect.

This is not hypothetical. On this same branch, integer-format textures were sampled with `GL_LINEAR`; GL requires `NEAREST` for integer formats, so on Mesa the texture was incomplete and read as **all-zero through `texelFetch`**, while NVIDIA tolerated it. The result: **every glyph in the engine was invisible on AMD**, while the font loaded (191 glyphs), `SlugFontProcessor` packed it (189 glyphs, 13535 band texels), `DrawString` emitted 852 quads across 4 draw calls, and **nothing logged an error**. Only a test asserting on rasterised pixels caught it.

Had the baselines been baked before that fix, a blank UI would now be the reference.

The window matters: every day these sit unverified, they accumulate authority — later diffs get judged against them, and a wrong baseline gets harder to distinguish from a real regression.

### What to do

On the NVIDIA machine, rebake and compare rather than overwrite:

1. `--olo-golden-rebase --olo-golden-vendor=nvidia` (or a scratch vendor name) — do **not** rebake into the shared baselines yet.
   These are **CLI flags**, not environment variables: `OLOENGINE_GOLDEN_REBASE` / `OLOENGINE_GOLDEN_VENDOR` were replaced in `OloEngine/tests/TestOptions.cpp`. Only the `=value` form parses — `--olo-golden-vendor nvidia` (space) hard-fails with "option needs a =value" rather than silently falling back to the shared baselines. Run from `OloEditor/` so `assets/...` resolves.
2. Diff each NVIDIA capture against its AMD counterpart in `OloEditor/assets/tests/{golden,visual}/amd/`.
3. Judge per image: last-bit rounding is agreement; a structural difference is a conformance bug in one of the two and must be diagnosed, not averaged away.

### Expected result, and what would falsify it

[...]

PR #741, opened 2026-08-02
fix(build): full-path /WHOLEARCHIVE unbreaks the clangcl link (#697) + property guards for the four smoke goldens (#734)

### LINKED ISSUE #697: clangcl preset cannot link: /WHOLEARCHIVE:usd_m is unresolvable by lld-link
The `clangcl` preset (and `clangcl-asan`, which inherits it) **cannot link at all** with the default `OLO_WITH_USD=ON`. Both presets are unusable out of the box.

## Symptom

```
lld-link: error: could not open 'usd_m': no such file or directory
ninja: build stopped: subcommand failed.
```

Everything compiles — 541/543 Windows-applicable engine TUs, zero errors, zero warnings from engine code. The failure is purely at link.

## Root cause

`OloEngine/CMakeLists.txt:262-264`:

```cmake
if(MSVC)
    target_link_options(OloEngine PUBLIC "/WHOLEARCHIVE:usd_m")
endif()
```

Two things combine:

1. **CMake sets `MSVC` TRUE for clang-cl** (it targets the MSVC ABI), so this branch is taken under the `clangcl` preset — the guard does not distinguish `link.exe` from `lld-link`.
2. **`/WHOLEARCHIVE:` resolves a bare library name differently between the two linkers.** MSVC `link.exe` looks a bare name up against the `/LIBPATH` search paths. `lld-link` does not — it expects the argument to identify a library already on the link line, or a full path, and errors out otherwise.

`target_link_directories(...)` on line 254 *does* put the right directory on the search path, and `usd_m.lib` really is present:

```
C:/Users/<user>/AppData/Local/OloEngine/usd-<hash>/install/Debug/lib/usd_m.lib
```

so this is purely a name-resolution difference, not a missing artefact.

## Suggested fix

`OloEngine_USD_LIB` (set in `cmake/vendor/OpenUSD.cmake:53,82`) is **already the full path** to `usd_m.lib`, and `link.exe` accepts a full path in `/WHOLEARCHIVE:` too — so this is likely a one-liner:

```cmake
if(MSVC)
    target_link_options(OloEngine PUBLIC "/WHOLEARCHIVE:${OloEngine_USD_LIB}")
endif()
```

Worth confirming against **both** linkers rather than assuming, since the path carries a `$<CONFIG>` genexpr (see the config-matched-USD comment on line 248) and `/WHOLEARCHIVE:` with an unquoted path containing spaces is its own trap. If a full path turns out not to work for one of them, the alternative is to branch on the linker instead of on `MSVC` — the clangcl cache has `CMAKE_LINKER=…/lld-link.exe`, so it is distinguishable.

Whatever the fix, keep the force-link: the comment on line 260-261 is correct that USD registers its Sdf file formats and schemas through static-init side effects, and without whole-archive the linker drops those objects and `UsdStage::Open` fails at runtime.

## Why it matters

- `clangcl` is one of the three documented presets in `CMakePresets.json` and `docs/ops/build.md`, and CLAUDE.md calls it out as the way to catch "clang-cl warnings with MSVC ABI".
- **`clangcl-asan` inherits `clangcl`**, so AddressSanitizer is unavailable too — that is the tool for exactly the memory bugs that are hardest to attribute after the fact. `docs/agent-rules/build-trees-and-windows-asan.md` documents a whole ASan workflow that currently cannot be run.
- It is silent in the sense that matters: it only bites whoever next configures a clangcl tree, and the error names a library that is present on disk, which sends you looking in the wrong place.

## Workaround

```powershell
cmake --preset clangcl -DOLO_WITH_USD=OFF
```

Verified: with USD off, `clangcl` configures, compiles and links `OloEngine-Tests` cleanly.

## Environment

- LLVM/clang-cl **21.1.8**, `Target: x86_64-pc-windows-msvc`
- `CMAKE_LINKER=C:/Program Files/LLVM/bin/lld-link.exe`
- CMake 4.2, Ninja Multi-Config, Windows 11
- Found while building the clangcl tree as a verification gate for #691 Phase 2; unrelated to that work.

## Score
```olo-score
capability: 5
craft: 1
stability: 3
decay: 3
effort: 1
confidence: 0.8
learning: 2
fun: 1
kano: table-stakes
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](../docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

<sub>Axis notes — `capability 5`: restores two presets and several dev workflows (clang-cl warning sweeps, ASan), rather than one subsystem. `stability 3`: real but dev/test-only, anchored on #446 — it breaks a verification path, it does not corrupt anything. `decay 3`: mild compounding, since clang-cl diagnostics and ASan findings accumulate unchecked while the preset is dark. `effort 1`: likely a one-line CMake change; the cost is the verification build, not the edit. `confidence 0.8`: the diagnosis is clear, but USD linking has a history of surprises in this repo (see `docs/agent-rules/asset-import-usd-alembic.md`).</sub>

### LINKED ISSUE #734: Renderer smoke goldens are weak guards — pin properties, not just pixels
The four fixed-function goldens under `OloEditor/assets/tests/golden/` (`GoldenImageTests.cpp`) currently assert one thing: "the image still matches the baseline within an RMSE threshold". Reviewing the AMD baseline candidates showed all four render **correctly** — but also that they would only catch an effect failing *completely*. A partial regression very likely slides under the threshold.

Measured on the current renders (128x128 each):

| golden | measured guard strength |
|---|---|
| `fxaa_hard_edge` | FXAA alters only **224 of 16384 px (1.37%)** vs the raw input, across **4 blend levels** |
| `scene_shadow_integration` | **8 unique colours**; two flat regions at 47.7% / 46.2% — no penumbra or falloff is pinned |
| `scene_splatmap_integration` | 641 unique colours; smooth, but nothing asserts the weights are *right* |
| `tonemap_reinhard_hdr_ramp` | 7595 colours; peak luma **225.9** — correct Reinhard behaviour, but nothing asserts it |

So e.g. an FXAA change that halved blend strength moves ~112 pixels by a few levels on a 16k-pixel image; a shadow that lost its penumbra changes almost nothing here. Both would pass.

### Proposal — add contract assertions next to the golden compare

Keep the goldens (they catch gross change cheaply), but have each test also pin the *property* the effect exists to provide. Each of these is derivable from the readback already in hand, so it is test-only work with no engine change.

- **FXAA** — assert a minimum fraction of pixels differ from the input, and that the blends are **symmetric**: the current output is `0->48 (x97)` / `255->207 (x97)` and `0->136 (x15)` / `255->119 (x15)`, i.e. complementary pairs. That symmetry is a real invariant of blending a black/white edge and is a much sharper signal than RMSE.
- **Reinhard tone map** — assert peak output `< 255` and monotonicity along the ramp. `x/(1+x)` asymptotically approaches 1 and never reaches it, so a passthrough or a broken operator saturates. This is the strongest invariant of the four and is currently unasserted.
- **Shadow** — assert the lit/shadowed attenuation ratio falls in a range (currently rgb(219,215,204) vs rgb(141,128,133), ~0.63) and that the boundary has non-zero width, so losing the penumbra fails loudly.
- **Splatmap** — assert each corner is dominated by its own layer and that weights sum to ~1, so a mis-indexed or renormalised splat fails even when the image still looks like a smooth blend.

### Why goldens alone are not enough here

An AMD/NVIDIA conformance bug fixed in this same session (integer-format textures sampled with `GL_LINEAR` read as all-zero on Mesa) made **every glyph in the engine invisible** while the font loaded, 189 glyphs packed, 852 quads were submitted, and nothing logged an error. Only a test asserting on rasterised pixels caught it. Image-compare tests are good at "did this change"; they are poor at "is this still doing its job".

### Note

Unrelated to guard strength, but worth recording while it is measured: on these four, AMD vs the existing baseline is byte-identical for `fxaa_hard_edge` and **RMSE 0.16-0.40, max |diff| = 1** for the other three — last-bit rounding only. That is evidence an unscoped (vendor-shared) comparison is viable for this set rather than needing a separate `golden/amd/` baseline.

### Scope

Test-only: `OloEngine/tests/Rendering/PropertyTests/GoldenImageTests.cpp`. No engine or shader change. The existing goldens stay as-is.

## Score
```olo-score
capability: 3
craft: 1
stability: 5
decay: 3
effort: 1
confidence: 0.8
learning: 2
fun: 2
kano: table-stakes
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](../docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

<sub>Axis notes — `capability 3`: self-contained, one area (the golden layer), but it is also the named mitigation for #735, so the vendor-baseline work leans on it. `craft 1`: test assertions, invisible from outside. `stability 5`: it does not fix a bug, it closes the detection gap for a class of bug that ships silently — the motivating incident (integer textures + `GL_LINEAR` making every glyph invisible on Mesa) logged nothing and passed every non-pixel test. `decay 3`: mild compounding — every additional baked baseline, including the per-vendor ones from #701, is another recording that inherits the weak guard. `effort 1`: test-only, all four invariants and their present-day measured values are already in the issue. `confidence 0.8`: the properties are certain; the unknown is threshold calibration across vendors, which is exactly what #735 is about.</sub>

drsnuggles8:
Two small, unrelated issues batched into one branch because neither justifies its own PR. They share no files — one commit each.

Closes #697
Closes #734

---

## #697 — `clangcl` could not link at all

`clangcl` (and `clangcl-asan`, which inherits it) failed every link with the default `OLO_WITH_USD=ON`:

```
lld-link: error: could not open 'usd_m': no such file or directory
```

Everything compiled. The error names a library that *is* present on disk, which sends you looking in the wrong place.

**Cause.** CMake sets `MSVC` TRUE for clang-cl (it targets the MSVC ABI), so the `if(MSVC)` guard around the USD force-link is taken under the clangcl presets too — and the two linkers resolve a bare `/WHOLEARCHIVE:` name differently. `link.exe` looks it up against `/LIBPATH`; `lld-link` opens it as a path relative to the working directory only, even when the archive is already on the link line by full path.

**Fix.** `OloEngine_USD_LIB` is already the full path, so it's a one-liner.

**Verified against both linkers, not assumed.** An isolated force-link probe — a static lib whose only content is a static-init side effect, plus a `main` that never references it — on LLD 21.1.8 and MSVC `link.exe` 14.51:

| `/WHOLEARCHIVE:` argument | `link.exe` | `lld-link` |
| --- | --- | --- |
| bare name, reachable via `/LIBPATH` | force-links | **link fails** |
| full path, lib also on the link line | force-links | force-links |
| full path, lib *not* on the link line | force-links | force-links |
| full path containing spaces | force-links | force-links |
| *(control)* no `/WHOLEARCHIVE` | object dropped | object dropped |

The control row is what proves the probe measures force-linking rather than incidental reachability.

Then confirmed end-to-end in the real tree, both directions: with the bare name the clangcl link reproduces the exact error above; with the full path, **clangcl and msvc both link `OloEngine-Tests` with USD on**.

**The force-link stays, and is load-bearing under both linkers in _different_ ways** — measured by relinking with the option removed:

- `link.exe` — links fine, `UsdStage::Open` then fails at **runtime**.
- `lld-link` — fails the link outright on `undefined symbol: __declspec(dllimport) ... UsdGeomGetStageUpAxis`, because `usd_m`'s headers declare its symbols `dllimport`.

So "it linked after I removed the flag" is a link.exe-only observation, and it's the silent-at-link/broken-at-runtime case. Worth knowing before anyone "fixes" a future whole-archive error by deleting the option.

Bonus: `OloEngine_USD_LIB` carries a `$<CONFIG>` genexpr that `target_link_options` expands per config, so the force-link is now config-matched instead of depending on `/LIBPATH` search order.

Docs: new gotcha 7 in `asset-import-usd-alembic.md` (linker matrix + both failure modes); a note in `build-trees-and-windows-asan.md` that its ASan recipe was unrunnable until this fix and no longer needs the `-DOLO_WITH_USD=OFF` workaround (and that its example build command needs an explicit `-jN`).

---

## #734 — the four smoke goldens were weak guards

They asserted one thing: "the image still matches the baseline within an RMSE/SSIM band". That catches an effect failing *completely* and little else. A baseline is also only ever a recording of one machine, so where a vendor baseline was baked separately (#735) the compare can pass on a frame nobody has verified is *correct*.

Each golden now also pins the property the effect exists to provide, computed from the readback it already holds. **Additive** — goldens unchanged, no engine or shader code touched.

| golden | property pinned |
| --- | --- |
| **Reinhard** | peak channel `< 245` **and** `> 200`, plus per-row monotonicity. `x/(1+x)` is asymptotic to 1 and can never reach it, so a correct frame *cannot* saturate; the floor stops a dim frame passing vacuously. |
| **FXAA** | bounded fraction of pixels blended, and blends are **complementary** — a black pixel lifts to `t`, its mirror drops to `1-t`, so the sorted multisets must pair to 255. |
| **Shadow** | lit/shadowed attenuation in range, ≥4 distinct colours, and the two flat regions must not cover the whole frame (the PCF penumbra has non-zero width). Orientation-free: regions are found by population, never by a hard-coded screen position. |
| **Splatmap** | blended HDR must equal `Σ wᵢ·layerᵢ` for the exact uploaded weights, texel by texel, plus edge dominance and a green:red separation of the two red-dominant layers. |

**Thresholds were measured, not guessed** — and each independently reproduces the numbers in the issue: peak 230 (theory 230.4), FXAA 1.3672% of 16384 px as 112 dark + 112 bright, shadow 8 colours / attenuation 0.619 / 93.85% flat, splatmap worst blend error 0.00135 against an fp16 quantum of ~0.0009.

**Every guard was confirmed to actually fail** — an assertion that can't fail is the anti-pattern here — by breaking the property in the real pipeline (scaffolding not committed):

| mutation | guard that fired |
| --- | --- |
| passthrough tone map | peak 255 vs 245 |
| V-shaped ramp | 8914 decreasing steps |
| zero texel size | 0% blended vs 0.5% floor |
| halved dark blends | pair 97 off by 68 |
| single-tap PCF kernel | 2 colours vs 4; 100% flat vs 98% |
| washed-out ambient | attenuation 0.906 vs 0.75 |
| swizzled splatmap channels | blend error 1.60 vs 0.01; all 4 edges mis-dominated |

Note the single-tap mutation left attenuation at **0.619** — losing the penumbra doesn't move the ratio at all. That's exactly why both shadow guards exist, and it's the issue's argument in miniature.

---

## Verification

- **clangcl** — configures, compiles and links `OloEngine-Tests` with `OLO_WITH_USD=ON`; new test code compiles clean under clang-cl (no new warnings).
- **msvc** — builds and links; goldens green.
- **Full suite: 5247/5254 passed**, 2 skipped, 10 disabled.
- **One pre-existing failure, unrelated to this branch:** `AtmosphereVisualEvidenceTest.DayNightWeatherMatrixRendersAndHoldsContracts` — `NightClear` RMSE 13.08 and `NightOvercast` 9.40 against a threshold of 8. Deterministic (identical values in isolation, so not a test-order state leak), and this branch touches only `GoldenImageTests.cpp`, one CMake link option and two docs. It looks like the GPU-baseline drift #735 is about. **No baseline was rebaked here**, per the handover.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OpenUSD linking for MSVC and lld-link builds, including configurations using clang-cl.
  - Enabled reliable USD plugin registration and reduced related build or runtime failures.

- **Tests**
  - Added validation for tone mapping, anti-aliasing, shadows, and splatmap blending to improve rendering quality checks.

- **Documentation**
  - Updated USD linking guidance and Windows AddressSanitizer build instructions, including explicit parallel build settings.

### DIFF
--- a/OloEngine/CMakeLists.txt
+++ b/OloEngine/CMakeLists.txt
@@ -259,8 +259,18 @@ if(OLO_WITH_USD)
 	target_compile_definitions(OloEngine PUBLIC OLO_WITH_USD=1)
 	# USD registers Sdf file formats + schemas via static-init side effects; without a
 	# whole-archive force-link the linker discards those objects and UsdStage::Open fails.
+	#
+	# FULL PATH, not the bare `usd_m` (#697). CMake sets MSVC TRUE for clang-cl too, so
+	# this branch is taken under the clangcl / clangcl-asan presets — and lld-link, unlike
+	# link.exe, will NOT resolve a bare /WHOLEARCHIVE: name against /LIBPATH. It killed
+	# every clangcl link with `could not open 'usd_m': no such file or directory`. A full
+	# path force-links under both linkers (verified LLD 21.1.8 / link.exe 14.51), and
+	# OloEngine_USD_LIB's $<CONFIG> genexpr makes it config-matched into the bargain.
+	# Do NOT delete the option to "fix" a whole-archive error — it is load-bearing under
+	# both linkers, silently at runtime for link.exe and loudly at link for lld-link.
+	# Full linker matrix + failure modes: docs/agent-rules/asset-import-usd-alembic.md.
 	if(MSVC)
-		target_link_options(OloEngine PUBLIC "/WHOLEARCHIVE:usd_m")
+		target_link_options(OloEngine PUBLIC "/WHOLEARCHIVE:${OloEngine_USD_LIB}")
 	endif()
 	# Ship USD's plugInfo resource tree next to every consuming executable — PlugRegistry needs
 	# it at runtime to map .usda/.usdc/.usdz -> file-format plugin (see UsdMeshImporter). The exe
--- a/OloEngine/tests/Rendering/PropertyTests/GoldenImageTests.cpp
+++ b/OloEngine/tests/Rendering/PropertyTests/GoldenImageTests.cpp
@@ -55,6 +55,7 @@
 #include <stb_image/stb_image_write.h>
 
 #include <algorithm>
+#include <array>
 #include <cerrno>
 #include <cmath>
 #include <cstdlib>
@@ -62,6 +63,7 @@
 #include <filesystem>
 #include <sstream>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 namespace OloEngine::Tests
@@ -633,6 +635,94 @@ namespace OloEngine::Tests
             }
             return out;
         }
+
+        // =====================================================================
+        // Property guards (#734)
+        //
+        // CompareOrBootstrap answers "does this frame still match its
+        // baseline?". It cannot answer "is the effect still doing its job?".
+        // A *partial* regression — FXAA blending only one side of an edge, a
+        // tone map that clips, a shadow that lost its penumbra, a splatmap
+        // whose layer weights got renormalised — can sit comfortably inside
+        // the RMSE/SSIM pass band while the invariant the effect exists to
+        // uphold is gone. And a baseline is only ever a recording of one
+        // machine: where a vendor baseline was baked separately (#735,
+        // assets/tests/golden/amd) the compare can pass on a frame nobody has
+        // verified is *correct*.
+        //
+        // So each golden below also asserts a property derived from the
+        // readback it already holds — one extra pass over the image, and a
+        // failure that says what broke instead of "the image moved".
+        //
+        // Index alignment used by the guards: glTextureSubImage2D uploads and
+        // glGetTextureImage reads back in the same GL row order, and a
+        // fullscreen quad drawn at the texture's own resolution samples texel
+        // centres 1:1 — so a CPU-authored input array and a readback of the
+        // frame it produced share the same texel index. Guards that cannot
+        // rely on that (the shadow mask, whose regions are placed by the
+        // authored depth map) are written to be orientation-free instead, so
+        // a flipped readback can never satisfy them by accident.
+        // =====================================================================
+
+        static f32 Rec601Luma(u8 r, u8 g, u8 b)
+        {
+            return 0.299f * static_cast<f32>(r) + 0.587f * static_cast<f32>(g) + 0.114f * static_cast<f32>(b);
+        }
+
+        struct ColorPopulation
+        {
+            u32 m_Rgb = 0; // packed 0x00RRGGBB
+            u32 m_Count = 0;
+        };
+
+        // Distinct-colour histogram, most populous first. Lets a guard find
+        // flat regions by population rather than by sampling a hard-coded
+        // screen position — which would bake a readback orientation into the
+        // assertion and quietly pass on a vertically flipped frame.
+        static std::vector<ColorPopulation> RankColorsByPopulation(const std::vector<u8>& rgba)
+        {
+            std::unordered_map<u32, u32> counts;
+            for (std::size_t i = 0; i + 3 < rgba.size(); i += 4)
+            {
+                const u32 key = (static_cast<u32>(rgba[i + 0]) << 16) |
+                                (static_cast<u32>(rgba[i + 1]) << 8) |
+                                static_cast<u32>(rgba[i + 2]);
+                ++counts[key];
+            }
+            std::vector<ColorPopulation> ranked;
+            ranked.reserve(counts.size());
+            for (const auto& [rgb, count] : counts)
+                ranked.push_back(ColorPopulation{ rgb, count });
+            std::sort(ranked.begin(), ranked.end(),
+                      [](const ColorPopulation& a, const ColorPopulation& b)
+                      { return a.m_Count > b.m_Count; });
+            return ranked;
+        }
+
+        // Mean RGB of a square patch centred on (cx, cy), in the readback's own
+        // index space (so it lines up with whatever CPU array authored the input).
+        static std::array<f32, 3> PatchMeanRgb(const std::vector<u8>& rgba, u32 width, u32 height,
+                                               u32 cx, u32 cy, u32 radius)
+        {
+            std::array<f64, 3> sum{ 0.0, 0.0, 0.0 };
+            u32 taps = 0;
+            const u32 x0 = cx > radius ? cx - radius : 0u;
+            const u32 y0 = cy > radius ? cy - radius : 0u;
+            const u32 x1 = std::min(cx + radius, width - 1u);
+            const u32 y1 = std::min(cy + radius, height - 1u);
+            for (u32 y = y0; y <= y1; ++y)
+            {
+                for (u32 x = x0; x <= x1; ++x)
+                {
+                    const std::size_t idx = (static_cast<std::size_t>(y) * width + x) * 4;
+                    for (u32 c = 0; c < 3; ++c)
+                        sum[c] += static_cast<f64>(rgba[idx + c]);
+                    ++taps;
+                }
+            }
+            const f64 n = taps > 0 ? static_cast<f64>(taps) : 1.0;
+            return { static_cast<f32>(sum[0] / n), static_cast<f32>(sum[1] / n), static_cast<f32>(sum[2] / n) };
+        }
     } // namespace
 
     // =========================================================================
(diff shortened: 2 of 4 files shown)


drsnuggles8:
@coderabbitai review

Addressed the MD040 finding in 45dee91f — the `lld-link` error block now carries a `text` fence, matching the fence the same file already uses for its other console-output block.

PR #743, opened 2026-08-03
feat(networking): assemble the server-authoritative multiplayer loop (#636)

### LINKED ISSUE #636: Networking: assemble the server-authoritative multiplayer loop + playable 2-client demo
## Problem

OloEngine's networking subsystem contains ~20 replication/prediction/interest/lag-comp classes and a working GameNetworkingSockets transport — all unit-tested **in isolation** — but they are **never assembled into a running server-authoritative loop**. The pieces exist; the game does not.

Verified dead seams (grep + read, 2026-07-11):

- **`NetworkManager::TickSnapshots()` has zero call sites.** `NetworkThread::ThreadFunc` (`NetworkThread.cpp:78`) only runs GNS callbacks + named-thread tasks. The header comment claiming it "invokes TickSnapshots()" is false. The server-side capture+delta+broadcast never runs.
- **`NetworkManager::SetActiveScene()` is never called in production.** `OloServer/src/OloServerApp.cpp` loads a scene and ticks `OnUpdateRuntimeFixed`, but never registers it with networking. `s_ActiveScene` is always null, so `TickSnapshots` would early-out anyway.
- **`SnapshotInterpolator::Interpolate(scene, dt)` is called only from tests.** Received snapshots pile into the buffer and are never applied to the client ECS — remote entities never move.
- **Prediction/reconciliation never fires end-to-end.** No `InputApplyCallback` is registered in production; the server never sends `InputAck`, so `ClientPrediction::Reconcile` is dead.
- **`RPC` is a bare `ENetworkMessageType` enum value** (`NetworkMessage.h:21`). No registry, marshalling, authority routing, or script bindings exist (`RegisterRPC`/`InvokeRPC`/`RpcDispatcher` = no matches anywhere).
- **Interest management is wired only into the separate MMO `ZoneServer`**, not the core loop — `TickSnapshots` broadcasts one shared delta to all clients with no per-connection baseline or relevance culling.
- **`LagCompensator` is never invoked** by the input/hit path (`ServerInputHandler::ProcessInput` does no rewind).
- **`EntitySpawn`/`EntityDespawn` are enum tokens with no lifecycle replication** — snapshots assume a static entity set; server-spawned entities never appear on clients.

No open GitHub issue tracks this.

## Why it matters

The engine advertises networking (dedicated `OloServer` target, MMO/lobby/P2P classes, Lua `Network.*` table) but **cannot run a single authoritative multiplayer match**. This is the highest-decay area in the codebase: a large, well-tested stack that produces nothing playable because the last-mile assembly and several keystone systems (RPC, spawn/despawn, interest-scoped replication, lag-comp integration) were never built. Bringing it to life turns a dead stack into the engine's flagship differentiator.

## What EXISTS today

- Transport: `NetworkServer`/`NetworkClient`/`NetworkConnection` (GNS), live-tested by `tests/Networking/NetworkIntegrationTest.cpp`.
- Replication: `EntitySnapshot` (Capture/CaptureDelta/Apply), `ComponentReplicator`, `ComponentInterpolationRegistry`, `SnapshotBuffer`, `SnapshotInterpolator`.
- Prediction: `ClientPrediction`, `ServerInputHandler`, `NetworkInputBuffer`.
- Lag comp: `LagCompensator` (rewind primitive).
- Interest: `NetworkInterestManager`, `SpatialGrid`, `NetworkPriorityQueue`, `RelevanceTier`.
- Components: `NetworkIdentityComponent` + `ENetworkAuthority{Server,Client,Shared}` (`Components.h:3509`), `NetworkInterestComponent`, `NetworkLODComponent`.
- Facade: `NetworkManager` (server capture+broadcast in `TickSnapshots`, `SendInput`, reconcile handlers in `Connect`).
- Lua: `Network.isServer/connect/startServer/…` (`LuaScriptGlue.cpp:2385`).

## Proposed approach

1. **Core loop driver** — drive `TickSnapshots()` from the NetworkThread (or a dedicated fixed-rate replication tick); register the active scene via `SetActiveScene()` from `OloServerApp` and the editor host; drive `SnapshotInterpolator::Interpolate()` per client frame.
2. **Authoritative ECS replication + entity lifecycle** — replicate spawn/despawn (wire the existing tokens), server-UUID→client-entity mapping, authority-gated writes.
3. **Client prediction + reconciliation end-to-end** — real `InputApplyCallback`, per-tick input sampling, server `InputAck`, replay-on-divergence.
4. **Registered RPC system** — typed marshalling over `FArchive`, authority routing (Server/Client/Multicast/target), reliability, dispatch over the `RPC` message type.
5. **RPC script bindings** — `RegisterRPC`/`InvokeRPC` in Lua (`LuaScriptGlue.cpp`) and C# (`ScriptGlue.cpp`).
6. **Interest-scoped core replication** — per-connection scoping via `NetworkInterestManager`/`SpatialGrid`/`RelevanceTier` + per-client delta baselines.
7. **Lag compensation integration** — rewind to the client's RTT-adjusted tick before authoritative hit checks, then restore.
8. **Dedicated-server integration** — `OloServerApp` registers its scene, spawns/despawns a player entity per connection, drives the replication tick.
9. **Playable 2-client demo + functional tests** — a `SandboxProject` scene running the full loop, plus `Functional`-axis convergence/authority/RPC/interest/lag-comp tests.

## Acceptance criteria

- [x] `TickSnapshots()` is driven and `SetActiveScene()` is wired from both `OloServerApp` and the editor host; a connected client's remote entities move via `SnapshotInterpolator::Interpolate()`.
- [x] Server spawns/despawns replicate to clients with a stable server-UUID→client-entity mapping; client writes to server-authoritative entities are rejected.
- [x] Client prediction + server `InputAck` + reconciliation converge end-to-end (replay of unacknowledged inputs on divergence).
- [x] A registered RPC system marshals typed args, routes by authority (Server/Client/Multicast/target), and selects reliability — no longer a bare enum.
- [x] RPCs are callable from Lua and C# scripts.
- [x] The core replication loop scopes snapshots per connection with per-client baselines and relevance culling (not one shared broadcast).
- [x] `LagCompensator` is invoked from the server input/hit path (rewind → check → restore).
- [x] `OloServer` runs the full authoritative loop: player-per-connection lifecycle + replication tick.
- [x] A playable 2-client demo scene exists under `SandboxProject`, plus `Functional` tests asserting cross-client convergence, authority rejection under the live loop, RPC round-trips, interest culling, and lag-comp hit registration.

```olo-score
capability: 5
craft: 4
stability: 2
decay: 5
effort: 13
confidence: 0.92
learning: 4
fun: 5
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary

Assembles OloEngine's networking stack into a running **server-authoritative multiplayer loop**, and builds the keystone systems that were never written: RPC, entity spawn/despawn replication, interest-scoped per-connection replication, and lag-compensation integration.

The stack was not broken — it was never wired. Before this change:

- `NetworkManager::TickSnapshots()` had **zero call sites**, and the header comment claiming `NetworkThread` invoked it was false.
- `SetActiveScene()` was never called in production, so the tick would have early-returned anyway.
- `NetworkServer::PollMessages()` / `NetworkClient::PollMessages()` had **no production call site**, so no message was ever received.
- `SnapshotInterpolator::Interpolate()` was called only from tests.
- `RPC` was a bare enum value.

## What this adds

**Core loop.** `NetworkManager::Tick(dt)` is the single game-thread entry point (poll → connection lifecycle → fixed-rate replication tick → interpolation), driven from `OloServerApp`, `OloRuntimeApp` and `EditorLayer`. `SetActiveScene()` is wired from all three, and cleared *before* a scene is released.

**Threading contract, corrected.** Replication is game-thread work — it captures, applies, spawns, destroys and runs gameplay handlers against the registry the game thread is iterating. GNS connection callbacks arrive on the network thread, so they are now *recorded* (`NetworkServer::DrainClientEvents`) and drained on the game thread. The false header comment is fixed.

**Two drivers, as objects.** `ServerReplicationDriver` and `ClientReplicationDriver` take their `Scene` and transport as parameters instead of being more statics on the singleton facade — which is what makes a two-client test possible in one process at all.

**Per-connection replication.** Each connection carries its own relevance set (`NetworkInterestManager`), its own known-entity set (driving spawn on entry / despawn on exit), and its own **acknowledged** delta baseline. Snapshots travel unreliably, so a `SnapshotAck` message advances the baseline only to state the client confirms applying — a baseline the client never received would otherwise poison every subsequent delta, permanently and only off localhost.

**Entity lifecycle.** `EntityLifecycle` + `NetworkSpawnRegistry` replicate spawn/despawn with a stable server-UUID → client-entity mapping. A spawn carries the entity's full initial state (reusing the snapshot record, so one hardened parser covers the attacker-controlled part). A despawn destroys **only** entities the client's own spawn path created, so a scene-authored prop leaving relevance is not deleted.

**Prediction that converges.** Three real bugs fixed: the client input tick was the *server's* counter (so every input carried tick 0 and the server dropped all but the first); `SendInput` recorded without applying (not prediction at all); and reconciliation replayed on top of the client's own drift because the interpolator deliberately skips predicted entities — it now snaps owned entities to authoritative state first, then replays. The interpolator also selects by **ownership** rather than authority, which was freezing every *other* player's pawn.

**RPC.** `RpcRegistry` / `RpcDispatcher` with `FArchive` marshalling (bool/int/float/string/vec3/entity), authority routing (Server/Client/Multicast) re-checked at both ends, reliability selection, and per-entity ownership gating. Callable from Lua (`Network.registerRPC` / `invokeRPC`) and C# (`OloEngine.Network`). Script-owned handlers are dropped at VM teardown — a `std::function` holding a `sol::protected_function` outlives its `sol::state` quite happily and then crashes.

**Lag compensation.** `ServerReplicationDriver::PerformLagCompensatedCheck` rewinds to the client's own view time — half-RTT **plus** the client's render delay — runs the check, restores.

**Demo + tests.** `SandboxProject/Assets/Scenes/NetworkDemo.olo` with `NetworkDemoController.lua`. Three new test files: `RpcMarshallingTest` and `EntityLifecycleTest` (transport-free contract tests), and `ServerAuthoritativeLoopTest` — a Functional-axis test that stands up a real server and **two** real clients over loopback and asserts cross-client convergence, authority rejection under the live loop, RPC round-trips, interest culling, and lag-compensated rewind.

**Lessons captured** in `docs/agent-rules/server-authoritative-networking-loop.md`.

Closes #636

## Verification

- `RpcMarshallingTest` (10) + `EntityLifecycleTest` (10): **pass**
- `ServerAuthoritativeLoopTest` (14, real server + two real clients over loopback): **pass**
- Broad regression set (`*Network*`, `*Snapshot*`, `*Replicat*`, `*Component*`, MMO/lockstep/P2P/persistence, …): **565 tests pass**
- `OloServer --scene SandboxProject/Assets/Scenes/NetworkDemo.olo` mounts the project, loads the demo scene, runs its Lua (`Network.useMovementInput` + two `Network.registerRPC` calls execute live) and listens with the replication tick at 20 Hz.

## Pre-existing bugs this surfaced

Three of these were in code that had **never executed** — the recurring theme of the issue:

1. `NetworkClient::PollMessages` dispatched handlers inside its own mutex, so any handler that replied self-deadlocked. `NetworkServer::PollMessages` already deferred dispatch outside the lock and documents why.
2. `GetClientPingMs` re-acquires the mutex `ForEachConnection` holds. Three call sites had this shape, including the editor's **Network Debug panel** — opening it against a live server froze the editor. Added `GetClientPingMsById` / `GetConnectionSnapshot` and converted every site.
3. `OloServer` mounted no `Project`, so scene-referenced Lua scripts never resolved — the dedicated server could host a world but run none of its server-side gameplay. Now mounts via `Project::NewInMemory` (`--project`, or inferred from the scene path), mirroring OloRuntime.

## Summary by CodeRabbit

* **New Features**
  * Added server-authoritative multiplayer with replicated entities, prediction, reconciliation, interpolation, interest scoping, lag compensation, and RPCs for C# and Lua.
  * Added a playable two-client network demo with hosting, connecting, movement, reset, and announcements.
  * Added configurable snapshot rates, project mounting, and replication diagnostics for servers and the editor.
* **Bug Fixes**
  * Improved connection handling and prevented networking-related editor or client hangs.
* **Tests**
  * Added coverage for replication, entity lifecycle, movement, ownership, RPCs, and malformed data.
* **Documentation**
  * Added networking integration and Windows build troubleshooting guides.


drsnuggles8:
All 24 review findings addressed in d7e4ee42 (plus 3e084c2d for the deadlock guard). Every one was valid — nothing dismissed.

### The two that mattered most

**Both authority gates keyed off the wrong thing.** `receivedOnServer` is not the same as "sent by a client", and that was wrong in two places:

- the ownership rule validates a *client* sender, so applying it to a server-originated call made the server unable to act on any client-owned pawn — its own check refusing it;
- the direction rule refuses a client pushing a Client/Multicast RPC, so applying it to the server's own local execution of a Multicast meant the broadcast reached every client **except the host**.

Both now key off `senderClientID == 0` ("the server originated this"). A forged client payload still arrives with its own non-zero id and is still refused. Two new unit tests pin the cases.

Worth noting: the second half only appeared *because* I took the suggestion to route the server-local Multicast through `ExecuteLocally`. The consolidation is right, but it moved a path under a guard that had never seen it — and the existing test caught it only because it asserts a *count* of recipients (server + 2 clients = 3); a "did the multicast arrive" check would have passed with 2.

That change also makes `senderClientID` load-bearing for authority, so I hardened its one production gap: `NetworkServer::PollMessages` defaulted an unresolvable sender to `0`, which the dispatcher would now read as server-origin. Untracked-connection messages are dropped instead. Issued client IDs start at 1, so `0` must never come off the wire.

**`ScriptOwned` was one flag for two VMs.** Tearing down Lua unregistered every live C# RPC, and a C# assembly reload did the same to Lua's. Replaced with `ERpcOwner{Native,Lua,CSharp}` + `ClearOwnedBy`, with the test rewritten to assert cross-VM independence.

### Everything else

Critical null-`MonoString*` guards in `Network_Spawn`/`RegisterRPC`/`InvokeRPC`; listen-server double-apply in `SendInput`; `Reset()` leaving stale per-client input ticks; scene `reload` keeping baselines for destroyed entities; the test fixture's use-after-free ordering; self-contained headers; `Encode` enforcing the string bound `Decode` does; payload-offset wrap guard; `System.UInt32` unboxing; the managed exception *message*; a `dt` clamp in the demo; `is_directory` validation; and documenting that C# handlers receive `long`/`double` regardless of what was passed.

### Verification

- 37 networking tests pass (2 new).
- 580-test regression sweep green (1 unrelated skip).
- `OloServer` still mounts the project, loads the demo scene, runs its Lua and ticks replication at 20 Hz.

drsnuggles8:
## TSan and UBSan failures — root-caused and fixed

Both were vendored-GameNetworkingSockets interactions, surfaced because `ServerAuthoritativeLoopTest` is the **first live-GNS test ever to reach these jobs** — `NetworkIntegrationTest` (and, for UBSan, `NetworkManagerTest` / `NetworkThreadDispatchTest`) have always been excluded. One of them turned out to be a genuine repo-level misconfiguration.

### UBSan — a real bug in `cmake/Sanitizers.cmake` (fixed at the root)

`UBSAN_CHECKS` names and documents **eight** checks, then the code passed `-fsanitize=undefined`, enabling **~20**. The variable was only used for `-fno-sanitize-recover`, so the enabled set never matched the documented intent.

That matters because the group includes `vptr`, and **`-fsanitize=vptr` requires RTTI**. GNS builds `-fno-rtti`, dropping that flag only when *its own* `SANITIZE_UNDEFINED` option is set — which our directory-scoped `add_compile_options` never sets. So GNS TUs were vptr-instrumented with no usable typeinfo, and every virtual call through one of its interfaces reported

```
member call on address 0x… which does not point to an object of type 'ISteamNetworkingUtils'
```

for an object that is exactly that type.

Verified with g++ on Linux rather than argued from first principles:

| | result |
|---|---|
| `-fsanitize=undefined`, mistyped virtual call | reproduces the CI diagnostic **verbatim** |
| narrowed curated set, same program | no report |
| `signed-integer-overflow` under narrowed set | still fires, still exits `1` with `-fno-sanitize-recover` |

Note CI sets `UBSAN_OPTIONS=halt_on_error=1`, which makes every report fatal regardless of the recover list — so the *enabled* set is the enforcement surface. The recover list is kept aligned so behaviour is unchanged if `halt_on_error` is ever dropped.

### TSan — false positive inside vendored GNS

The cycle is real as a graph, unreachable as a deadlock:

* create (`ConnectByIPAddress` → `BInitConnection`): **ConnectionLock → TableLock**
* destroy (`GameNetworkingSockets_Kill` → `KillConnections`): **TableLock → ConnectionLock**

Both are the same rank (`k_nOrder_ObjectOrTable`) and **both paths hold `SteamNetworkingGlobalLock` for their whole duration**. GNS's own lock checker permits precisely this (`steamnetworkingsockets_lowlevel_misc.cpp`, `AboutToLock`: *"If we hold the global lock, it's OK"*), and `BInitConnection` documents why creation is safe — the connection isn't in the table yet, so no other thread can hold its lock. TSan's detector is a pairwise lock-order graph over mutex **addresses**; it models neither the dominating global lock nor object reachability.

### Why the test is still excluded from both jobs

The vptr fix is at the root, but it does **not** make GNS UBSan-clean: GNS deliberately disables the alignment check for its own build (its vendored `FindUBSan.cmake` adds `-fno-sanitize=alignment` to the stock `arsenm/sanitizers-cmake` flags; its `FindTSan.cmake` is unmodified, so that is considered, not boilerplate), while `alignment` is one of our eight fatal checks.

So `ServerAuthoritativeLoopTest` joins the live-GNS suites already excluded — with comments in the workflow recording *why*, so the next person finds an investigation rather than a blanket skip.

### Coverage that is unaffected

`build` (Windows) **passed with this suite running** — it isn't excluded there, so the two-client loop does get real CI execution. ASan + LSan passed via the existing `OLO_ASAN_ENABLED` self-skip. The transport-free networking tests continue to run under all three sanitizers.

PR #747, opened 2026-08-03
feat(rhi): heap-bindless binding model + make the bindless path survive a full suite run (#691 Phase 3)

drsnuggles8:
Phase 3 of #691 — the heap-bindless binding model — plus the work needed to make
that path survive a full test run.

## Result

| Configuration | Before | After |
|---|---|---|
| `OLO_RHI_BINDLESS=1`, full suite | 13 failed | **1 failed** |
| Heap off (what CI runs) | 1 failed | **1 failed** (5304 passed) |

Both configurations now fail the *same single test*,
`AtmosphereVisualEvidenceTest.DayNightWeatherMatrixRendersAndHoldsContracts`,
which fails with the heap **off** too — its AMD goldens are committed
`UNVERIFIED — see #735`. So the bindless path is now at exact parity with the
default path.

## What Phase 3 adds (earlier commits on this branch)

- A second descriptor kind: storage images, with their own reserved null. A null
  is needed **per kind** — `image2D(samplerHandle)` is undefined exactly the way
  `image2D(0)` is.
- `HeapBindingSeam` — the one place a bind forks into "write an offset" or "bind
  the old way", reachable from both render passes and the context-free compute
  systems.
- Publish-mode (`PublishTextureOffsetAndBind`) for engine-global slots whose
  consumers are not bound at publish time and can be of *both* kinds at once.
- `RetireResource` as the destruction counterpart to `InvalidateResource`.

## What the final commit fixes

None of the 13 failures were defects in #740's shader conversions.

1. **`HeapFixture::TearDown` shut down the process-wide heap singleton and never
   restored it.** A non-GPU unit test. Every test in that file passes either way —
   they drive the heap directly — so the damage was invisible at the crime scene
   and landed on six visual-evidence suites that run later and pass in isolation.
   A shader's bindless-or-not variant is fixed at **compile** time and cached, so
   switching the heap off afterwards does not rebuild those programs; it only
   stops the seam publishing the offsets they still read.
2. **The offset table was never re-based across a heap epoch bump.** Slots a pass
   did not re-stage kept indices minted by the *previous* heap. Latent rather than
   the cause of the 13. The scratch also started zeroed, pointing the image region
   at a sampler descriptor.
3. **`~OpenGLRendererAPI` tore the heap down at atexit**, with the GL context
   already gone — an access violation inside the driver with nothing left to log
   through. Now a neutral `RendererAPI::ShutdownGpuResources()` virtual. A virtual
   rather than a `dynamic_cast` in `RenderCommand.h`, because that header is
   included across most of the engine and naming the concrete GL type there would
   drag `glad/gl.h` behind it — undoing what this issue's include ratchet measures.
4. **`GLStateGuard::kUboSlots` had drifted** (48, while the registry had grown to
   56), leaving the heap offset table — the one binding every converted pass
   depends on — outside the leak detector built to catch exactly this.
5. **The post-process property harnesses bound their inputs with raw
   `glBindTextureUnit`**, which a bindless-variant program cannot see. They now
   compile slot-based via `ScopedSlotBasedShaders`. Teaching them to bind through
   the heap was tried first and abandoned — the heap's lifetimes assume a frame
   these harnesses never run, and each variant produced a different
   order-dependent failure. Coverage does not move: `BindlessHeapGpuTest` proves
   the seam with real texel readback, and the visual-evidence suites drive these
   same shaders through the real passes with the heap on.

## Known residual

One `GL_INVALID_OPERATION: Not a valid texture` is logged at **process shutdown**
under bindless only (0 under heap-off). It fails no test and is strictly better
than the crash it replaced — the context is still alive, so it is a logged error
rather than a fault. It indicates a texture-destruction path that does not call
`RetireResource`, which is worth finding properly rather than silencing with a
liveness guard. Not blocking; noted for a follow-up.

## Scope

Phase 3 is **not** closed by this PR — bucket 1 (the ~172 ordinary sampler-bind
conversions) remains reverted, because the recipe defects it exposed are what
this branch has been fixing. #691 stays open; Phases 4–9 remain. The ratchet
counter `sweep_bind_texture_sites` is unchanged at 87.

## Notes for review

Three diagnostic lessons are recorded in ADR 0011 amendments (33)/(34) and
`docs/agent-rules/rhi-abstraction-boundary.md` §4d, because each cost real time:

- A failure set whose **membership moves between runs** is evidence about
  ordering, not about the individual tests.
- A minimal repro that passes is not exoneration when the thing under test *is*
  order.
- Audit the identity/no-op tests when converting a pass.
  `MotionBlurStaticTest.ZeroVelocityIsIdentity` was green throughout while its
  velocity buffer never reached the shader — a texture the shader cannot see
  reads as zero velocity, which is precisely what it asserts.

## Summary by CodeRabbit

* **New Features**
  * Added bindless storage-image support for compute and rendering workflows.
  * Expanded heap-based texture and image binding across lighting, post-processing, terrain, fluid, cloud, ocean, and environmental effects.
* **Bug Fixes**
  * Improved GPU resource cleanup during shutdown, replacement, and invalidation.
  * Prevented stale bindings and descriptor conflicts.
* **Tests**
  * Added GPU and lifecycle coverage for storage images, rebinding, retirement, and shader compatibility.
* **Documentation**
  * Updated rendering and shader guidance for storage-image workflows.


drsnuggles8:
## CodeRabbit triage

Thanks — two of these were real bugs I had shipped, including one I had written up
as a "known residual" in the PR description. Worked through all 15.

### Fixed

| Finding | Note |
|---|---|
| `~OpenGLFramebuffer` never retires attachment descriptors (🔴 Critical) | Correct, and this **was** the known residual in the PR description. I had fixed the `Invalidate()`/resize path — which produced the visible bug, a flickering viewport and a crash on maximise — and missed the destructor, whose only symptom is one `Not a valid texture` at shutdown. Same defect, quieter symptom. |
| `s_BindlessPrograms` entries outlive their GL programs (🟠 Major) | Correct, and my bug. Registration had no paired retirement across ~13 `glDeleteProgram` sites; GL reissues program names, so a later slot-based program inherits `true`, the seam records an offset and skips the bind, and the pass renders with empty sampler bindings and **no diagnostic**. Added `Shader::UnregisterProgram` at every deletion site — unconditionally, since erasing an unregistered id is a no-op and per-site reasoning is how you miss one. |
| CloudShadowMap: retire before deleting the texture (🟠 Major) | Correct. Fixed one level down, in `OpenGLRendererAPI::DeleteTexture(ResourceHandle)`, so it covers every system that mints via `CreateTexture2DHandle` rather than only this one. `OpenGLTexture*`/`OpenGLFramebuffer` keep their own retire because they manage GL names directly. |
| `g_OloHeapOffsets` out of sync with the engine constants (🟠 Major) | Correct and worse than reported: the shared include was `[16]` (texture slots only) while the engine writes 18, so **every image unit indexed past the end of the array**. Cause: the revert commit rolled `BindlessHeap.glsl` back along with the shader conversions, undoing the growth *and* deleting the `OLO_HEAP_IMAGE*` macros — so the storage-image descriptor kind shipped with no GLSL support. Both restored. It survived because `BindlessHeapGpuTest` declares its own inline copy at the correct size: a test that restates a shared declaration cannot detect that the shared one is wrong. |
| Two shaders created outside the slot-based scope (🟠 Major) | Correct — my edit matched only the `m_Shader =` form and missed ToneMap and BloomDownsample. |
| `ScopedHeapInput` documents behaviour that does not exist (🟠 Major) | Correct. When I reverted the heap-participation approach I left an RAII-shaped type with a defaulted destructor and docs promising a scoped release. Replaced with a plain `BindSlotBasedInput()` function. |
| `readonly` listed as a valid `OLO_HEAP_IMAGE` qualifier (🟡 Minor) | Correct — it is the one qualifier that fails (`error C7504`, initialising a readonly local is a write). Doc now states the slot-path fallback. |

### Not a defect

**HZBGenerator `Persistent` on a pooled resource (🟠 Major).** `m_HZBTexture` is
created by the generator itself via `Texture2D::Create` and cached across frames
(reused when the size is unchanged) — it never comes from `context.Resolve*`, so
it is not a render-graph pool resource and `Persistent` is the correct lifetime.
On resize the texture is recreated and the old descriptors are dropped by the
`RetireResource` in the texture destructor. The `BindImageOrOffset` +
`Persistent` shape matches the GTAO rule's *form* but not its precondition.

### Accepted but deferred

Both are narrow, need care plus a full two-configuration re-verification, and I
did not want to make late unverified changes to a tree that is currently green.
Logging them rather than silently dropping them:

- **`OpenGLDescriptorHeap` image access when already resident.** Real: on the
  `RefCount == 0` path the code records `residency.Access = wanted` without
  enforcing it in GL when the handle is already resident. Narrow, because the
  release path goes non-resident at zero, so that branch is defensive rather
  than hot. Enforcing it needs a non-resident/re-resident transition, not just a
  second `glMakeImageHandleResidentARB` (which would be `INVALID_OPERATION`).
- **DDGI / FluidComposite explicit `FlushHeapOffsets`.** Not currently a defect —
  staged offsets are published by the next consumer's own flush, and the publish
  path binds as well, so slot-based consumers are unaffected either way. Adding
  the flush would match the "publish last" discipline and remove the dependency
  on consumer behaviour, which is worth doing on its own merits.

### Verification

Full suite, both configurations, after these changes — reported in a follow-up
comment once it completes. Prior to the fixes: bindless 5304 passed / 1 failed
(`AtmosphereVisualEvidenceTest`, which fails heap-off too, goldens committed
`UNVERIFIED — see #735`), heap-off identical.

**Result (ab01f8a9):** full suite, both configurations — **5304 passed / 1 failed**
in each, the same `AtmosphereVisualEvidenceTest`. `Not a valid texture` errors
under bindless went **1 → 0**, confirming the framebuffer destructor retire was
the cause of the residual noted in the PR description.

Issue #726, opened 2026-07-31
Editor: observer camera — freeze the render frustum and fly around to inspect culling


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

There is no way to inspect what culling actually did. We have `GPUFrustumCuller`, `OcclusionCuller`, `GPUDrivenOcclusionPass`, `DeferredGPUOcclusionPass`, two-phase virtual-geometry occlusion (#682) and shadow cluster culling — and every one of them can only be judged from inside the camera that did the culling, where a wrongly-culled object is invisible *by definition*. A grep for `FrozenFrustum` / `ObserverCamera` / `FreezeCulling` across the repo finds nothing outside one test file. #652 (shadow cull used the camera frustum instead of the light's) and #650 (light-facing face culled) are exactly the class of bug this makes obvious in one glance.

**Describe the solution you'd like**

Freeze the culling camera, then detach a second free-flying camera to look at the frozen result from outside:

- One toggle freezes the view/projection used for culling, LOD selection and Hi-Z; rendering continues from the observer.
- The frozen frustum draws as wireframe (via the shader-debug-draw work).
- Anything culled by the frozen camera stays culled and is therefore visibly missing when you fly around it — which is the whole point.
- Exposed over MCP alongside the existing `olo_camera_*` tools so an agent session can drive it.

**Describe alternatives you've considered**

Wireframe / overdraw / cluster-ID debug views — all render from the culling camera, so they cannot show what is not there.

**Additional context**

Reference: `D:\repos\Timberdoodle` — described in `README.md:42`, implemented across `src/camera.cpp` / `.hpp` and `src/application.cpp`, with the observer coordinate space threaded into the debug draws (`src/shader_shared/debug.inl:6-8`, `DEBUG_SHADER_DRAW_COORD_SPACE_NDC_VIEW_CAMERA`). Apache-2.0. Their README: *"invaluable when debugging various artifacts that were difficult to reason about from the point of view of the main camera"*.

Related: #725 (frustum wireframe + observer coord space), #607 (MCP capability gaps), #682, #652, #650, #654.

**Acceptance criteria**

- Freezing the culling camera and flying away shows the culled set from outside, with the frozen frustum drawn.
- An intentionally over-culling change is visibly diagnosable in one screenshot.
- Available over MCP.

## Score
```olo-score
capability: 3
craft: 1
stability: 5
decay: 1
effort: 2
confidence: 0.8
learning: 3
fun: 5
[...]

PR #749, opened 2026-08-04
feat(renderer): GPU-pushable shader debug draws (#725)

### LINKED ISSUE #725: Renderer: GPU-pushable shader debug draws (line / AABB / sphere / cone / box from any shader)

Shaders cannot draw. `Renderer/ShaderDebugUtils.h` is 15 lines and does one thing — toggle the shader cache. Debug geometry is CPU-only (`Renderer3DUtilityDraws.cpp`), which is exactly backwards for the direction the renderer is going: the cull decisions, cluster bounds, probe placements, page allocations and LOD choices that most need visualising are all *computed on the GPU* and never come back. Today the only way to see why a cluster was culled or where a probe relocated to is to add a debug colour output, reload, and screenshot — a loop `docs/agent-rules/volumetric-cloud-debugging.md` describes as the standard instrument precisely because nothing better exists.

A per-frame debug draw buffer any shader can append to:

- One ring buffer per primitive type — **line, circle, rectangle, AABB, box (8 explicit corners), cone, sphere** — each with `{ draw_indirect, capacity, request_count, data_ptr }`.
- A `coord_space` tag per entry: **world space / main-camera NDC / observer-camera NDC** (the last one pairs with the observer-camera issue).
- Appended from **both CPU and GPU** through the same buffer, so a CPU-side bound and its GPU-computed counterpart can be drawn together and compared.
- Drawn at end of frame with one indirect call per type.
- **Per-type overflow flags** read back so "I drew nothing" is distinguishable from "I overflowed the buffer" — the failure mode that otherwise wastes an hour.

Debug colour output into a spare render target — what we do now; one scalar per pixel, no world-space geometry, and it perturbs the pass being debugged.

**Additional context**

Reference: `D:\repos\Timberdoodle\src\shader_shared\debug.inl` (the whole contract is 130 lines — struct-per-primitive at :10-76, `ShaderDebugOutput` overflow flags at :84-96, the `DECL_DEBUG_DRAWS` macro at :98-114), rendering in `src/rendering/tasks/shader_debug_draws.hlsl` / `.inl`, GLSL helpers in `src/shader_lib/debug.glsl`. Apache-2.0. Their README calls the debug-draw + observer-camera pair *"invaluable"* for the culling work, and it is a stated dependency of essentially every GPU-driven feature they built after it.

This is a force multiplier for the VSM, virtual-texturing, DDGI-v2, GPU-quadtree and triangle-culling issues — all of which are otherwise debugged blind. Related: #607, #519, #316.

**Acceptance criteria**

- A compute shader can emit a world-space AABB that appears in the viewport in the same frame.
- Overflowing a channel raises a visible flag rather than silently dropping draws.
- At least one existing GPU-driven pass (cluster cull or DDGI probes) ships a debug visualization built on it.
- Zero cost when disabled.

## Score
```olo-score
capability: 8
craft: 3
stability: 5
decay: 3
effort: 3
confidence: 0.8
learning: 5
fun: 8
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #725.

Any shader can now draw. The cull decisions, cluster bounds, probe placements and page allocations that most need visualising are computed on the GPU and never came back — the only instrument was "add a debug colour output, reload, screenshot" (the loop `docs/agent-rules/volumetric-cloud-debugging.md` documents precisely because nothing better existed). Now:

```glsl
#define OLO_DEBUG_DRAW_AABB
#include "include/DebugDrawCommon.glsl"
...
OloDebugDrawAABB(worldMin, worldMax, vec3(1.0, 0.0, 0.0), OLO_DEBUG_SPACE_WORLD);
```

and it appears in the viewport **in the same frame**, depth-tested against real scene geometry.

Seven primitives — line, circle, rectangle, AABB, box (8 explicit corners), cone, sphere — one append channel each at SSBO 46..52, drawn by `ShaderDebugDrawPass` with one `glDrawArraysIndirect` per type as the last SceneColor writer.

## The buffer contract

Full detail in [`ShaderDebugDrawTypes.h`](OloEngine/src/OloEngine/Renderer/Debug/ShaderDebugDrawTypes.h). Three properties are load bearing.

**The channel buffer IS the indirect draw.** Each channel opens with a 32-byte header whose first 16 bytes are a GL `DrawArraysIndirectCommand`, read at offset 0 of that same buffer. The primitive count never travels through the CPU — which is what makes a same-frame visualization possible at all.

**Two counters, and the overflow flag falls out for free.**

```glsl
uint slot = atomicAdd(RequestCount, 1u);   // UNCLAMPED
if (slot >= Capacity) return;              // dropped, but COUNTED
Entries[slot] = entry;
atomicAdd(InstanceCount, 1u);              // only accepted pushes draw
```

The order is the contract. Bumping `InstanceCount` only on the accepted path is what keeps the indirect draw in range — it can never cover an unwritten slot. And `RequestCount > Capacity` is exactly "you overflowed and N draws were dropped", with no separate flag word and no extra atomic on the overflow path. That distinction — "I drew nothing" vs "I overflowed and the rest was silently thrown away" — is the hour-wasting failure the issue calls out.

**Coordinate spaces**: world / main-camera NDC / observer-camera NDC. The observer camera is #726 and does not exist yet, so the pipeline feeds that space the *main* camera's inverse view-projection. That makes it an exact identity round-trip to `MainCameraNDC` rather than a source of garbage geometry, and leaves one line to change when #726 lands.

## Zero cost when disabled

Helpers open with a plain, non-atomic `Capacity == 0u` test and return, so a push site costs one scalar load — no atomic, no write, no bandwidth. Nothing is cleared, uploaded, read back or drawn, and the pass declares no graph resources at all.

The channels do stay **allocated and bound** while disabled, and that is load bearing rather than sloppy: reading an unbound SSBO is undefined in GL (the spec permits program termination), so the guard read needs a real buffer underneath it. 7 × 32 bytes is the price of the guard being safe.

## Channels are opt-in per primitive

Not politeness — storage blocks are a scarce per-stage resource. GL guarantees only 8 per stage and NVIDIA's compute limit is 16, and `VirtualClusterCull.comp` already declares 9 of its own. A header that declared all seven unconditionally would have failed to link there. `OLO_DEBUG_DRAW_ALL` exists for the draw-side shader, which by definition needs them all.

## Shipped consumer

`VirtualClusterCull.comp` emits each cluster's world-space cull sphere colour-coded by which test decided it — drawn (green) / frustum-culled (red) / cone-culled (blue) / Hi-Z occluded (yellow) — so "why did that cluster disappear" is answerable directly instead of by elimination.

Two details worth copying into the next consumer: it emits on **phase 1 only** (phase 2 re-tests an already-classified subset and would double-emit, so the phase-2 dispatch sets the uniform to 0 *explicitly* — uniform state is per program and persists across dispatches), and **perspective only** (the shadow cascades bind the same program with `u_OrthoMode = 1`). It is stride-sub-sampled, because a Nanite-class cut has far more clusters than any channel holds.

## Reachable from everywhere it needs to be

`RendererSettings::ShaderDebugDrawEnabled`, the Renderer Settings panel (with a live per-channel overflow readout), the F3 overlay (the one reachable in a shipped build), and a new `olo_shader_debug_draw` MCP tool. That last one matters: an agent debugging a GPU-driven pass has to be able to flip this itself, or it is back to the debug-colour-output loop.

## Verification

**Live editor** (Sponza, Deferred path): 197–988 spheres pushed by the cull compute and drawn in the same frame, correctly occluded by the geometry they describe, from several camera angles. Zero GL errors in `OloEngine.log`.

`GLStateGuard` earned its keep during bring-up — it caught two genuine leaks (a left-behind `GL_LEQUAL` depth func and cull-face disabled), both now restored via `ResetOpaqueForwardDrawState()`. Its remaining diff is only the deliberate program/VAO unbinds.

**Tests** (61 green):

| test | what it pins |
|---|---|
| `ShaderDebugDrawContractTest` | struct sizes/offsets, binding derivation, the segment-count literals cross-checked against the shader **source**, and the two-counter overflow protocol |
| `ShaderDebugDrawExpansionTest` | the primitive→segment expansion, **analytically** (every ring point at `radius` and in-plane; an AABB's 12 edges each axis-aligned and full-extent) rather than against a recorded blob — a blob would pass for a wrong-but-stable implementation |
| `ShaderDebugDrawGpuPushTest` | dispatches a real compute shader through the production include and confirms the overflow protocol **on hardware**: 500 pushes into a 64-slot channel → `requested 500, drawn 64, dropped 436` |
| `ShaderDebugDrawVisualTest` | all seven primitives reach the composited frame; a buried sphere is depth-culled while a control sphere in open air is not; disabled draws nothing |

That control sphere is deliberate. The first version of the depth-test assertion checked only that the buried sphere was invisible — which passes just as happily when the whole feature is broken and nothing draws. Adding the control turned a vacuous pass into a real one (and immediately exposed a bug in the test's own colour classifier, which was counting green pixels as orange).

Also bumps `kHighestKnownSSBOBinding` in `ShaderReflectionBindingTest`, which was still pointing at slot 44.

## Acceptance criteria

- [x] A compute shader can emit a world-space AABB that appears in the viewport in the same frame — verified live and by `ShaderDebugDrawGpuPushTest` + `ShaderDebugDrawVisualTest`.
- [x] Overflowing a channel raises a visible flag rather than silently dropping draws — `RequestCount > Capacity`, surfaced in the settings panel, the F3 overlay, the MCP response, and a warn-once log line; pinned on hardware.
- [x] At least one existing GPU-driven pass ships a debug visualization built on it — the cluster cull.
- [x] Zero cost when disabled — one scalar load per push site, nothing per frame on the CPU, no graph declaration.

## Notes for the reviewer

`AtmosphereVisualEvidenceTest.DayNightWeatherMatrixRendersAndHoldsContracts` fails on this branch, and it is **not** from these changes. I neutralised this feature's three always-on effects (channel allocation, pass creation, the fingerprint term) and rebuilt: the RMSE came back bit-identical (13.0777068470475). It is pre-existing golden drift on the base commit. Rest of the suite: 5491/5496 pass, 4 environmental skips.

Docs: [`docs/agent-rules/gpu-debug-draws.md`](docs/agent-rules/gpu-debug-draws.md) — how to push, how to read the overflow flag, what the coord-space tags mean, and the traps that cost time here (the `SHADER_STORAGE|COMMAND` barrier whose absence presents as "the first frame after a push draws nothing, and then it works"; why a channel must never be read directly — the `GL_DYNAMIC_COPY` VIDEO→HOST migration trap; why "zero cost when disabled" still requires the buffers to stay bound).

## Summary by CodeRabbit

* **New Features**
  * Added GPU-driven shader debug drawing for lines, shapes, and virtual-geometry cluster bounds.
  * Added controls for overlays, line width, culling outcomes, and sampling stride.
  * Added per-channel draw, drop, and overflow statistics.
  * Added render-tool support for configuring and inspecting shader debug draws.

* **Documentation**
  * Added GPU debug-drawing integration and troubleshooting guidance.
  * Clarified headless MCP write-consent configuration and documented additional render tools.

* **Tests**
  * Added coverage for rendering, GPU pushing, overflow handling, buffer contracts, and shader bindings.


drsnuggles8:
Thanks — one of these was a real bug that my own live verification had failed to isolate. Addressed in cecfd72d.

## The important one: the enable was not invalidating the fingerprint

You were right, and my comment asserting the opposite was the giveaway. `HashPassState` hashes the pass pointer and `IsReadyForExecution()`, never `IsEnabled()` — its own comment even says per-pass enables are expected to be hashed separately, exactly as `PostProcess.OverdrawDebugView` is. Since `ShaderDebugDrawPass::Setup` declares nothing while disabled, the enable gates a graph declaration, so the caches short-circuited and the pass stayed undeclared.

Worth recording *why* I did not catch this: I did verify the toggle live in the editor and saw the geometry appear — but in that same session I had also opened a scene and switched rendering path, **both of which move the fingerprint on their own**. The observation was real; the isolation was not.

Fixed by hashing `data.Settings.ShaderDebugDrawEnabled`, and pinned by a new test, `ShaderDebugDrawVisualTest.EnablingAfterTheGraphCacheIsWarmStillDeclaresThePass`. Its shape is the point: it runs frames with the feature **off first** so the blackboard and frame-graph caches are warm, then changes only the enable. Enabling from a cold start passes either way, so the warm-up is the load-bearing part.

I checked it is not vacuous — with the hash line removed it fails at `0 vs 50` matching pixels, then passes again with it restored.

## Also fixed

| finding | what changed |
|---|---|
| Unbounded CPU-push capacity / `u32` byte overflow | Clamped at **every** sizing path, including inside `EnsureChannelCapacity` so `channel.Capacity` and the allocation can never disagree — that disagreement was the actual out-of-bounds `SetData`. Byte math now in `u64`. `SetChannelCapacity` shares the same ceiling instead of a second literal. |
| `ClearCpuEntries` without `CpuMutex` | Locked on the disabled path and in `Shutdown`. You are right that the `IsEnabled()` early-out only makes a race *unlikely* — a worker already inside `push_back` when the toggle flips is not excluded, and "unlikely" is not an invariant. |
| Missing `<array>` | Added directly; it was arriving transitively through `ShaderDebugDrawTypes.h`. |
| Recomputed inverse view-projection | Now reuses `data.InverseViewProjectionMatrix`, which `PrepareFrame` already computed this frame. |
| `DisabledDrawsNothing` order dependence | Warm-up frame moved **before** the pushes, plus an `ASSERT_FALSE(ShaderDebugDraw::IsEnabled())` so the rest of the test cannot run vacuously. Your reasoning about *why* it could break was not quite right (`BeginFrame` drops the staged entries before the graph executes, so the pushes could not have drawn) — but the conclusion was: the test was inheriting enable state from whichever test ran before it, and that is not a property worth depending on. |
| `stbi_write_png` return discarded | Asserted. The rendering rules make the PNG the evidence, so a silent write failure must not leave the suite green. |
| `EXPECT_TRUE` for segment counts | `EXPECT_EQ`, so a drift prints the expected-vs-actual numbers that are the entire diagnosis. |
| F3 overlay described as carrying the full counters | Corrected — it carries the overflow lines only. |

## Skipped, with reasons

**`ApplyRendererSettings()` on the enable change** — wrong mechanism. That function clamps MSAA and applies the wireframe/path state; it is not what rebuilds the graph, and no other debug toggle uses it for that (`ShowBoundingBoxes` does not call it; `OverdrawDebugView` goes through the fingerprint). The fingerprint hash above is the engine's documented route for "an enable gates a declaration", and it is now verified end to end.

**Synchronising the staged readback in the GPU probe** — `glGetNamedBufferSubData` is a blocking, synchronising read, and GL commands execute in order, so the `CopyBufferSubData` issued before it has completed by the time it returns. The `SHADER_STORAGE | BUFFER_UPDATE` barrier already orders the shader's atomic writes ahead of that copy. Adding a fence would not change the guarantee.

## Suite state

62 targeted tests green. Full suite 5491 pass, 4 environmental skips, and two failures neither of which is from this branch:

- `AtmosphereVisualEvidenceTest` — pre-existing golden drift, bisected by neutralising this feature's three always-on effects and rebuilding: bit-identical RMSE (13.0777068470475).
- `ModelLoadDeterminism.Backpack_FreshVsCache` — failed once in-suite, passes in isolation; an order-dependent mesh-cache-write flake. Nothing here touches `Model` or `MeshCache`.

The fingerprint trap is now written up in `docs/agent-rules/gpu-debug-draws.md`, since "works only after you happen to also change something else" is a failure mode worth naming.

PR #750, opened 2026-08-05
feat(renderer): offline reference path tracer for ground-truth GI/BRDF validation (#709)

### LINKED ISSUE #709: Renderer: offline reference path tracer for ground-truth GI / BRDF validation

We have no ground truth. CLAUDE.md is explicit that rendering changes can pass every contract test and still be wrong on screen, and the response has been golden images plus multi-angle screenshot review — both of which answer "did it change?" but never "is it *correct*?". The PBR probe shaders under `OloEditor/assets/shaders/tests/` (`PbrFurnaceProbe`, `PbrBrdfProbe`, `PbrReciprocityProbe`, ...) test analytic invariants, which catches gross BRDF errors but cannot validate multi-bounce GI, energy conservation across the full pipeline, or whether DDGI's irradiance is actually right.

An offline (non-realtime) path tracer sharing the engine's scene, materials and BRDF, rendering a converged reference image for a given camera pose. Not a shipping feature — a validation instrument:

- Reuses `include/PBRCommon.glsl`'s BRDF so a divergence between raster and reference *is* the bug.
- Converges over N frames into an accumulation buffer; dumps PNG/EXR.
- Drives a test mode: render the same scene through the raster pipeline, compare against the converged reference within a tolerance.

Useful immediately for the DDGI v2 and denoiser issues, both of which are hard to evaluate without a reference.

Keep relying on golden images — they lock in whatever we shipped, including bugs; a golden rebase after a "looks fine" change silently blesses a regression.

**Additional context**

Reference: `D:\repos\Timberdoodle\src\rendering\path_trace\` — `path_trace.hlsl` / `.inl` plus a vendored BRDF library under `path_trace/kajiya/` (`brdf.hlsl`, `layered_brdf.hlsl`, `brdf_fg.hlsl`, `quasi_random.hlsl`, `pack_unpack.hlsl`, `ray_cone.hlsl`). Apache-2.0. Their tracing uses hardware RT; ours would trace against the existing `Renderer/BoundingVolumeHierarchy` in compute, which is slower but fine for an offline reference.

Related: #258 (screenshot tests in CI), #549/#563 (visual-test order dependence — a reference image is order-independent by construction).

**Acceptance criteria**

- A converged reference for a Cornell-box-style scene matches the analytic furnace test within tolerance.
- At least one existing GI feature (DDGI or SSGI) is validated against the reference and any divergence is either fixed or documented.
- Runs headless so it can gate CI on a small scene.

## Score
```olo-score
capability: 5
craft: 3
stability: 5
decay: 1
effort: 5
confidence: 0.8
learning: 8
fun: 8
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #709.

## The problem

We had no oracle. CLAUDE.md is explicit that a renderer change can pass every contract test and still be wrong on screen, and the answer so far has been golden images plus multi-angle screenshots — both of which answer *"did it change?"* but never *"is it correct?"*. Golden images are worse than neutral on that question: a rebase after a "looks fine" change silently blesses a regression.

This adds an unbiased offline path tracer that shades with the engine's **own** BRDF and computes transport by brute force, so a raster-vs-reference divergence *is* the bug.

## What's here

`OloEngine/src/OloEngine/Renderer/PathTracing/`

| file | what |
|---|---|
| `ReferenceBRDF.h` | function-for-function port of `PBRCommon.glsl` |
| `PathSampler.h` | stateless Owen-scrambled Sobol' |
| `ReferenceScene.{h,cpp}` | instance TLAS over the existing per-mesh `BoundingVolumeHierarchy` |
| `PathTracer.{h,cpp}` | NEE + power-heuristic MIS + Russian roulette, film + display transform |

29 new tests (23 headless, 6 GPU-gated with clean SKIPs). Full run below.

### The BRDF port is deliberately unfaithful to textbook math

It reproduces **what the renderer computes**, not what it should — the 1e-4 NDF denominator clamp (which collapses the specular lobe below roughness ≈0.27 rather than making it a mirror), the non-height-correlated Smith, `kD = 1 - F` at the half vector. Each quirk is pinned by a test so a "cleanup" surfaces there rather than as a silent brightness change.

`ReferenceBRDFGpuParityTest` renders a new probe shader that calls the production `cookTorranceBRDF` over a (roughness × metallic × N·L) grid and diffs it against the C++ port at 1% relative tolerance — **49,152 samples, all matching**. Without that guard, editing `PBRCommon.glsl` would leave the reference integrating the old formula: still converging beautifully, still "passing", measuring a renderer that no longer exists.

### Determinism is a contract, not a nicety

Renders are **bit-identical at any thread count**, asserted as an exact hash. A reference reproducible only up to thread scheduling would force its own gate to use a tolerance, and that tolerance would have to be loose enough to swallow scheduling noise — exactly the band a real regression hides in.

It rests on two things: `PathSampler` is stateless (pure function of pixel/sample/dimension/seed), and a pixel's samples are summed by one thread in ascending order. `ASinglePixelReproducesTheFullRenderExactly` pins it in the form that does *not* depend on how many workers happen to be running.

### Why CPU, not compute

The issue suggested tracing against the BVH in compute. A GPU tracer's only gain here is speed — an offline reference is offline by definition — and it costs a scene TLAS upload, a compute traversal, a dependency on a live GL context (which kills "runs headless so it can gate CI", the actual point), and a collision with the in-flight #691 RHI work.

## Acceptance criteria

**1. Converged Cornell-box reference matches the analytic furnace test.** ✅

A single plane under a uniform environment, viewed along its normal, is a convex single-scatter configuration, so its converged radiance is *exactly* the directional albedo `∫f(l,v)(n·l)dω` — the identical quantity `PbrFurnaceProbe.glsl` estimates on the GPU. Measured against an independent uniform-hemisphere estimate:

| roughness | metallic | analytic | traced |
|---|---|---|---|
| 0.10 | 0 | 0.96130 | 0.96139 |
| 0.10 | 1 | 0.03472 | 0.03475 |
| 0.25 | 1 | 0.89252 | 0.89309 |
| 1.00 | 0 | 0.97222 | 0.97235 |
| 1.00 | 1 | 0.30683 | 0.30657 |

Same `[0.6, 1.05]` energy band `PbrBrdfTest.FurnaceIntegralWithinEnergyBounds` applies on the GPU, so the two deliver the same energy *verdict*, not merely two similar numbers. The multi-bounce Cornell furnace converges to 0.858 with no energy created.

**2. DDGI validated, divergence documented.** ✅ — see below.

**3. Runs headless so it can gate CI.** ✅ — 23 of 29 tests need no GL; the whole headless set is ~5 s of work (65 s under MSVC Debug, which is ~40× slower).

## Finding 1 — the instrument caught a bug in itself before it ever ran

`PdfGGX` was first built on the same `DistributionGGX` the BRDF evaluates, on the reasonable-sounding grounds that the sampled PDF and the evaluated D then can't drift apart. That is exactly backwards: `ImportanceSampleGGX` inverts the **unclamped** GGX CDF and has never heard of `EPSILON`. Below roughness ≈0.27 the clamp caps D right where the sampler concentrates, so the PDF under-reports and `f/pdf` explodes:

| roughness | metallic | analytic | clamped PDF | correct PDF |
|---|---|---|---|---|
| 0.10 | 0 | 0.961 | **2.179** | 0.962 |
| 0.10 | 1 | 0.035 | **1.000** | 0.035 |
| 0.50 | 1 | 0.860 | 0.860 | 0.860 |

A reference returning 2.18 for a directional albedo is *creating energy* — it would have blessed every too-bright specular bug it exists to catch. Note the shape: invisible above roughness 0.5, converges beautifully, looks like a plausible image. **A PDF describes the sampler, never the integrand.**

## Finding 2 — DDGI: an air-fitted probe volume kills the entire bounce term

`DDGIReferenceParityTest` compares **irradiance at a probe** rather than pixels — `DDGI_BlendIrradiance.glsl` pins its storage convention in a header comment ("the atlas stores full irradiance E", via `E = π·Σ(wL)/Σ(w)`), so the atlas texel and the traced integral are the same number in the same units. No calibration constant.

| probe | DDGI | ref (4 bounces) | ref (direct only) | DDGI/full | DDGI/direct |
|---|---|---|---|---|---|
| (0, 0.8, 0) | 0.230 | 0.446 | 0.246 | 0.52 | **0.93** |
| (0, 3.2, 0) | 0.238 | 0.443 | 0.237 | 0.54 | **1.003** |
| (−3, 3.2, 0) | 0.342 | 0.560 | 0.341 | 0.61 | **1.005** |
| (3, 3.2, 0) | 0.097 | 0.236 | 0.098 | 0.41 | **0.989** |

DDGI's **direct** transport is essentially exact. It contributes **zero** multi-bounce.

The mechanism is a two-line chain, invisible from either end:

- `DDGI_Relight.glsl` computes the bounce as `ddgiSampleIrradiance(u_PrevIrradiance, …, hitPos, …)` — irradiance at the cached **hit point**.
- `ddgiSampleIrradiance` opens with `if (!ddgiIsInsideVolume(worldPos)) return vec3(0.0);`.

Every cached hit point is **on a surface**. A volume fitted to the interior air excludes every wall, floor and ceiling — every surface the bounce was supposed to come from. And air-fitted is the *natural* authoring: it is what `DDGITest.olo` and `DDGIVisualEvidenceTest` both do.

Confirmed by construction — `DDGIReferenceParityWideVolumeTest` runs the identical room and reference, changing only the bounds:

| volume | DDGI/direct | DDGI/full |
|---|---|---|
| fitted to the air | 0.93 – 1.005 | 0.41 – 0.61 |
| enclosing the walls | **1.74** | **1.12** |

**Documented, not fixed**, and filed separately: `ddgiIsInsideVolume` is also the guard that keeps the feedback contractive (ADR 0007), so the fix is a design call for #632's owner, not a patch.

### A methodological note

DDGI's radiance cache is diffuse-only *by design* (a probe hit point has no view direction). So the reference is rendered in **both** shading models: DDGI is asserted against the Lambertian one, isolating its transport, and the Lambertian-vs-`cookTorranceBRDF` gap is reported separately. Folded together, their sum would have been reported as "DDGI's error" and the factor of two would have been far harder to attribute. That gap measures **0.7%** here — negligible, and now *known* to be negligible rather than assumed.

## Verification

```
29 tests from 7 test suites ran. PASSED 29.
52 tests (DDGI* + Pbr*) ran. PASSED 52.   # existing goldens + PBR probes unaffected
```

Visual evidence: `PathTracer_CornellBox.png` / `…_DirectOnly.png` at gate resolution, plus an opt-in converged 192×192 via `OLO_PATHTRACER_EVIDENCE=1`. I inspected a 384² render directly — red/green bleeding, dark ceiling ring around the one-sided emitter, soft block shadow, no fireflies.

## Notes for review

- The headless set is 65 s under Debug — ~5 s of real work; `ParallelFor` runs inline because the test process starts no `FScheduler` workers. Budgets are single constants at the top of each file with the measured margin recorded beside every threshold.
- `CornellFurnaceCreatesNoEnergy`'s sample count is a **correctness** parameter, not a quality one: at 128 spp the correct render peaks at 1.069 purely from variance. The max of a Monte Carlo image is a noise statistic, so the energy claim is made on the frame mean.
- Lessons in `docs/agent-rules/reference-path-tracer.md`, linked from CLAUDE.md's companion-guides list.

## Summary by CodeRabbit

- **New Features**
  - Added a deterministic CPU reference path tracer supporting physically based materials, multiple light types, environment lighting, indirect illumination, and emissive surfaces.
  - Added scene construction, ray intersection, visibility testing, sampling, film output, and irradiance estimation capabilities.
  - Added documentation covering path-tracing workflows, GPU parity, DDGI validation, and configuration guidance.

- **Tests**
  - Added comprehensive CPU and GPU validation for rendering determinism, BRDF behavior, energy conservation, lighting, color bleeding, and DDGI irradiance parity.


drsnuggles8:
All 12 addressed in `becbb81c`. 29/29 tests still pass with identical measured values, and the existing DDGI golden + PBR probe suites are unaffected.

Two were real correctness bugs, and both were worth catching:

**Mirrored transforms passed the uniform-scale check** — correct, and it is the sharper version of a check I had already written. `glm::length` is never negative, so a reflected basis reads as a clean uniform scale while flipping winding, inverting every geometric normal, and making a one-sided emitter fire from its back face. Rejected now by a determinant test, next to the non-uniform one.

**The determinism test's divide vs. `Render`'s reciprocal multiply** — a genuine latent flake in the file's own load-bearing assertion, and exactly the kind of thing that would have surfaced later as an unreproducible CI failure. Fixed to mirror `Render`'s arithmetic form rather than its algebra.

Also done: unconditional TLAS pushes behind `static_assert(s_TraversalStackSize >= s_MaxDepth + 2)` + a runtime assert; `m_Built` asserts on both query entry points (which catch *mutated-after-`Build()`*, not just never-built); DDGI ranking moved onto the direct-only reference with its divisor guarded; three stale comments corrected.

### Two places I did something different from the suggestion

**`Render` "does not resize the film".** The finding is right that code and comment disagreed, but the proposed fix — resize to "the configured render dimensions" — has nothing to resize to: `PathTracerSettings` carries no width/height, and the film's extent *is* the render resolution by design. So I corrected the header contract and added the diagnostic the finding was really pointing at: a default-constructed film was a silent no-op return that looks identical to "the scene is black". It now logs.

**The 96×96 doc conflict.** Real conflict, but the proposed wording attributes 96×96 to `PathTracerFurnaceTest`, which doesn't render that either — nothing does, since I cut the budgets from 203 s to 65 s before opening the PR. Replaced with a per-test table of the actual figures, plus the measured ~40× MSVC Debug penalty and why the budgets look small.

### On the Debug-runtime finding

Already measured and acted on before review: 203 s → 65 s. Now recorded per-test in `docs/agent-rules/reference-path-tracer.md` §7 so the next person doesn't re-derive it. I deliberately did **not** add Debug/Release-conditional sample counts — that would make the gate's coverage depend on build configuration, and `CornellFurnaceCreatesNoEnergy`'s sample count is a *correctness* parameter, not a quality one (at 128 spp the correct render peaks at 1.069 purely from variance; the max of a Monte Carlo image is a noise statistic).

### On the second camera angle

Took the point — CLAUDE.md does require multiple inspected angles. My first raking pose was bad: ~18% of the frame landed outside the room. The one I kept sits *inside*, looking down across the floor, where the block's left face reads visibly red and its right face green — the clearest available read on whether indirect light carries wall albedo. Head-on covers what that pose cannot (the emitter, and the ceiling that is lit by bounce alone). I rendered and looked at both rather than assuming.

The HiRes pair is gitignored: it regenerates only under `OLO_PATHTRACER_EVIDENCE=1`, so a committed copy would silently go stale against the tracer it is evidence for. The gate-resolution PNGs stay committed — every test run rewrites those.

### One note for human reviewers

`PbrBrdfParityProbe.glsl` was excluded from this review by the `!**/*.glsl` path filter. It is the probe that makes the GLSL↔C++ anti-drift guard work — without it, editing `PBRCommon.glsl` would leave the reference silently integrating the old BRDF. Worth a manual look.

PR #795, opened 2026-08-14
feat(platform): Steamworks integration — achievements, Cloud, rich presence (#644)

### LINKED ISSUE #644: Platform: Steamworks SDK integration (achievements, Steam Cloud, overlay/rich-presence)
## Problem

There is no Steamworks SDK integration. For a polished Steam release the engine lacks achievements, Steam Cloud saves, the Steam overlay hooks (rich presence), and Steam auth/ownership.

Surfaced by an engine capability survey (what can ship on Windows/Steam today).

## Important scope note

This is **not** required to publish on Steam — a game can ship on Steam with zero Steamworks integration (Windows build uploaded via SteamPipe; the overlay attaches automatically to most apps). This issue is the **polish layer** expected of a real Steam title, not a launch blocker.

## Current state

- No `SteamAPI_Init`, achievements, overlay, or Steam Cloud anywhere in the engine.
- **Do not confuse with GameNetworkingSockets** — Valve's GNS transport lib *is* integrated (`OloEngine/CMakeLists.txt`, `Networking/Core/NetworkManager.cpp`, `Networking/Transport/*`), but that is the engine's own netcode transport, not Steamworks platform integration.
- Prior context: issue #317 (CLOSED) dealt with a `SteamNetworkingSockets` + MSVC ASan startup crash — again networking, not platform integration.

## Proposed work (suggest slicing)

1. **Foundation** — vendor the Steamworks SDK, `SteamAPI_Init/Shutdown`, `steam_appid.txt` for dev, run `SteamAPI_RunCallbacks` on the game loop. Graceful degradation when Steam isn't running (mirror the C#-scripting-unavailable pattern — log + disable, don't crash).
2. **Achievements / stats** — `ISteamUserStats` wrapper + script bindings (`Steam.UnlockAchievement("id")`), editor-side achievement definitions.
3. **Steam Cloud** — route the existing SaveGame system's files through Steam Cloud (auto-cloud or ISteamRemoteStorage).
4. **Rich presence / overlay** — set rich-presence strings; confirm overlay renders over the OpenGL context.
5. Consider Windows-only guard (matches the stated target: Windows + Steam only).

## Notes

Lowest priority of the shipping gaps for *getting a game on Steam*, but the highest for making it feel like a finished Steam product. Good candidate to defer until a real game project needs it.

🤖 Filed from a capability survey with [Claude Code](https://claude.com/claude-code)

## Score
```olo-score
capability: 3
craft: 3
stability: 1
decay: 1
effort: 5
confidence: 0.5
learning: 3
fun: 3
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #644 — Steamworks platform integration: achievements, Steam Cloud, rich presence and overlay state.

## What is and is not verified

Verification here is **three-tiered**, which is unusual for this repo and is the most important thing to read before reviewing. It follows from one fact: **this repository is public and the Steamworks licence forbids redistribution**, so the SDK can never be vendored and can never reach a CI runner.

| Tier | Covers | Where |
|---|---|---|
| **CI, every push** | The OFF path, **and the ON path** — compiled, linked and executed against a hand-written stub SDK | `steam-stub.yml` + the normal suite |
| **This machine only** | The real-SDK `OLO_WITH_STEAM=ON` build, and live behaviour against App ID 480 | evidence below |
| **Not verified at all** | Steam **overlay rendering** (OpenGL or Vulkan); achievement **unlock** against live Steam | — |

I am not claiming the overlay works. It hooks the swapchain, there are two RHI backends, and verifying it needs the game launched *through* Steam. Not done.

## Live evidence — App ID 480 (Spacewar), real Steam client

Editor launched with `cwd = OloEditor/`:

```
[Steam] Initialized. AppID=480, user='drsnuggles'.
```

That proves, in order: the process started at all (so `steam_api64.dll` resolved at load), `SteamAPI_InitEx` reached the running client, `steam_appid.txt` was found, and `ISteamUtils::GetAppID()` / `ISteamFriends::GetPersonaName()` returned live data. The editor then ran ~1.5 min with the per-frame `SteamAPI_RunCallbacks` pump: no Steam warnings, log advancing normally.

A direct SDK probe (scratchpad tool, not committed; read-only w.r.t. achievements) settled what the docs would not:

```
EnabledForAccount : yes      FileWrite  : ok
EnabledForApp     : yes      FileRead   : 21 bytes, content MATCHES
GetQuota          : ok       FileDelete : ok
  total bytes     : 4096

ACH_WIN_ONE_GAME              known=yes unlocked=no
ACH_TRAVEL_FAR_ACCUM          known=yes unlocked=no
ACH_DEFINITELY_NOT_A_REAL_ID  known=NO  unlocked=-
```

- **`ISteamRemoteStorage` works on 480** — so the SaveGame mirror is built on a mechanism verified against real Steam, not assumed.
- **The `NotFound` mapping is correct against real Steam**: genuine ids report `known=yes`, a made-up id reports `known=NO`. That distinction is what turns a typo'd achievement id into a loud warning instead of silence.
- **Spacewar's cloud quota is 4096 bytes.** Real saves are far larger, so an end-to-end save round-trip through Steam Cloud at realistic size **cannot** be tested on 480 — that needs a real App ID. This is exactly why the cloud/local reconciliation rules are covered by fake-backed unit tests rather than the live probe.

Nothing was left on the account: cloud file deleted, rich presence cleared, no achievement unlocked.

## The shape of the change

All real Steamworks calls live in **exactly one TU** (`Platform/Steam/SteamworksBackend.cpp`) behind `ISteamBackend`. That is load-bearing: it makes the logic that actually has bugs testable against a fake with no Steam client, keeps the stub SDK a sufficient stand-in for CI, and means `OLO_WITH_STEAM=0` changes nothing about the link surface. Enforced by a CI grep *and* structurally — the SDK include dir is `PRIVATE`, so the header does not even resolve elsewhere.

**Two fakes at two levels, not redundant** — please do not collapse them in review. `FakeSteamBackend` stands in for the *backend* so the engine's logic can be tested; the stub SDK stands in for *Valve's headers* so the one Valve-calling TU compiles and links in CI.

**Degradation is the contract.** Compile-time (`OLO_WITH_STEAM=0` → `NullSteamBackend`) and run-time (`SteamAPI_InitEx` fails → warn, disable, **return**). `Application.cpp` throws `std::runtime_error` when AudioEngine or NetworkManager fail; Steam deliberately does not join that club, and that is asserted in CI on both paths.

`SteamManager` is a process-level static manager, not an ECS component, so this change is clear of the entire hand-maintained cross-binding touch-point set.

## Five bugs found while building it — four would have shipped silently

All documented in the new `docs/agent-rules/steamworks-platform-integration.md`.

1. **`STEAMWORKS_SDK_ROOT` one level too high silently disabled Steam.** The zip has an outer folder, so this is the likeliest user error — and the message explaining the correct layout only fired when the option was *already* forced ON, so the person who made the mistake was guaranteed not to see it. Now warns and prints the corrected path in native separators.
2. **The single-TU CI guard was red on day one** — GameNetworkingSockets also ships under `steam/`.
3. **Every executable needs `steam_api64.dll` at process load**, including `OloServer` and `OloEngine-Tests` which never call Steam: the SDK ships an *import* library and Windows resolves imports before `main`. Symptom is a bare dialog with no log line.
4. **`set_source_files_properties` silently did nothing** from the wrong CMakeLists — directory-scoped, no error, the flag just never reaches the compiler. Replaced with a `#pragma` in the file.
5. **The SDK include directory hijacked GameNetworkingSockets' headers engine-wide.** The serious one. The two SDKs share *file names*; putting `<sdk>/public` on the engine include path made the netcode's `<steam/isteamnetworkingutils.h>` resolve to Valve's copy while `<steam/steamnetworkingsockets.h>` still came from GNS — half of each SDK in one TU, 12 × `C2365` reported *inside a Steamworks header* from files that never mention Steam. Fixed by rooting the include dir at `<sdk>` and including `"public/steam/steam_api.h"`, so no `steam/` prefix is resolvable at all.

On #5, my first diagnosis was **wrong**: I blamed unity builds and wrote a `SKIP_UNITY_BUILD_INCLUSION` entry before finding `OLO_ENABLE_UNITY_BUILD:BOOL=OFF` in the cache — MSBuild was just interleaving parallel output. The bogus fix is reverted and the doc explicitly warns against that same wrong turn.

## Steam Cloud

Local stays the source of truth; the atomic-rename write is untouched and the cloud copy is a mirror pushed afterwards. Four reconciliation rules, each a player-visible failure if wrong:

- **post-write mirror** — the save completes on a *worker* thread and `SteamManager` is game-thread-only, so it hops via `EnqueueGameThreadTask`. That is the one place this could have introduced a data race. A cloud failure can never downgrade a good local save's result.
- **load falls back to cloud only when there is no local file**, then restores to disk and re-enters the normal load path, so checksum and version gating apply identically.
- **`EnumerateSaves` unions in cloud-only slots** — otherwise a fresh machine shows an empty load menu, which looks exactly like "cloud is broken". Local always wins, so a cloud copy can never shadow newer progress.
- **`DeleteSave` removes the cloud copy** — otherwise the deleted save reappears on the next enumerate.

## Developers must obtain the SDK themselves

Documented in `README.md`, `docs/ops/build.md` and `CLAUDE.md`. It is **free** — the Steamworks SDK Access Agreement needs only an ordinary Steam account; the $100 Steam Direct fee is for *publishing*. Steam is OFF by default and **the whole engine builds, runs and tests without it**. `vcpkg.json` does not change and cannot.

## Test results (this machine, real SDK)

`43 Steam` + `38 SaveGame` + `152 Lua/coverage` passed, `0 failed`. The 16 skips are the stub-SDK ON-path tests, which correctly skip in a real-SDK build and run in CI instead.

## Rot risk, named

Guarded code CI never compiles bitrots silently — Trap 3 in `docs/agent-rules/vcpkg-dependency-management.md`. That is why the stub SDK **links** rather than merely compiling: CI builds a genuine `OLO_WITH_STEAM=1` binary and runs the ON-path branch tests, including the init-fails case.

If the stub job goes red because a Steamworks call is missing, add the declaration to `StubSDK/public/steam/steam_api.h` and the canned implementation to `StubSDK/SteamStubSDK.cpp`. **That is the mechanism working — do not delete the job.**

A repo secret smuggling the SDK onto the runner was considered and rejected: it is a licence question I should not answer unilaterally, and secrets are unavailable to fork PRs.

## Summary by CodeRabbit

- **New Features**
  - Added optional Steamworks integration with graceful fallback when unavailable.
  - Added achievements, rich presence, overlay status, user identity, and Steam Cloud support.
  - Added Steam Cloud save synchronization, including cloud-only recovery and deletion.
  - Added Lua access, a demonstration script, and a sample Steam scene.

- **Documentation**
  - Added Steamworks setup, build, runtime, and troubleshooting guidance.

- **Tests**
  - Added coverage for Steam services, Lua bindings, cloud saves, and unavailable-Steam scenarios.
  - Added automated validation using a Steamworks stub.


drsnuggles8:
### CI evidence — the stub-SDK mechanism is proven working

The `stub-build` job passed on this PR (61 min), and the log confirms every link in the chain rather than just a green tick:

```
-- OLO_WITH_STEAM_STUB_SDK=ON implies OLO_WITH_STEAM=ON — enabling it.
OK: building OLO_WITH_STEAM=ON against the stub SDK.
[  PASSED  ] 59 tests.        <- the full Steam* filter
[  PASSED  ] 16 tests.        <- SteamStubOnPathTest, none skipped
OK: the ON path was genuinely exercised.
```

The last line is an explicit assertion in the job that it did **not** go vacuously green — without it, a broken stub wiring would show as "all tests skipped, job green", which is the exact failure a job like this is prone to.

Worth noting the inversion, because it is the whole design in one line: locally with the **real SDK**, 43 tests pass and **16 skip**. In CI with the **stub**, all 59 run — and those 16 are precisely the ones that skip locally, including the `SteamAPI_Init()`-returns-false branch. The two halves are complementary, not redundant.

### Trigger narrowed after that verification

`stub-build` now runs on **master, weekly (Wed), and on demand** — not on every PR. The cheap `single-valve-tu` guard (~20 s, no build) still runs on **every PR**, since it catches a mistake made in the PR under review.

Rationale: the build job defends against guarded code *rotting over months*, not "this PR broke it". Per-PR would add ~60 min of runner time to every PR in this repo permanently, for a feature exactly one machine can exercise for real.

Deliberately verified **before** narrowing — narrowing first would have removed the only opportunity to see the job pass on a PR at all.

Residual risk, stated plainly: a PR can now merge green and break the Steam path, discovered on the next master run rather than in review. Given the alternative cost that is the better side of the trade, and it should be revisited if the project ever ships on Steam.

### Also filed

#796 — the `Video (FFmpeg backend)` Linux job has failed **every run for 10+ weeks** (bare `--parallel` OOM-killing the runner, with an unpinned-compiler failure queued behind it). Found because this PR's new workflow hit one of the same two faults. Scored, with the diagnosis and fix; not touched here since it is unrelated to #644.

Issue #793, opened 2026-08-14
Scripting: visual scripting — editor canvas, inspector and debugger (follow-up to #634)


drsnuggles8:
## Problem

[#634](https://github.com/drsnuggles8/OloEngineBase/issues/634) shipped the **runtime half** of visual scripting: the data model, the flow VM, a 70+ node standard library, the `.olovs` asset + YAML serializer, `VisualScriptComponent` / `VisualScriptSystem` with every cross-binding touch-point, the GameplayEventBus bridge, and Lua/C# interop — all headless and tested (see the PR that closes #634, and [ADR 0014](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/adr/0014-visual-script-execution-model.md)).

**What is not there is any way to author a graph.** A `.olovs` today is hand-written YAML. That slice was deliberately deferred rather than ground into one unreviewable diff — see the "Size reality" note in #634 — and this issue carries the rest so it is not an unscored blind spot.

## What exists today (verified against the merged code, not assumed)

| Piece | State |
|---|---|
| `VisualScriptGraph` / `VisualScriptNode` / `VisualScriptLink` / `PinType` / `PinValue` | Done. Authoring helpers (`AddNode`, `AddLink`, `RemoveNode`, `RemoveLink`) already on the graph. |
| Node registry with category / display name / tooltip, sorted stably | Done — the context-search menu can read it directly. |
| `CheckLinkCompatibility(source, target)` | Done — the exact function a link-drag needs for its type-compat feedback. |
| `VisualScriptInstance::PeekOutput(graph, node, pin)` | Done — the hook a pin-value watch reads. |
| `VisualScriptSystem::FindInstance` / `GetLog` / `CollectErrors` | Done — the debugger's data sources. |
| `VisualScriptEditorPanel` | **Absent.** |
| Debugger (node highlight, breakpoint/step, pin-value watch) | **Absent.** The VM has no step hook. |
| ContentBrowser create/icon/double-click-open for `.olovs` | **Absent** — deliberately, since there is no panel to open into. |

## Scope

[...]

Issue #770, opened 2026-08-10
Decal normal mode renders nothing — shader output location doesn't match its draw-attachment map


drsnuggles8:
`Decal_GBuffer_Normal.glsl` declares its fragment output at `layout(location = 0)`:

```glsl
layout(location = 0) out vec4 gNormalRoughAO;
```

but `DecalRenderPass::ExecuteOnGBuffer` (`DecalRenderPass.cpp:379`) gives the
normal mode the draw-attachment map `{NoAttachment, 1, NoAttachment,
NoAttachment, NoAttachment}` — i.e. fragment output location 0 maps to *no*
attachment, and attachment 1 (the target) is fed by location 1, which the
shader never writes. The other three G-Buffer decal modes (albedo, RMA,
emissive) agree between shader and map; only normal disagrees.

Result: a normal-mode decal draws, passes depth, and writes nothing. This is
**not** a Vulkan regression — the same mismatch is present on OpenGL, where
`glDrawBuffers` with `GL_NONE` at index 0 discards the write silently.

Found during the #691 Phase 7 Vulkan port; Vulkan's validation layer states it
verbatim:

```
vkCmdDrawIndexed(): Inside [VK_SHADER_STAGE_FRAGMENT_BIT] it writes to
[Output variable, Location 0, "gNormalRoughAO"] but there is no
VkRenderingInfo::pColorAttachments[0] and this write is unused.
```

(warning severity, so it does not trip the zero-error gate).

**Current state:** `VulkanPassSuite.DecalGBufferModeMatrixMasksItsTargetRenderTargets`
pins the *observed* behaviour, so the fix will fail that assertion loudly and
the test can be updated in the same change.

**Fix:** move the output to `layout(location = 1)` (matching the map), or change
the map to `{1, NoAttachment, ...}`-style so location 0 lands on the normal RT.
Prefer whichever keeps the four modes consistent with each other. Needs a
visual check — a normal decal has never actually rendered, so nobody has seen
what it should look like.

Found during the #691 Phase 7 Vulkan port (PR #769).

## Score
```olo-score
capability: 3
craft: 5
stability: 5
decay: 1
effort: 2
confidence: 1.0
learning: 1
fun: 5
kano: table-stakes
blocked_by: []
blocks: []
```
[...]

PR #799, opened 2026-08-15
feat(scripting): visual scripting — node-graph gameplay logic (runtime VM + editor canvas)

### LINKED ISSUE #634: Scripting: visual scripting — node-graph gameplay logic (runtime VM + editor canvas)
## Problem

OloEngine can already author **audio** graphs (SoundGraph), **material** graphs (ShaderGraph), and **animation-blend** graphs (AnimationGraph) — but there is **no way to author gameplay logic as a node graph**. Gameplay scripting is text-only: C# (`Scripting/C#/ScriptEngine.cpp`) and Lua (`Scripting/Lua/LuaScriptEngine.cpp`), driven per-entity in `Scene::OnUpdateRuntime`. A designer who wants "when the player enters this trigger, spawn three enemies, then fire a quest event" must write code. Every mature engine (Unreal Blueprints, Unity Visual Scripting, Godot) ships a node-graph gameplay layer; OloEngine has the three *ingredients* (a proven node-editor UI pattern, a proven node-VM pattern, an event bus, a dependency-ordered scheduler) but has never assembled them into a general gameplay system.

## Why it matters

- **Non-programmer authoring.** Designers and technical artists get first-class gameplay authoring without touching C#/Lua — the single biggest capability multiplier a tooling flagship can add.
- **The seams already exist.** `GameplayEventBus` (synchronous Subscribe/Publish-by-type), `Scene::GetGameplayScheduler()` (Reads/Writes/After/Before topological scheduling), and the `ShaderGraph`/`SoundGraph` prior art mean this is an *assembly-and-invention* epic, not a research spike — the risk is scoped.
- **Interop, not replacement.** Wired to the existing C#/Lua engines and the event bus, a graph and a text script can co-drive one entity; visual scripting extends the scripting story rather than forking it.

## What exists today (verified, not assumed — the #440 discipline)

| Piece | Where | What it is / is NOT |
|---|---|---|
| SoundGraph node VM | `OloEngine/src/OloEngine/Audio/SoundGraph/SoundGraph.h`, `NodeProcessor.h`, `SoundGraphFactory.cpp` | Topological process order → flattened compiled-ops, typed stream/event endpoints, block-rate `Process(numFrames)` on the audio thread. **Audio-DSP only** — no control flow, no ECS. Prior art for the VM's compiled-plan idea. |
| ShaderGraph editor canvas | `OloEditor/src/Panels/ShaderGraphEditorPanel.{h,cpp}`, `Renderer/ShaderGraph/*` | Hand-rolled ImGui canvas: pan/zoom, pins, links, context-search menu, copy/paste, undo/redo, auto-layout. **Compiles to GLSL.** The UI pattern to mirror. |
| AnimationGraph | `Animation/AnimationGraphAsset.h`, `AnimationGraphSystem.cpp` | State machine for **animation blending only**. |
| Text scripting | `Scripting/C#/ScriptEngine.cpp` (`ScriptComponent`), `Scripting/Lua/LuaScriptEngine.cpp` (`LuaScriptComponent`), driven in `Scene.cpp:1043/1053/2532` | The interop target. |
| GameplayEventBus | `Gameplay/GameplayEventBus.h` | Per-scene synchronous Subscribe/Publish-by-type. Event-node integration seam. |
| Gameplay scheduler | `Scene::GetGameplayScheduler()` (`Scene.cpp:2276`), `SystemScheduler.{h,cpp}` | Dependency-ordered system execution. Where `VisualScriptSystem` registers. |

**Confirmed genuinely absent:** grep for `Blueprint|GameplayGraph|ScriptGraph|LogicNode|EventGraph|VisualScript` finds only an ADR reference — no gameplay node-graph data model, VM, editor, or component exists. No open issue tracks it (checked `gh issue list`).

## Proposed approach (the sub-parts — each is itself effort-5-to-8-scale)

1. **Data model + asset + serialization** — `VisualScriptGraph` (nodes / pins / links), a first-class split between **exec pins** (control-flow wires) and **typed data pins** (bool/int/float/vec2/vec3/string/entity/AssetHandle), a graph-scoped **variable/blackboard** table, and reusable **sub-graph functions/macros**. `VisualScriptAsset : Asset` with loader + AssetManager registration + hot-reload, plus a YAML serializer round-tripping nodes/links/variables/default pin values (reference `ShaderGraphSerializer` / `SoundGraphSerializer`).
2. **Runtime execution VM** — a flow-driven interpreter (distinct from the audio-block SoundGraph VM): exec-token propagation through the flow nodes, lazy pull-evaluation of data pins with per-tick memoization, per-entity instance state, and **latent/async** nodes (Delay, timeline, WaitForEvent) that suspend and resume across frames. Compile to a flattened execution plan with cycle/dangling-link validation and a per-tick node-budget guard against runaway loops.
3. **ImGui node-editor canvas panel** — a new `VisualScriptEditorPanel` modeled on `ShaderGraphEditorPanel` (pan/zoom, pin hit-test, link drag with type-compat checks, context-search create menu, copy/paste, undo/redo, auto-layout), extended to render **exec-wires** (thick, directional) vs **data-wires** (thin, pin-type-colored) and a **live debug overlay** highlighting the executing node and evaluated pin values at runtime.
4. **Standard node library (40+ nodes)** — Events (OnBeginPlay, OnUpdate, OnCollision/Trigger, custom event, EventBus subscribe); Flow (Branch, Sequence, For/ForEach/While, Gate, DoOnce, DoN, FlipFlop, Delay); Math/Logic (arithmetic, compare, boolean, vector make/break/dot/cross, lerp, random, clamp); Variables (Get/Set); ECS (Get/Set component field, Has/AddComponent, DestroyEntity, Get/SetTransform); Spawn (Instantiate prefab); Queries (Raycast, FindEntityByTag, overlap); Utility (Print/Log, Format). A type-name-keyed node registry/factory (mirror `SoundGraphFactory`) shared by the editor menu and the serializer.
5. **ECS + scheduler integration** — a `VisualScriptComponent` (graph handle + per-instance variable overrides) and a `VisualScriptSystem` registered in the gameplay scheduler with declared Reads/Writes/After/Before; per-entity VM lifecycle (instantiate/OnBeginPlay → tick → teardown on stop). Thread the component through **every** CLAUDE.md cross-binding touch-point — `Components.h`, hand-written `SceneSerializer` block (non-trivial variable table) + `kComponentsCustomSerialize`, SaveGame `Serialize()`+`RegisterAll()`, OnComponentAdded/Removed, `OLO_PROPERTY`, Lua/C# expose, editor undo `operator==`.
6. **GameplayEventBus + C#/Lua interop** — event nodes that Subscribe/Publish through `GameplayEventBus` by payload type; 'Call C# Method' / 'Call Lua Function' nodes bridging into `ScriptEngine`/`LuaScriptEngine`; a script-callable 'Custom Event' entry so text scripts can trigger graph flow; and graph variables exposed read/write to both engines.
7. **Debugging + tests** — live per-node execution highlight, breakpoint + single-step, pin-value watch; plus the test pyramid: VM-determinism + per-flow-node contract tests, a serialization round-trip test, and **Functional** cross-subsystem tests driving a real graph through `Scene::OnUpdateRuntime` (spawn a prefab, mutate a component, react to a collision, fire a quest/inventory event).

## Acceptance criteria

- [ ] `VisualScriptGraph` data model with exec pins, typed data pins, blackboard variables, and sub-graph functions; `VisualScriptAsset : Asset` with loader + AssetManager registration + hot-reload; YAML serializer round-trips a graph identically (save → load → save).
- [ ] Runtime flow VM: exec-token control flow, data-pin pull-evaluation, per-entity instance state, latent Delay/WaitForEvent suspend-resume, compiled execution plan with cycle/dangling validation and a runaway-loop budget guard.
- [ ] `VisualScriptEditorPanel` with pan/zoom/pins/links/context-search/copy-paste/undo-redo/auto-layout, exec-vs-data wire rendering, and a live runtime debug overlay.
- [ ] 40+ standard nodes across Events / Flow / Math-Logic / Variables / ECS / Spawn / Queries / Utility, behind a shared type-name registry/factory.
- [ ] `VisualScriptComponent` + `VisualScriptSystem` registered in `Scene::GetGameplayScheduler()` with correct Reads/Writes; component threaded through all CLAUDE.md cross-binding touch-points (build links, scenes persist, save-games round-trip).
- [ ] Event nodes bridge `GameplayEventBus`; call-C#/call-Lua nodes and a script-callable custom-event node work; graph variables are readable/writable from C# and Lua.
- [ ] In-editor debugger: node highlight, breakpoint/step, pin-value watch.
- [ ] Contract tests (VM determinism + each flow node), serialization round-trip test, and Functional cross-subsystem tests proving a graph spawns a prefab, mutates a component, reacts to a collision, and fires a gameplay event through real `OnUpdateRuntime` — all classified per the testing-architecture rules.

```olo-score
capability: 5
craft: 5
stability: 5
decay: 2
effort: 13
confidence: 0.9
learning: 5
fun: 5
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
Adds node-graph gameplay logic to OloEngine: a data model, a runtime flow VM, a ~70-node
standard library, the `.olovs` asset, ECS + scheduler integration, C#/Lua interop, and an
ImGui node editor with a live debugger.

The execution model is genuinely new for this codebase and is written up in
**[ADR 0014](docs/adr/0014-visual-script-execution-model.md)** — read that first if you only
read one thing. The recurring gotchas are in
**[docs/agent-rules/visual-script-vm.md](docs/agent-rules/visual-script-vm.md)**.

## Why not reuse the SoundGraph VM

It was the obvious move and it does not fit. SoundGraph is cycle-free by topological sort,
has no control flow (every node runs every block), is block-rate on the audio thread, has one
instance per playing sound, and cannot suspend. Gameplay logic needs the opposite of all five.
What transferred is the *layering* — `Asset → compiled plan → per-entity instance` — and that
is what makes instancing a vector resize over shared state.

## The execution model, in three sentences

**Control flow** is an exec token pushed by synchronous descent: triggering an exec output runs
that whole branch to completion right there, which makes `ForLoop`/`Sequence` five lines each
instead of a continuation record apiece. **Data** flows by lazy pull, memoized per *exec step*
rather than per tick — per-tick would make a `Set` invisible to a later `Get` in the same tick.
**Latent nodes** (`Delay`, `WaitForEvent`) park a resume record and are resumed by the tick loop
or by an event.

## What's here

| Sub-part | |
|---|---|
| Data model + `.olovs` asset + YAML serializer | `PinType`/`PinValue`, exec-vs-data pin split, blackboard variables, sub-graph functions, byte-identical round-trip |
| Runtime flow VM | exec push, data pull, per-entity instances, latent suspend/resume, compile-time cycle + dangling + type validation, runaway-loop budget, exec-depth cap |
| Node library | ~70 nodes across Events / Flow / Math / Logic / Vector / Variables / Entity / Queries / Utility / Functions / Scripting |
| ECS + scheduler | `VisualScriptComponent` + `VisualScriptSystem`, registered `After("Scripts")`, `ReadsWrites(kLocalTransforms)`, unmarked (join-all) |
| Interop | `GameplayEventBus` both directions, `visual_script` Lua table, `VisualScript` C# class, call-Lua / call-C# nodes |
| Editor | `VisualScriptEditorPanel` + a reusable `GraphCanvas` widget, live debugger |
| Tests | 30 VM/registry contract cases, 9 serializer cases, 6 Functional |

## Three design calls worth reviewing

**1. `GameplayEventBus` was not changed.** It has no unsubscribe, so a per-entity subscription
would dangle the moment that entity died. Subscribing **once per scene** and fanning out to
whichever instances are alive at delivery time closes that window with no API change to a header
Quest/Inventory/Progression all include.

**2. Everything that could escape the tick is queued.** Entity create/destroy goes through
Scene's existing deferred command queue, and `VisualScriptSystem::Update` brackets itself with
`FlushPendingEntityCommands()` exactly as `Scene::UpdateScripts` does — so a graph spawn is
visible to physics and rendering in the *same* tick. Bus publishes go to an outbox drained
between iterations. Component add/remove is applied inline, and that is safe **only because**
the system iterates a snapshot of entity UUIDs rather than a live EnTT view; the comment says so
at the call site.

**3. `GraphCanvas` is new code, not an extraction.** This is the eighth graph editor in a repo
with seven hand-rolled canvases and zero shared code. Extracting from `ShaderGraphEditorPanel`
first would have been the cleanest end state but edits a working 1,507-line panel whose only
net is `ShaderGraphCommandTest.cpp`. Writing the viewport half as new code means no existing
panel was touched, the widget starts with a real consumer, and migrating the others becomes
incremental. ADR 0014 records all three options and why.

## Cross-binding — verified by diff, not by assumption

`VisualScriptComponent` was threaded through every touch-point in CLAUDE.md's *Definition of
done*. I built OloHeaderTool standalone and diffed its output rather than guessing:

| Generated file | Contains it? | |
|---|---|---|
| `AllComponents.Generated.inl` | ✅ | scene copy / prefab / `HasComponent<T>` |
| `OnComponentAdded.Generated.inl` | ✅ | generated no-op; nothing needed on add |
| `OnComponentRemoved.Generated.inl` | ❌ | hand-written body + `kComponentsCustomOnRemove` — the VM instance must die with the component |
| `SceneSerialize/Deserialize` | ❌ | `std::map<std::string, PinValue>` is unrecognised, so the classifier skips it — **hand-written blocks are the only path**, and `kComponentsCustomSerialize` is correctly NOT needed |
| `SaveGameComponentCapture/Restore` | ✅ | `Serialize()` overload + `RegisterAll()` added |
| `McpFieldRegistry.Generated.inl` | ✅ | `Graph` + `Enabled` |

Total codegen delta: **84 added lines, nothing removed** — no other component was disturbed.

Also checked: the `.scenebin` fast path treats an uncovered component as *"this entity
round-trips as YAML"*, so the hand-written block is reached. No silent drop.

## What the self-review caught

I reviewed my own diff at high effort before opening this. It found **six real bugs**, none of
which the test suite noticed — the "green but wrong" class the agent-rules keep warning about.
All six are fixed here, with three regression tests added:

| Bug | Why the tests were green |
|---|---|
| Bus-published events delivered **twice** — the outbox drain queued locally *and* published, and this system subscribes to `VisualScriptCustomEvent` itself, so the publish came straight back | The test counted only the non-graph subscriber; only a listening *graph* can see the double |
| With 2+ functions, `Function.Entry`/`Return` resolved against the **wrong** signature — matched by NodeId, which is graph-**local**, so every function's Entry is id 1 | Every test used exactly one function graph |
| `Flow.WaitForEvent`'s `Payload` output was permanently dead — resume jumps to the exec targets and never re-runs the body | Nothing read that pin |
| A parked `Wait For Event` was **lost** if the tick's node budget was already spent (record erased before an `ExecuteFrom` that silently bails) | Needs a budget-blown graph *and* a pending wait |
| `Event.OnEndPlay` never fired when the component was disabled or removed, contradicting its own tooltip | Nothing asserted teardown |
| Alt+click wire-cutting missed visibly-drawn wires at any zoom ≠ 1 (`DistanceToWire` used unzoomed control points, `DrawWire` scaled them) | Geometry mismatch, invisible without a zoomed click |

A seventh was found by **looking at the running editor**: the right-click node-search menu could
not be opened *at all*. `GraphCanvas` latched `m_IsPanning` on right-button **press**, and the
panel suppresses its own input while panning, so the release that opens the menu was never seen.
My own header comment claimed the two gestures "do not collide" — they did. Panning now starts
only after a 4px drag threshold.

## Honest limitations

- **No node-by-node single-step.** The debugger breaks *before* a node and offers Resume and
  Step-Tick. Exec descent has no resumable continuation, so stepping one node would mean
  re-running the tick and repeating every side effect before the breakpoint. The tooltip says
  this; ADR 0014 explains it; #793 tracks a continuation-based VM.
- **No recursion in sub-graph functions.** One set of value slots per function per instance;
  re-entry is refused with a reported error rather than silently sharing them.
- **`Add/RemoveComponent` covers a curated 16-component table**, not everything. An EnTT type
  is not reachable from a runtime string without a name→type dispatch and the engine has none
  (the generated MCP registry is editor-side). Adding a row is the whole cost; an unknown name
  is reported, never silently ignored.
- **No `Get/Set arbitrary component field` node** — same reason; tracked in #793.

## Review guide

**Where I'd look hardest**

1. `VisualScriptVM.cpp` — `ExecuteFrom` / `EvaluateOutput` / `DispatchEvent`. The memo-stamp
   rule (bumped per exec **step**, not per tick) and the latent resume path are where the
   subtle bugs were, and two of the six above lived here. ADR 0014 §3 and §5 explain the
   intent; if the code and the ADR disagree, the code is wrong.
2. `VisualScriptSystem.cpp::Update` / `DrainOutbox` — the ordering is load-bearing (BeginPlay
   for every new instance **before** the inbox drain, both before any OnUpdate) and the
   double-dispatch bug lived in the outbox. The two `FlushPendingEntityCommands()` calls must
   stay where they are.
3. `Nodes/EntityNodes.cpp` — `Entity.AddComponent`/`RemoveComponent` apply **inline**, which is
   safe *only* because the system iterates a UUID snapshot rather than a live EnTT view. If
   anyone ever changes that iteration, these two nodes must move to the deferred queue first.

**What I verified, and how** — full suite 5792/5797 (one pre-existing machine-local flake,
`EASUVisualEvidenceTest.GTAOSurvivesRuntimeUpscaleSwitch`, named in the flake notes with a
prior occurrence on an unrelated MCP-only diff; passes in isolation). 56/56 VisualScript tests.
Codegen verified by building OloHeaderTool standalone and **diffing** its output rather than
assuming — 84 added lines, nothing removed. The editor was driven live over MCP: Window menu,
panel render, context search with filtering, node creation, sidebar details, `[compiles]` and
dirty indicators all confirmed on screen (screenshots in the PR thread).

**Least confident about** — the exec-vs-data **wire rendering** and the **live debugger overlay**
are the two things I could not get on screen. Injected clicks drove menus fine, but drags onto
9px pin targets inside a floating ImGui panel reported success and did nothing, so I never got
two nodes wired in the running editor. The drawing code is straightforward and the geometry is
shared with the hit-test I did fix, but *I have not seen a wire drawn*. Treat that as unverified.
Gaps logged to #607.

**Deliberately not tested** — the C# bridge (`VisualScript_*` internal calls) has no automated
test; Mono is not up in the Functional harness. The Lua half is covered end-to-end.

Closes #634

## Summary by CodeRabbit

- **New Features**
  - Added a Visual Script editor for creating, linking, debugging, compiling, and saving graph-based scripts.
  - Added Visual Script assets with content-browser support, scene assignment, per-entity variables, execution toggles, and drag-and-drop workflows.
  - Added runtime support for gameplay events, physics contacts, entity actions, math, flow control, scripting bridges, and scene/save-game persistence.
  - Added a sample patrol-and-announce Visual Script.

- **Documentation**
  - Added Visual Script execution-model and VM guidance.
  - Documented build-lock requirements and automated build enforcement.

---

## Since the first push

Three commits on top of `28453f8d`.

### `fix(scripting)` — review round (#634)

Two review passes over the initial commit — a local `/code-review high` and an
external round — found ~36 issues. All fixed bar one dismissal, with three
regression tests for the ones a reader could not otherwise see. The ones that
mattered most:

- **Use-after-free in `Scene::OnRuntimeStop`** — the visual-script system was
  destroyed 54 lines before `GameplayEventBus::Clear()`, while the bus still held
  its subscription. The bus has no unsubscribe, so the fix is ordering.
- **Events queued during `OnBeginPlay` were dropped**, because the queue drained
  before the begin-play pass rather than after.
- **`OnEndPlay` skipped** on disable, remove and hot-reload — three separate paths.
- **Latents parked at budget exhaustion were lost**, so a graph that ran out of
  step budget mid-await never woke.
- **Non-static C# methods invoked with a null `this`** — `ScriptEngine` now asks
  Mono and surfaces a managed throw instead of swallowing it.
- **The right-click context menu could not be opened at all**: `GraphCanvas`
  latched `m_IsPanning` on button *press*, so the release never read as a click.
  A 4px drag threshold separates the two. Found only by driving the live editor.

One finding dismissed: renaming the `s_`-prefixed panel constants to `k`. This
repo's convention is `s_PascalCase` for statics (CLAUDE.md → Conventions), and the
neighbouring `ShaderGraphEditorPanel` already uses `s_GridSize` / `s_NodeWidth` /
`s_PinRadius`.

### `chore(build-lock)` — ride-along, unrelated to #634

Not part of this issue; landed here rather than in its own PR at the repo owner's
request. The cross-worktree build mutex from #788 was being bypassed, and a build
whose launching session died kept the lock for ~45 minutes past the point its work
had merged. Two fixes:

1. A **`PreToolUse` hook** (`scripts/claude-build-lock-guard.py`) denies any agent
   tool call running `cmake --build` / `ninja` / `msbuild` that skips
   `build-lock.ps1`. A permission rule cannot do this — `deny` degrades to a
   prompt, which an unattended session approves. A build tool only counts in
   statement position, so `Get-Process -Name cl,link,MSBuild` is not a build; the
   24-case matrix ships in the script (`--selftest`).
2. **Parent watch** in `build-lock.ps1`: the build runs as a child process while
   the launcher is polled, pinned by `(pid, StartTime)` because Windows recycles
   pids. Launcher gone ⇒ the build tree is killed and the lock released. Every
   ambiguous reading fails open.

### Merge of `master`

`origin/feature/visual-scripting-634` had `master` merged into it (bringing #691
Vulkan Phase 8 and the Steamworks integration); that is merged in here, without
conflicts.

### Verification against the merged tree

Everything below was re-run *after* the merge, not carried over from before it.

- **Build**: `OloEngine-Tests` + `OloEditor`, Debug, 0 errors.
- **Full suite**: 5867 tests, **5844 passed**, 22 skipped (Steam stub — no SDK on
  the path for this run), 1 failed:
  `EASUVisualEvidenceTest.GTAOSurvivesRuntimeUpscaleSwitch`. That one is a known
  pre-existing failure being fixed in the decal/GTAO/R32F worktree (#770/#772) —
  it asserts `upMean >= nativeMean * 0.5` and gets `upMean = 0` with the render
  graph reporting `GTAO=n/a`, which is unrelated to anything here.
- **Live editor** (driven over the MCP diagnostics server, screenshots read back):
  0 `[error]` lines in the whole editor log; `olo_shader_errors` returns 0;
  `[ScriptEngine] Initializing` with no "assembly unavailable" warning, so C#
  scripting loaded; `[VisualScript] Standard node library registered: 84 node
  types`; the Visual Script Editor panel opens and draws its sidebar, canvas grid
  and the `On Begin Play` node with `[compiles]` green; and right-clicking the
  canvas opens the **Add Node** search menu — the specific interaction that the
  pan-latch bug made impossible, re-confirmed fixed on the merged build.

### Second review round

Eight findings fixed, six regression tests, three skipped with reasons.

Most of the fixes are one class: **narrowing a floating value that does not fit
the destination is undefined behaviour**, not a saturating or wrapping result —
and every site involved is fed by a script or a hand-edited YAML scalar.
`PinValue::AsInt` and `IntFromStorage` cast f32 (reach ~3.4e38) straight to i64
(stops at ~9.2e18) behind an `isfinite` check that cannot catch it; the Lua glue
checked the *narrowed* result instead of the source; `FromLua` cast unguarded and
decided integral-ness with `==` on a double, which CLAUDE.md bans. All of them now
route through `IsRepresentableAsFloat` / `IntFromFloat`.

The other five:

- **C# bridge memory safety** — `GetMethod` matches on *arity* alone, so
  `void F(string)` came back for a one-argument lookup and had the caller's `f32*`
  reinterpreted as a `MonoString*` inside `mono_runtime_invoke`. Corruption, not a
  catchable managed error. Now refused via `IsMethodSingleFloatParameter`.
- **Teardown** — `Sync` replaced an instance with `insert_or_assign` and erased one
  on compile failure, neither firing `OnEndPlay`; and clearing the graph field left
  the old graph ticking, so clearing it read as doing nothing.
- **Debugger** — a breakpoint set `m_Paused` and returned, but that unwinds only one
  `ExecuteFrom` frame, so a Sequence's later pins kept running behind a frozen canvas.
- **Latent nodes in functions** — `CallFunction` pops the return frame the moment
  `ExecuteFrom` returns, so a `Delay` inside a function silently yielded nothing.
  Now rejected at compile time; latents in the event graph are untouched.
- **Editor** — the override inspector hid its stale-override list when the graph
  handle resolved to nothing, making those overrides invisible yet still serialized.

Skipped, each checked against the code rather than declined on preference: the
`DestroyEntity` cleanup the reconciliation sweep already performs (with `EndPlay`);
the runtime-stop ordering, which already mirrors startup and where
`ScriptEngine::OnRuntimeStop` does not unload the assembly; and publishing node
descriptors as `shared_ptr`, where the hazard needs a re-registration that has no
call site — every registration runs once inside `call_once` with distinct names.

### Verification

Re-run against the merged tree, after `master` (accessibility, #458) was merged in:

- **Build**: `OloEngine-Tests` + `OloEditor`, Debug, 0 errors.
- **Full suite**: 5927 tests, **5904 passed**, 22 skipped, 1 failed —
  `EASUVisualEvidenceTest.GTAOSurvivesRuntimeUpscaleSwitch`, a known pre-existing
  failure being fixed in the decal/GTAO/R32F worktree (#770/#772) and unrelated to
  anything here.
- **Live editor** over the MCP diagnostics server: 0 `[error]` lines, 0 shader
  errors, C# scripting loaded, 84 node types registered, the panel drawing, and
  right-click opening the Add Node menu.

### CI, and one ride-along outside this issue

`ci(asan)` in this branch is **not** part of #634. The clang-cl ASan job had been
red since the accessibility merge (`e14a8dfd`, #458) — on master itself, not just
here — with `AccessibilitySettingsTest.LoadOfACorruptFileFailsWithoutCorruptingSettings`
dying on SEH 0xC0000005. That is the LLVM bug this job already documents and works
around in nine other places (#497): ASan redzones the compiler-emitted MSVC EH
metadata and the EH runtime reads across it, so any throw unwinding into a catch
crashes. The new test takes a real throw+catch through yaml-cpp and was never added
to the exclude list. One regex entry, matching the existing precedents; the three
sibling accessibility tests still run, and the throw path stays covered by the MSVC
and Linux ASan jobs.

It landed here because it was what kept this PR's Sanitizers gate red, and the same
line unblocks master. Happy to move it to its own PR if preferred.

All five checks are green on `fc6c6c47`: Pre-commit, Windows, Sanitizers,
Steam (stub SDK), SonarCloud.


drsnuggles8:
## 🤖 Self-review @ `28453f8d46fa5bbb2147aa619a2f66d6db257fc8`

Reviewed the full diff at **high** effort before opening the PR. It found **six real bugs**, none of which the test suite noticed. All fixed in this commit, with three regression tests added.

- **Findings: 6 · Fixed: 6 · Dismissed: 0**
  1. *Bus-published events delivered twice* — `DrainOutbox` queued locally **and** published; this system subscribes to `VisualScriptCustomEvent` itself, so the publish came straight back as a second delivery and every listening `Custom Event` node fired twice. Now the bus route is exclusive. Regression test: `VisualScriptPublishesToEventBusTest.ABusPublishedEventReachesListeningGraphsExactlyOnce` (the pre-existing test counted only the non-graph subscriber, which cannot see the double).
  2. *Wrong function signature with 2+ sub-graph functions* — `OwningFunctionGraph` matched by `NodeId`, but NodeIds are graph-**local**, so every function's Entry node is id 1 and the second function resolved against the first's parameter list. Now matched by address. Regression test: `VisualScriptVMTest.TwoFunctionsResolveTheirOwnSignatures`.
  3. *`Flow.WaitForEvent`'s `Payload` output permanently dead* — resuming jumps straight to the resume pin's exec targets and never re-runs the node body, so nothing ever filled it. Added `PublishResumePayload`. Regression test: `VisualScriptVMTest.WaitForEventPublishesTheArrivingPayload`.
  4. *Parked waits lost on budget exhaustion* — the latent record was erased before an `ExecuteFrom` that silently bails when the tick's node budget is spent, so the branch was gone forever. Now the dispatch returns early and leaves the record parked.
  5. *`Event.OnEndPlay` never fired on component disable/remove*, contradicting its own tooltip. All removal paths now funnel through `EndAndErase`.
  6. *Alt+click wire-cutting missed drawn wires at any zoom ≠ 1* — `DistanceToWire` used unzoomed bezier control offsets while `DrawWire` scaled them, so the hit curve bowed away from the drawn one.

A **seventh** was found by driving the running editor rather than by reading code: the right-click node-search menu could not be opened *at all*. `GraphCanvas` latched `m_IsPanning` on right-button **press**, and the panel suppresses its own input while panning, so the release that opens the menu was never seen. My own header comment asserted the two gestures "do not collide" — they did. Panning now starts only past a 4px drag threshold.

**Not fixed, stated plainly:** I could not get **wire rendering** or the **live debugger overlay** on screen. Injected clicks drove menus and list rows reliably, but drags onto ~9px pin targets inside a floating ImGui panel returned `ok` and did nothing, so I never wired two nodes in a running editor. The drawing code is small and shares its geometry with the hit-test I *did* fix, but I have not seen a wire drawn — treat it as unverified. Tooling gaps logged to #607.

Issue #713, opened 2026-07-31
Renderer: GPU prefix-sum / parallel-scan primitive


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

There is no GPU scan primitive in the engine — the only `PrefixSum` hit anywhere in `OloEngine/src` is an unrelated one in `Renderer/Commands/CommandBucket.cpp`. Every compaction we do therefore uses a global `atomicAdd` to allocate output slots: `LightCulling.comp` (both the shared-memory and the global allocation), `Particle_Compact.comp`, `Fluid_Compact.comp`, `InstanceFrustumCull.comp`, `VirtualClusterCull.comp`. That works but serializes on the counter under contention and — more importantly — produces **non-deterministic output ordering**, which makes compaction-heavy passes awkward to test and to reason about.

**Describe the solution you'd like**

A reusable work-group + device-level exclusive scan (`include/` GLSL helper plus a small C++ dispatch wrapper): subgroup scan where available, shared-memory Hillis-Steele fallback, then a device-level pass over per-group totals. Retrofit the compaction passes above onto it where the ordering guarantee or the contention relief is worth it.

**Describe alternatives you've considered**

Status quo (`atomicAdd` everywhere) — fine at low contention, degrades at high, and the ordering non-determinism is a standing test problem.

**Additional context**

Reference: `D:\repos\Timberdoodle\src\rendering\tasks\prefix_sum.glsl` + `prefix_sum.inl` (Apache-2.0). Cheap, self-contained, and unblocks the cleaner formulations of several other items on the backlog.

**Acceptance criteria**

- Headless unit test over randomized inputs including non-power-of-two sizes.
- At least one existing compaction pass converted, with deterministic output order verified.

## Score
```olo-score
capability: 5
craft: 1
stability: 3
decay: 1
effort: 1
confidence: 1.0
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #818, opened 2026-08-16
Terrain: move LOD quadtree selection and seam resolution to the GPU (#714)

### LINKED ISSUE #714: Terrain: move LOD quadtree selection and seam resolution to the GPU

Terrain LOD runs entirely on the CPU, every frame. `Terrain/TerrainQuadtree.cpp` does a recursive `SelectNode()` descent with per-node screen-space error, then `ResolveNeighborLODs()` walks the selected set doing `FindLeafAt()` point queries through an `unordered_set` to fill each node's `NeighborLODs[4]` for crack-free stitching, then the result is uploaded as `TerrainChunkLODData` per chunk. That is single-threaded pointer-chasing proportional to the visible node count, plus a per-frame upload, and it caps how fine the quadtree can practically get.

Do the whole descent on the GPU with a persistent worklist:

- `CullingState { pendingNodeCount, nextPassNodeCount, finalNodeCount }` in an SSBO, plus two ping-pong node buffers.
- One dispatch per tree level: each thread takes a pending node, frustum-culls it, compares distance-to-bounds against `deviation * lodMaxDistanceMultiplier`, and either `atomicAdd`s 4 child coords into the next-pass buffer or `atomicAdd`s itself into the visible list.
- A 1-thread kernel between levels swaps the counters and writes the next `glDispatchComputeIndirect` args plus the instance count for the terrain draw.
- **Seams without CPU point queries:** a `LODLevelMap` kernel writes the selected level per texel by walking the split map top-down, then a `LODSeamMap` kernel samples the four neighbours at the node's step size and packs the four `max(0, neighbour - center)` deltas into one `uint` per node. The vertex shader unpacks that to snap edge vertices.

Everything stays on the GPU — no readback, no per-frame LOD upload.

Multithreading the existing CPU descent — helps the descent, not the upload or the `FindLeafAt` hash lookups, and still bounds quadtree depth by CPU budget.

**Additional context**

Reference: `D:\repos\Hollow-TerrainSystem\Packages\PhotoTerrain\Shaders\TerrainCulling.compute` — `CSSelectLODPatches` at :43, `CSComputeIndirectArgs` at :144, `CSComputeLODMap` at :163, `CSComputeLODSeams` at :187; C# driver in `Runtime/Core/Rendering/PhotoTerrainRendering.Culling.cs`; per-patch error precompute in `Shaders/Editor/ComputeTerrainLODErrors.compute`; shared patch struct in `Shaders/Includes/TerrainInstance.hlsl`. MIT. Their reported figure is ~6 ms to recompute the full LOD tree for a 2049^2 terrain. Also worth lifting: `Runtime/Utility/UnsafeFixedQuadTree.cs` and `BoundsTree.cs`.

Related: #613 (floating-origin terrain streamer), #469/#428 (terrain collision — heightfield updates share the tile identity).

**Acceptance criteria**

- `TerrainQuadtree::SelectLOD` / `ResolveNeighborLODs` no longer run per frame; terrain draws from a GPU-built indirect list.
- No cracks at LOD boundaries, verified from ground level and from above.
- Measured CPU frame-time reduction on a large terrain scene; quadtree depth can be raised beyond today's practical limit.

## Score
```olo-score
capability: 5
craft: 5
stability: 1
decay: 1
effort: 5
confidence: 0.8
learning: 8
fun: 8
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #714.

Moves terrain LOD selection and seam resolution off the CPU. `TerrainQuadtree::SelectLOD`'s recursive descent, the `unordered_set`-backed `FindLeafAt` neighbour queries in `ResolveNeighborLODs`, and the per-chunk `TerrainChunkLODData` upload no longer run per frame; terrain draws from a GPU-built indirect list.

## How it works

A persistent worklist, one dispatch per tree level, nothing read back:

| kernel | job |
|---|---|
| `TerrainNodeSelect.comp` | pops a pending node, frustum-culls it, tests screen-space error, `atomicAdd`-appends either its 4 children or itself to the visible list |
| `TerrainCullArgs.comp` | 1 thread: swaps the counters, writes the next `DispatchComputeIndirect` args **and** the draw's `instanceCount` |
| `TerrainLODMap.comp` | split map → selected level per texel |
| `TerrainSeamMap.comp` | packs four `max(0, mine − neighbour)` deltas per visible node |

Terrain then draws as **one instanced indirect call** over a shared unit patch grid. The vertex stage reconstructs each instance's terrain rect from its node coord and snaps edge vertices to the coarser neighbour's spacing.

Per the issue's note on #713: the worklist uses `atomicAdd` compaction, not a prefix scan. An unordered append is all a worklist needs and no consumer here reads the ordering.

## Acceptance criteria

**1. `SelectLOD` / `ResolveNeighborLODs` no longer run per frame; terrain draws from a GPU-built indirect list.** Done. The CPU descent remains only as a fallback when compute is unavailable or `OLO_TERRAIN_CPU_LOD=1` forces it.

**2. No cracks, from ground level and from above.** Proven two ways.

*Exhaustively, in `TerrainGPUQuadtreeTest`*: every `+X` boundary in a selected set, over 15 camera poses × 4 split thresholds, comparing the two edges' vertex sets for **exact** equality — both sides are integer multiples of the same dyadic step, so a tolerance would only hide a real crack. It also asserts at least one boundary joined two *different* levels; that guard fired for real during development, when every pose used the shipped 8 px threshold and the descent always reached leaves, so the check was passing vacuously.

*Visually*: background pixels inside the terrain body, live editor, measured.

| view | GPU path |
|---|---|
| ground level across a LOD boundary | **0.0000%** |
| from above | **0.0000%** |

**3. Measured CPU frame-time reduction.** Same scene, same pose, 25 samples each, only the LOD path differs (`OLO_TERRAIN_CPU_LOD=1` for the CPU column):

| | GPU path | CPU path | delta |
|---|---|---|---|
| **cpuMs** | **2.037** ± 0.136 | **4.482** ± 0.173 | **−2.45 ms (−54.6%)** |
| frameTimeMs | 3.878 ± 0.364 | 5.786 ± 0.253 | −1.91 ms |
| gpuMs | 2.163 ± 0.459 | 3.660 ± 0.110 | −1.50 ms |

Separation is ~14× the sample standard deviation, well clear of the frame-to-frame noise floor. Scene: `TerrainGpuLodTest.olo`, 2048² world over a 1024² heightmap, ground-level camera.

**Depth ceiling:** the GPU tree is no longer clamped to `TerrainLODConfig::MAX_LOD_LEVELS` — that cap existed because the CPU descent is a per-frame single-threaded walk. This is a structural change, not a measured one: at the scene sizes tested the chunk-derived depth (4) is still below the old cap (5), so nothing here demonstrates the extra headroom being used.

## Two pre-existing bugs this uncovered

**No sandbox scene set `TessellationEnabled`**, so the quadtree LOD path had *zero* runtime coverage — not in `AssetSceneLoadTest`, not in any visual test. `TerrainGpuLodTest.olo` now enables it.

That hid a real defect: `FindChunkForNode` maps a selected node to the **one chunk at its centre**, which is correct only while every selected node is a leaf. Select one level up and the node covers 2×2 chunks while three go undrawn. The committed evidence PNGs are the same scene and pose with only the LOD path differing — the CPU frame leaves **80.1%** background, the GPU frame **0.000%**. The GPU path removes the mapping entirely: a visible node *is* the drawn primitive.

## Supporting changes

- **RHI:** `DispatchComputeIndirect` (GL `glDispatchComputeIndirect` / Vulkan `vkCmdDispatchIndirect`) did not exist. `DrawBoundElementsIndirect` gained an explicit `topology` so terrain can draw `PatchList` indirectly — explicit rather than a default argument, since a default on a virtual resolves from the static type.
- **`TerrainQuadtree`:** node bounds now come from an exact min/max height pyramid shared with the GPU descent. The previous per-node strided resample (every `extent/16` texels plus corners) could miss a spike and produce an AABB that clipped its own terrain. Sharing the array is also what lets `TerrainGPUQuadtreeTest` assert an *identical* selected-node set rather than a fuzzy overlap.
- **`BuildFromPyramid`** and a `std::span` overload of `BuildHeightPyramid` let the selection math be tested with no GL context. Populating a `TerrainData` uploads a heightmap texture, so the L1 tests were faulting through a null GL entry point when run in isolation.
- **`OLO_TERRAIN_CPU_LOD=1`** forces the CPU path — the A/B lever the perf table above uses.

## Notes for review

- **Shadow casters deliberately stay on the chunk meshes.** The GPU visible list is culled against the *camera* frustum; replaying it per cascade would drop every caster that is off-screen but still shadowing. A correct fix is a second descent with the light's frustum — its own change.
- **A `.comp` never goes through shaderc on the GL backend.** `OpenGLComputeShader::Compile` hands GLSL straight to `glShaderSource`, so `glslc` passing says nothing about whether the driver accepts it. All four kernels compiled under `glslc` *and* `glslangValidator -G` while NVIDIA rejected them over a parameter named `packed` — a GLSL reserved keyword. Worse, the failure is an infinite hang, not an error: `OLO_CORE_ASSERT` → `MessageBoxA` with no headless suppression, so the test process sits at ~0% CPU forever. Both are written up in the agent-rules doc.

## Testing

- Full suite: **5949 tests, 5927 passed, 45 skipped, 0 failures.**
- `TerrainGPUQuadtreeTest` — 8 L1 tests: node addressing, pyramid containment, CPU/GPU selection parity (60 comparisons), LOD-map correctness, seam bounds, and the crack proof.
- `TerrainGPUQuadtreeVisualEvidenceTest` — 3 L8 tests: top-down coverage, ground-level silhouette, and the GPU-vs-CPU coverage A/B.
- Live editor verification over MCP from ground level and above, plus the frame-time table.

## Summary by CodeRabbit

* **New Features**
  * Added GPU-driven terrain level-of-detail selection, frustum culling, seam handling, and indirect rendering.
  * Added a procedural terrain test scene with tessellation, lighting, and shadows.
  * GPU-driven terrain can be enabled or disabled at runtime, with automatic CPU fallback when unavailable.

* **Bug Fixes**
  * Improved terrain edge matching to reduce visible cracks between different detail levels.

* **Documentation**
  * Added guidance covering GPU terrain LOD behavior, rendering requirements, and troubleshooting.

* **Tests**
  * Added CPU validation and visual coverage tests for terrain LOD, seams, and rendering.


drsnuggles8:
## Review response

Thanks — several of these were real. Summary of what changed and what I declined.

### Fixed

| finding | what it actually was |
|---|---|
| **A/B test passes when readback fails** | Real, and narrower than described: `ASSERT_` inside the `void Capture` helper returns from the *helper*, so the test continued on an empty buffer. With both captures empty the comparison still fails, but with only the **CPU** one empty it passes. All four call sites now use `ASSERT_NO_FATAL_FAILURE`. |
| **Extend the crack proof to Z** | Real gap. The seam kernel packs `+X/-X` and `+Z/-Z` through separate paths and the vertex stage snaps them in separate branches, so an X-only sweep left the Z arithmetic unproven. The test now sweeps both axes, each with its own anti-vacuous cross-level guard. |
| **Reject non-finite height samples** | Real, and a repo rule I had missed (`CLAUDE.md`: validate every float read from a file with `std::isfinite`). A NaN height propagates through the whole pyramid, and a NaN AABB makes every frustum comparison false — so the node tests visible at *every* level and the descent splits the entire tree. Samples are now skipped, with a finite fallback if every sample in a node is bad. |
| **Validate `maxDepth` in the public overload** | Taken. The pyramid is `4^(maxDepth+1)/3` entries, so an unchecked depth allocates before any other guard runs. |
| **Validate `m_TargetTriangleSize`** | Taken. It is user-editable and round-trips through YAML; every comparison against NaN is false, so a non-finite threshold silently stops the descent splitting at all — one root-sized patch. Falls back to the config default. |
| **Reject partial GPU-driven state** | Taken. With only one of the two handles supplied the dispatcher took the non-indirect path while the VAO was the shared *unit* grid — one garbage patch, drawn silently. Now fails the draw. |
| **Correct the misleading Vulkan comment** | Correct, my error. `ResolveIndirectBuffer` validates the resource *kind*; the usage bit comes from `VulkanStorageBuffer::CreateBuffer`. Comment rewritten. |
| **Compile-time bounds for `kMaxDepth`** | Taken — the packed coord gives the level 4 bits and x/y 14 bits each, and violating either is silent truncation rather than a compile error. |

### Declined, with reasons

**Renumber section 6a.** The suffix is deliberate: 6a is about the same compile-route topic as 6. Renumbering churns every cross-reference for no reader benefit.

**Capture the discarded parameters in `MockRendererAPI`.** `Record()` takes only a name; adding per-parameter capture is new mock machinery for a path nothing asserts on yet. Worth doing when a test needs it.

**`ShaderReflectionBindingTest` accepts slot 57.** True, and a real (small) loss of strictness versus the old bound of 53. I kept the simple bound rather than an exclusion list because 57/63 are `OLO_VULKAN`-only and this test does not compile that branch, so a stray declaration at 57 on the default route is already impossible to *reach*; the comment now documents exactly that.

**Per-frame cost scales as `4^maxDepth`.** Correct observation and worth recording: the split-map clear and the LOD-map dispatch are both proportional to the tree, not to the visible set. At the depths reachable today (4–6) that is negligible; at depth 10 the LOD map alone is ~1M texels/frame. It does bound how far the "deeper quadtree" headroom is actually usable, and it is the first thing to fix if anyone pushes the depth up — noted in the agent-rules doc rather than pre-optimised here.

**Move `PollOverflow` to the start of `Dispatch`.** Reasonable — reading the previous frame's flags before the reset would avoid a sync on this frame's work. Deferred: it currently runs once per 240 frames and only until the first warning, so the stall is rare enough not to justify reordering the state reset late in review.

### Out of scope (raised separately)

**Shadow casters still use the chunk meshes.** Investigated properly rather than deferred on assumption: `ShadowRenderPass` draws every terrain caster into every cascade with **no per-cascade culling at all** — for meshes and voxels too, not just terrain. So the choice here is between the current camera-frustum list (off-screen casters missing) and `GetAllChunks()` (correct, but ~256 chunks × 4 cascades of extra draws, unmeasured). Neither belongs in this PR; the real fix is per-cascade light-frustum culling in the shadow pass, which is shared machinery and pre-existing behaviour rather than a regression from #714.

Issue #702, opened 2026-07-31
Renderer: Virtual Shadow Maps — sparse page-table directional shadows replacing fixed-res CSM


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

Directional shadows are a fixed 4-cascade `Texture2DArray` at a single authored resolution (`Renderer/Shadow/ShadowMap.h`, `ShadowSettings::Resolution`). Resolution is therefore a global compromise: near-field contact detail and far-field coverage trade against each other, every cascade is fully redrawn every frame regardless of what changed, and `MaxShadowDistance` defaults to 200 m because extending it just thins the same texel budget. This is also the structural cause behind the shadow cost issues we have already hit (#651 Nanite shadow at 5.7x the geometry pass, #522 full-cost pass when nothing casts).

**Describe the solution you'd like**

A GPU-driven Virtual Shadow Map for the directional light:

- **Virtual page table (VPT) + physical page table (PPT)** with per-page allocated / visible / dirty flags; N concentric clip levels (Timberdoodle uses 16 x 4096^2, each level 2x the previous, first level ~2 m wide -> ~65 km total).
- **Manual GPU allocation, not sparse-texture APIs.** Sparse/tiled resources cannot be driven indirectly — knowing which pages need backing would require a GPU readback round trip, which produces shadow pop-in. Allocation, freeing and page marking all run on the GPU.
- **Per-page caching.** A drawn page keeps its contents and allocation into the next frame. Requires clip frusta to move in page-sized increments and the light matrix to slide along a plane parallel to its near plane, so cached texels stay valid after translation.
- **2D wraparound addressing** so pages entering the clip frustum reuse the VPT slots of pages leaving it.
- **Hierarchical Page Buffer (HPB)** — a HiZ-shaped structure built by 2x2 *max reduction of the dirty flag* (not depth), used to cull geometry that overlaps no page being redrawn this frame. Without it the many-clip-level draw is unaffordable.
- Invalidation of cached pages touched by dynamic objects (a dedicated drawlist).

**Describe alternatives you've considered**

Raising `ShadowSettings::Resolution` / `AtlasResolution` — buys a constant factor at quadratic memory cost and does nothing for the redraw-everything-every-frame problem. Cascade count tuning — same trade, just moved around.

**Additional context**

[...]

PR #820, opened 2026-08-17
Virtual Shadow Maps — sparse page-table directional shadows (#702)

### LINKED ISSUE #654: Nanite / virtualized-geometry roadmap: gaps vs UE5.8
Roadmap tracking issue for the virtualized-geometry (Nanite-style) system built under #629. This captures the gaps between our implementation and **Unreal Engine 5.8's** Nanite, derived from a source-grounded comparison of both codebases. It is a roadmap, not a committed sprint — items are prioritized by value-for-effort **in OloEngine's context** (some UE features are deliberately out of scope).

## Where we stand

Our system is a **visibility-buffer, software-raster + hardware-MDI hybrid, deferred-only** Nanite on plain OpenGL 4.6 — no mesh shaders, no vendor lock-in. On the *core* virtualized-geometry loop we are at parity with UE5.8. The gaps are in the advanced surrounding features, not the fundamental technique.

**Already at parity:** bounded-triangle cluster generation; the LOD DAG (group simplify, monotone absolute error, watertight crack-free cut); GPU cluster culling (frustum + normal cone); 64-bit atomic visibility buffer (plus a portable 2×32 fallback UE doesn't need); compute software rasterizer for micropolys; hardware MDI raster fallback; deferred material resolve from the visibility buffer; baked/serialized cluster data (`.omesh` sidecar with cook fingerprint); masked-material handling (forced to HW path, same rule as UE); cull stats + debug views (ClusterId / LOD / Overdraw).

## Status — verified against the code on 2026-09-09

A `/start-work` sweep re-checked **every box below against `master`** rather than trusting the
checkboxes, per `CLAUDE.md` (*docs and TODOs lie; verify against the code*). One Tier-1 item had
already shipped and one Tier-3 item's justification had gone stale. Both are corrected in place.

This roadmap carries `effort: 13` — the rubric's own "split it" verdict — and had **no children
filed**, so none of its work was visible to `issue_scores.py rank`. **Every remaining item is now
filed and scored**: #1143 (Tier 1), #1144 (ray tracing), #1149-#1151 (Tier 2), #1152-#1155 (Tier 3).
The Tier-3 four are expected to sort low — that is the correct outcome for "defer or skip" work, and
sorting low is still strictly better than being invisible. **This issue is now an index, not a
backlog**: it holds the UE5.8 comparison and the tiering rationale; the work lives in the children.

| item | state on `master` |
|---|---|
| T1 two-pass occlusion vs prev-frame HZB | **done** (#682, #726) |
| T1 persistent-thread cull / multi-view cull | open, filed as **#1143** |
| T2 Virtual Shadow Maps for VG | open, filed as **#1149** — VSM exists engine-wide; only the VG side is missing |
| T2 skinned / skeletal Nanite | open, filed as **#1150** — hard rejection at `VirtualMeshBuilder.cpp:736` and `:1016` |
| T2 on-disk page streaming | open, filed as **#1151** — `LoadPage` copies from the in-memory `Packed` payload |
| T3 WPO / tessellation / spline meshes / programmable raster | open, filed as **#1152**, **#1153**, **#1154**, **#1155** — no code for any |
| T3 ray-tracing fallback mesh | open, filed as **#1144**, and no longer correctly Tier 3 |

### Tier 1 — high value, feasible in GL 4.6, do first

- [x] **Two-pass occlusion culling with a persistent previous-frame HZB.** Today the cull only occludes against *same-frame* ScenePass depth (`VirtualClusterCull.comp` `IsOccluded`, built in `VirtualGeometryPass`), so an all-Nanite scene gets no occlusion benefit and the test is off by default. Adopt the standard Nanite two-pass scheme (cull vs prev-frame HZB → draw → build HZB → re-test disoccluded). Directly targets the "FPS cost even at zero drawn clusters" seen in the #629 stress scene.
  - **Already shipped** — verified in code 2026-09-09. `VirtualGeometryPass.cpp:337-345` reads `Renderer3D::GetRetainedOcclusionHZB()` — the pyramid built at the tail of the previous `EndScene` from that frame's FINAL depth, so it already contains virtual geometry — and gates a `twoPhase` flag on it (#682). Two-phase is deliberately OFF while the culling camera is frozen (#726), since the rebuild would destroy the pyramid phase 1 tested against. The HZB cull itself remains off by default.
- [ ] **Single persistent-thread cull over all instances × views.** We currently dispatch the cull *once per instance per cascade/face* (`VirtualGeometryPass` / `VirtualGeometryShadow::RenderCascade`) and re-cull serially. With many instances × cascades this is the dominant cost. Collapse to one work-queue dispatch (UE `NaniteHierarchyTraversal.ush` persistent-thread model). Highest-leverage perf change.
  - **Filed as #1143**, together with the multi-view item below.
- [ ] **Multi-view cull (view array in one pass).** Cull main view + all shadow cascades/faces from one packed view array (UE `FPackedView` / `NANITE_MAX_VIEWS_PER_CULL_RASTERIZE_PASS`). Prerequisite for efficient shadows and VSM; couples with the persistent-thread item.
  - **Filed as #1143** — the two restructure the same dispatch and touch the same files, so splitting them would half-migrate the cull.

### Tier 2 — large but architecturally correct

- [x] **Virtual Shadow Maps.** Replace classic CSM/atlas shadow casting for virtual geometry (`VirtualGeometryShadow`) with a page-table VSM (per-page on-demand draw commands). The proper pairing with Nanite; heavy but the marquee feature. — #1149, PR #1167. **Clip levels only**: the local-light LAYERS still need the multi-view cull (#1143), so with `LocalLights` on a virtualized caster casts the sun's shadow and not a lamp's. The classic CSM + atlas path is unchanged and still what runs with VSM off (the default).
  - **Filed as #1149. Note the roadmap text below is imprecise: **VSM already exists engine-wide** (`Renderer/Shadow/VirtualShadowMap.cpp`, `Renderer/Passes/VirtualShadowMapMarkPass.cpp`); only the virtual-geometry side is missing, so this is an integration, not a build.**
- [ ] **Skinned / skeletal Nanite.** Remove the build-time hard rejection of skeleton/morph/bone sources (`VirtualMeshBuilder`); add a skinning permutation to the cull/raster path (UE `r.Nanite.AllowSkinnedMeshes`, `NaniteEncodeSkinning`). Unlocks animated characters on the virtualized path.
  - **Filed as #1150.**
- [ ] **On-disk page streaming.** Our streaming is CPU-resident-payload → VRAM residency (data baked whole into `.omesh`, loaded up front). Add true disk→CPU root+streaming page paging (UE `NaniteStreamingManager`) for scenes exceeding memory.
  - **Filed as #1151.**

### Tier 3 — advanced, defer or skip

- [ ] **World Position Offset** (vertex animation on virtual geometry).
  - **Filed as #1152.**
- [ ] **Tessellation + displacement.**
  - **Filed as #1153.**
- [ ] **Spline meshes / landscape integration.**
  - **Filed as #1154.**
- [ ] **Ray tracing fallback mesh** — **Filed as #1144, and no longer Tier 3.** The original text read *"blocked until OloEngine has RT/Lumen at all; N/A for now"*, which is stale: #978 (BLAS/TLAS manager), #1055, #1056 and #1057 have all closed, so the RT tier exists. Verified 2026-09-09 — `RayTracingScene.cpp:166-170` excludes virtualized-cluster entities from GPU Scene (counted in `GPUSceneUnsupportedCategory`, so the exclusion is loud, not silent), which makes virtual geometry invisible to every ray-traced effect in the engine.
- [ ] General **programmable raster** beyond masked-to-HW.
  - **Filed as #1155.**

**Explicitly out of scope** (deep UE-specific investments with little payoff here): Nanite Voxels/SVO, Assemblies, Curves/hair, Nanite Translucency, Material Cache, mesh-shader barycentric raster path.

## Evidence

Comparison grounded in `D:/repos/UE_5.8` (`Engine/Source/Runtime/Renderer/Private/Nanite/`, `Engine/Shaders/Private/Nanite/`, `Engine/Source/Developer/NaniteBuilder/`, `Engine/Shaders/Private/VirtualShadowMaps/`) and our `OloEngine/src/OloEngine/Renderer/VirtualGeometry/` + `OloEditor/assets/shaders/**/Virtual*`.

Related: #629 (virtualized-geometry epic), #607 (MCP diagnostics — cluster/LOD inspection tools).

## Score
```olo-score
capability: 5
craft: 8
stability: 1
decay: 3
effort: 13
confidence: 0.5
learning: 8
fun: 5
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
Implements **#702** in full: sixteen concentric clip levels of a 4096² virtual shadow map over a configurable physical page pool of 64² pages, allocated and cached entirely on the GPU. No readback round trip — a readback would reintroduce exactly the pop-in VSM exists to remove.

## Does it work?

**Yes, on both backends, verified by eye.** Live editor, four poses of `VirtualShadowMapTest.olo`, CSM vs VSM.

**Reading the table below: CSM is NOT broken.** It casts correctly — the headless evidence test *requires* the CSM baseline to darken the floor before it will conclude anything about VSM, and it does, with margin. The rows are the inherent fixed-resolution artifacts of four cascades stretched over the scene's `MaxShadowDistance: 600`, which the test scene sets **deliberately**: "shadow distance well past the 200 m default" is an acceptance criterion of #702, and it is precisely the regime a fixed texel budget cannot serve without thinning everywhere — thin features drop below a cascade texel, and bias plus front-face culling detach what remains. The table is the value gap the feature exists to close, shown in a deliberately CSM-hostile configuration. At the 200 m default, CSM's rows would look considerably better; VSM's would not change.

| | CSM (4 cascades over the same 600 m) | VSM (GL **and** Vulkan-forward) |
|---|---|---|
| thin post (0.18 m) | no shadow (below cascade texel size) | crisp, full length |
| near cube / sphere | no contact shadow (bias + texel size) | attached at the base |
| mid box at ~15 m | shadow lost | attached, complete |
| far pylon at ~70 m | detached, floating | attached |
| clip-level banding | — | none observed |

**Measured, not forecast** (test scene, 1024×768): `91.8 MB` physical pool. Peak `1094` pages resident, `1094` drawn on the warm-up frame, and **a static camera settles to `0` pages redrawn** — acceptance criterion #2. `0` allocation failures, `0` cull overflows.

**Vulkan status, precisely:** full verification ladder — the shaders compile with `OLO_VULKAN` (harness-controlled test; those branches had never been parsed by anything before), a full VSM frame executes on a real device (`VulkanPassSuite.VirtualShadowMapRunsAFullFrameOnVulkan`: 291 pages allocated, 0 failed, real depth in the pool, zero validation errors), and the editor comparison above confirms the shadows land where CSM's do — nothing mirrored, which settles the one-line y-flip composition in `VirtualShadowRasterStage.glsl`.

## Bugs this work surfaced and fixed (each invisible to every prior test)

1. **`VirtualShadowMapMarkPass::Setup` early-returned while VSM was off — frozen in forever by the frame-graph fingerprint cache** (`91acb7a`). Enabling VSM after the first frame marked nothing, allocated nothing, rendered unshadowed with no error anywhere. Written up as [`virtual-shadow-map-page-cache.md` §5](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/virtual-shadow-map-page-cache.md) — the rule generalises to any pass with a runtime toggle. Guarded headlessly by `VirtualShadowMapMarkPassTest`.
2. **`HEAP_IMAGE_SLOT_BASE` drifted from its hand-written GLSL twin** (`6d4eba8`). Adding `TEX_VSM_PHYSICAL` moved the derived base 66→67; every bindless storage image then read a *sampler* descriptor through an image declaration. Now pinned headlessly, both mirrors.
3. **Three raw bind sites pushed the RHI ratchet 26→29** (`6d4eba8`). Routed through the heap seam; the image binds moved after their shader's `Bind()`, which is load-bearing (the seam forks on the program in flight).
4. **`ImageFormat` had no `R32UI` member** (`e73e81f`). On Vulkan, `CreateTexture2DHandle(R32UInt)` returned the null handle — VSM's entire pool didn't exist there, with one warn-once line as the only symptom. Wired end to end, mirroring `R32I`.
5. **VSM restored a degenerate `{0,0,0,0}` viewport** (`e73e81f`) — harmless on GL, `VUID-VkViewport-width-01770` per frame on Vulkan.
6. **The MCP `virtualshadowmaps` set-handler reported the *requested* state, not the effective one** (`c0cedb5`, via review). Apply built the reply before `SetSettings` ran, so a refused enable answered `"on"` while CSM rendered — the CPU-side twin of the Vulkan lie in (4). Both directions of that lie are now closed.

Also surfaced, **filed rather than fixed** (pre-existing, not VSM): the Vulkan editor's *deferred* path renders a frozen frame — light edits never reach the screen with VSM out of the picture entirely (**#823**, full evidence chain). Live Vulkan lighting verification currently requires `renderpath=forward`; that is how the comparison above was run.

## Review (CodeRabbit)

**18 of 22 findings fixed** (`c0cedb5`): the MCP effective-state bug above, a missing `BufferUpdate` barrier before the stats readback, `IsReady()` on the graphics shaders (a non-null *failed* shader passed the load check), epsilon float compares in `SetSettings`, GPU-side bounds checks on the cull's CPU-supplied batch records (capacities added to the C++/GLSL mirror test), the double linear batch scan replaced by a hash-map lookup, count-buffer creation order, caster-budget truncation logging, derived strides in the clear kernel, unified VISITED-cycle docs, and a batch of test hardening (sampler binding literals pinned, fixture `TearDown` restore, exact 13-stage count, the GPU test's own table declaration pinned).

**4 answered on the PR, not changed**: the cull's z-rejection matches the raster's own z clip exactly (no depth clamp exists in the facade — an x/y-only cull would emit draws the raster clips anyway; real depth-clamp support is a facade-extension follow-up); the CSM fallback for terrain/foliage/voxel casters is the documented §7 limitation behind default-off; the access-specifier indent is the repo-wide clang-format output; the PCH note folded into the include fix.

## Testing

Full suite **5992 ran, 5970 passed, 0 failed, 22 skipped** (post-merge with #819 *and* #818).

- `VirtualShadowMapTest` — CPU contract tests (L1): wrap predicate, clip-level agreement, per-level depth equality, offset/delta saturation, the C++/GLSL constant mirror (now including the cull capacities), and the two hand-written sampler binding literals pinned to their layout constants.
- `VirtualShadowMapMarkPassTest` — headless guard for bug 1, both halves (`Setup` and `IsEnabled`).
- `VirtualShadowMapVisualEvidenceTest` (L8) — CSM vs VSM on a scene it builds; golden-free, differential: both cast, same darkness-weighted centroid (within 2.5% of frame width; measured 1.1%), comparable mass (measured 1.11× — VSM peter-pans less), plus the caching claim via a peak-vs-settled split. Restores CSM from `TearDown`.
- `VirtualShadowMapVulkanShaders` — all 11 shaders / exactly 13 stages compile with `OLO_VULKAN` for the Vulkan 1.4 target; the harness proves the macro reaches the preprocessor (its failure mode is passing).
- `VulkanPassSuite.VirtualShadowMapRunsAFullFrameOnVulkan` — the first shadow-pass case in that suite; full frame on a real device under the zero-validation-error gate.
- `BindlessShaderPipeline.HeapImageBaseMatchesTheBindingLayout` — pins every mirror of bug 2, including the GPU test's own inline table declaration.

## Binding-space merges (#819, then #818)

Both landed mid-flight and collided; the later arrival moves, so VSM has been renumbered twice:

| | final |
|---|---|
| `UBO_VIRTUAL_SHADOW` / `_DRAW` | **81 / 82** (#713 took 79, #714 took 80) — `UBO_BINDING_LIMIT` 83, **one slot left** under the GL 4.6 minimum of 84, static_assert armed |
| the ten VSM SSBOs | **68–77**, contiguous above the reserved pull slots (#713 owns 54–56, #714 owns 58–62/65–67) |

The #818 SSBO overlap (eight of VSM's ten buffers) **auto-merged silently** — both blocks compile side by side and only the runtime would have noticed, as terrain culls and shadow pages swapping buffers mid-frame. Worth remembering as a class: binding-number collisions between parallel branches are invisible to git and to the compiler.

## Debugging it later

`VirtualShadowMapSettings::DebugMode` has seven views. Start at **7 (Shadow factor)** — the number the lit pass actually receives; it splits "the page table is wrong" from "the term is computed and then dropped" in one frame.

## Deliberately not in this PR

- **Default is OFF.** VSM covers static and skinned *mesh* casters; terrain, foliage, voxel and virtualized-geometry casters still use CSM (documented in §7 of the agent-rules doc).
- VSM settings don't persist to disk (runtime toggle only).
- #823 (Vulkan deferred frozen frame) — pre-existing, filed with repro.
- Depth-clamp support in the facade (would let the VSM cull/raster keep casters beyond ±DepthRange along the light axis).
- Retrofit of #713's prefix-sum onto the allocator's `atomicAdd` compaction.
- On merge: tick #654's VSM checkbox (**do not close #654** — umbrella); closing #702 is the maintainer's call.
- Two MCP capability gaps logged on #607 (a `DebugMode` lever, `olo_virtual_shadow_map_stats`).


drsnuggles8:
**Merged `master` in (`59e6133`) — #819's GPU prefix-sum collided on bindings, so VSM moved.**

| | was | now |
|---|---|---|
| `UBO_VIRTUAL_SHADOW` / `_DRAW` | 79, 80 | **80, 81** (`UBO_BINDING_LIMIT` 82, still inside the GL 4.6 minimum of 84) |
| the ten VSM SSBOs | 54–56, 58–62, 64–65 | **58–62, 64–68** (still skipping the reserved vertex-pull 57 and bone-pull 63) |

VSM moved because it is the later arrival — #819's numbers are already named by shipped shaders.

The SSBO overlap was worth more than a mechanical renumber. #819 documents 54–56 as **scratch**, rebound before every scan dispatch; VSM binds its ten **once** in `BindWorkingSet()` and then runs nine kernels off them. Sharing would have meant any scan dispatch landing between the bind and a later VSM kernel swapped the page table out from under it — visible only when particle compaction and the shadow pass interleave, which is not a bug that reproduces on demand.

Full suite on the merge: **5946 passed, 0 failed, 22 skipped** — VSM's tests and #819's prefix-sum tests green together.

drsnuggles8:
**Correction to the Vulkan claim in the description, and a test that closes half of it (`7e21599`).**

I wrote that the Vulkan path "compiles and is structurally complete but has never been executed". The second half was true. **The first half was not knowable, and I should not have written it.**

`OLO_VULKAN` is defined in exactly two places in the engine — `VulkanShader.cpp` and `VulkanComputeShader.cpp` — so an `#ifdef OLO_VULKAN` branch reaches a compiler *only* when a live Vulkan backend loads that shader. Nothing drives VSM on Vulkan. So the vertex-pull branches in `VSM_Depth`/`VSM_DepthSkinned` and the `gl_FragCoord` flip in `VirtualShadowRasterStage` had never been **parsed by anything**. Not untested at runtime — unparsed. A missing semicolon would have shipped with the entire GL suite green.

They do compile: all 11 VSM shaders / 13 stages, `OLO_VULKAN` defined, Vulkan 1.4 target env, using `VulkanShader`'s own option set spelled out option for option.

The harness is controlled, because this test's failure mode is *passing*: had `OLO_VULKAN` not reached the preprocessor, every shader would have compiled its **OpenGL** branch and the suite would have gone green while the Vulkan code stayed exactly as unparsed as before. So a probe `#error`s when the macro is absent, and a second assertion compiles that probe *without* the macro to prove the `#error` is live rather than inert. (The first version of the test also failed with 14 errors that were entirely its own fault — it resolved `#include`s across the whole file with one `seen` set, starving the fragment stage. It now splits stages first, like the engine's loaders do.)

**Still not proven: that VSM renders correctly on Vulkan.** The fork is a claim about two conventions cancelling (`fragCoord.y_vulkan = RES - fragCoord.y_gl`) and only a rendered frame settles it. `VulkanPassSuite` runs 30+ passes on a real device but has **no shadow pass at all**, so covering this means writing the first shadow-pass Vulkan case — not done here. The docs now say "the Vulkan shaders compile" and never "both backends are supported".

The general gap is named and left: every shader in the engine with an `OLO_VULKAN` branch has this same exposure. That is #691's debt, not #702's.

Full suite: **5949 passed, 0 failed, 22 skipped.**

drsnuggles8:
**Vulkan: VSM did not work there at all. Two engine gaps, both fixed (`e73e81f`), plus the first shadow-pass test in `VulkanPassSuite`.**

Running the editor under `--rhi=vulkan` found what compiling could not.

### 1. `ImageFormat` had no `R32UI` member

`RawFormatToImageFormat` mapped `RHI::Format::R32UInt` to nothing, so `CreateTexture2DHandle(res, res, R32UInt)` returned the **null handle** — VSM's entire physical pool did not exist on Vulkan and every `imageAtomicMin` went nowhere. VSM still initialised and still logged its 91.8 MB. The only surface symptom was one warn-once line per shader:

```
[RHI/Vulkan] 'VSM_ClearDirtyPages' image binding 0 has no staged heap slot — slot 0
```

which reads like a VSM binding bug rather than a missing enum three layers down. VSM is the first system in the engine to want an R32UI texture — the page pool stores raw float depth *bits*, because `imageAtomicMin` has no float form — which is why the hole sat there. `R32UI` is now wired end to end, mirroring `R32I` at every site.

### 2. VSM restored a degenerate viewport

`GetViewport()` answers `{0,0,0,0}` when nothing has set one, which is the normal case because VSM runs at the top of the frame. GL shrugs; Vulkan raises `VUID-VkViewport-width-01770` once per frame.

### The test

`VulkanPassSuite.VirtualShadowMapRunsAFullFrameOnVulkan` — the first shadow-pass case in that suite. Real API, frame order, real device. Measured: **291 pages allocated, 0 failed, 2 draw instances, 341 760 pool texels holding real depth, zero validation errors, zero unimplemented-facade hits.**

Two things the *test* got wrong first, both now recorded in comments because both produce "the raster is broken" from a healthy raster: it ran outside a recording bracket (90 silent no-op dispatches — the graph opens the bracket for every other test, and VSM is driven directly), and it used a constant `0.5` scene depth, which on a perspective buffer unprojects to ~0.2 units from the eye, marking pages at the *camera* while the caster sat 13 units away.

### What is still not proven

**No side-by-side of a Vulkan frame against a GL one.** I tried and could not get a meaningful one: in `VirtualShadowMapTest.olo` under Vulkan the frame comes back flat — no skybox and no CSM shadow — **with VSM off**, so the baseline the comparison rests on is not casting either and CSM-vs-VSM frames are byte-identical. I have *not* root-caused that and I am not claiming it is pre-existing #691 debt; the default scene renders sky and water fine on Vulkan, so Vulkan is not broadly broken, but that is a weak control and I am not dressing it up as more.

So the honest line, which the docs now carry: **VSM executes correctly on Vulkan; the y-flip composition is argued and proven to run, not confirmed by eye.**

OpenGL is separately re-verified live after all of this — four poses, CSM vs VSM differing, hashes byte-identical to the earlier verified run, so none of the Vulkan work regressed it.

Also fixed: my verification script only de-duplicated *within* a technique, so when the Vulkan toggle silently did nothing it reported four clean captures of what were really CSM frames. It now compares CSM against VSM per pose and fails loudly on identical frames — that is how gap (1) surfaced.

Full suite: **5950 passed, 0 failed, 22 skipped.**

drsnuggles8:
**VSM is now confirmed by eye on Vulkan (`f77bd6c`) — and the earlier "flat frame" is a filed engine bug, #823, not a VSM problem.**

Root-causing the flat Vulkan frame with the capture tools instead of guessing:

| evidence | result |
|---|---|
| `ShadowMapCSMCascade0` capture on Vulkan | **real depth, correct silhouettes** — cascades render |
| sun `CastShadows` off (deferred) | frame unchanged → not the shadow term |
| sun `Intensity` 2.6 → 20 (deferred) | frame **byte-identical** → the deferred frame is frozen |
| sun `Intensity` 2.6 → 20 (forward) | frame changes → the light pipeline itself is fine |
| `renderpath=forward` | **sun, CSM shadows, skybox all correct on Vulkan** |

So: the Vulkan editor's **deferred** path renders a frozen frame — light edits never reach the screen, with VSM out of the picture entirely. Filed as **#823** with the full evidence chain (including the empty G-Buffer captures that need distinguishing from a capture-aliasing artefact). It is what made VSM look broken on Vulkan for a whole round: a CSM-vs-VSM comparison on a frozen path returns byte-identical frames.

**On the forward path, the real comparison ran — and passed.** Four poses of `VirtualShadowMapTest.olo` under `--rhi=vulkan`, CSM vs VSM (`PBR_MultiLight` carries the same VSM branch):

- the thin post casts a full crisp shadow (CSM loses it entirely, same as GL);
- near-prop contact shadows attach; the mid sphere and far pylon shadows attach instead of floating;
- **the shadows land in the same places as CSM's — nothing mirrored, no clip-level banding.**

That last point is what settles the one-line y-flip composition in `VirtualShadowRasterStage.glsl` — the docs' "argued but not confirmed by eye" caveat is now closed. The remaining honest caveat is inherited, not VSM's: live Vulkan verification of any lighting feature currently requires `renderpath=forward` until #823 is fixed.

Verification ladder for the Vulkan half, complete: shaders compile with `OLO_VULKAN` (harness-controlled) → full frame executes on a real device (zero validation errors) → **confirmed by eye in the editor, four poses, against CSM in-frame**.

drsnuggles8:
**CodeRabbit triage: 18 of 22 fixed (`c0cedb5`), 4 answered below. Plus a second master merge (`580c6cc`) — #818 collided on both binding namespaces.**

### Fixed (highlights; the commit message carries the full list)

- **MCP `virtualshadowmaps` now reports the *effective* state** — Apply built the reply before `SetSettings` ran, so a refused enable answered `"on"` while CSM rendered. This is the CPU-side twin of the lie the Vulkan round hit; both directions are now closed.
- **The cull bounds-checks CPU-supplied batch records** against `VSM_MAX_BATCHES` / `VSM_MAX_DRAW_INSTANCES` (both now in the C++/GLSL mirror test) — malformed records count as overflows instead of scribbling.
- **`BufferUpdate` barrier before the stats readback**; `IsReady()` on the graphics shaders; epsilon float compares in `SetSettings`; count-buffer creation reordered; caster-budget truncation now logs.
- **The double linear batch scan became one hash-map lookup** (worst case was kMaxCasters × kMaxBatches, twice, per frame).
- Tests: the two hand-written sampler binding literals are pinned against their layout constants; the evidence fixture restores CSM from `TearDown`; the Vulkan shader test asserts the **exact** 13-stage count; the heap pin now covers `BindlessHeapGpuTest`'s own inline table declaration.

### Answered, not changed

- **Cull z-rejection (Major):** the suggested x/y-only rejection would *desynchronize* the cull from the raster — no depth clamp exists in the facade (GL or Vulkan pipeline state), so the raster clips z at exactly the same ±`DepthRange` planes. The cull dropping what the raster would clip anyway is self-consistent; a caster >4 km along the light axis from the render origin is outside the documented scene-scale envelope (§2 of the agent-rules doc). Adding real depth-clamp support is a facade extension — legit follow-up, not a quick win.
- **CSM fallback for terrain/foliage/voxel casters (Major):** deliberate, documented (§7), and why VSM is default-OFF. Running both systems per frame is the cost the design declined; a scene relying on those caster types keeps CSM.
- **Access-specifier indent:** matches the repo-wide clang-format output (`ShadowRenderPass.h` et al. are identical) — changing one file would fight the formatter.
- **PCH include note:** folded into the include fix rather than tracked separately.

### Merge with #818 (terrain GPU LOD)

Third binding renumber this PR has absorbed, same rule each time — the later arrival moves. Terrain-cull owns UBO 80 → VSM to **81/82** (limit 83; **one slot left** under the GL 4.6 minimum of 84, static_assert armed). Terrain owns SSBOs 58–62/65–67, overlapping *eight of VSM's ten* — and that overlap **auto-merged silently**; only the runtime would have noticed, as terrain culls and shadow pages swapping buffers mid-frame. VSM now sits contiguously at **68–77**, above the reserved pull slots. GLSL twins renumbered with the C++.

Full suite post-everything: **5992 ran, 5970 passed, 0 failed, 22 skipped** — including the Vulkan full-frame test and the new pins.

(CI note: the earlier TSan and Sonar-build failures were `HTTP 429` infra during the GitHub degradation — the glad generator's spec download got rate-limited before any code compiled. Sonar re-run was already requested; the new push supersedes the rest.)

Issue #828, opened 2026-08-17
Fresh configure with STEAMWORKS_SDK_ROOT set produces a test binary that cannot start (0xC0000135)


drsnuggles8:
Found while cross-checking the AMD golden baselines (#735) in a **fresh worktree**. Filed separately rather than widening that PR.

## Symptom

On a machine with `STEAMWORKS_SDK_ROOT` set, a **fresh** configure + build of `OloEngine-Tests` produces an executable that fails to start. The build reports failure only at the very end, from gtest discovery:

```
CMake Error at .../GoogleTestAddTests.cmake:233 (message):
  Error running test executable.
    Path: .../build/OloEngine/tests/Debug/OloEngine-Tests.exe
    Working directory: .../OloEditor
    Result: Exit code 0xc0000135
...
error MSB3073: ... exited with code 1 [OloEngine-Tests.vcxproj]
```

`0xC0000135` is `STATUS_DLL_NOT_FOUND`. Running the exe by hand just prints nothing and exits with that code — **nothing in the message mentions Steam**, which is what makes this expensive to diagnose.

The missing DLL is `steam_api64.dll`:

```
dumpbin /DEPENDENTS OloEngine-Tests.exe   # ours: 30 deps, includes steam_api64.dll
                                          # a Steam-OFF build: 29 deps, no steam_api64.dll
```

## Root cause — configure ordering, not the Steam block

`OloEngine/CMakeLists.txt`:

- **line 49** — `add_subdirectory(tests)`, which runs `olo_copy_steam_runtime(OloEngine-Tests)` (`OloEngine/tests/CMakeLists.txt:1049`)
- **line 369** — `set(OLO_STEAM_RUNTIME_DLL "..." CACHE INTERNAL ...)`

`olo_copy_steam_runtime` (`cmake/CommonProperties.cmake:182`) is guarded by:

```cmake
if(OLO_WITH_STEAM AND NOT OLO_WITH_STEAM_STUB_SDK AND DEFINED OLO_STEAM_RUNTIME_DLL)
```

On a **fresh** cache the variable is not defined yet at line 49, so the guard is false and **no POST_BUILD staging rule is generated at all**. The engine still links `steam_api64.lib` (an *import* library), so the binary carries a hard import on a DLL nothing ever stages.

A **second** configure populates the cache, the rule appears, and everything works — which is exactly why this only bites fresh worktrees and never a long-lived build tree.

Confirmed by counting the staging rule in the generated project:

```
[...]

Issue #754, opened 2026-08-07
AtmosphereVisualEvidenceTest: NightClear/NightOvercast goldens drift (RMSE 13.08 / 9.40 vs threshold 8) — permanently red


drsnuggles8:
**Describe the bug**
`AtmosphereVisualEvidenceTest.DayNightWeatherMatrixRendersAndHoldsContracts` fails on **every** run, in **both** the slot-based and heap-bindless configurations, on two of its matrix cells:

```
AtmosphereVisualEvidenceTest.cpp(408): error: Expected: (rmse) <= (kGoldenRmseThreshold),
    actual: 13.077706847047567 vs 8
    Capture 'NightClear' drifted from its golden (RMSE 13.077706847047567)

AtmosphereVisualEvidenceTest.cpp(408): error: Expected: (rmse) <= (kGoldenRmseThreshold),
    actual: 9.4036103562132709 vs 8
    Capture 'NightOvercast' drifted from its golden (RMSE 9.4036103562132709)
```

**It is deterministic, not flaky.** The two RMSE values are identical *to the last digit* across runs taken days apart, on different builds, in both configurations. This is a fixed offset between what the renderer produces and what the committed goldens hold — not noise, and not a threshold that is marginally too tight.

**The failure is confined to NIGHT, and that is the diagnostic lead.** `NightClear` and `NightOvercast` fail; `NightStorm` passes, as do all of Dawn, Dusk and Noon. Whatever changed is specific to the night sky path (star field, moon, night-side scattering) and appears to be *suppressed* by heavy overcast — which is presumably why `NightStorm` survives.

**Root cause (unknown — two candidates, and they have opposite fixes)**
1. **The goldens are stale.** They were last written by 91e1e108 (#633, atmosphere/weather director), and a later change altered night rendering without rebasing them. Fix: confirm the current output is *correct*, then rebase.
2. **Night rendering regressed** after those goldens were captured. Fix: find and fix the regression — rebasing would silently bless the bug.

**Do not rebase without looking at the images first.** RMSE 13 on a night sky is not subtle; per `CLAUDE.md`'s rendering-verification rule, open `Atmosphere_NightClear.png` against its golden and decide which of the two it is. A blind `OLOENGINE_GOLDEN_REBASE=1` here would convert a possible rendering regression into a permanent, invisible baseline.

**To Reproduce**
```
[...]

PR #829, opened 2026-08-17
test(golden): cross-check the AMD baselines against NVIDIA, and audit every recorded baseline (#735)

### LINKED ISSUE #735: URGENT: AMD golden baselines are unverified — cross-check against NVIDIA before they are trusted

drsnuggles8:
Cross-checks the AMD/Mesa golden baseline set (committed unverified in `5d776ba2`) against a fresh NVIDIA capture on the RTX 4090 dev box, and leaves behind the tooling that turns the next such check into a CI-gated assertion instead of a session.

## Verdict

**All 17 AMD baselines stand — no rebake, no defect.**

The measurement had to separate three things a naive diff conflates. Fresh NVIDIA capture at `d780bd6e4`:

| comparison | isolates | result (RMSE, 0..255) |
|---|---|---|
| run 1 vs run 2 | **noise floor** | **≤ 0.30**; 12 of 17 byte-identical |
| fresh NVIDIA vs committed shared | **temporal drift** | goldens **byte-identical**; Atmosphere day cells **0.87–2.60** |
| fresh NVIDIA vs committed AMD | **vendor** | **0.10–0.72**; the 4 goldens max **1 LSB** |

**The naive reading is backwards.** Diffing the committed AMD set against the committed shared set gives 0.23–2.59 and a textbook `InfiniteGrid` depth-tie heatmap — which reads as cross-vendor variance and is not. Fresh NVIDIA sits *closer to AMD* than to the shared baselines: both vendors drifted away from a shared set last rebaked 2026-07-16, **together**. On the worst cell (`DuskClear`), the same frame compared both ways:

| | pixels with any channel delta > 4 | 8×8 blocks | max delta |
|---|---|---|---|
| vs stale shared baseline (**drift**) | 11 718 / 921 600 | 560 | 70 |
| vs AMD baseline (**vendor**) | **5** / 921 600 | 4 | 39 |

The baselines' provenance confirms it: the shared **day** cells date from `91e1e108f` (2026-07-16) and carry the drift, while the **night** cells were rebaked in `f7e2db2f0` (2026-08-11, the #754 fix) and show essentially none — `NightClear` 0.016, `NightOvercast` 0.004. Drift tracks bake date, as it must if it is temporal rather than vendor.

Controls: the Atmosphere fixture is unchanged since the AMD bake (only the env-var→CLI-flag migration), and `AtmosphereSky.glsl`'s GL path is untouched (the diff since is a Vulkan-only `#ifdef` branch plus comments).

### The star field — the one claim a single vendor cannot check

`f4fef24b` replaced `fract(sin(...))` with an integer PCG hash because the old one placed stars in genuinely different *locations* on NVIDIA and Mesa. Confirmed bit-exact: **99.45–99.92%** positional agreement on the `NightClear` star mask across every threshold from +5 to +90 above the sky floor. For scale, shifting one mask by a **single texel** drops agreement to 88%. `NightOvercast`/`NightStorm` hold zero star pixels — the cloud deck occludes the sky, as intended.

### Correctness, not just agreement

A small RMSE measures *portability*, never *correctness* — two vendors running the same wrong shader agree perfectly. So #734's property guards were run against the recordings: **23/23 pass on the AMD set and 23/23 on the shared set**, with derived quantities agreeing to three significant figures (shadow attenuation 0.6162 vs 0.6190, night sky luma 58.00 vs 57.99, dawn warmth 1.0534 vs 1.0534).

## What ships here

1. **`GoldenBaselineAuditTest.cpp` — the first L9 *code*** (that layer was workflow-only). Works on the *committed* PNGs — the shared set and every vendor subdirectory, **discovered rather than listed** — with no GPU, so it gates every CI run rather than only the nightly owning that hardware. Two jobs:

   - **Property audit** — runs #734's guards against each recorded set, closing the gap `GoldenImageTests.cpp` names in its own #734 comment: *"where a vendor baseline was baked separately (#735) the compare can pass on a frame nobody has verified is correct."*
   - **Cross-vendor parity** (`VendorSetsAgreeWithTheSharedSet`) — compares each vendor set against the shared one. This is the check per-vendor scoping structurally cannot make: each set is otherwise only ever compared against itself, so a divergence both sets keep to themselves is invisible forever.

2. **Parse-time whitelist for `--olo-golden-vendor` / `--olo-perf-machine`.** Both become path components at their use sites, and two of the three consumers concatenated them unchecked — while `AtmosphereVisualEvidenceTest`'s validator claims to "mirror `GoldenImageTests::GoldenBaselineDir`", the mirror it names. Since `--olo-golden-rebase` *writes* to the composed path, a separator would drop PNGs outside the asset tree.

3. **Removal of 8 committed `.actual.png`/`.diff.png`** L10 escalation artefacts, all byte-identical to the baselines beside them, plus a `.gitignore` rule.

4. **`docs/agent-rules/vendor-golden-baseline-crosscheck.md`** — the method and the result.

### Why parity instead of dropping vendor scoping

The obvious conclusion from these numbers is "the vendors agree, so delete the per-vendor sets and let AMD compare against the shared baselines". That trades one property for another: a per-vendor baseline gives a **tight self-comparison** (no vendor term eating the threshold), a shared baseline gives **cross-vendor conformance**. Keeping both sets *and* comparing them gets both, needs no GPU and no AMD hardware, and touches no workflow.

The parity check is two-tier, because the two metrics fail in opposite directions:

| tier | frames | gate | why |
|---|---|---|---|
| 1 | the 4 renderer goldens | **pixel** — RMSE < `kRmsePassBelow`, max delta ≤ 2 | procedural and drift-free, so a pixel gate means what it says (measured: 1 LSB) |
| 2 | the 13 Atmosphere captures | **derived** — per-band luma delta < 2.0, plus a loose RMSE ≤ 8.0 backstop | pixel distance here tracks bake-date skew more than vendor (measured 2.60 apart, only 0.72 of it vendor) |

Tier 2 is the lesson of this whole investigation turned into a gate. Band luminances are **vendor-sensitive but drift-insensitive** — across a month of bake-date skew they agreed to 0.43/255 while pixel RMSE ranged to 2.60 — so they gate tightly without firing every time one set is rebaked and the other isn't. Both bars are expressed as *"the vendor set would pass the shared set's own gate"* rather than as invented constants, so they move if and only if those gates move.

## Follow-up, not done here

**Rebake the shared Atmosphere set at HEAD.** It carries 0.87–2.60 of accumulated drift. Nothing is red because of it (threshold 8.0), but that drift is why tier 2's backstop cannot be tightened, and it is the same normalised-drift pattern #754 was about. Left out because it wasn't caused by this change and would add 13 large binary diffs.

Also filed **#828** separately (fresh configure with `STEAMWORKS_SDK_ROOT` set produces a test binary that cannot start) — hit while doing this, unrelated to goldens.

## CI note

The five red checks are **inherited from master, not from this PR**. All five are the same single test — `AssetContentValidity.SandboxAssetRegistryDeserialisesAndPathsResolve` — which fails identically on master at `d780bd6e4`, this branch's base ([master run 32060140380](https://github.com/drsnuggles8/OloEngineBase/actions/runs/32060140380)). It reports 10 missing `SandboxProject/Assets/cache/embedded/emb_*_basecolor.png` files that are registered in `AssetRegistry.oar` but not tracked in git. Being fixed separately.

## Review guide

**Where I'd look hardest**

1. **The tier-2 thresholds** in `VendorSetsAgreeWithTheSharedSet` — `kBandLumaDelta = 2.0` and the `kAtmosphereRmse255 = 8.0` backstop. This is the judgement call of the PR: too tight and the check fires on bake-date skew (which would be the exact confusion this task existed to resolve), too loose and it stops catching anything. Current margins are 4.6× and 3.1× against measured values, but those come from *one* cross-vendor pair.
2. `CoversTheCommittedBaselineSets` — the guard that stops the file passing vacuously, so its own edge cases matter most. It deliberately treats an **empty** vendor directory as "not a baseline set", because `CompareOrBootstrap` calls `fs::create_directories` on the composed path *unconditionally* before checking whether the baseline exists — so any `--olo-golden-vendor=<v>` run leaves one behind, including `cross-vendor.yml`'s llvmpipe.
3. `GoldenBaselineAuditTest.cpp` thresholds generally — mirrored from #734's guards by hand. If one moves there and not here, this file silently stops matching what it claims to mirror. I did not solve that structurally: the #734 guards close over GPU readbacks and cannot be called from a CPU-only test.

**What I verified, and how**

- **The audit fails when it should.** Six deliberately-broken baselines — a saturated ramp, an unblended FXAA edge, a hard-thresholded shadow, a swizzled splatmap, an unreadable PNG, and two structurally-broken sets bracketing the good ones alphabetically. Each fired with the diagnostic naming that defect class, while the real sets passed in the same run. The bracketing pair specifically proves the `EXPECT`+`continue` refactor works: both are reported, where the old `ASSERT` reported only the first.
- **The whitelist rejects what it should**: `../../evil`, `..\..\evil`, `C:evil`, `..`, `.`, `` all exit 2; the four real workflow values (`amd`, `llvmpipe`, `ci-llvmpipe-windows`, `selfhosted-amd-navi10-linux`) are accepted. The `C:evil` case is why it is a whitelist — drive-relative paths contain no separator yet `fs::path` treats them as rooted, so `base /= value` silently discards `base`.
- **The captures touched nothing authoritative**: both runs wrote only to scratch vendor directories; `git status` showed no modified tracked PNG.
- All numbers come from a reimplementation of the engine's own `ComputeRgbSsim` / `ComputeRgbRmse` / `Rgba8Rmse`, cross-checked against all four `GoldenImageSsimTest` expectations before use.

**Least confident about**

The tier-2 gate is calibrated against a single vendor pair at a single point in time. If a second vendor set lands (llvmpipe on a software rasteriser, say), 2.0 on band luma may well be too tight — llvmpipe is a different kind of difference from AMD-vs-NVIDIA. I'd rather that be discovered by the check firing and being loosened deliberately than by picking a number now that tolerates everything.

Also: everything here compares fresh NVIDIA at HEAD against an AMD *recording* from 2026-08-01, so it cannot see a divergence introduced on the AMD side since. The parity test is what makes that cheap to re-check on the next AMD bake.

**Deliberately not tested**

`RequirePlainPathComponent`'s rejection path has no automated test: it calls `std::exit(2)`, which needs a gtest death test, and there is no death-test precedent anywhere in this suite. Verified manually instead of introducing the pattern for a ride-along fix.

Closes #735

## Summary by CodeRabbit

- **New Features**
  - Added automated, GPU-independent validation for shared and vendor-specific rendering baselines.
  - Added rendering property, cross-vendor parity, and coverage checks for visual test assets.
  - Added safer validation for vendor and performance configuration values.

- **Documentation**
  - Added guidance for cross-vendor baseline verification, troubleshooting visual differences, and capturing test artifacts.
  - Updated testing documentation to describe continuous baseline auditing in CI.

- **Chores**
  - Excluded generated comparison images and heatmaps from version control.


drsnuggles8:
## 🤖 Self-review @ `013478072297e86424006b7b3414b5d237cfcc90`

Reviewed the PR diff at **high** effort before opening, per `docs/process/task-loop.md` Phase 3.

- **Findings:** 3 · all real, all fixed in `013478072`:
  - **Empty vendor directory was a false positive.** `CompareOrBootstrap` calls `fs::create_directories` on the composed path *unconditionally*, before checking whether the baseline exists — so any `--olo-golden-vendor=<v>` run leaves an empty `golden/<v>/` behind, including `cross-vendor.yml`'s llvmpipe and the scratch-directory workflow this PR's own doc recommends. The guard now fires on "PNGs present but none readable" instead, with an unreadable-PNG negative control proving that branch fires.
  - **In-loop `ASSERT_*` aborted the whole test**, so one malformed vendor set stopped every *later* set being audited — losing coverage exactly when something is already wrong. Now `EXPECT` + `continue`, verified with two broken sets bracketing the good ones alphabetically: both are reported in a single run, where the old code reported only the first.
  - **The doc asserted a mechanism I had not verified** — that a missing vendor directory "is not an error". It is a hard fail (`EXPECT_TRUE` on `m_Passed`; `ASSERT_NE` in `AtmosphereVisualEvidenceTest`). The conclusion held, the stated reason was invented. Corrected and marked unconfirmed from the runner.
- **Also self-caught on recount:** 12 of 17 captures are byte-identical between runs, not 13; and the goldens' margin is 2.5× on `kRmsePassBelow` / 12× on `kRmseFailAbove` (one sentence had attributed the 12× to the pass gate).
- **Dismissed:** 1 — `LoadBaseline` clearing stb's sticky flip flag. This file is the first *un-GPU-gated* caller, so the pre-existing-pattern argument needed checking rather than assuming; every setter (`AssetSerializer`, `Model`, `TextureCompression`, `OpenGLTexture`, `VulkanTexture`, `IBLPrecompute`) sets it immediately before its own load, so nothing depends on it persisting across tests.

Re-ran Phase 2 after the fixes: build exit 0, audit 10/10 green with no GPU, negative controls still fire.

drsnuggles8:
## Added: cross-vendor parity @ `02aab6f99`

Per review discussion, this now does **both** rather than dropping per-vendor scoping: each vendor keeps a tight self-comparison, *and* the sets are compared against each other (`VendorSetsAgreeWithTheSharedSet`). Both baseline sets are committed, so the comparison needs no GPU and no AMD hardware — it gates every CI run and touches no workflow.

Why this is better than the drop-scoping option I originally recommended: a per-vendor baseline gives a **tight self-comparison** (no vendor term eating the threshold), a shared baseline gives **cross-vendor conformance**. Keeping both sets and comparing them gets both properties instead of trading one for the other.

**Verified with a control that isolates exactly what parity adds** — a vendor set diverged just enough that the property audit *passes* it while parity catches it:

| check | result on the diverged set |
|---|---|
| `ReinhardRampBaselinesAreToneMapped` | **PASS** (peak 234 sits inside the 200–245 guard band) |
| `AtmosphereBaselinesHoldTheirPhysicalContracts` | **PASS** (still bright, still blue, night still far darker) |
| `VendorSetsAgreeWithTheSharedSet` tier 1 | **FAIL** — RMSE 0.0133 > 0.004, max delta 4 > 2 |
| `VendorSetsAgreeWithTheSharedSet` tier 2 | **FAIL** — sky-band luma delta 4.0 > 2.0 |

Without the cross-comparison that set would be greenlit indefinitely. All 11 audit tests pass on the real sets, no GPU, in ~3 s.

**One bug caught in my own code while doing this:** `LoadPairForCompare` called `LoadForAudit` before its definition (`error C3861`). Fixed by reordering — which also restored an explanatory comment my insertion had orphaned from its function.

**On the red checks:** all five are the same single test, `AssetContentValidity.SandboxAssetRegistryDeserialisesAndPathsResolve`, which fails identically on master at `d780bd6e4` (this branch's base) — [master run 32060140380](https://github.com/drsnuggles8/OloEngineBase/actions/runs/32060140380). It reports 10 missing `SandboxProject/Assets/cache/embedded/emb_*_basecolor.png` files that are registered in `AssetRegistry.oar` but not tracked in git. Inherited, being fixed separately — nothing from this PR.

drsnuggles8:
## Review pass + master merge @ `440118cb3`

### Merged master (not rebased)
Per `CLAUDE.md` → *Committing and publishing* and `task-loop.md` §5d — **merge master in, never rebase, never force-push** — and this branch is already pushed, so a real rebase would need a gated force-push. Clean merge, no conflicts; verified both sides' edits to `CLAUDE.md` and `docs/agent-rules/README.md` survived.

This clears the five red checks. The fix rode along in `83ea8966d` (titled *"unbreak the Video (FFmpeg backend) Linux job"*), which is why looking for the missing `emb_*_basecolor.png` files said "not fixed" — it removed the **stale registry entries** instead of restoring the files:

| ref | stale `cache/embedded/emb_*` refs in `AssetRegistry.oar` |
|---|---|
| `d780bd6e4` (old base) | **10** |
| `origin/master` | **0** |

### Review findings — 2 fixed, 1 declined with reason

**Fixed — empty vendor directory made parity fail spuriously.** Reproduced it: with the only vendor set empty, the check died on *"nothing was cross-compared"*. Not hypothetical — that is the drop-scoping end state plus the empty directory `CompareOrBootstrap` creates as a side effect (it calls `fs::create_directories` on the composed path *unconditionally*, before checking whether the baseline exists), which is exactly what `cross-vendor.yml` does with llvmpipe. `DiscoverBaselineSets` now drops PNG-less directories.

The filter is **"holds a PNG", not "holds a readable PNG"**, deliberately: a truncated baseline must still count as a set so the coverage guard can report it, instead of vanishing from discovery and being policed by nobody. Verified both directions — the empty case now **SKIPs**, and a directory whose only PNG is unreadable is **still caught**.

**Fixed — strict `<` where the mirrored gate is inclusive.** `CompareOrBootstrap` gates on `m_Rmse <= kRmsePassBelow`; at exactly 0.004 this test failed where the real compare passes. Now `EXPECT_LE`.

**Fixed — duplicated filename lists.** Defined once at namespace scope. Coverage now also requires the shared **Atmosphere** set, which closes a real gap: the per-cell audits skip what they cannot load, so only a *total* loss tripped the `audited > 0` floor and a partial one (9 of 13) was invisible.

**Declined — "centralize the RMSE constants with the shared test helper used by GoldenImageTests and AtmosphereVisualEvidenceTest".** No such helper exists: `kRmsePassBelow` is a function-local `constexpr` inside `CompareOrBootstrap`, `kGoldenRmseThreshold` is TU-local. Centralizing means hoisting constants out of two test files this PR has no other reason to touch. Instead I took the finding's own fallback and corrected the claim it was propping up — code and doc now say the values are **copies, not references**, with *"if either gate moves, move it here too."*

### Verification
- 11/11 audit + SSIM tests green on the real sets, no GPU.
- Empty-vendor-dir reproduction now SKIPs (was FAIL).
- Unreadable-PNG control still caught; parity divergence control still fires on both tiers (RMSE 0.0133, sky-band luma delta 4.0).
- One knock-on caught and fixed: the coverage loop's comment still claimed empty directories were tolerated *there*, which the discovery filter had made untrue.

Issue #823, opened 2026-08-17
Vulkan editor: the deferred path renders a frozen frame (light edits never reach the screen)


drsnuggles8:
Found during #702's Vulkan verification (PR #820). **Not** a VSM bug — reproduces with VSM never enabled, and it is what made VSM look broken on Vulkan for a whole verification round.

## Repro (RTX 4090, driver 98.352.0, this branch = master + #820)

1. `OloEditor.exe --rhi=vulkan`, open `Scenes/VirtualShadowMapTest.olo`
2. `olo_renderer_settings_set renderpath=deferred`
3. Change the sun: `DirectionalLightComponent.Intensity` 2.6 → 20, or `CastShadows` on/off

**Observed:** the viewport frame is byte-identical across all of it (same MD5 per pose, `e0d15729e117` at the near pose — including across editor *processes*). The scene renders dim, IBL-only: no directional light contribution, no shadows, no skybox.

**Expected / control:** switch `renderpath=forward` in the same session and everything works — sun responds to intensity, CSM cascades render *and* show on the floor, skybox visible. The same deferred scene on **OpenGL** is fully correct.

## What is ruled out

- CSM cascade depth **is rendered** on Vulkan (`olo_render_capture_target` on `ShadowMapCSMCascade0` shows correct silhouettes).
- `DeferredLightingPass` appears in the live graph as executed (`culled: false` in `olo_render_frame_breakdown`), and its shader compiled (0 entries in `olo_shader_errors`).
- The lights UBO is fed (no `no published occupant` warning for it — the only null-block warnings are VSM's own bindings while VSM is off, which correctly read as zeros/disabled).
- Not the shadow term: `CastShadows=false` changes nothing either.

## The suspicious part

`olo_render_capture_target` on the G-Buffer under the deferred path returns **uniform buffers with no silhouettes** — `GBufferAlbedo` solid white, `GBufferNormal` solid black, `GBufferResolved` solid white — while the viewport clearly shows geometry. Either the G-Buffer genuinely never receives geometry on Vulkan (and the visible frame is stale output of an earlier lit pass being re-presented, which would explain the frozen hash), or the capture path resolves the wrong physical resource for transient G-Buffer targets under Vulkan (the `render-graph-transient-aliasing.md` archetype). Distinguishing those two is the first step.

[...]

PR #831, opened 2026-08-18
feat(renderer): virtual shadow pages for point and spot lights (#703)

### LINKED ISSUE #703: Renderer: virtual shadow pages for point and spot lights (VSM follow-up)

Local-light shadows are packed into one 4096^2 atlas by priority (`Renderer/Shadow/ShadowAtlas.h`): ranks 0-1 get atlas/4 tiles, ranks 2-5 atlas/8, the rest atlas/16, and a point light gets *half* tier resolution per cube face because six full-tier faces would exhaust the atlas area. Beyond ~16 lights or 32 entries, casters are simply starved â€” `AtlasCasterRecord::Allocated == false` exists precisely to record that. Tile size is also decided by a heuristic score, not by what the shading actually needs.

Extend the VSM page table (#702) to point and spot lights: a cube-compatible page-table image with mips, array layers of `6 * MAX_POINT_LIGHTS + MAX_SPOT_LIGHTS`, sharing the same physical memory block, allocator, caching and invalidation as the directional clip levels. Pages are allocated on demand from what is actually visible, so a light far from the camera costs the handful of pages its on-screen footprint needs rather than a fixed tile â€” and the starvation case disappears.

Growing `AtlasResolution` and `kMaxShadowedLights` â€” pushes the wall back at quadratic memory cost, still allocates by heuristic rather than need.

**Additional context**

Reference: `D:\repos\Timberdoodle\src\rendering\virtual_shadow_maps\cull_and_draw_point_pages.hlsl`, `gen_point_dirty_bit_hiz.hlsl`, and the point/spot resource setup in `vsm_state.hpp:282-299` (cube-compatible `R32_UINT` page table with `log2(res)+1` mips) plus the per-face camera construction in `vsm_state.hpp:50-200`. Apache-2.0.

**Acceptance criteria**

- More simultaneously-shadowed local lights than the current 16-light / 32-entry budget at equal VRAM, with no starved casters in a representative scene.
- Per-light shadow detail scales with screen footprint, not with an authored priority rank.

## Score
```olo-score
capability: 5
craft: 8
stability: 3
decay: 1
effort: 5
confidence: 0.5
learning: 8
fun: 8
kano: performance
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) Â· score = confidence Ã— (capability + craft + stability + decay) / effort, derived by the picker.</sub>

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

drsnuggles8:
Extends the #702 page table to point and spot lights. They read the **same** page table, physical pool, allocator and eviction policy as the directional clip levels — the sharing is the point of the issue, not an implementation detail.

A point light owns six consecutive **layers** (its cube faces, in the atlas path's `+X,-X,+Y,-Y,+Z,-Z` order), a spot owns one, and both come from one flat 256-deep pool. A layer is assigned per light identity, so a light keeps its cached pages while it stands still.

## Acceptance criteria, measured

**1. More simultaneously-shadowed local lights at equal VRAM, no starved casters.**

Measured live in the editor on the new `VirtualShadowMapLocalLights.olo` (22 casting local lights), via `olo_shadow_atlas_layout`:

| | candidates | allocated | starved |
|---|---|---|---|
| Atlas | 22 | 13 | **9** |
| VSM pages | 22 | **22** | **0** |

There is no score, no rank and no tile tier — a layer costs nothing until a page under it is on screen. `AtlasCasterRecord::Allocated` is still filled by the VSM path precisely so "was anything starved" stays answerable with the same question either way.

**2. Detail scales with screen footprint, not an authored rank.**

A layer is **mipped**; the mip is chosen per texel from the camera and light distances (`vsmLocalMipForDistances`), so a distant light resolves to mip 5 and spends one page on a whole cube face. `DetailScalesWithScreenFootprintNotWithAnAuthoredRank` pins it: two lights identical in every authored property, 30x apart in camera distance, resolve to different mips with the far one costing at least 16x fewer pages for a full face.

## Page counters

| | pool | resident | drawn (settled) | failed |
|---|---|---|---|---|
| 28 lights, headless evidence test | 4096 | 2697 | 0 | 0 |
| 1 light, isolation test | 4096 | 694 | 0 | 0 |

`drawn = 0` at rest is the local page cache working, the same way #820 reported it for the sun.

## The one new binding

`SSBO_VSM_LOCAL_LIGHTS = 78`, called out loudly per the binding-space warning. Everything else rode existing buffers: local pages appended to the page table, local pyramids to the HPB, local requests sharing the request ring behind a flag bit, local draws sharing the instance/command buffers at disjoint ranges. Only the ~40 KB of per-layer projections could not — past the GL 4.6 minimum UBO size (16 KB), so not expressible as a uniform block. `kHighestKnownSSBOBinding` moved 77 to 78, which is the guard that makes such an addition deliberate.

## Bugs fixed on the way

Each of these left the frame rendering and wrong:

* **`ShadowSettings::Enabled` did not disable VSM at all** — predates this work (#702). `ShadowRenderPass` early-outs but `VirtualShadowMapMarkPass` does not, so marking continued and the lit pass kept sampling. Found because a test needed a shadows-off control and did not get one. Now gated by `SetSuppressed`.
* **`AnyShadowsRequested()` had no third source** — a lamp-lit scene with no sun skipped the whole shadow pass. Unshadowed, nothing in the log.
* **`VSM_InvalidatePages` had no local half** — a mover dragged its shadow behind it under a lamp, which reads as an animation bug.
* **The local invalidation dispatched up to 262,144 work groups** against a guaranteed per-dimension maximum of 65,535. Now 2D. Invisible on NVIDIA (~2^31 per axis); on a conforming implementation at the minimum the dispatch is rejected and every local invalidation that frame is silently dropped.
* **`Terrain_PBR` (non-Forward+ loop) and `Terrain_Voxel`** still gated on `u_AtlasEntryCount`, which is zero under VSM local lights — both would have stopped receiving local shadows entirely.
* **`LocalDetailBias` was inverted** — the slider sharpened where three comments said it blurred. Invisible at the default, where both spellings agree.

## Deliberately not converted

`DDGI_Relight.glsl` and `compute/FroxelFogScatter.comp` stay on the atlas, so probe irradiance and light shafts lose local-light occlusion under VSM. Both are compute passes that never call `BindForSampling`; reading the page table there would rely on the buffers happening to still be bound from the shadow pass — true today, unverified, and exactly the ordering dependency that breaks silently. Noted in both shaders and in the agent-rules doc.

## Review guide

**Where I would look hardest**

1. `OloEditor/assets/shaders/compute/VSM_CullLocalCasters.comp:106-140` — the clip-space frustum test. A layer is a perspective, so the directional cull's "project eight corners and divide by w" is not transferable: a corner behind the near plane has `w <= 0` and the divide mirrors it through the origin. The symptom would be a caster vanishing *only when it gets close to the light*. The same trap and fix are in `VSM_InvalidatePages.comp`.
2. `OloEditor/assets/shaders/include/VirtualShadowLocalRasterStage.glsl` — the sub-rect scale and the second Vulkan y-flip. Flipping about `VSM_LOCAL_VIRTUAL_RESOLUTION` instead of the instance's raster resolution is correct at mip 0 and writes nothing at every coarser mip, on Vulkan only.
3. `VirtualShadowMap.cpp` `RenderCasters` run sizing — the directional runs are laid out first and the local ones take what is left, because dropping the local half costs an unshadowed lamp while dropping the directional half replaces CSM with nothing.

**What I verified, and how**

* 19 L1 contract tests (`VirtualShadowMapLocalTest`) — mip heuristic direction and monotonicity, meta encoding disjointness, page-index injectivity, cube-face parity against `ShadowMap::BuildPointLightFaceMatrices`, and the layer pool's stability / LRU / class-change behaviour.
* Vulkan shader-compile suite extended to the 3 new shaders and the second y-flip composition.
* 3 GPU evidence tests: single-light isolation (atlas vs pages vs shadows-off, frame-difference, rect-free), the starvation pair, and four azimuths around a point light each measured against its own no-shadow control.
* Live editor over MCP: the table above, plus zero `olo_shader_errors`, plus before/after captures and a close pass for cube-face seams (none found).
* Full suite 5989/6015 before the review fixes; the two failures I owned (`ShaderReflectionBinding`, `BindlessShaderPipeline`) are fixed and green. `AssetContentValidity` fails on 10 missing `Assets/cache/embedded/*.png` — a generated cache directory, registry untouched by this branch.

**Least confident about**

The `LocalDetailBias = 2.0` default. It is calibrated against `ShadowAtlas::TileSizeForRank`'s own "half tier resolution per cube face" precedent, and it is the difference between a 4096-page pool overflowing (2181 failed) and not (0) at 28 lights — but it is one number tuned against one scene, and a scene with many close, mutually-overlapping lamps will still pressure the pool and report `PagesFailed`. The degradation path is the documented one (fall back to coarser mips), not a cliff.

Also: the local raster bounds the *useful* area with the sub-rect trick but not the *rasterized* one, so a caster projecting outside its face still generates fragments that the fragment stage discards. Paid only on frames where the layer has dirty pages. Bounding it properly means one draw per (batch, mip); worth doing if a profile says so, and documented rather than done.

**Deliberately not tested**

Skinned casters into local layers have no dedicated test — the code path mirrors the directional one and compiles on both routes, but no frame in this PR contains an animated caster under a lamp. Vulkan is compile-verified only for the local path; per the page-cache doc the Vulkan editor's deferred path renders a frozen frame (#823), so a live Vulkan check would need the forward path and was not run.

Closes #703

## Summary by CodeRabbit

* **New Features**
  * Added virtual shadow map support for local point and spot lights.
  * Added renderer controls, usage statistics, and starvation warnings for local-light shadows.
  * Added a demonstration scene featuring multiple point and spot lights.
* **Bug Fixes**
  * Improved local-light shadow allocation, invalidation, culling, eviction, and rendering.
  * Ensured shadow-disabled mode suppresses virtual shadow map processing correctly.
* **Tests**
  * Added CPU, GPU, visual-output, and shader-compatibility coverage.
* **Documentation**
  * Expanded guidance for local-light shadow behavior, perspective culling, and page-cache handling.


drsnuggles8:
## 🤖 Self-review @ `5fe5aa3806fce6ade64b57f19ad91ffd74305db4`

Reviewed the full diff at **high** effort before the PR existed (Phase 3), then re-verified after fixes.

**Findings: 4 · all real, all fixed**

- `VSM_InvalidatePages` local dispatch was `invalidations × layers` = up to **262,144 work groups** against the guaranteed per-dimension maximum of **65,535**. Split into 2D. This box (NVIDIA, ~2³¹ per axis) could never have shown it; on a conforming implementation at the minimum the dispatch is rejected and every local invalidation that frame is silently dropped, reintroducing trailing shadows.
- `Terrain_Voxel.glsl` had no local-shadow path at all, so voxel terrain under a lamp would render fully unshadowed once `u_AtlasEntryCount` goes to zero. Wired.
- `Terrain_PBR.glsl` had only its Forward+ half converted; its MultiLightUBO loop (taken when Forward+ is off) still gated on `u_AtlasEntryCount`. Wired.
- `FroxelFogScatter.comp` shadows local lights only through the atlas — undocumented, unlike the equivalent DDGI limitation. Now documented in the shader and in the agent-rules doc, and deliberately left unconverted: it never calls `BindForSampling`, so reading the page table there would rely on an unverified binding-order assumption.

**Dismissed: none.**

**Separately caught by the suite, both mine, both fixed**

- `ShaderReflectionBinding` — `kHighestKnownSSBOBinding` pinned at 77; SSBO 78 is exactly the deliberate-addition guard. Bumped.
- `BindlessShaderPipeline` — the new local raster stage declares a slot-based storage image and needed the same allowlist entry its directional twin has.

**Not mine:** `AssetContentValidity` fails on 10 missing `Assets/cache/embedded/*.png` — a generated cache directory, `AssetRegistry.oar` untouched by this branch.

**Cleared on inspection:** `SSBO_VSM_LOCAL_LIGHTS = 78` has no collision (UBO 78 is a separate namespace); `VSM::Statistics` matches its GLSL twin member-for-member; `EndLocalLights` → `BeginFrame` preserves `Params4.y/z`; the `b_LocalRasterMip` reset/atomicMin/read ordering across setup → HPB build → raster is correct; the batch-cap halving keeps two-commands-per-batch inside `kMaxBatches`.

Given the diff size (~5.6k lines, 44 files) and that three of the six real bugs here were found by the GPU or by reading a PNG rather than by an assertion, `/code-review ultra` on this branch would be worth the spend before merge.

drsnuggles8:
@coderabbitai review

drsnuggles8:
## Update: master merged + self-review findings fixed

**Merged `origin/master` (#827).** That prunes 10 stale `Assets/cache/embedded/*.png` entries from `AssetRegistry.oar` (8152 → 7392 bytes) — the files are gitignored under `**/cache/**` and can never exist on a clean checkout, so the registry was referencing phantoms. `AssetContentValidity` is now **26/26**, including `SandboxAssetRegistryDeserialisesAndPathsResolve`, which was the only genuinely failing test in the full suite.

I had earlier called that failure environmental on the grounds that the cache directory is generated. The generated part was right; the conclusion was not — the defect was in the registry, and I should have checked whether master had moved before classifying it.

Merged rather than rebased, per `CLAUDE.md` → *"Resolve by merging master in, never rebase, never force-push"*. Both files the two sides touched (`CLAUDE.md`, `docs/agent-rules/README.md`) kept both sets of edits.

**Self-review findings, all nine verified against current code and fixed.** None were skipped.

*Correctness:*
- `RegisterLocalLight` rejects non-finite descriptions. A NaN `Range` slips through `<= 0.0f` and poisons the light **permanently**, not for a frame: projections go non-finite so every cached page is garbage, and the pose compare can never match again (`NaN != NaN`), so the slot re-invalidates itself forever.
- `GetLocalLightCount()` gated on `AreLocalLightsActive()` — the head list is only rebuilt while the system is on, and this is the third source `AnyShadowsRequested()` consults, so a stale non-zero held the shadow pass open for unregistered lights.
- `DestroyResources` clears the layer pool, matching `ResetPageState`.

*Dead code:* the duplicated `vsmLocalFootprintHasDirtyPage` call collapsed to one shared test.

*Tests that could not fail:* `EXPECT_EQ(0u, 0u & kRequestLocalBit)` (true of every bit → now asserts the flag/mip fields are disjoint, with `kRequestMipMask` added as the missing C++ mirror); `owned.insert(layer).second` (loop variable, unique by construction → now per-slot ownership counting); float equality on axis components plus a dead `? 0.0f : 0.0f` ternary; and a binding-limit assertion that named shader-storage bindings for what is the `GL_MAX_UNIFORM_BUFFER_BINDINGS` minimum, now using `VulkanBindingState::kMaxBufferBindings`.

*Stale comments:* the `EndFrame` dispatch claim, the Vulkan harness stage counts (nine/two → ten/four), and a scene description left over from before that fixture moved to the surround arrangement.

**Verified:** 54/54 across `VirtualShadowMap*`, `ShaderReflectionBinding`, `BindlessShaderPipeline`; `AssetContentValidity` 26/26; all touched shaders re-validated under `glslc` on both the GL and `OLO_VULKAN` routes.

Issue #629, opened 2026-07-11
Renderer: Nanite-style virtualized micro-poly geometry (cluster LOD DAG + compute software rasterizer)


drsnuggles8:
## Problem

OloEngine can **build** GPU meshlets on the CPU but cannot **render** virtualized geometry. `MeshOptimization::GenerateMeshlets()` produces `MeshletData` with per-meshlet bounding cones (`OloEngine/src/OloEngine/Renderer/MeshOptimization.{h,cpp}`), yet nothing consumes it — grep confirms it is referenced only by itself and `MeshOptimizationTest.cpp`. There are no mesh/task shaders, no `glDrawMeshTasks`, no cluster hierarchy, no streaming, no compute raster. Triangle throughput is therefore bounded by classic per-mesh discrete LOD (`Renderer/LOD.h`, distance-based mesh-swap with visible LOD-pop cracks) plus **instance**-level Hi-Z occlusion (`Passes/GPUDrivenOcclusionPass.h`, #431). Film-quality source meshes must be manually decimated; there is no seamless sub-pixel LOD.

This epic delivers a **Nanite-style virtualized micro-poly geometry pipeline** — chosen over GPU ray tracing because OpenGL 4.6 has **no hardware RT API** (RT would be a pure-compute reimplementation that fights the platform and duplicates already-shipped SSR/GTAO/contact shadows), whereas virtual geometry maps perfectly onto GL 4.6 DSA (compute, SSBOs, 64-bit atomics, `glMultiDrawElementsIndirectCount`, persistent mapping) and Nanite's real primary path is a **compute software rasterizer, not mesh shaders**.

## Why it matters

- **Seamless, crack-free LOD** at the cluster level replaces popping mesh-swap LOD — the single biggest visual-quality gap in the geometry pipeline.
- **Fixed triangle cost regardless of source density**: sub-pixel triangles rasterize faster in compute than through the hardware rasterizer (the core Nanite insight), so million-triangle imports become viable.
- **Builds on existing seeds**: the dead meshlet+cone builder, meshoptimizer simplification, and the two-phase HZB cull (extended from instance- to cluster-granularity) are all reused rather than reinvented.
- Flagship, multi-week, single-worktree scope — nothing intermediate renders, so it cannot be sliced into independently-mergeable chunks.

## Proposed approach

[...]

PR #837, opened 2026-08-18
fix(asset): cook embedded textures without a project so the mesh cache can persist them (#791)

### LINKED ISSUE #791: Mesh cache round-trip drops embedded albedo textures — DeccerCubes textured variants fail on a warm cache
## Problem

`AllVariants/DeccerCubesLoaderFixture.LoadsWithMeshesAndFiniteBounds` fails for the two textured
variants — `TexturedComplex` and `TexturedEmbedded` — **whenever the engine's repo-relative mesh
cache (`OloEditor/assets/cache/mesh/`) is warm**. With a cold cache all 7 variants pass.

```
DeccerCubesLoadingTest.cpp(143): error: Value of: sawAlbedoMap
  Actual: false
Expected: true
No material in TexturedComplex carries an albedo map.
```

So the **first** load of these `.gltf` files resolves their albedo textures correctly and writes an
`.omesh`; the **second** load reads that `.omesh` back and the materials come up with no albedo map.
The cached representation is not carrying (or not re-resolving) the imported material textures.

Note both failing variants are the ones whose textures are **embedded** in the glTF (base64 buffer /
embedded images). The test's own comment names the historical failure mode here — Assimp's `*N`
embedded-texture URI being fed to the filesystem texture loader instead of
`aiScene::GetEmbeddedTexture`. A cache round-trip that stores the URI rather than the resolved
texture would reproduce exactly this shape.

## Reproduction (deterministic, no parallelism needed)

```powershell
Remove-Item -Recurse -Force OloEditor\assets\cache\mesh   # cold
cd OloEditor
..\build\OloEngine\tests\Debug\OloEngine-Tests.exe --gtest_filter='AllVariants/DeccerCubesLoaderFixture.LoadsWithMeshesAndFiniteBounds/*'
#   -> [  PASSED  ] 7 tests.
..\build\OloEngine\tests\Debug\OloEngine-Tests.exe --gtest_filter='AllVariants/DeccerCubesLoaderFixture.LoadsWithMeshesAndFiniteBounds/*'
#   -> [  PASSED  ] 5 tests.  [  FAILED  ] 2 tests  (TexturedComplex, TexturedEmbedded)
```

Measured on `master` @ `d396ff03`, Windows / MSVC Debug:

| binary | mesh cache | result |
|---|---|---|
| `master` @ `d396ff03` | cold | **7/7 pass** |
| `master` @ `d396ff03` | warm | 5 pass, **2 fail** |
| `feature/test-isolation-races-789` | cold | **7/7 pass** |
| `feature/test-isolation-races-789` | warm | 5 pass, **2 fail** |

Identical on both, so it is pre-existing and unrelated to #789 — it surfaced there only because two
back-to-back full-suite runs left the cache warm for the second.

## Why CI is green

The GitHub runners start from a clean checkout with no `assets/cache/mesh/`, so every CI run is the
**cold** case. This only bites a developer (or an agent) running the suite twice on one machine —
where it looks like a flaky test, because the first run of the day passes.

## Related

`DeccerCubesLoaderFixture` is **not** in `OLO_MESH_CACHE_TESTS`
(`OloEngine/tests/CMakeLists.txt`), the `RESOURCE_LOCK` list for tests that share this cache. Adding
it there would serialize these cases against the other cache users, but it would **not** fix this:
the failure reproduces with a single process and no parallelism at all. The bug is in what the
cache stores or restores, not in concurrent access to it.

See also `docs/agent-rules/testing-architecture.md` §6.2.

---

Found while verifying #789 (test temp-dir isolation); filed rather than fixed there to keep that
PR's scope to temp paths.

## Score
```olo-score
capability: 3
craft: 3
stability: 8
decay: 1
effort: 2
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #791.

`AllVariants/DeccerCubesLoaderFixture.LoadsWithMeshesAndFiniteBounds` failed for
`TexturedComplex` and `TexturedEmbedded` whenever `OloEditor/assets/cache/mesh/` was warm.
The first run of the day passed, so it read as a flaky test — and **CI can never catch this
class**, because a runner starts from a clean checkout and is always cold.

## Root cause

`Model::CookEmbeddedTexture` returned early unless an active project *with an
`EditorAssetManager`* was mounted. Those two variants are exactly the ones whose images live
in glTF **bufferViews** rather than sidecar URIs:

| variant | `images[]` | warm result (before) |
|---|---|---|
| `Textured`, `MergedAtlas` | `uri: T_*.png` | fine |
| `TexturedComplex`, `TexturedEmbedded` | `bufferView: N` | **albedo dropped** |

Without the cook the texture stayed an in-memory decode with no path and no handle, so
`ImportedMaterialCodec::Describe` wrote an **empty `TextureRef`** into the `.omesh` —
indistinguishable from a slot that was legitimately never set. The warm load resolved a
material with no albedo map. Every step was individually correct; the information was
destroyed at the first one.

The cook needed a **directory to write to** but was gated on a **registry to register with**.
Only the first is required, because `RealizeTexture` resolves the recorded path *before* it
consults the handle. The two are now separated:

* directory ← `MeshCache::GetEmbeddedTextureCacheDirectory()`, which uses the same asset-root
  fallback as the `.omesh` itself (project asset dir, else CWD-relative `assets`), so a cooked
  texture always lands under the root that references it;
* registration ← best-effort. Losing it costs the packed runtime a handle; returning early
  additionally cost the `.omesh` its path, the strictly worse outcome.

With a project mounted the cooked path is byte-identical to before, so handle stability across
re-imports is unchanged.

## Also in this PR

**A writer-side guard.** `ImportedMaterialCodec::CanPersistEveryTexture()` + a check at the
cache-write site. `Describe()` cannot fail, so the *writer* must ask whether every texture that
is set describes to something resolvable; if not the material table is omitted and the next load
re-imports from source. A cache that is slow is a cost; a cache that is wrong on the second run
only is a bug that gets diagnosed as flakiness.

**`.omesh` v4 → v5**, no new sections — purely to invalidate existing files, same as the v3 bump.
A writer fix does not repair what is already on disk: every machine that had run the suite held
`.omesh` files with the empty references, still valid by the header check. That is the
"fixed on a clean machine, still broken on mine" shape. Costs one cold re-import.

**Ride-along: the packed metallic-roughness map had the same defect.** `Model` packs the
legacy/specular workflow's separate metallic + roughness files into one texture in memory, so it
had no path either and **every warm load of `backpack.obj` silently lost its MR map**. Now cooked
through the same extracted helper.

> Kept in the same commit rather than split out (contrary to the usual ride-along rule) because it
> shares the extracted `CookRgbaToAssetCache` helper — splitting would leave an intermediate commit
> that was never built. Named explicitly in the commit message. Happy to reconstruct the split
> history if preferred.

This moves **where the stbi row-flip happens**: `LoadSingleChannelImage` no longer flips, because
the cooked `.png` is flipped again by `Texture2D::Create(path)` on load. **This is the part of the
diff worth reviewing hardest** — verified quantitatively below.

## Verification

Every set run **twice** (cold then warm) — run 2 is a different experiment than run 1.

**The reproduction, before and after:**

| | run 1 (cold) | run 2 (warm) |
|---|---|---|
| baseline (fix reverted, test kept) | 7/7 pass | **5 pass, 2 FAILED** |
| with fix | 7/7 pass | **7/7 pass** |

Warm run with the fix: **7 loads from cache, 0 Assimp re-imports, 0 unresolved-texture warnings,
0 dropped material tables** — so it resolves through the cooked path, not via the new guard's
slow fallback.

**Full suite: 6038 passed / 0 failed, identical cold and warm** (6063 run, 25 skipped GPU/Steam).
The warm full-suite run is the one that matters here and had never been done before.

**Live editor over MCP** (`DeccerCubesTest.olo`, `ModelLoadingTest.olo`), cold vs warm:

| entity | cold vs warm frame |
|---|---|
| `DeccerCubes_Textured` | SHA256 **identical** |
| `DeccerCubes_Textured_Complex` | SHA256 **identical** |
| `DeccerCubes_Textured_Embedded` | SHA256 **identical** |
| `DeccerCubes_Merged_Atlas` | SHA256 **identical** |
| `Backpack` | SHA256 **identical** |

The backpack renders with metal reading as metal (canteen, axe head) and leather/canvas as rough —
an inverted MR map would swap those.

**MR orientation, checked against the source maps rather than trusting the render:**

| comparison | mean abs diff |
|---|---|
| cooked MR **G** vs `roughness.jpg` as-is | **0.004** |
| …vs vertically flipped | 54.703 |
| cooked MR **B** vs `specular.jpg` as-is | **0.004** |
| …vs vertically flipped | 62.717 |

R=0, A=255 per the packing convention. The cooked file is genuinely top-down; the single GL flip
happens at load, as for any image file. No double flip.

## Tests

* **`DeccerCubesLoaderFixture.SurvivesAWarmMeshCacheRoundTrip`** — invalidate, load cold, assert the
  `.omesh` was written, load warm, compare **GPU pixel readback** (a handle to the *wrong* texture
  would satisfy a bare "has an albedo map" check). Reproduces the failure **in one process**, so CI
  can finally see this class. Verified to fail on the unfixed engine and pass with it.
* **`AssetContentValidity` gained an `embedded` cache-kind entry.** It caught the new `mr_` file
  immediately — which also revealed the directory had **no entry at all**: the pre-existing `emb_`
  files never reached the repo tree before, because the old cook only ran with a project and wrote
  into the project's asset dir.
* Both DeccerCubes cases joined `OLO_MESH_CACHE_TESTS` (`RESOURCE_LOCK`) — the new case invalidates
  a shared repo-relative cache entry the other case reads. Note `OloEngine/tests/CMakeLists.txt` is
  a known collision point for parallel branches; the change there is two lines.

## Docs

`docs/agent-rules/cache-stored-unresolvable-reference.md` — the generalizable rule: a cache that
stores a **name** (handle, path, URI, cross-file index) must refuse to store one nothing can
resolve. Covers why CI is structurally blind to this, and the two things to copy when writing a
warm-path test for another cache. Linked from both indexes and `CLAUDE.md` as
`docs/agent-rules/README.md` requires. Distinct from the staleness species (#505/#530/#542), where
the artifact is *out of date*; here it was current and internally consistent, with a field pointing
at nothing.

## Note for reviewers

Loading a model with embedded textures in the editor adds cooked-texture entries to
`SandboxProject/AssetRegistry.oar`, which point at git-ignored `cache/` files. That churn is
pre-existing (#629) and deliberately **not** committed here — committing it would give a fresh
clone dangling entries. This PR adds the two `mr_*` entries to that same pre-existing churn.

## Summary by CodeRabbit

* **New Features**
  * Embedded and packed metallic-roughness textures are now cached as reusable PNG assets.
  * Texture processing works even when no project is mounted.

* **Bug Fixes**
  * Prevented cached models from silently losing unresolved textures.
  * Improved consistency between first-time imports and cached reloads.
  * Corrected texture orientation handling for single-channel images.

* **Compatibility**
  * Updated the mesh cache format; existing version 4 caches will be regenerated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -105,6 +105,7 @@ you are in.
 - [component-serializer-codegen.md](docs/agent-rules/component-serializer-codegen.md) — when a component round-trips for free, when to annotate a field, and when to hand-write *and* exclude.
 - [scene-binary-sidecar.md](docs/agent-rules/scene-binary-sidecar.md) — the `.scenebin` fast path: generated, hybrid-covered, and how it gets invalidated.
 - [binary-format-versioning.md](docs/agent-rules/binary-format-versioning.md) — versioning a fixed-order binary archive: gate each new field, don't trust the header check to exclude old data.
+- [cache-stored-unresolvable-reference.md](docs/agent-rules/cache-stored-unresolvable-reference.md) — a cache that stores a *name* (handle, path) must refuse to store one nothing can resolve; the failure only shows on the second load, so CI never sees it.
 - [scene-copy-must-carry-scene-level-settings.md](docs/agent-rules/scene-copy-must-carry-scene-level-settings.md) — `Scene::Copy()` drops scene-level settings structs the instant Play starts, invisibly to headless tests.
 - [floating-origin-rebase-subsystems.md](docs/agent-rules/floating-origin-rebase-subsystems.md) — the four subsystems holding world-space state outside the rebased set, each needing a different fix.
 - [asset-degradation-and-constructor-preconditions.md](docs/agent-rules/asset-degradation-and-constructor-preconditions.md) — a precondition asserted in a constructor delegates safety to ~30 call sites, and they will disagree.
--- a/OloEngine/src/OloEngine/Asset/MeshCache.cpp
+++ b/OloEngine/src/OloEngine/Asset/MeshCache.cpp
@@ -263,6 +263,23 @@ namespace OloEngine
                     OLO_CORE_INFO("MeshCache: Cleared animation cache");
                 }
             }
+
+            // The cooked-texture cache is derived data like the other two: every file in it is
+            // reproducible from the model that produced it. Leaving it behind would make "clear
+            // cache" silently partial — and the .omesh files that reference these textures by
+            // path are being deleted right above, so keeping them would strand them anyway.
+            if (auto embeddedDir = GetAssetRoot() / kEmbeddedTextureCacheSubdir; std::filesystem::exists(embeddedDir, ec))
+            {
+                std::filesystem::remove_all(embeddedDir, ec);
+                if (ec)
+                {
+                    OLO_CORE_ERROR("MeshCache: Failed to clear embedded-texture cache: {}", ec.message());
+                }
+                else
+                {
+                    OLO_CORE_INFO("MeshCache: Cleared embedded-texture cache");
+                }
+            }
         }
 
         void InvalidateCache(const std::filesystem::path& sourcePath, const std::string& prefix)
@@ -398,6 +415,7 @@ namespace OloEngine
 
             accumulateDirectorySize(GetAssetRoot() / kMeshCacheSubdir);
             accumulateDirectorySize(GetAssetRoot() / kAnimationCacheSubdir);
+            accumulateDirectorySize(GetAssetRoot() / kEmbeddedTextureCacheSubdir);
 
             return totalSize;
         }
--- a/OloEngine/src/OloEngine/Asset/MeshCache.cpp
+++ b/OloEngine/src/OloEngine/Asset/MeshCache.cpp
@@ -17,6 +17,7 @@ namespace OloEngine
     {
         constexpr const char* kMeshCacheSubdir = "cache/mesh";
         constexpr const char* kAnimationCacheSubdir = "cache/animation";
+        constexpr const char* kEmbeddedTextureCacheSubdir = "cache/embedded";
 
         // Returns the project asset directory if a project is active, or falls back
         // to the CWD-relative "assets/" for headless / test scenarios.
@@ -89,6 +90,13 @@ namespace OloEngine
         std::filesystem::path GetAnimationCacheDirectory()
         {
             auto dir = GetAssetRoot() / kAnimationCacheSubdir;
+            EnsureDirectoryExists(dir);
+            return dir;
+        }
+
+        std::filesystem::path GetEmbeddedTextureCacheDirectory()
+        {
+            auto dir = GetAssetRoot() / kEmbeddedTextureCacheSubdir;
             EnsureDirectoryExists(dir);
             return dir;
         }
--- a/OloEngine/src/OloEngine/Asset/MeshCache.h
+++ b/OloEngine/src/OloEngine/Asset/MeshCache.h
@@ -23,6 +23,18 @@ namespace OloEngine
         // Returns the animation cache directory path (assets/cache/animation/).
         std::filesystem::path GetAnimationCacheDirectory();
 
+        // Returns the embedded-texture cache directory path (assets/cache/embedded/),
+        // where Model cooks a glTF/FBX embedded bitmap into a real .png so it gains a
+        // path (and, when a project is mounted, a registry handle).
+        //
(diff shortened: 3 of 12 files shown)


drsnuggles8:
@coderabbitai review

Issue #811, opened 2026-08-16
Build: OLO_WITH_VULKAN=OFF is documented as an escape valve but is neither buildable-without-SDK nor tested


drsnuggles8:
Found during #691 Phase 9's OloServer decision (ADR 0011 amendment (86)).

## The claim vs the build

ADR 0011's Consequences and `rhi-abstraction-boundary.md` both describe
`OLO_WITH_VULKAN=OFF` as the escape valve that lets `OloServer` and a lean CI
"build without a Vulkan SDK present".

That is not what the option does. `OloEngine/CMakeLists.txt` calls
`find_package(Vulkan REQUIRED ...)` **ungated**, deliberately — the SDK is also
where the *shader* toolchain lives (shaderc, glslang, SPIRV-Tools,
SPIRV-Cross), which every build needs regardless of backend. So OFF removes the
volk/VMA/vulkan-headers ports and the backend object code, but the SDK is still
a hard configure-time requirement.

Amendment (86) corrected the ADR text to say so. This issue is about the two
things the correction leaves open.

## 1. No CI job builds OFF

Grepped every workflow: nothing sets `OLO_WITH_VULKAN`. Three test files carry
`GTEST_SKIP` branches for the OFF configuration that have therefore presumably
never executed. A compile break in the OFF path would reach master unnoticed —
and the `#if OLO_WITH_VULKAN` self-guards in ~40 Platform/Vulkan TUs are
exactly the kind of thing that rots without a build.

Cheapest fix: one configure+build job (no tests needed) on the existing Windows
or Linux runner.

## 2. It is a build-tree-wide option, so it cannot serve its stated purpose

`OloServer` is built from the same tree as the editor and runtime, so OFF
cannot be set "for the server" — it is all or nothing per configure. If a
genuinely lean server build is wanted (the deployment container is
ubuntu + libstdc++6), that needs either a separate configure or decoupling the
shader toolchain from the Vulkan SDK discovery.

Worth deciding explicitly whether that lean build is a goal at all. If it is
not, the honest move is to delete the escape-valve language rather than leave a
capability documented that nobody can use.

## Score
```olo-score
capability: 1
craft: 1
stability: 3
decay: 3
effort: 1
confidence: 0.8
learning: 1
fun: 1
kano: table-stakes
blocked_by: []
blocks: []
```
[...]

PR #840, opened 2026-08-19
fix(build): resolve the Steamworks SDK paths before any add_subdirectory (#828)

### LINKED ISSUE #828: Fresh configure with STEAMWORKS_SDK_ROOT set produces a test binary that cannot start (0xC0000135)

drsnuggles8:
Closes #828

A fresh configure on a machine with `STEAMWORKS_SDK_ROOT` set produced an `OloEngine-Tests.exe` that would not start. The build only failed at the very end, in gtest discovery, with exit `0xC0000135` (`STATUS_DLL_NOT_FOUND`) — an error naming nothing about Steam, which is what made it expensive. The missing DLL was `steam_api64.dll`.

## Root cause

Two independent facts combined.

**A CMake variable is only visible to a subdirectory added after it is set.** `OloEngine/CMakeLists.txt` processed `add_subdirectory(tests)` at line 49 but resolved `OLO_STEAM_RUNTIME_DLL` at line 335/369. So when `OloEngine/tests/CMakeLists.txt` called `olo_copy_steam_runtime(OloEngine-Tests)`, the variable did not exist yet.

**The guard read "unset" as "nothing to do."** Its `DEFINED OLO_STEAM_RUNTIME_DLL` term was meant to say *"a stub build has no DLL to stage"* — it also silently absorbed *"the path has not been computed yet"*, a completely different situation. The copy step was simply never generated.

Only a **fresh** configure was affected: the `CACHE INTERNAL` write from a previous run made the second configure pass, which is why this read as "works on my machine" and why the known local workaround was to configure twice. CI never caught it because the runners have no Steamworks SDK, so `OLO_WITH_STEAM` is OFF there and the branch is never entered.

## The fix — two parts

**1. Move the resolution above every `add_subdirectory()`.** The path resolution and validation moved out of `OloEngine/CMakeLists.txt` into a new `cmake/Steamworks.cmake`, included from the root `CMakeLists.txt` alongside the other configuration modules. The **target wiring** — include directories, the import library, the compile definitions — stays in `OloEngine/CMakeLists.txt`, because it needs the `OloEngine` target to exist.

I weighed this against simply reordering within `OloEngine/CMakeLists.txt` and chose the split deliberately: reordering fixes the symptom and leaves the trap armed, since any subdirectory added above the Steam block later reintroduces it. A block that only computes variables has no ordering requirement of its own, so it can be hoisted to the earliest point that makes all four consumers (`OloEngine/tests`, `OloEditor/src`, `OloRuntime/src`, `OloServer/src`) correct. A generator-expression approach was considered and rejected — the DLL is a plain file path, not a target, so making it an imported target would still need that target defined before `tests/`, which is the same ordering problem wearing a different hat.

**2. Make the silent case loud.** `olo_copy_steam_runtime` now `message(FATAL_ERROR ...)`s when Steam is on with the real SDK and the path is unset, naming Steam, `0xC0000135`, and the ordering requirement. If someone re-arms this, they get a sentence at configure instead of a status code at test discovery. It is safe in CI, which never enters that branch.

## Verification — from a clean configure

**A rebuild in an already-configured tree proves nothing here**, because the cache papers over the bug. Everything below is from a deleted or separate build tree.

The cheap check needs no build at all — grep the generated project for the staging step:

| target | before fix | after fix |
|---|---|---|
| `OloEngine-Tests.vcxproj` | **0** refs | 6 refs |
| `OloEditor.vcxproj` | 6 refs | 6 refs |
| `OloRuntime.vcxproj` | 6 refs | 6 refs |
| `OloServer.vcxproj` | 6 refs | 6 refs |

All four call sites of the shared function were checked, not just `OloEngine-Tests`.

Three configures, each from a deleted or separate tree:

| configure | result |
|---|---|
| `msvc`, Steam ON, `build/` **deleted first** | all four targets stage the DLL |
| `-DOLO_WITH_STEAM=OFF` | no staging, no warnings, no fatal |
| `-DOLO_WITH_STEAM_STUB_SDK=ON` (the CI path) | stub message, no staging, no fatal |

Then the end-to-end check, built from that clean-configured tree through the build mutex:

- `cmake --build build --target OloEngine-Tests --config Debug --parallel 6` — exit 0, log shows `Staging steam_api64.dll next to OloEngine-Tests`
- `steam_api64.dll` present next to `OloEngine-Tests.exe`
- `dumpbin /DEPENDENTS` → **30** dependents including `steam_api64.dll`, matching the figure in the issue (a Steam-OFF build has 29, without it)
- the exe **starts**: `--gtest_list_tests` exits 0 and enumerates 7183 lines. That is the same operation that previously died with `0xC0000135`, since gtest discovery works by running the executable.

## ⚠ Coordination with #811

Per the handover, `feature/vulkan-off-build-coverage-811` is editing `OloEngine/CMakeLists.txt` concurrently. **Lines this PR touches in that file: the Steamworks block at 277-372, which becomes 277-303** (target wiring only — the resolution half moved out). Nothing else in that file is touched, and the Vulkan region is untouched, so whichever lands second should rebase cleanly.

Root `CMakeLists.txt` gains exactly one `include(cmake/Steamworks.cmake)` plus its comment, in the existing includes block.

## Docs

New `docs/agent-rules/configure-time-variable-visibility.md`, since the shape generalises past Steam: a guard whose single condition cannot distinguish *"this does not apply here"* from *"this has not been computed yet"*. Linked from both indexes (`docs/agent-rules/README.md` under "it silently drops something" and "ordering and lifetime", plus the Build & deps roster; and `CLAUDE.md`'s companion guides). Cross-linked from `build-trees-and-windows-asan.md`, which documents the **other** cause of `0xC0000135` in this repo (a missing ASan runtime on PATH) — a reader who hits that code can now tell the two apart with `dumpbin /DEPENDENTS`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Steamworks setup during configuration to ensure required runtime files are correctly found and staged.
  * Configuration now reports clear errors for unsupported platforms, missing SDK settings, or incomplete SDK installations.
  * Prevented builds from silently omitting required Steamworks runtime files.

* **Documentation**
  * Added troubleshooting guidance for missing Steamworks runtime DLLs and related Windows startup errors.
  * Documented configuration requirements and verification steps for fresh builds.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -76,6 +76,7 @@ you are in.
 - [build-trees-and-windows-asan.md](docs/agent-rules/build-trees-and-windows-asan.md) — never build the msvc and clangcl trees concurrently; plus per-user `mspdbsrv` stalls, the codegen build-graph wiring, the local ASan recipe, and (§5e) why both memory job pools vanish on Linux CI so a bare `--parallel` is a runaway there and nowhere else.
 - [static-archive-4gib-ceiling.md](docs/agent-rules/static-archive-4gib-ceiling.md) — a .lib cannot exceed 4 GiB, `LNK1248` under-reports the overshoot, and no CI job had ever built a `/GL` object.
 - [vcpkg-dependency-management.md](docs/agent-rules/vcpkg-dependency-management.md) — read before adding, bumping or removing a dep: the CRT triplet mismatch is heap corruption, and three of the five traps are silent.
+- [configure-time-variable-visibility.md](docs/agent-rules/configure-time-variable-visibility.md) — a CMake variable resolved *below* the `add_subdirectory()` that consumes it, behind a guard that reads "unset" as "nothing to do": correct on every configure but the first.
 - [asset-import-usd-alembic.md](docs/agent-rules/asset-import-usd-alembic.md) — the importer/exporter registry seam, and vendoring OpenUSD / Alembic / MaterialX into a static-everything build.
 
 **Renderer**
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,13 @@ endif()
 # Includes
 include(cmake/SetupConfigurations.cmake)
 include(cmake/CommonProperties.cmake)
+# Resolves OLO_STEAM_{SDK_ROOT,INCLUDE_DIR,IMPORT_LIB,RUNTIME_DLL} from the two options above.
+# MUST stay ABOVE every add_subdirectory() below — those variables are read from four different
+# subdirectories (OloEngine/tests, OloEditor/src, OloRuntime/src, OloServer/src) and a variable
+# set after a subdirectory is added is invisible inside it. That ordering trap is issue #828; the
+# file's header comment carries the full story. It touches no targets, so it has no ordering
+# requirement of its own and can sit here with the other pure-configuration includes.
+include(cmake/Steamworks.cmake)
 include(cmake/Sanitizers.cmake)
 # Must come AFTER SetupConfigurations.cmake (it overrides that file's MSVC
 # debug-info default) and BEFORE the target subdirectories below (it sets the
--- a/OloEngine/CMakeLists.txt
+++ b/OloEngine/CMakeLists.txt
@@ -279,74 +279,12 @@ endif()
 # macro, compiling to empty TUs when OFF — the same convention as OLO_WITH_VULKAN above and the
 # OLO_WITH_* import formats.
 #
-# The API surface comes from exactly one of two mutually exclusive places:
-#   * the REAL SDK, resolved from $ENV{STEAMWORKS_SDK_ROOT}. Never vendored: the licence forbids
-#     redistribution and this repo is public, so developers install it themselves (it is free —
-#     see docs/ops/build.md). This mirrors how the Vulkan SDK is found above, NOT how Mono is
-#     vendored in-tree at :432 — Mono may live in-tree only because it is redistributable.
-#   * our hand-written STUBS (OLO_WITH_STEAM_STUB_SDK), which let CI compile, link and RUN the
-#     ON path that the real SDK can never reach. See the root CMakeLists for why that matters.
+# Only the TARGET WIRING lives here. Resolving the SDK paths themselves
+# (OLO_STEAM_{SDK_ROOT,INCLUDE_DIR,IMPORT_LIB,RUNTIME_DLL}) moved to cmake/Steamworks.cmake,
+# included from the root CMakeLists ABOVE every add_subdirectory() — because those variables are
+# consumed by subdirectories, and this file adds tests/ near its top, long before this line.
+# See issue #828 and that file's header comment; do not move the resolution back down here.
 if(OLO_WITH_STEAM AND NOT OLO_WITH_STEAM_STUB_SDK)
-	# Both halves matter. WIN32 is TRUE for 32-bit Windows too, and everything below hard-codes
-	# the x64 redistributables (steam_api64.lib / steam_api64.dll) — so a 32-bit configure would
-	# sail past a WIN32-only guard and fail much later at link, with an error that says nothing
-	# about pointer size. Reject it here, where the message can.
-	if(NOT WIN32 OR NOT CMAKE_SIZEOF_VOID_P EQUAL 8)
-		message(FATAL_ERROR
-			"OLO_WITH_STEAM=ON is only wired for 64-bit Windows, the stated target for issue #644 "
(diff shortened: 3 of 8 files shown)


drsnuggles8:
@coderabbitai review

Issue #841, opened 2026-08-19
OpenGL: 115 shader programs survive Renderer::Shutdown(); Vulkan releases the same set cleanly


drsnuggles8:
Found while verifying #839. Not part of that issue's families — those were scene-owned
meshes/textures and are fixed — this is a separate, GL-only population.

## Symptom

`OloEditor.exe --rhi=opengl`, open the start scene, select an entity, close the window.
`RendererMemoryTracker::Shutdown()` (the new teardown survivor report added in #839) says:

```
[Teardown] 115 GPU allocation(s) (3218124 bytes) survived the renderer — each is a
           process static, or a Ref that outlived Renderer::Shutdown()
[Teardown]   surviving Shader 'FullscreenBlit' (3284 bytes)   created at OpenGLShader.cpp:1622
[Teardown]   surviving Shader '__WarmupBoot' (18016 bytes)    created at OpenGLShader.cpp:1622
[Teardown]   surviving Shader '__Fallback' (5648 bytes)       created at OpenGLShader.cpp:1622
[Teardown]   surviving Shader 'Renderer2D_Quad' (12952 bytes) created at OpenGLShader.cpp:1622
[Teardown]   surviving Shader 'PBR_MultiLight' (272948 bytes) created at OpenGLShader.cpp:1622
   ... 110 more, ALL of ResourceType::Shader
```

**Every one of the 115 is a shader. Zero textures, zero buffers, zero vertex arrays** — so
this is specifically shader-program lifetime, not the asset-graph leak #839 dealt with.

## Why it is real, not a reporting artefact

The tracker's alloc/dealloc calls are properly paired for this class: `OpenGLShader`'s
constructor path tracks at `OpenGLShader.cpp:1619` and `~OpenGLShader` calls
`OLO_TRACK_DEALLOC(this)` (guarded by `m_TrackedAllocation`, which is exactly the async
"never finalized" case). So a shader still tracked when `RendererMemoryTracker::Shutdown()`
runs — the last statement of `Renderer::Shutdown()` — means the `OpenGLShader` **object has
not been destroyed yet**, with `ShaderLibrary::Clear()`, `ShutdownFallbackShader()`,
`ShaderWarmup::Shutdown()` and every `s_Data.*Shader.Reset()` already behind it.

The names make that concrete: `__WarmupBoot` and `__Fallback` are released by
`ShaderWarmup::Shutdown()` / `ShaderLibrary::ShutdownFallbackShader()`, and the
[...]

Issue #814, opened 2026-08-16
Vulkan: the fullscreen-triangle VAO survives teardown in any session that never initialises Renderer3D


drsnuggles8:
Found by running `OloRuntime --rhi=vulkan` for the first time (the runtime's
Vulkan path was gated off until #801 un-gated it). It violates a Phase 8 exit
criterion — "clean shutdown: zero VMA leaks and zero surviving device objects" —
and the in-tree forensics #794 built are what caught it, with a full creation
stack.

## Symptom

```
[Vulkan] surviving VertexArray (1 stream(s)) created at:
  MeshPrimitives::GetFullscreenTriangle
  ShaderWarmup::RenderProgressFrame
  Renderer2D::Init
[Vulkan] 1 shader/vertex-array object(s) survived full teardown
[Vulkan] 2 VMA allocation(s) still alive at allocator teardown (80 bytes)
```

The VMA dump names them: `vmaCreateBuffer (VulkanVertexBuffer)` 64 B and
`vmaCreateBuffer (VulkanIndexBuffer)` 16 B.

## Root cause — a cross-subsystem ownership split

`MeshPrimitives::GetFullscreenTriangle()` caches a static
`s_FullscreenTriangleVA`. Its only release site is `MeshPrimitives::Shutdown()`,
and that has exactly one caller: `Renderer3DLifecycle.cpp:692`, inside
**`Renderer3D::Shutdown`**.

But the triangle is *created* from the **2D / warmup** path —
`Renderer2D::Init` → `ShaderWarmup::RenderProgressFrame` →
`GetFullscreenTriangle`. So any session that runs the warmup screen without
initialising Renderer3D never releases it. The runtime run above confirms it:
`"Shutting down Renderer3D"` appears **0 times** in its log (it exits before
3D init when no scene is found), while the test suite's log has it once — which
is why the suite never saw this.

## Why it stayed invisible

- GL has no allocator-teardown assertion, so the same leak is silent there.
- The Vulkan runtime path had never been run by anyone (Phase-4 gated until
  #801), and the editor always initialises Renderer3D, so the editor's teardown
  is clean.

## Fix options

1. Release it where it is created — give `ShaderWarmup`/`Renderer2D` teardown
   the reset, or
2. Call `MeshPrimitives::Shutdown()` from a subsystem-neutral place (the
   `Renderer::Shutdown` that owns both 2D and 3D), which is probably the real
[...]

PR #845, opened 2026-08-19
fix(renderer): sweep the lazy-static release class — eight more members (#839)

### LINKED ISSUE #839: Vulkan teardown: 133 MB and 102 allocations survive shutdown — the lazy-static release class #814 only fixed one member of
Found by running `OloEditor.exe --rhi=vulkan`, loading a sandbox scene and closing the window. The Phase 8 forensics (#794) report it; **OpenGL leaks the same objects silently**, so this is not a Vulkan regression — Vulkan is just the only backend with an allocator-teardown assertion.

## Symptom

```
[Vulkan] 64 shader/vertex-array object(s) survived full teardown — see the lines above for owners
[Vulkan] 3 texture/storage-buffer object(s) survived full teardown
[Vulkan] 102 VMA allocation(s) still alive at allocator teardown (139196096 bytes) — dumping detailed stats
```

**~133 MB**, in three families. Every survivor carries a creation stack.

## Family 1 — 64 vertex arrays: every mesh the scene loaded

Two per mesh, both inside `MeshSource::Build`:

- `OloEngine/src/OloEngine/Renderer/MeshSource.cpp:122` — the mesh vertex array
- `OloEngine/src/OloEngine/Renderer/MeshSource.cpp:190` — `BuildShadowIndexBuffer`

Every one of the 64 stacks runs through scene deserialization:

```
MeshSource::BuildShadowIndexBuffer   (MeshSource.cpp:190)
MeshSource::Build                    (MeshSource.cpp:143)
MeshPrimitives::CreateSphere/CreateCube
CreateMeshFromPrimitive              (SceneSerializer.cpp:205/207)
DeserializeEntityComponents          (SceneSerializer.cpp:1977)
SceneSerializer::DeserializeEntity   (SceneSerializer.cpp:6100)
SceneSerializer::Deserialize         (SceneSerializer.cpp:6406)
EditorLayer::LoadEditorSceneFile     (EditorLayer.cpp:3583)
EditorLayer::OpenScene               (EditorLayer.cpp:3570)
```

So ~32 meshes built by `File > Open Scene`, none released before the device goes away. This is the family that needs an actual ownership decision rather than a one-line move — see "Scope" below.

## Family 2 — 2 font atlas textures: `Font::GetDefault()` has no release site at all

```
SlugFontProcessor::Process           (SlugFontProcessor.cpp:607 and :618)
Font::LoadFromMemory                 (Font.cpp:223)
Font::LoadFromFile                   (Font.cpp:82)
Font::Font                           (Font.cpp:22)
Font::Create                         (Font.cpp:307)
Font::GetDefault                     (Font.cpp:270)
UITextComponent::UITextComponent     (Components.h:3612)
entt::uninitialized_construct_using_allocator<UITextComponent>
```

Note where it is constructed from: **an ECS component constructor**. Adding a `UITextComponent` to an entity lazily builds the default font atlas. Grepping `Font.cpp` for a release site finds only the getter at line 246 — there is no reset, no shutdown hook, nothing. It is a lazy static that is never freed in any session, on any backend.

## Family 3 — 1 BRDF LUT

```
EnvironmentMap::GenerateBRDFLutWithConfig  (EnvironmentMap.cpp:309)
EnvironmentMap::GenerateIBLWithConfig      (EnvironmentMap.cpp:216)
EnvironmentMap::GenerateIBLTextures        (EnvironmentMap.cpp:131)
EnvironmentMap::CreateFromCubemap          (EnvironmentMap.cpp:96)
Scene::LoadAndRenderSkybox                 (Scene.cpp:6622)
Scene::RenderScene3D                       (Scene.cpp:9773)
Scene::OnUpdateEditor                      (Scene.cpp:4618)
```

Created on the first editor frame that renders a skybox.

## This is a known class whose sweep never happened

All three are the shape written up in [`docs/agent-rules/lazy-static-release-ownership.md`](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/lazy-static-release-ownership.md) — *"Release a shared lazy static from the teardown that always runs, not from the one that happened to draw with it."*

That document was produced by **#814 / PR #832** (the fullscreen-triangle VAO released only from `Renderer3D::Shutdown()`), which fixed exactly one member of the class. Nobody then swept for the others. It also already establishes the fix and its safety:

- release belongs in `Renderer::Shutdown()`, which owns both sub-renderers and runs whether or not 3D ever came up;
- moving a release *later* in `Renderer::Shutdown` is safe on both backends, because `VulkanContext::Shutdown()` runs a second final `FlushAll()` immediately before `VulkanDevice::Shutdown()`.

## Secondary symptom — shutdown looks like a hang

The log is silent from `12:43:25` (pipeline cache saved) to `12:45:46` (final VMA dump): **over two minutes**. That is the leak reporter symbolising 67 stacks of ~12 frames each through dbghelp. It only runs *because* there are leaks, so fixing the leaks removes the stall — but until then, closing the Vulkan editor looks like it hung, which is its own bad signal and easy to misfile as a deadlock.

## Scope

1. **Sweep the class, do not just fix three instances.** The value here is finding every lazy/cached GPU resource whose release site is conditional or missing. Two of the three below are mechanical once found; the search is the work.
2. **`Font::GetDefault()`** — give it a release and call it from `Renderer::Shutdown()`. Simplest member.
3. **`EnvironmentMap` IBL / BRDF LUT** — same treatment.
4. **Scene-deserialised meshes** — the genuine design question: who releases meshes built by `CreateMeshFromPrimitive`, and does that happen before device teardown? This may be a teardown-*ordering* problem (asset/scene layer outliving the RHI device) rather than a missing release, in which case the fix is ordering, not a new shutdown hook. Decide and state which.
5. **A guard so it cannot silently regress.** The reason this survived is that GL says nothing and no test asserts on teardown. A headless test that brings the renderer up, loads a scene, tears down and asserts zero survivors would catch the whole class — and would have caught #814 too.

## Acceptance criteria

- `OloEditor --rhi=vulkan`, open a sandbox scene, close: **zero** surviving shader/vertex-array objects, zero surviving texture/storage-buffer objects, zero live VMA allocations at allocator teardown.
- Shutdown no longer stalls (follows from the above — the reporter stops running).
- A test guards the invariant so the next lazy static is caught automatically.
- Verified on **both** backends: GL must not regress, even though it reports nothing.

## Score
```olo-score
capability: 1
craft: 1
stability: 5
decay: 3
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

Scoring notes: **Stability 5** — no data loss, but 133 MB and 102 allocations surviving teardown is a real correctness failure in a core path, and it currently masquerades as a two-minute hang. **Decay 3** — genuinely compounding: every new lazily-created GPU cache joins the class, and the class already grew silently past the one instance #814 fixed. **Effort 3** — two members are mechanical, the mesh-ownership question and the regression guard are not. **Confidence 0.8** — the playbook exists and two fixes are known-shape; the scene-mesh ownership is the unknown.

Related: #814 / PR #832 (the first member of this class), #794 (the forensics that report it), #823 (a separate open Vulkan editor bug).

drsnuggles8:
Closes #839.

#814 fixed one member of the lazy-static release class and
[the write-up it produced](docs/agent-rules/lazy-static-release-ownership.md)
predicted there were others. There were **eight**, and the useful finding is that
they were not all the same shape. "The release site is in the wrong teardown" —
the shape #814 documented — turned out to be the *rarest* of three species.

| species | what it looks like | members found here |
|---|---|---|
| **(a)** release in a **conditional** teardown | a release call exists and runs, just not in every session that can create the object | the fullscreen triangle (#814) — and nothing else |
| **(b)** **no release site at all** | nobody ever wrote one; the object is immortal | the default font's two Slug atlas textures, the GL cubemap staging PBO, `SceneHierarchyPanel`'s `Ref<Scene>` **and its per-component undo-snapshot maps**, `Scene`'s cached default `Material` |
| **(c)** a release site **nothing calls** | someone wrote the teardown — one with a comment saying when to call it — and never wired it up | `VideoSystem::Shutdown()`, `GPUTimerQueryPool::Shutdown()` |

Species (c) is the one worth internalising. Both read as *finished* code: a public
`Shutdown()`, a correct body, and in `VideoSystem`'s case the comment *"Release the
global fullscreen player. Call on engine/scene shutdown."* Grepping "is this
released?" finds the function and stops. The question that finds the bug is
**"who calls it?"**, and it has to be asked separately.

## The big one was not a teardown-ordering bug

The issue's headline — 64 vertex arrays, two per mesh, plus the BRDF LUT — is not a
misplaced release. Those `MeshSource`s are owned by the deserialised scene through
`MeshComponent::m_MeshSource`, the BRDF LUT through `EnvironmentMapComponent`, the
albedo maps through the model components. They survived because **the scene
survived** — `SceneHierarchyPanel.cpp` kept a file-scope `Ref<Scene>` that
`DrawComponents()` set on every inspector frame and nothing ever cleared.

**The A/B that proves it.** Same binary, same scene, `--rhi=vulkan`, close the window;
the only variable is whether an entity is selected in the inspector:

| | vertex arrays | textures | VMA allocations | bytes | exit |
|---|---|---|---|---|---|
| no entity selected | 0 | 0 | 0 | 0 | 0 — *"Context shut down cleanly"* |
| one entity selected | 12 | 6 | 30 | 147,330,896 | **3 — abort** |

Selecting an entity is the entire difference: it is what makes the Properties panel
call `DrawComponents()` at all. Two consequences the issue did not have:

* the failure is **worse than a leak** — closing the Vulkan editor *aborts*
  (`vmaDestroyAllocator`), it does not merely leak and exit;
* it is **conditional on a UI interaction**, so "I couldn't reproduce it" may only
  mean you did not click the same thing. The no-selection run looks perfectly clean.

This also settles the ownership question the issue asked to decide explicitly, and
the answer is **not** teardown ordering. `~Application` already detaches the layers
and calls `Project::Unload()` *before* `Renderer::Shutdown()`, deliberately and with
a comment saying why (#691 Phase 8). The order was right; the object had one more
owner than the order accounted for. Reordering teardown to "fix" a stray strong
reference would have converted a leak into a use-after-free — the failure direction
the issue warned about.

## Where each release went

Each at the narrowest teardown that is unconditional for every session that can
create the object, and that still runs while the graphics context is alive:

| release site | member |
|---|---|
| `Renderer::Shutdown()` | `Font::ShutdownDefault()` |
| `OpenGLRendererAPI::ShutdownGpuResources()` | `OpenGLTextureCubemap::ShutdownSharedResources()` |
| `Renderer3D::Shutdown()` | `GPUTimerQueryPool` (3D-only by construction) |
| `~Application` (both the destructor and the exception path) | `VideoSystem::Shutdown()`, `Scene::ReleaseSharedRenderDefaults()` |
| the end of the call | `SceneHierarchyPanel`'s file-scope state, via a scope guard rather than a trailing assignment |
| `SceneHierarchyPanel::SetContext` + its new destructor | `DrawComponent<T>`'s per-instantiation undo-snapshot maps |
| deleted outright | `McpToolsRender`'s duplicate static default `Material` |

### One of these has a trap worth reading

The GL staging PBO is a persistent `GL_PIXEL_UNPACK_BUFFER` that `glNamedBufferData`
grows to the largest cubemap face the session ever uploaded, and that nothing ever
deleted. It is **GL-only**, so the Vulkan forensics that found everything else were
blind to it — *a leak class found on Vulkan still needs a GL-side pass*.

Its release does **not** go in the `if (GetAPI() == OpenGL)` branch already sitting in
`Renderer::Shutdown()` next to `OpenGLFramebuffer::ShutdownSharedResources()`, which is
where it obviously belongs. `OpenGLTextureCubemap.h` includes `<glad/gl.h>` (that
framebuffer header does not), so calling it from `Renderer.cpp` would pull the GL loader
into an engine-core translation unit and undo the `sweep_glad_includes == 0` property
`RHIBoundaryRatchetTest` pins — **and the ratchet would not have caught it**, because it
counts *direct* `#include` lines naming `glad/gl.h`. An indirect one through a backend
header breaks the property with the counter still reading zero.

## Guard

`RendererShutdown.EveryProcessStaticGpuResourceHasAReleaseSite`. The two existing tests
in that file both start from a teardown *body*, so they only ever see facilities somebody
already thought to release — structurally blind to species (b). This one starts from the
**declaration**: every process static holding a GPU-owning `Ref` (seeded roster) and every
lazily created raw GL handle must be reset somewhere in its own translation unit.

Red-checked: **4 findings against the unfixed tree, 0 against the fixed one.** Neither a
declaration's own `= nullptr` initializer nor an out-of-class static member *definition*
(`Ref<Shader> ShaderLibrary::s_Fallback = nullptr;`) counts as a release — both textually
contain `= nullptr`, and without excluding them the check would have been silently vacuous
for every class static in the engine.

It deliberately does **not** cover species (c). A prototype that tried to resolve callers
textually through singleton accessors, aliases and delegating bodies was too noisy to keep,
and a curated roster of "teardowns that must be called" drifts by construction. That is
said out loud in the test and the doc rather than left as an implied guarantee.

Species (c) is caught empirically instead — and **GL now has an instrument for it**.
`RendererMemoryTracker::Shutdown()` used to `clear()` its live-allocation map without
looking at it, discarding the type, size, name, file and line of every survivor at exactly
the moment it had them. It now reports first. (`Initialize()` had no callers either, which
left the `m_IsShutdown` latch *its own comment exists to clear* permanently set after the
first `Renderer::Shutdown()` in a process; `Renderer::Init()` now calls it.) Limit stated
in the code: `TrackAllocation` is called from the OpenGL resource classes only, so on
Vulkan the report is empty and the #794 forensics remain the instrument there.

### The eighth member, and the guard's blind spot

With the other seven fixed, Vulkan still reported **1 surviving vertex array and 2
allocations (18 KB)** — a `Ref<Model>` held by `DrawComponent<T>`'s
`static std::unordered_map<u64, EditState>` of undo snapshots, whose `EditState::snapshot`
is a **copy of the component**. Drawing an entity in the inspector once pinned its GPU
resources for the life of the process. Each instantiation now registers a clearer (a
function-local static of a template is otherwise unreachable from outside), invoked from
`SetContext` — which also drops stale cross-scene undo state — and from a new panel
destructor that runs while the renderer is still alive.

**The guard cannot catch this one, by construction.** It looks for
`static Ref<GpuOwningType> name;`; here the Ref is a member of a struct that is the
`mapped_type` of a static map, three type layers down. Deciding whether an arbitrary
static's type *transitively* owns GPU memory is a type-system question, not a text one.
The test and the doc both say so rather than implying coverage they do not have.

## Verification

`OloEditor.exe --rhi=vulkan`, start scene, **entity selected** (the reproducing condition),
window closed with WM_CLOSE so the real `~Application` → `Renderer::Shutdown` →
`~VulkanContext` path runs:

| | vertex arrays | textures | VMA allocations | bytes | shutdown |
|---|---|---|---|---|---|
| baseline | 12 | 6 | 30 | 147,330,896 | **abort** |
| after the first seven members | 1 | 0 | 2 | 18,208 | **abort** |
| **this PR** | **0** | **0** | **0** | **0** | `[Vulkan] Context shut down cleanly`, exit 0 |

Run under `cdb` with `sxe av` armed: **no access violation**, and the run reaches the #794
forensics, which print nothing. That last point matters — see below.

**GL does not regress.** Same scene and interaction on `--rhi=opengl`: exit 0, and the new
`RendererMemoryTracker` report shows **zero** surviving textures, buffers or vertex arrays.

**Test suite:** `RendererShutdown.*` green (3/3), including the new guard; full suite result
below.

### Two things that cost time and are worth knowing

**The VMA leak assert aborts, and an abort discards spdlog's buffered tail.** So
`OloEngine.log` stopped several lines *before* the forensics that name the survivors, and an
intermediate run looked like a possible fresh crash in `~VulkanContext`. It was not: `cdb`
showed the full report on the inherited console and proved there was no AV. A nonzero exit
with a log that stops early is the abort — read the console, not the file. Had I trusted the
log, the eighth member would have been missed and this reported as fixed at "1 surviving
vertex array".

**A dirty scene blocks WM_CLOSE on a native MessageBox** (`OnWindowClose` →
`ConfirmDiscardChanges` → `MessagePrompt::YesNoCancel`), which no headless session can click
— a 12-minute "hang" until diagnosed. The engine already ships `OLO_EDITOR_UNSAVED_PROMPT`
for it, the same way `OLO_EDITOR_AUTOSAVE_RECOVERY` covers the other modal; the repro
harness just was not setting it.

## Found while verifying, filed separately

The new GL-side report immediately paid for itself: the first OpenGL session showed **115
shader programs surviving `Renderer::Shutdown()`** (3.2 MB) — and nothing else, no textures
or buffers. `OpenGLShader` pairs its alloc/dealloc tracking correctly and the survivors
include `__WarmupBoot`, `__Fallback` and the `Renderer2D_*` set whose teardowns had all
already run, while **Vulkan releases the identical set cleanly**. That is a backend
divergence in shader lifetime rather than one of this issue's scene-owned families, so it is
**#841** rather than more scope here.

The report was deliberately made terse before shipping — one ERROR summary with a per-type
breakdown plus six WARN examples — because two dozen ERROR lines on every editor exit, and at
the end of every test run (`OloEngineTest` calls `Renderer::Shutdown()`), is noise nobody
reads.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of graphics resources during normal shutdown and error recovery.
  * Prevented lingering editor undo data when panels are closed or scenes change.
  * Added diagnostics for GPU allocations that remain active during shutdown.
  * Improved cleanup of shared textures, fonts, materials, timers, and rendering resources.

* **Documentation**
  * Added guidance for identifying and safely releasing lazily created graphics resources.

### DIFF
--- a/OloEditor/src/MCP/McpToolsRender.cpp
+++ b/OloEditor/src/MCP/McpToolsRender.cpp
@@ -4686,7 +4686,12 @@ namespace OloEngine::MCP
                 // The engine's stand-in when neither an override nor an imported
                 // material exists. Constructed exactly like Scene's cached default
                 // (Scene.cpp::GetDefaultMaterial), which is file-static there.
-                static const Ref<Material> s_EngineDefault =
+                //
+                // Deliberately NOT static: this is a per-call temporary. As a function
+                // static it was a process-wide Ref<Material> with no release site — the
+                // same shape as everything #839 swept — and caching buys nothing here,
+                // since an MCP tool handler runs at agent speed, not per frame.
+                const Ref<Material> engineDefault =
                     Material::CreatePBR("Default", glm::vec3(0.8f, 0.8f, 0.8f), 0.0f, 0.5f);
 
                 const Material* overrideMaterial = entity.HasComponent<MaterialComponent>()
@@ -4739,14 +4744,14 @@ namespace OloEngine::MCP
                     // rule the renderer no longer follows would make this tool lie to the
                     // next person debugging a material.
                     const Material& resolved =
-                        ResolveSubmeshMaterial(overrideMaterial, meshSource.get(), index, *s_EngineDefault);
+                        ResolveSubmeshMaterial(overrideMaterial, meshSource.get(), index, *engineDefault);
                     const Material* material = &resolved;
                     std::string_view source = "engine default material";
                     if (material == overrideMaterial)
                     {
                         source = "MaterialComponent (override)";
                     }
-                    else if (material != s_EngineDefault.get())
+                    else if (material != engineDefault.get())
                     {
                         source = "MeshSource imported material (per-submesh)";
                     }
--- a/OloEditor/src/MCP/McpToolsRender.cpp
+++ b/OloEditor/src/MCP/McpToolsRender.cpp
@@ -4746,7 +4746,7 @@ namespace OloEngine::MCP
                     const Material& resolved =
                         ResolveSubmeshMaterial(overrideMaterial, meshSource.get(), index, *engineDefault);
                     const Material* material = &resolved;
-                    std::string_view source = "engine default material";
+                    std::string_view source;
                     if (material == overrideMaterial)
                     {
                         source = "MaterialComponent (override)";
@@ -4754,6 +4754,10 @@ namespace OloEngine::MCP
                     else if (material != engineDefault.get())
                     {
                         source = "MeshSource imported material (per-submesh)";
+                    }
+                    else
+                    {
+                        source = "engine default material";
                     }
 
                     const PODMaterialData data = Renderer3D::CreatePODMaterialDataForMaterial(*material, RHI::NullResource);
--- a/OloEditor/src/Panels/SceneHierarchyPanel.cpp
+++ b/OloEditor/src/Panels/SceneHierarchyPanel.cpp
@@ -81,9 +81,41 @@ namespace
 
 namespace OloEngine
 {
+    // Every DrawComponent<T> instantiation owns its own static map of undo snapshots, and a
+    // snapshot is a COPY of the component — so the ModelComponent one owns a Ref<Model> and
+    // the texture-bearing ones own Ref<Texture2D>. There is no way to reach a function-local
+    // static of a template from outside, so each instantiation registers a clearer here the
+    // first time it runs, and ClearComponentEditSnapshots() calls them all.
+    //
+    // Note this is the same leak species as s_DrawComponentScene further down this file, but
+    // wrapped in a container: a scan that looks for `static Ref<GpuType>` declarations cannot
+    // see a Ref that lives inside a std::unordered_map's mapped_type. It was the last object
+    // still alive at vmaDestroyAllocator after the rest of the #839 sweep landed.
+    static std::vector<void (*)()>& DrawComponentSnapshotClearers()
+    {
+        static std::vector<void (*)()> clearers;
+        return clearers;
+    }
+
     SceneHierarchyPanel::SceneHierarchyPanel(const Ref<Scene>& context)
     {
         SetContext(context);
+    }
+
+    SceneHierarchyPanel::~SceneHierarchyPanel()
+    {
+        // The panel dies with EditorLayer during Application's m_LayerStack.Clear(), i.e.
+        // before Renderer::Shutdown() — so anything the snapshots pin is released while
+        // the graphics context is still alive, which is the deadline that matters (#839).
+        ClearComponentEditSnapshots();
+    }
+
+    void SceneHierarchyPanel::ClearComponentEditSnapshots()
+    {
+        for (void (*clear)() : DrawComponentSnapshotClearers())
+        {
+            clear();
+        }
     }
 
     void SceneHierarchyPanel::SetContext(const Ref<Scene>& context)
@@ -91,6 +123,11 @@ namespace OloEngine
         m_Context = context;
         m_SelectionContext = {};
         m_SelectedEntities.clear();
+        // The snapshots are keyed by entity UUID and hold copies of the OUTGOING scene's
+        // components; keeping them across a scene swap both leaks that scene's GPU
+        // resources and lets a stale "before" state feed the undo stack of a UUID that
+        // happens to repeat in the new scene.
+        ClearComponentEditSnapshots();
     }
 
     void SceneHierarchyPanel::OnImGuiRender()
@@ -1107,9 +1144,42 @@ namespace OloEngine
         return modified;
     }
 
-    // File-scope state for undo integration in DrawComponent (set by DrawComponents)
+    // File-scope state for undo integration in DrawComponent (set by DrawComponents).
+    //
+    // Scoped to ONE DrawComponents() call by DrawComponentStateScope below, and that
+    // matters: s_DrawComponentScene is a strong Ref, and leaving it set between frames
+    // pinned the last scene the Properties panel drew for the rest of the process —
+    // every MeshSource, texture and cubemap in it — long past m_LayerStack.Clear().
+    // On Vulkan that made closing the editor ABORT rather than exit — every MeshSource
+    // and texture in the scene was still live at vmaDestroyAllocator (measured: 12 vertex
+    // arrays, 6 textures, 147 MB on the sandbox start scene; 64 vertex arrays on the
+    // scene in #839). On GL it was silent. Note the leak only happens once an entity has
+    // been selected, because that is what makes the Properties panel call DrawComponents
+    // at all. A Ref parked in file-scope state "until the next frame overwrites it" is a
+    // leak on the frame that never comes.
     static CommandHistory* s_DrawComponentCmdHistory = nullptr;
     static Ref<Scene> s_DrawComponentScene = nullptr;
+
+    // Clears both on every exit path from DrawComponents(), so the invariant "these are
+    // live only for the duration of the call" is enforced by the type system rather than
+    // by a trailing assignment somebody can jump over.
+    struct DrawComponentStateScope
+    {
+        DrawComponentStateScope(CommandHistory* history, const Ref<Scene>& scene)
+        {
+            s_DrawComponentCmdHistory = history;
+            s_DrawComponentScene = scene;
+        }
+        ~DrawComponentStateScope()
+        {
+            s_DrawComponentCmdHistory = nullptr;
+            s_DrawComponentScene = nullptr;
+        }
+        DrawComponentStateScope(const DrawComponentStateScope&) = delete;
+        DrawComponentStateScope& operator=(const DrawComponentStateScope&) = delete;
+        DrawComponentStateScope(DrawComponentStateScope&&) = delete;
+        DrawComponentStateScope& operator=(DrawComponentStateScope&&) = delete;
+    };
 
     // Compile-time stable component name extraction (no RTTI dependency).
     // Uses a probe type in the OloEngine namespace to calibrate __FUNCSIG__/__PRETTY_FUNCTION__.
@@ -1503,6 +1573,16 @@ namespace OloEngine
                         alignas(alignof(T)) unsigned char snapshotBytes[sizeof(T)]{};
                     };
                     static std::unordered_map<u64, EditState> s_EditStates;
+                    // Register this instantiation's map for bulk clearing exactly once. A
+                    // capture-less lambda can name an enclosing function's static directly
+                    // (statics are not captured), so it converts to the plain function
+                    // pointer the registry holds. See DrawComponentSnapshotClearers (#839).
+                    [[maybe_unused]] static const bool s_ClearerRegistered = []
+                    {
+                        DrawComponentSnapshotClearers().push_back([]
+                                                                  { s_EditStates.clear(); });
+                        return true;
+                    }();
                     auto& editState = s_EditStates[static_cast<u64>(entity.GetUUID()) ^ typeid(T).hash_code()];
 
                     // Take a snapshot once per idle→edit cycle (not every frame)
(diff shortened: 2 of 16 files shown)


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
All three addressed in 0b07c09e — thanks, the Major one was a genuine hole.

**Scan all supported C++ extensions (Major).** Correct, and it landed on the guard itself,
which makes it the most useful finding here: the test enumerated only `.cpp`/`.h`, so a
GPU-owning static in any other C++ file was never examined and the test passed anyway —
the exact "a scan that stops checking" failure this PR documents, reproduced inside the
guard written to prevent it.

One correction to the specifics: `.cc`, `.hpp` and `.hh` **do not exist** anywhere under
`OloEngine/src` or `OloEditor/src`, so those three alone would have changed nothing today.
What does exist is **15 `.inl` files** (the OloHeaderTool-generated component lists plus
hand-written ones like `DebugLevers.inl`). Both are now covered by a seeded roster —
`.cpp/.cc/.cxx` and `.h/.hpp/.hh/.inl` — and `SiblingTranslationUnits()` probes every
implementation/header counterpart instead of the single `.h`<->`.cpp` pair.

Red-checked rather than assumed: planting `static Ref<Texture2D> s_RedCheckInlLeak;` in
`DebugLevers.inl` now fails the test and names the file, where the previous scan never read
it. The scan reads 1763 files (was 1748) and still reports zero, so the widening introduced
no false positives.

**Function-pointer alias + `[[nodiscard]]` (Minor).** Applied as suggested —
`using SnapshotClearer = void (*)();`, used for the vector and the loop variable.

**Close the material-source chain (Minor).** Applied, with one change: rather than adding an
`else` that restates the initializer, I dropped the initializer and made it a complete
if/else-if/else, so the default is stated exactly once.

Issue #847, opened 2026-08-20
Test: CrossShaderUBOMemberOffsetsAgree compares ("", 0) for every member and cannot fail


drsnuggles8:
Found while adding a C++<->GLSL offset test for the DDGI UBO blocks in #842. Not fixed there
because it is pre-existing, outside that PR's subject, and repairing it may surface real drift in
unrelated blocks that deserves its own change.

## The test cannot currently fail

`ShaderUBOSizeConsistency.CrossShaderUBOMemberOffsetsAgree` builds a per-block signature as a list
of `(member_name, member_offset)` pairs and compares that signature across every shader declaring
the block. **Both halves of every pair are constant**, so every signature is identical and the
comparison always succeeds:

1. **Offsets are always 0.** It calls
   `refl.get_member_decoration(res.type_id, m, spv::DecorationOffset)`. For a uniform-buffer
   resource, spirv-cross keeps member decorations on `base_type_id`, not `type_id`; the lookup
   finds nothing and returns the default, `0`. Confirmed empirically: the same call against a
   DDGI block returned 0 for all 23 members, while `base_type_id` returned the correct std140
   offsets (16, 32, 48, 64, 68, ...).

   Note `get_type(res.type_id)` *does* return a usable struct — member count and
   `get_declared_struct_size` are both correct — which is why the sibling size test works and why
   this is easy to miss.

2. **Names are always empty.** `ShaderHarness::CompileStageToSpv` does not call
   `SetGenerateDebugInfo()`, so shaderc emits no `OpMemberName` and `get_member_name` returns `""`
   for every member.

So two shaders whose `CameraMatrices` blocks had genuinely swapped members produce byte-identical
signatures and compare equal. That is precisely the bug the test's own header comment describes
("shader A has `mat4 u_View; mat4 u_Projection;` while shader B has them swapped ... same total
size, identical bindings, `glLinkProgram()` doesn't catch it").

## Fix

The offset half is one word:

```cpp
// OloEngine/tests/Rendering/ShaderUBOSizeConsistencyTest.cpp, ~line 451
refl.get_member_decoration(res.base_type_id, m, spv::DecorationOffset)
```

`get_type(res.type_id)` at ~line 442 should move to `base_type_id` too, for consistency.

**Expect this to go red.** The test has never compared real offsets, so any block that has drifted
between two shaders since it was written has been passing. Triage what it reports before assuming
[...]

Issue #824, opened 2026-08-17
Build: glad's configure-time spec fetch has no cache/retry, hit a 15+min GitHub 429


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

While measuring a clean `build-clang/` build for #822, the `glad-generate` custom command
(`OloEngine/vendor/CMakeLists.txt:265`, via glad 2.x's configure-time Python generator) failed
repeatedly with `HTTP Error 429: Too Many Requests` fetching
`https://raw.githubusercontent.com/KhronosGroup/OpenGL-Registry/main/xml/gl.xml`. The rate limit
held for **15+ minutes** across several retries (the response's `Expires` header suggested a
5-minute edge-cache TTL, but the block persisted well past that — likely an origin-level throttle,
not just a CDN cache).

Root cause: `glad_add_library()`'s custom command (`OloEngine/vendor/glad-src/cmake/*.cmake:242-252`)
has **no cache and no retry** — every fresh `glad-build/` directory (any new worktree, any CI
runner, any developer who deletes the directory) re-fetches the live Khronos spec from GitHub raw
content over the network, unauthenticated. There is also no `DEPENDS` on the custom command's
`OUTPUT`s, so once it succeeds once in a given build tree it is never re-attempted — meaning this
is purely a **first-configure** risk, but every fresh tree pays it.

This is a plausible, currently-undocumented contributor to CI flakiness on **any** job that
configures a fresh `build-clang`/Linux tree (`clangcl`, `linux-*` presets) — worth checking if it
ever explains a run that failed at configure/early-build with no diagnosis, the way #796 tracks a
different-cause instance of "job fails during vendor setup."

**Describe the solution you'd like**

`glad_add_library()` (`OloEngine/vendor/glad-src/cmake/*.cmake:183`) already supports a
`REPRODUCIBLE` keyword, which adds `--reproducible` to the generator invocation and uses glad's
bundled local spec instead of fetching from Khronos — no network call at all. Since this repo pins
a single fixed API surface (`gl:core=4.6`, stable for years), the downside (missing the
very latest spec revision) is low. Proposed: add `REPRODUCIBLE` to the
`glad_add_library(glad STATIC API gl:core=4.6 ...)` call.

Verified locally: reconfiguring with `REPRODUCIBLE` added produces a working build (confirmed via
a full `OloEngine-Tests` build) with no network dependency for this step.

**Describe alternatives you've considered**

[...]

Issue #718, opened 2026-07-31
Renderer: shared power-of-two atlas allocator (shadow atlas, VT cache, impostor atlas)


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

`Renderer/Shadow/ShadowAtlas.h` implements a bespoke shelf packer with three hard-coded tile tiers (atlas/4 for ranks 0-1, atlas/8 for ranks 2-5, atlas/16 for the rest) and no free/reuse — every frame is a fresh pack, and a caster whose tiles do not all fit is skipped whole. Impostor baking (`Renderer/Impostor/ImpostorBaker`) does its own layout, and virtual texturing will need a third. None of them supports freeing a region and reusing it, which is exactly what a cached/persistent atlas requires.

**Describe the solution you'd like**

One reusable power-of-two atlas allocator: allocate a square region of size 2^k, free it, coalesce buddies, query occupancy. Retrofit `ShadowAtlas` (which then gains the ability to *keep* a tile across frames rather than repacking — a prerequisite for any shadow caching) and the impostor atlas onto it.

**Describe alternatives you've considered**

Leaving each subsystem with its own packer — cheap now, but the VSM and VT work both want persistent allocation with free/reuse, so this gets written eventually either way; writing it once is strictly better.

**Additional context**

Reference: `D:\repos\Hollow-TerrainSystem\Packages\PhotoTerrain\Runtime\Utility\AtlasAllocatorPD2.cs` (MIT). Sibling utilities in the same folder worth a look while porting: `IndexingBinaryMinHeap.cs`, `BoundsTree.cs`.

**Acceptance criteria**

- Allocator has headless unit tests covering allocate / free / coalesce / fragmentation.
- `ShadowAtlas` uses it with no behavioural regression (`ShadowAtlasPackingTest` still green) and can hold an allocation across frames.

## Score
```olo-score
capability: 5
craft: 1
stability: 3
decay: 1
effort: 2
confidence: 1.0
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #715, opened 2026-07-31
Terrain: Adaptive Virtual Texturing (feedback buffer, indirection map, page cache)


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

The engine has **no virtual texturing of any kind** — a grep for `VirtualTexture` / `IndirectionMap` / `FeedbackBuffer` / `PageTable` across `OloEngine/src` returns nothing. Terrain surfacing is splat-mapped: `Terrain/TerrainMaterial` keeps up to 8 layers with splatmaps and `Terrain_PBR.glsl` / `Terrain_GBuffer.glsl` blend them per pixel with triplanar + height blending. That means per-pixel cost scales with layer count, unique texel density is bounded by splatmap resolution, and there is a hard ceiling on how many distinct materials a terrain can carry.

**Describe the solution you'd like**

An adaptive virtual texture for terrain surfacing — one indirection lookup plus one cache sample per pixel, regardless of how many layers went into it:

- **Feedback buffer** — the terrain shader writes the page it wanted (id + mip) into a small feedback target.
- **Analysis** — read back and reduce the feedback into a unique page-request list off the render thread (our `Task`/`Async` job system is the equivalent of the reference's Unity Jobs usage).
- **Page cache** — LRU/clock over a physical cache texture, with tiles rendered on demand by compositing the splat layers once into a scratch target and blitting into the cache.
- **Indirection map** — a small texture mapping virtual page -> physical cache tile, updated incrementally (delta lists rather than full rebuilds), with a mip chain so a missing fine page falls back to a coarser resident one instead of popping.
- **Adaptive part** — variable-size virtual images allocated from an atlas so density follows what the camera actually looks at, rather than a uniform grid.
- Cache tiles stored BC-compressed (see prerequisite below).

**Describe alternatives you've considered**

Raising splatmap resolution and layer count — quadratic memory for the first, linear per-pixel cost for the second, and neither removes the density ceiling.

**Additional context**

[...]

PR #849, opened 2026-08-20
build: route the default path at the cached tree, and earn a second concurrent build slot

drsnuggles8:
Two build improvements, both driven by metrics the new build queue recorded from three live worktrees this afternoon rather than from reasoning about what ought to help.

## What the measurement said

Three worktrees (#715 slice 2, #718, #823) built through the mutex on 2026-08-20:

| time | worktree | target | jobs | wait | build | exit |
|---|---|---|---|---|---|---|
| 13:16 | #718 | `OloEngine-Tests` | 11 | 0.0m | **42.4m** | 1 |
| 13:31 | #823 | `OloEditor` | 11 | **29.3m** | 14.5m | 0 |
| 13:32 | #718 | `OloEngine-Tests` | 12 | 13.8m | 1.1m | 0 |

Two things stood out, and neither was what I expected going in.

**The compiler cache was doing nothing.** ccache reported *exactly* the call count from the previous day's test builds — zero new activity — and no worktree had ever configured `build-cached`. Every command aimed at `build/`. 58 minutes of building at a 0% hit rate, while a measured 12m04s → 3m23s speedup sat unused.

**Memory was never the binding constraint; serialisation was.** Free memory stayed between 31 and 40 GB all afternoon. 43 minutes of waiting against 58 minutes of building, including one build that waited 13.8 minutes to run for 1.1 — a 12.8:1 ratio.

That second point reframed the cross-tree link semaphore. I had it queued as a memory-safety fix, but with a hard 1-slot mutex there is never more than one linker across all trees — the 3-linker/4.7-GB-free disaster only happened when the mutex was *bypassed*. The semaphore is not a safety fix. It is the mechanism that makes relaxing the mutex safe.

## (1) Route the default build path at the cached tree

Not a tooling failure — `dev-cached` was named in all three worktree handovers and still lost, because `build/` is what CLAUDE.md's examples showed and what the VS Code tasks ran. The path of least resistance was the uncached one.

- Four VS Code tasks: `configure-dev-cached`, `build-cached-tests-debug`, `build-cached-oloeditor-debug`, `run-cached-tests-debug`. The build tasks go through `build-lock.ps1`, unlike the existing `build-*` tasks.
- CLAUDE.md's build examples lead with the cached tree; the skill doc matches.
- `build-lock.ps1` nudges on every uncached build. A nudge, not a rewrite — switching trees mid-task costs one cold build, so it is the caller's call.

The `build/` tree is unchanged and still correct; it is what the VS solution and debugger use.

## (2) Cross-tree link semaphore, then bounded concurrency

Every link now takes a permit from an OS-named semaphore that **every build tree on the machine** contends for, closing the gap CLAUDE.md has flagged as "does not exist yet" since the original measurement. A kernel semaphore rather than a lock file, because the kernel reclaims the permit however the holder exits — no dead-holder case to get wrong. It **fails open**: unthrottled is "slower, maybe tight on memory"; a stuck throttle is "no build in any worktree ever completes again".

The mutex then becomes an N-slot gate (default 2), and **a second slot is earned, not default**. All three must hold:

- the build targets the **cached Ninja tree**, which carries the link semaphore and the compiler cache. A Visual Studio tree carries neither — that generator ignores `CMAKE_<LANG>_LINKER_LAUNCHER`, Ninja job pools and `CMAKE_<LANG>_COMPILER_LAUNCHER` alike, so nothing bounds its link steps;
- **every** current holder is also a cached tree — a cached build starting beside a `build/` build is still the unsafe pair, and the arriving build is the only party positioned to notice;
- free memory is at least `-ConcurrentMinFreeGB` (default 24).

Lane ceiling is divided among active builds, so two concurrent builds get 6 each rather than 12. Every uncertainty answers *no*. `-MaxConcurrent 1` restores the original behaviour exactly.

This is why the two halves reinforce each other: concurrency is something a tree earns *by being throttleable*, and the tree that qualifies is also the fast one.

## Verification

The link semaphore, in isolation:

- exit codes propagate exactly (0 / 1 / 42) — a failing link must fail the build;
- 4 fake links: **3.3s** at 99 permits, **6.3s** at 2, **13.1s** at 1;
- a hard-killed permit holder let the next linker start in **2.3s**, not at the 1800s timeout — the kernel reclaim works;
- a real `clang-cl` invocation forwarded `/Fe:out.exe main.c` through the wrapper and produced a working binary (arg forwarding is the risk: linker command lines carry `-o`, `/OUT:` and sometimes a bare `--`, which is why the wrapper has no `param()` block).

`LinkSemaphore.cmake`, configured against a throwaway project: launcher set under Ninja, **empty** under Visual Studio with an explanatory message, and `FATAL_ERROR` on a zero slot count.

Build concurrency, in an **isolated git repo** so the live lock was untouched (three real builds were running throughout):

| case | elapsed | result |
|---|---|---|
| cached + cached | 8.3s | CONCURRENT |
| uncached + uncached | 14.2s | SERIALISED |
| uncached 1st, cached 2nd | 14.0s | SERIALISED |
| cached 1st, uncached 2nd | 14.1s | SERIALISED |
| cached ×2, `-MaxConcurrent 1` | 14.1s | SERIALISED |

Plus: a 9999 GB free-memory floor refused with its reason, and the second slot took `-j6` while the first held `-j9`.

**Not verified here:** a real end-to-end OloEngine build through the semaphore. Three worktrees are mid-task on the mutex and I did not want to take an hour of their queue. The wrapper is proven on a real toolchain invocation and fails open, so the risk is bounded — but the first cached build after this merges is the real test, and worth watching.

## Ride-along fix, separate commit

`scripts/BuildCleanEnv.ps1` ended in a **bare `--parallel`**, which CLAUDE.md is explicit is never acceptable: it forwards the omission to the native tool, and Ninja's default is `cores + 2` = **30** here, so omitting the number *raises* the width on a host that has already been OOM-killed once. Every `build-*` VS Code task routes through that script. Now takes `-Jobs`, honouring `CMAKE_BUILD_PARALLEL_LEVEL` (which the lock exports) with a `-j6` fallback.

## Also included

`docs(start-work)`: the step-2 sweep now runs `issue_scores.py lint`. `rank` only lists issues carrying an `olo-score` block, so an unscored issue is not ranked low — it is **absent**, and `/start-work` cannot see it. Issues filed with `gh issue create --body ...` bypass the template that seeds the block, which is how worker sessions file findings. Measured today: three open issues were unscored and one (#847, a test that structurally cannot fail) scored **6.0 — rank #2**, above everything the sweep surfaced on its own.

Refs #824.

## Summary by CodeRabbit

- **New Features**
  - Added VS Code tasks for configuring, building, and testing cached Debug builds.
  - Enabled concurrent cached builds when memory and capacity allow.
  - Added shared link throttling to coordinate parallel builds.
  - Added configurable build parallelism and link concurrency limits.

- **Documentation**
  - Updated work-selection guidance to score all unscored issues and include results in reports.
  - Clarified cached-build usage, concurrency rules, Windows editing safeguards, and Visual Studio limitations.


drsnuggles8:
## Correction: the permit-reclamation claim in the description was wrong

Review caught three real defects, pushed as `f756e08f`. One of them invalidates a verification claim in the PR body above, so recording it explicitly rather than quietly editing it.

**The body says:** *"a hard-killed permit holder let the next linker start in 2.3s, not at the 1800s timeout — the kernel reclaim works."*

**That test was invalid.** A Windows semaphore's count has no owner — `ReleaseSemaphore` is never implicit, so a killed holder never gives its permit back. The test passed only because it had a *single* holder: killing it dropped the last handle, the kernel destroyed the named object, and the next process created a fresh one with a full count. It measured object destruction, not permit reclamation.

Re-measured with a second process keeping the object alive:

```
next linker took 15.2s  ->  PERMIT LEAKED (waited out the fail-open timeout)
```

This is not hypothetical here: `build-lock.ps1` kills whole build trees via `Stop-ProcessTree` when it detects an orphaned build, so every orphan cleanup would have burned a permit permanently, silently degrading the throttle to nothing while still looking healthy.

**Fixed** by using N named **mutexes** with `WaitAny` instead of one semaphore. A mutex is owned by a thread, so a dead owner hands the next waiter an `AbandonedMutexException` *together with the acquisition* — the automatic reclamation a semaphore only appeared to have. Same scenario, re-run:

```
next linker took 3.3s, exit 0  ->  PERMIT RECLAIMED
[link-semaphore] reclaimed permit 1 from a process that died holding it
```

That change surfaced a second issue: `WaitAny` caps its handle count and **throws** past the cap, which fails open — so a nonsense-high `OLO_LINK_SEMAPHORE_SLOTS` would have silently disabled the throttle rather than loosening it. Now clamped to **63**, not 64: an STA thread reserves one, and clamping to 64 still threw `must be less than or equal to 63`.

## The other two

**Slot 0 was admitted unconditionally**, which is not the same as "nothing else is building". With A in slot 0 and B in slot 1, A finishing frees slot 0 while B is still linking — and an uncached build would have taken it with no check at all, landing exactly the unsafe pair the rule exists to prevent. Admission is now derived from the actual held state of every slot rather than the index, and a held slot whose holder record cannot be read is refused rather than assumed safe. Regression-tested: uncached C now waits 23.3s for B instead of starting beside it.

**A linker that failed to launch could report success**, because `$LASTEXITCODE` keeps its previous value when the executable never starts, and that value can be 0.

One deliberate deviation from the suggested fix here: the review proposed setting `$code = 1` when `$?` is false and otherwise retaining `$LASTEXITCODE`. That would be wrong — `$?` is `$false` for *any* non-zero exit status, so a linker legitimately returning 42 would be reported as 1. Instead `$LASTEXITCODE` is cleared to a sentinel before the call and the launch throw is caught, so "still null" means "nothing native ran". Both cases retested: `42 -> 42`, missing executable `-> 1`.

## Re-verification after the changes

Admission matrix re-run in an isolated git repo (the live lock untouched — three real builds were running): **7/7 correct**, including the new freed-slot-0 case and the `-ConcurrentMinFreeGB` floor. Link wrapper: exit codes `0/1/42` preserved, launch failure exits 1, throttling still 3.3s / 6.3s / 12.4s at 99 / 2 / 1 permits, clamp reports and works. Both scripts parse clean.

`CLAUDE.md` is corrected to describe the mutex behaviour, including why the original measurement was misleading — that seemed worth keeping as a documented trap rather than just silently swapping the mechanism.

Issue #624, opened 2026-07-11
Texture compression: higher-quality BC6H, BC5 auto-select, GPU compute encoder (post-#440 follow-ups)


drsnuggles8:
**Is your feature request related to a problem? Please describe.**
The offline texture-compression pipeline (#440) now ships BC7 (albedo/emissive, sRGB), BC5 (tangent-space normals), and BC6H (HDR/IBL, unsigned) via `Renderer/TextureCompression.{h,cpp}`, the `.olotex` container, the `glCompressedTextureSubImage2D` upload path, and asset-pack auto-cook in `AssetPackBuilder` (honoring `BuildSettings::m_CompressAssets`). Several higher-effort or lower-priority extensions were deliberately deferred to keep #440 a shippable slice. This issue tracks them.

**Describe the solution you'd like**
Deferred scope from #440, roughly in priority order:

1. **Higher-quality BC6H encoder.** The current encoder emits only BC6H **mode 11** (1 subset, 10-bit endpoints) with a least-squares endpoint refinement — ~38 dB peak-normalized PSNR on a curved HDR gradient (the flat/smooth case is near-lossless). Multi-subset modes (0–9, two partitions) and a proper PCA/covariance endpoint fit would close the gap on high-variance blocks. Also add the **signed** BC6H variant (`GL_COMPRESSED_RGB_BPTC_SIGNED_FLOAT`) for source data that carries negatives.

2. **BC5 auto-selection for normal maps in the pack auto-cook.** Auto-cook currently picks BC7 (LDR) / BC6H (HDR) but never BC5 — BC5 stays an opt-in via a pre-cooked `.olotex`, because the filename heuristic can't reliably distinguish a 2-channel tangent-space normal from other linear data (roughness/AO/height). A reliable signal (asset metadata / per-texture import setting) would let normals ship as BC5 (better quality + smaller than BC7 for xy data).

3. **GPU compute BC encoder** (the "fast path" the original #440 proposed). The current CPU encoders are fine for a cook step but a compute-shader encoder would speed large bakes.

4. **Linear-space / sRGB-aware mip generation.** The mip downsample is a box filter in the stored space (gamma-naive for sRGB, matching `glGenerateMipmap`'s default). A gamma-correct (linear-space) mip build would remove the slight darkening on sRGB mip chains.

5. **Per-channel / packed-format heuristics** (ARM/ORM channel packing, per-channel format selection).

**Describe alternatives you've considered**
[...]

PR #857, opened 2026-08-20
feat(terrain): publish the VT indirection map by delta, not by rebuild (#715 slice 2)

drsnuggles8:
**Slice 2 of 4** on issue #715 (terrain adaptive virtual texturing). Slice 1 shipped in #844 and left
the issue open for slices 2–4; **this PR does not close it.**

## What changed

Slice 1 republished the whole indirection map whenever residency changed: clear every mip, re-stamp
every resident page, re-propagate every level. That is ~1.33 × `VirtualPagesWide²` texel writes plus
as many again in reads, to express a change of typically under a dozen texels. This replaces it with
the reference's delta list (PhotoTerrain's `IndirectionMapDelta` / `PopulateIndirectionMap.compute`,
MIT), which writes only the texels that changed and re-propagates only their descendants.

The point is that the delta is **equivalent** to the rebuild, not merely cheaper. Three rules make it
so, and each fails quietly:

1. **An eviction is an ENTRY, not an absence.** With the clear pass gone, "page P is no longer
   resident" has to be written as an explicit all-zero texel — byte-for-byte what the clear kernel
   used to leave. There are two sources of unmapping and only one notifies: the LRU eviction
   listener, and `Invalidate()`, whose `DeallocateObject` does **not** fire the listener, so it
   writes its own unmaps by hand.
2. **The coarse→fine propagation still runs, over the DESCENDANTS of every change.** A texel that
   changed at mip *m* invalidates a 2^k × 2^k block at every finer mip *m−k*.
   `VTIndirectionDelta::GetFillRect` derives that in one walk: level *l*'s rect is (level *l+1*'s
   rect, doubled) ∪ (what changed at *l*). Stopping one level short reintroduces the page pop the
   mip chain exists to prevent — and it presents as a streaming bug, not an indirection bug.
3. **One entry per texel, last write wins.** Two updates to one texel inside a single dispatch race.

The whole-map rebuild survives, but only for a map whose contents are **not known** rather than
merely stale — fresh texture storage, and a delta that outgrew its upload buffer. It is expressed
*as* a delta (every resident page mapped, every fill rect full) rather than as a second code path, so
there is one write loop and one fill loop and the flag only changes the clear pass and the
rectangles.

While in there, the reference's own `AVT_WriteMissingPixels` guard (`maxTouchedMip > 0`) turns out to
be wrong — a frame whose only change is at mip 0 never repairs an unmapped mip-0 texel. Noted in the
guide; not reproduced here.

## Measured

The handover asked for a before number and warned not to promise a speedup. Both paths are measured
in the same build, on the same traffic, via the `OLO_TERRAIN_VT_FULL_REBUILD` lever.

**RTX 4090, Debug, at the sandbox scene's 256-page map** (`TheDeltaPublishesTheSameFrameAsTheFullRebuild`
with its map temporarily raised for the measurement):

| | texels touched per publish | best GPU sample |
|---|---|---|
| whole-map rebuild | 174,905 (fixed) | 0.064 ms |
| delta | 17,053 – 32,763 | 0.029 ms |

Published on **~55% of frames** during ordinary camera movement. At the evidence test's own 32-page
map it is 2,873 → 192–1,085 texels and 0.033 → 0.014 ms.

Two things worth being honest about:

- **The publish is dispatch-bound at these sizes.** 0.064 ms for 175k texels on a 4090 is not
  bandwidth. Most of the 2.3× comes from issuing ~7 dispatches instead of ~18, not from writing fewer
  texels. In absolute terms both numbers are under a tenth of a millisecond.
- **The delta is ~10× off its ideal texel count**, because the fill rect is a bounding box and LRU
  evictions scatter: the pages *arriving* are camera-local, but the pages *leaving* are wherever the
  LRU tail happens to be. Tracking N rectangles per level would recover that at up to one dispatch
  per change per level — very likely a regression at this scale, which is why it was left alone.
  §4 of the guide records this so the next person does not "fix" it blind.

What the slice buys unconditionally is that the cost stops being a function of `VirtualPagesWide²` on
every residency change, which is what makes slice 3's larger and variable-size virtual images
affordable.

GPU milliseconds here are the **lowest** resolved sample of each path, not the latest. A single
timestamp pair around a handful of small dispatches also catches whatever else the GPU was doing —
measured, the same rebuild reported anywhere from 0.05 ms to 2.9 ms across five runs. The texel
counts are deterministic; prefer those.

## Verification

**Headless equivalence, with two negative controls.** The guard this slice exists for is
`TheDeltaProducesTheSameMapAsAFullRebuildOverRandomTraffic`: both paths driven over a randomised
insert/evict sequence against a cache far smaller than the working set, compared texel for texel at
every mip, every frame. The delta and the rect derivation come from production code; only the three
kernels are modelled.

That test is worth nothing if it cannot fail, so two more deliberately publish a broken delta and
require the comparison to **reject** it:

- `TheEquivalenceCheckFailsWhenTheDeltaOmitsItsUnmaps` — and then names the damage: the evicted
  page's texel still addresses the old tile, still marked direct, so the fill pass will not repair it
  either.
- `TheEquivalenceCheckFailsWhenThePropagationStopsOneLevelShort` — the child still names a page that
  is no longer resident.

**GPU-side A/B.** The above compares a CPU *model* of the kernels, and this slice changed two of
them; a rect that is right in C++ and wrong in GLSL, or an std430 header whose two declarations
drifted, would pass. So `TheDeltaPublishesTheSameFrameAsTheFullRebuild` renders the same scene
published both ways and compares the frames against a floor measured from a second delta-path run
through the identical invalidate-and-reconverge cycle — so page-to-tile assignment, bake order and
analysis-completion order vary in both, and the publish path is the only difference. Measured
**0.0023/255 mean, max 16 on a handful of pixels**, against 0.23/255 for the VT-vs-splat comparison
that already ships. Anti-vacuous in three directions: both frames must be lit, the delta runs must
not have fallen back to a rebuild, and the forced run's publishes must *all* have rebuilt.

**Visual, multi-angle and in motion.** Slice 1's three committed evidence goldens (`converged`,
`vt_on`, `motion_vt`) still match to within run-to-run noise (mean 0.0008–0.058/255, max 11–13) under
the delta path — an independent confirmation that was not reverse-engineered from the new test — so
they are left untouched rather than rebased. The two new PNGs are the delta/rebuild pair. Live in the
editor on `TerrainVirtualTextureTest.olo` from four poses (overview, oblique, ground level, grazing):
coherent terrain, no wrong-tile patches, no shader errors in `OloEngine.log`.

`TerrainVirtualTexture.*` 33/33, the four `TerrainVirtualTextureVisualEvidenceTest` cases and the
`TerrainGPUQuadtree*` neighbours all green; `DebugLevers.*` / `CVar*` /
`ShaderUBOSizeConsistency.*` green for the new lever and the widened std430 header.

## Diagnostics added

- `OLO_TERRAIN_VT_FULL_REBUILD` — a registered debug lever, so it appears in the startup log, in
  `olo_debug_levers`, is settable at runtime by name (`olo_cvar_set`, the editor console) and is the
  same switch as the terrain panel's new checkbox. First thing to reach for when the terrain looks
  wrong: if forcing the rebuild fixes the frame, the bug is in the delta.
- `Stats` gains what each publish cost (texels stamped / re-propagated), how many publishes and how
  many of them rebuilt, the frame denominator, and a GPU-ms figure per path. All shown in the terrain
  panel.

## Two things found on the way, neither fixed here

- **A ccache direct hit leaves ninja with an empty header-dependency record**, so that TU never
  rebuilds when a header changes. 506 of 1408 objects in a fresh `build-cached` are in that state;
  it swallowed a `DebugLevers.inl` edit in this session and surfaced only as a link error, which is
  the *lucky* outcome — the same mechanism silently produces objects compiled against different
  versions of one header. Worked around here with a `--target clean` before the verification builds.
  Reported separately; it is a build-system change with a cache-invalidating config decision in it,
  not something to slip into a terrain PR.
- **There is no read-only route to the VT counters over MCP.** Reading them meant screenshotting the
  editor's Properties panel. Logged to the MCP follow-up tracker.

## Scope

Deliberately **not** in this PR: slice 3 (adaptive / variable-size virtual images, unblocked now that
#718 has merged) and slice 4 (BC-compressed cache tiles, still blocked on the GPU compressor in
item 3 of #624). Doing 2 and 3 together would make one unreviewable PR out of two separately
verifiable changes.

Companion guide `docs/agent-rules/terrain-virtual-texturing.md` gains §4 for the delta's invariants,
the measured bounding-box finding, and the reference's mip-0 bug; sections renumbered and the
`agent-rules/README.md` cross-reference updated with it.

## Summary by CodeRabbit

* **New Features**
  * Terrain virtual-texture indirection updates now publish incrementally, reducing texel work during resident-page changes.
  * Evictions explicitly clear mappings, preventing stale page references.
  * Full rebuilds remain available automatically when needed or through a diagnostic toggle.
  * Added terrain virtual-texturing telemetry, including publish counts, GPU timings, and texel activity.

* **Bug Fixes**
  * Improved consistency between incremental updates and full rebuilds.

* **Documentation**
  * Updated terrain virtual-texturing guidance and failure-mode documentation.


drsnuggles8:
## 🤖 Self-review @ `a2c3f1ec`

Reviewed the PR diff at high effort, plus a pass over the two GLSL kernels against their C++ twin.

**Findings: 3 · Fixed: 3**

- **Timing queries leaked on a partial creation failure.** `EnsureTimingQueries` assigned handles
  slot by slot and called `DestroyTimingQueries()` the moment one came back invalid — which releases
  only what had been *adopted*, stranding every handle past the bad one with nothing able to reach
  them. Now the whole batch is validated before anything is adopted, and released out of the array
  on failure. Instrumentation is allowed to be absent, not to leak.
- **`Stats::m_IndirectionTexelsWritten` did not count the clear pass**, so the reported cost of the
  path this PR replaces was roughly half of what it actually is (1 508 rather than 2 873 texels at
  the evidence test's map size). A measurement that flatters the new path is worse than none;
  counted now.
- **The GPU-ms figure was the latest sample, and it was not a number I could stand behind** — the
  same rebuild reported 0.05 ms to 2.9 ms across five runs of the same test, because one timestamp
  pair around a handful of small dispatches also catches whatever else the GPU was doing. Changed to
  the lowest sample per path, which is stable to ±0.002 ms across runs; the deterministic texel
  counters carry the assertion.

**Dismissed:**

- *The interleaved 16-byte CPU write between dispatches that read the same buffer* (now two per
  publish rather than one). Kept: it is slice 1's existing pattern, GL orders it, and the Vulkan
  backend gives the same guarantee through #691 Phase 8's write snapshots. The alternative is a
  bound range per dispatch, which needs a `glBindBufferRange` the RHI facade does not expose.
  Comment now names the guide rather than just asserting GL is fine.
- *`Destroy()` resets the delta with a default-constructed config*, so `m_PagesWide` briefly reads
  256 for an object that owns nothing. Harmless — the eviction listener is detached first and
  `Configure()` resets it again — and a second clearing API to avoid it would be worse.
- *The fill rectangle being a bounding box rather than an exact set.* Measured, not assumed: it is
  ~10× off the ideal texel count. Tightening it to N rectangles per level trades back the dispatch
  count that most of the win actually comes from, so it stays, documented, with the measurement
  attached.

One thing outside the diff, found while verifying it and reported separately as **#858**: a ccache
*direct* hit leaves ninja with an empty header-dependency record, so 506 of 1408 objects in this
tree never rebuild when a header changes. It swallowed a `DebugLevers.inl` edit mid-session. Every
build behind the verification in the PR description was taken after a `--target clean`, for that
reason.

drsnuggles8:
@coderabbitai review

(The first attempt hit the OSS review rate limit and never started — re-requesting now that it has reset.)

drsnuggles8:
@coderabbitai review

drsnuggles8:
## 🤖 Self-review @ `2ff63225` — CodeRabbit round

All five findings taken, none dismissed. Verified each against the current code first; the two
documentation ones were overclaims I wrote, and the three code ones are real.

- **The rebuild's trigger list contradicted itself.** §4 said the rebuild runs on "two occasions and
  no others" and that anything else reaching it is a delta bug — while §4 itself documents
  `OLO_TERRAIN_VT_FULL_REBUILD` as a third trigger forty lines later. Now stated as three, with the
  lever marked intentional so a rebuild count is not misread while it is set. (The `.cpp` comment
  already had this right, which is how the two drifted.)
- **"The delta path can never be the slower one" was unconditional** about a bounding box that
  degenerates to O(`VirtualPagesWide²`). Made precise rather than deleted, because the direction is
  still right and worth knowing *why*: the delta skips the clear pass outright and issues fewer
  dispatches, so it does less GPU work than a rebuild of the same map — but in the degenerate case
  the two are the same order, and the delta adds up to `MipCount-1` extra 16-byte header writes the
  rebuild's fill loop does not need. Same treatment for "the cost stops being a function of
  `VirtualPagesWide²`": true of the clear pass, true of the fill only down to its bounding box. §4
  now says which, and flags that this is what slice 3 has to plan against.
- **`offsetof` → named constants.** Two SonarCloud failures, and correct on the merits: `glm::uvec4`
  is a class type, so `offsetof` on the header is conditionally-supported at best. I flagged this
  as a risk in the first self-review and kept it — wrong call. The replacement is pinned by
  `static_assert(kFillParamsOffset + sizeof(glm::uvec4) == sizeof(VTIndirectionHeader))`, so the
  constants cannot drift from the struct, and it catches a member added at *either* end — which is
  the failure `offsetof` would not have caught anyway. `<cstddef>` dropped with it.
- **The GPU A/B did not validate `deltaB`.** It validated the two subject frames but not the one
  used as the noise floor — and a blank floor *inflates* the bound instead of failing, so a
  genuinely wrong rebuild frame would have passed underneath it. Exactly the anti-vacuous check I
  applied to the other two and missed on the one where it matters most.
- **`resident.size() > 0` does not prove an eviction happened.** It proves pages were mapped. Rule 1
  — an eviction is an *entry*, not an absence — is the reason that test exists, and an eviction is
  the only thing that exercises it. Now counted in the listener and asserted non-zero, so a future
  config change that stops the cache overflowing fails loudly instead of quietly testing half the
  delta.

Re-verified at this head: `TerrainVirtualTexture.*` 33/33, all four
`TerrainVirtualTextureVisualEvidenceTest` cases green on GPU (delta 645 texels / 0.015 ms against
rebuild 2 873 / 0.035 ms, published on 83 of 160 frames), pre-commit clean.

PR #860, opened 2026-08-21
fix(renderer): break OpenGLShader self-ref cycle causing 115 leaked programs

### LINKED ISSUE #841: OpenGL: 115 shader programs survive Renderer::Shutdown(); Vulkan releases the same set cleanly

drsnuggles8:
## What changed

`ShaderResourceRegistry::m_Shader` was a strong `Ref<Shader>` pointing back to the `OpenGLShader`
that embeds it (`OpenGLShader::m_ResourceRegistry` is a by-value member). Every file-backed shader
therefore held a hidden strong reference to itself, so its refcount could never reach zero —
`Renderer::Shutdown()`'s entire release sequence (`ShaderLibrary::Clear()`, every
`s_Data.*Shader.Reset()`, `ShutdownFallbackShader()`, `ShaderWarmup::Shutdown()`) would run
correctly and still leave the object alive, because one owner (the shader's own registry) was never
dropped. `VulkanShader` never calls `InitializeResourceRegistry`, which is why only the OpenGL
backend showed the leak — Vulkan released the identical set cleanly on the identical teardown path.

Fixed by storing a non-owning `const Shader*` instead: the registry can never outlive the shader
that embeds it (it isn't separately heap-allocated), so a raw back-pointer is safe by construction.
`GetShader()` goes through `WeakRef<Shader>::Lock()` rather than reconstructing a `Ref` directly —
a self-review finding (see below) — so a caller reaching the registry through the raw-pointer-keyed
`ShaderResourceRegistry::Find()` map (as `CommandDispatch::SetShaderResource` already does) can
never resurrect a `Ref` during the shader's own release, matching the liveness-checked pattern the
codebase's sibling raw-pointer path (`ShaderRegistry::LockAll`) already uses.

## Verification

- **Live editor, both backends** — start scene, select an entity, close. Before this fix: 115
  surviving `Shader` GPU allocations reported by `RendererMemoryTracker::Shutdown()` on OpenGL.
  After: **zero** survivors of any type on OpenGL, matching the already-clean Vulkan teardown.
  (Evidence captured to `.claude/skills/run-oloengine/shots/gl-final.log` /
  `vulkan-after-fix.log` in the worktree — not part of this diff.)
- **New regression test** — `ShaderResourceRegistrySelfRefTest.cpp`: creates a shader through a
  local `ShaderLibrary`, confirms `InitializeResourceRegistry()` ran, drops every external
  `Ref<Shader>`, and asserts a `WeakRef<Shader>` to it goes dead. This is the isolated,
  in-process version of the live-editor check above.
- **Full suite** — `OloEngine-Tests`: 6245/6268 passed, 23 environmental skips (Steam stub,
  headless MCP attach, etc. — all pre-existing skip conditions unrelated to this change), 0
  failed.

## Review guide

**Where I'd look hardest**
1. `OloEngine/src/OloEngine/Renderer/ShaderResourceRegistry.h:95-98,105-112` — the actual fix.
   The invariant "this registry can never outlive its shader" is enforced by comment, not the
   type system (a self-review finding — see Least confident below).
2. `OloEngine/src/OloEngine/Renderer/ShaderResourceRegistry.cpp:24` — the constructor's matching
   `m_Shader(shader.get())`. This constructor has zero call sites in the tree today (verified by
   grep across the whole repo), so it's unexercised by anything except symmetry with the header.

**What I verified, and how** — named above: the live before/after teardown logs on both backends,
the new isolated regression test, and the full suite run. I did not rely on the test suite alone
for a lifetime/teardown-ordering change — `docs/testing.md`'s own guidance is that a green suite
was what shipped this exact leak in the first place (it's how #839's sweep found the 115
survivors: the suite was green throughout).

**Least confident about** — the safety of the non-owning raw pointer rests on "`ShaderResourceRegistry`
is always a by-value member of exactly one `OpenGLShader`, for its whole life." That's true today
(verified: no other constructor call site, no copy/move path since `RefCounted`'s atomic member
makes the owning `OpenGLShader` non-copyable/non-movable), but it's an invariant a future refactor
could violate silently. A self-review pass (4 parallel finder agents at `/code-review` scope) flagged
this and specifically caught that `GetShader()`'s original implementation reconstructed a `Ref<Shader>`
from the raw pointer with no liveness check — a real (if currently unreachable) resurrection-during-release
race, since `ShaderResourceRegistry::Find()` already hands out raw registry pointers across an
unsynchronized global map. Fixed by routing through `WeakRef<Shader>::Lock()`, which is now in this
diff.

**Deliberately not tested** — I did not add a test asserting the by-value-member invariant itself
(e.g. via `static_assert` on non-copyability propagating through the ownership chain), since the
existing `RefCounted` atomic-member mechanism already makes this structurally true for any class in
this hierarchy, not something specific to `ShaderResourceRegistry` that would need its own guard.

## A note on how long this took to verify

The fix above was correct from the first edit, but confirming that live took an unusually long
investigation: several rebuild-and-relaunch rounds each seemed to show the leak persisting (a
different session's survivor count, then live refcount tracing showing every survivor pinned at
1 with no reachable second owner, then a live session reproducibly double-freeing an *unrelated*
shader on close). All of it turned out to be stale/ODR-inconsistent state in the `dev-cached`
incremental build tree — a live debugger attach on a hung (assert-blocked) process showed a call
stack that was structurally impossible under the actual (fixed) header, which is what identified
it as a build problem rather than a code problem. A clean rebuild made both symptoms disappear
immediately. Captured as a new postmortem for the next person who hits this shape:
`docs/agent-rules/incremental-build-odr-staleness.md` (added in this PR).

Closes #841

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed shader resource registry ownership behavior to prevent self-referential shader leaks and unsafe cleanup during teardown.
  * Added regression coverage confirming shaders are released after external references are removed.

* **Documentation**
  * Added troubleshooting guidance for stale incremental builds and misleading runtime results.
  * Documented self-referential shader ownership patterns and updated companion guide indexes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,6 +78,7 @@ you are in.
 - [vcpkg-dependency-management.md](docs/agent-rules/vcpkg-dependency-management.md) — read before adding, bumping or removing a dep: the CRT triplet mismatch is heap corruption, and three of the five traps are silent.
 - [configure-time-variable-visibility.md](docs/agent-rules/configure-time-variable-visibility.md) — a CMake variable resolved *below* the `add_subdirectory()` that consumes it, behind a guard that reads "unset" as "nothing to do": correct on every configure but the first.
 - [asset-import-usd-alembic.md](docs/agent-rules/asset-import-usd-alembic.md) — the importer/exporter registry seam, and vendoring OpenUSD / Alembic / MaterialX into a static-everything build.
+- [incremental-build-odr-staleness.md](docs/agent-rules/incremental-build-odr-staleness.md) — when a correct fix's live behavior makes no logical sense, suspect the `dev-cached` incremental build before the code: a struct-layout change to a by-value member with an inline `=default` destructor can link without error while a stale copy of that destructor survives in another TU.
 
 **Renderer**
 
--- a/OloEngine/src/OloEngine/Renderer/ShaderResourceRegistry.cpp
+++ b/OloEngine/src/OloEngine/Renderer/ShaderResourceRegistry.cpp
@@ -21,7 +21,7 @@ namespace OloEngine
     } // namespace
 
     ShaderResourceRegistry::ShaderResourceRegistry(const Ref<Shader>& shader)
-        : m_Shader(shader)
+        : m_Shader(shader.get())
     {
     }
 
--- a/OloEngine/src/OloEngine/Renderer/ShaderResourceRegistry.h
+++ b/OloEngine/src/OloEngine/Renderer/ShaderResourceRegistry.h
@@ -77,16 +77,42 @@ namespace OloEngine
             return m_Initialized;
         }
 
-        // @brief Set the associated shader for this registry
+        // @brief Set the associated shader for this registry.
+        //
+        // Stores a non-owning back-pointer, deliberately NOT a Ref<Shader>: this
+        // registry is always a BY-VALUE member of the OpenGLShader it describes
+        // (OpenGLShader::m_ResourceRegistry), so its lifetime is strictly bounded
+        // by the shader's — it is constructed and destroyed as part of the shader
+        // object and can never outlive it or be reused for a different one. A
+        // Ref<Shader> here would be a self-referential strong cycle (the shader
+        // owning a strong reference to itself through its own member), which
+        // inflates every shader's refcount by one for its entire life: no amount
+        // of external Ref<Shader>::Reset() at teardown can ever bring the count to
+        // zero, so the shader is never destroyed and its GL program never deleted
+        // (issue #841 — 115 OpenGL shader programs surviving Renderer::Shutdown()
+        // while the identical Vulkan path, whose VulkanShader never calls
+        // InitializeResourceRegistry, released the same set cleanly).
         void SetShader(const Ref<Shader>& shader)
         {
-            m_Shader = shader;
+            m_Shader = shader.get();
         }
 
-        // @brief Get the associated shader
+        // @brief Get the associated shader.
+        //
+        // Goes through WeakRef<Shader>::Lock() rather than constructing a
+        // Ref<Shader> from m_Shader directly: Lock() atomically checks that
+        // the object is still live (under the same mutex Ref<T>::DecRef's
+        // release path uses) before incrementing, so a caller that reached
+        // this registry via the raw-pointer-keyed ShaderResourceRegistry::Find()
+        // map — as CommandDispatch::SetShaderResource already does — can never
+        // resurrect a Ref during the shader's own release. A plain
+        // Ref<Shader>(m_Shader) would skip that check entirely (see
+        // docs/agent-rules/intrusive-refcount-weakref-races.md, case 2).
+        // const_cast is safe: m_Shader is const only because SetShader() takes
+        // its Ref<Shader> by const&, not because the pointee is actually const.
         Ref<Shader> GetShader() const
         {
-            return m_Shader;
+            return m_Shader ? WeakRef<Shader>(const_cast<Shader*>(m_Shader)).Lock() : nullptr;
         }
 
         // Resource discovery from reflection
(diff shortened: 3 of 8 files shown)


drsnuggles8:
@coderabbitai review

Issue #813, opened 2026-08-16
RHI/Vulkan: mesh shaders — replace virtual geometry's software cluster rasteriser where the hardware allows


drsnuggles8:
Follow-up to #691 (roadmap: #812). Wanted explicitly, not speculative.

## Why this engine specifically

Most engines adopting mesh shaders have to invent the data first. This one
already has it: `VirtualGeometry` builds cluster hierarchies with LOD
simplification and boundary locking (see
`docs/agent-rules/cluster-lod-simplification.md`), culls them on the GPU, and
**software-rasterises** the visibility buffer — including a two-pass occlusion
scheme against an HZB, and a 64-bit `atomicMin` fast path where the driver
exposes `GL_ARB_gpu_shader_int64` + `GL_NV_shader_atomic_int64` (issue #629).

A cluster hierarchy plus GPU culling *is* the mesh-shader input format. The
task/mesh split maps onto the existing cull → raster split almost directly.

## Scope

- `VK_EXT_mesh_shader` on the Vulkan backend, behind the ADR 0010 capability
  gate's existing shape (refuse-or-degrade, decided explicitly — not a silent
  fallback).
- The software rasteriser **stays**. It is the GL path, it is the fallback for
  devices without mesh-shader support, and it is the reference the mesh-shader
  output must match. This is an added path, not a replacement.
- Golden/visual parity against the software path is the acceptance test:
  virtual geometry already has visual evidence tests
  (`VirtualGeometryVisualEvidenceTest`, `VirtualGeometryRasterParityTest`) and
  the second one exists precisely to compare rasterisation strategies.

## Known traps before starting

- **This is a second shader compile route** for the affected family (amendment
  (19)/(24)'s rule: a new route needs a variant-keyed cache entry, or a warm
  cache silently serves the wrong SPIR-V). It interacts with #805 — decide the
  route taxonomy once rather than twice.
- Mesh shaders and the `OLO_VULKAN` vertex-pull branch are alternatives, not
  layers: a mesh-shader pipeline has no vertex input *and* no pull stream, so
  the §5 "minimise PSO axes" reasoning needs re-checking for this family.
- Device support is narrower than the ADR 0010 floor. NVIDIA/AMD desktop yes;
  the capability gate must treat it as optional, unlike `descriptor_heap`.
- Measure before converting: the software rasteriser exists because it beat the
  hardware path for *small* clusters. The interesting question is the crossover
[...]

PR #861, opened 2026-08-21
Baked GI for static geometry: lightmap bake, path-traced probe volumes, runtime sampling (#439)

### LINKED ISSUE #439: Renderer: baked GI / lightmap path for static geometry
## Problem
GI is realtime-only (SSGI + IBL). There's no baked lightmap / irradiance-volume path for static geometry. It looks fine, but you lose a major perf lever and ground-truth bounce lighting for static-heavy scenes.

## Why it matters (Tier 3)
A baked path gives high-quality, cheap indirect lighting for the static world (terrain, buildings, props), freeing the realtime budget for dynamic objects. Standard for shipping open-world titles.

## Proposed approach (compute/offline)
- Offline lightmap bake (UV2 unwrap + path-traced or irradiance bake — can run in compute) for static meshes; runtime samples the lightmap.
- And/or baked irradiance volumes (probe grid) for dynamic objects sitting in static lighting.
- The bake step lives in the asset-cook pipeline (`AssetPackBuilder` / `BuildGamePanel`).

## Acceptance criteria
- A static scene shows baked indirect bounce; dynamic objects sample baked probes; runtime cost lower than full SSGI for the same look.

---
**Tier 3 — visual quality / perf lever.**

## Score
```olo-score
capability: 5
craft: 13
stability: 1
decay: 1
effort: 8
confidence: 0.5
learning: 5
fun: 5
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
Baked GI for static geometry: an offline CPU lightmap bake (xatlas UV2 unwrap + the
reference path tracer as the bake kernel) with runtime sampling in the forward PBR
path, a path-traced bake mode for the existing SH light-probe volumes so dynamic
objects sit in matching baked lighting, and a staleness gate that refuses to sample
a bake the scene has drifted from.

## Design in one paragraph

Every texel stores `PathTracer::EstimateIrradiance` at its surface point — indirect
irradiance E in the tracer's physical units, the same convention the DDGI atlas pins
against the same oracle. Punctual direct lighting is excluded *by construction* (delta
lights only reach the estimator through path-vertex NEE, never the query point), so
direct light + shadows stay fully realtime and the lightmap replaces exactly the
probe/IBL diffuse term in the existing ambient ladder, with identical energy-split
semantics. Texel seeds derive from atlas coordinates, so a bake is bit-deterministic
at any thread count — which is what the whole verification story leans on.

## The pieces

- **UV2 as a parallel `MeshSource` stream** (attribute 3, static VAOs only), not a
  `Vertex` member — the 32-byte vertex contract, every Vulkan pull stride, and the
  #813 contact surface stay untouched. Persisted as an appended `.omesh` v6 section;
  remapped through `MeshOptimization` beside bone influences. xatlas vendored (CPM,
  SHA-pinned, `XA_MULTITHREADED=0` for determinism).
- **`LightmapBaker`** in two thread-safe stages: `Prepare` (game thread — in-place
  unwrap, area-driven region sizing, `AtlasAllocator` packing with degrade-before-
  drop, UV2-space rasterization) and `BakeTexels` (any thread — reads only frozen
  jobs + the built `ReferenceScene`; ParallelFor + pull-dilation).
- **`ReferenceSceneBuilder`** — the ECS→`ReferenceScene` adapter `ReferenceScene.h`
  always promised: verbatim Scene.cpp light packing, shared `ResolveSubmeshMaterial`
  precedence, uniform-scale instances shared / non-uniform pre-transformed (winding
  flipped under mirroring), UUID-sorted determinism.
- **`LightmapAsset` (`.olmap`, `AssetType::Lightmap = 40`)** — framed zlib sections,
  version-range header, all-or-nothing finiteness-validated decode, full asset-system
  registration; stored via `CreateOrReplaceAsset` so asset packs carry it.
- **`SceneLightmapRuntime`** — the staleness gate: an FNV-1a bake key over the static
  set, geometry proxies, transforms, resolved material factors, lights and bake
  settings, recomputed at resolve; mismatch ⇒ the lightmap is never sampled
  (`Enabled=0` upload + warning). The resolve **self-heals**: a lightmap-static mesh
  that lost its UV2 stream (primitives recreated at load) is re-unwrapped
  deterministically onto the baked layout, so only genuine changes read as stale.
- **Runtime plumbing** — new `LightmapData` UBO at the documented-free binding 1;
  atlas published at documented-free TEX 16 via `PublishTextureOffsetAndBind` (the
  DDGI shared-include mechanism); per-draw regions ride `InstanceData` (224→240 B)
  through the Color/Custom FrameDataBuffer pattern so auto-batching preserves them.
- **`PBR_MultiLight_Skinned` gains the probe ladder it never had** — skinned dynamic
  meshes now sample baked/realtime probe volumes at all (the issue's dynamic-object
  criterion was structurally unserved on forward).
- **`BakeVolumePathTraced`** — ground-truth SH probes from the same oracle, matching
  the shipped SH convention exactly (which turned out to store band-limited
  *radiance*, not irradiance — a pre-existing π-scale divergence from DDGI/lightmap,
  now numerically pinned by tests and documented rather than latent).
- **Editor** — Build ▸ Bake Lightmaps (game-thread prepare, background texel bake,
  mailbox completion, progress + cancel), inspector "Lightmap Static" checkbox, a
  path-traced probe-bake button, `LightmapTest.olo` acceptance scene + TEST_SCENES
  entry, and `docs/agent-rules/baked-lightmap-pipeline.md` linked from both indexes.

## Ride-along engine fixes (own commits)

- `EditorAssetManager::ReloadData` self-deadlocked the game thread on every
  hot-reload of an existing tracked asset (`SerializeAssetRegistry` re-locking
  `m_RegistryMutex`) — found by cdb against a live wedge during re-bake verification.
- `Ref<T>` const-propagation compile error in the evidence path (caught by review).

## Acceptance criteria, with evidence

1. **A static scene shows baked indirect bounce** — `LightmapBakeParityTest` (5
   headless tests: bit-exact same-seed oracle re-derivation, independent-seed region
   means in [0.85, 1.18], the red/green colour-bleed ordering, memcmp determinism);
   `LightmapVisualEvidenceTest` (GPU, ON/OFF differential + 4 committed evidence
   PNGs); and a live editor session: Build ▸ Bake Lightmaps on `LightmapTest.olo`
   (6 entities, 59,802 texels, 64 spp, ~7 s), room visibly bounce-lit with correct
   bleed, log trail from click to resolve.
2. **Dynamic objects sample baked probes** — `LightProbePathTracedBakeTest` (6
   headless tests; E_SH/E = 0.952–1.037 against the oracle across all nine probe/
   normal points), plus the skinned probe branch closing the forward gap.
3. **Runtime cost lower than full SSGI for the same look** — measured over MCP GPU
   pass timings on the RTX 4090: SSGI costs **~0.43 ms/frame** (pass-total 1.409 →
   0.974 ms toggling `SSGIEnabled` on `SSGITest.olo`, deferred, ~1411×942), while the
   lightmap's marginal cost is **~0.02 ms** (scene pass 0.407 → 0.426 ms on the
   corridor with the bake resolved) — ≈20× cheaper, with multi-bounce and off-screen
   bounce SSGI structurally cannot produce. Staleness semantics carry their own
   suite (`SceneLightmapStalenessTest`, 6 tests).

## Self-review (8-angle /code-review) — what it changed

The full batch of confirmed findings was folded in before opening this PR; the
severest ones, so reviewers can check the fixes rather than rediscover the bugs:

- **Black-texel light leak (shader):** `sampleLightmapIrradiance` now returns
  vec4 with COVERAGE in `.a` as the branch signal — the old `dot(rgb,rgb) > 0`
  branch made a validly-baked pure-black texel (enclosed room) fall through and
  glow with sky IBL. Plus un-premultiply at chart edges.
- **One ambient ladder** — the four-rung source ladder is hoisted into
  `include/AmbientLadder.glsl`, one definition for both forward shaders (the
  skinned copy's "photometrically continuous" comment overclaimed; the include
  now documents the SH radiance-vs-E units caveat instead).
- **Resolve() cost + spam:** the per-frame O(scene) bake-key recompute, the
  per-frame WARN while stale, and the per-frame xatlas retry on a
  persistently-failing mesh are gone — throttled recheck (15 frames),
  once-per-episode warn latches, failed-unwrap memo cleared on `Invalidate()`.
  The Scene.cpp call is now unconditional so clearing `LightmapAsset` actually
  drops the resolved state (the old gate skipped exactly that call).
- **Bake-key parity (key format v2):** lights are hashed per type in the exact
  terms `ReferenceSceneBuilder` consumes — raw `Translation` for point/spot
  (was: composed `transform[3]`, which false-stales on re-parenting), no
  position for directional, and `!(intensity > 0)` lights invisible to the key.
  Two new staleness tests pin it. **v2 stales pre-review bakes** — re-bake.
- **Unwrap knobs deleted:** `LightmapBakeSettings::UnwrapResolution/Padding`
  could bake UVs the self-healing resolve can't reproduce (same counts, same
  key, wrong charts — silent). `Prepare()` hard-codes the shared constants.
- **Cross-region dilation bleed:** dilation is now region-aware (per-entry
  rects, ping-pong buffers) with a new adversarial regression test; a
  singular/NaN world transform is rejected up front (NaN defeated the old
  `len < 1e-6f` guard); vertices are pre-transformed once per entity.
- **Hostile `.olmap` headers:** allocation is bounded by actual remaining
  bytes and deflate's 1032:1 expansion ceiling BEFORE any vector is sized (a
  ~100-byte file could previously demand a 32 GiB texel buffer); three hostile
  crafted-file tests added. The framed-zlib code is now one shared helper
  (`Serialization/ZlibSection`) used by the lightmap, mesh, and animation
  serializers — byte-identical output. `.omesh` UV2 reads get an isfinite scan
  (bad stream dropped, mesh kept).
- **Editor bake lifecycle:** the bake is pinned to the scene captured at start
  (a mid-bake scene switch previously attached the result — and a wrong key —
  to whatever scene was active at completion); `.olmap` hot-reload invalidates
  the runtime (it cached the old GPU atlas forever); the file name gains a
  scene-path hash suffix so two same-named scenes stop overwriting each
  other's bake.
- **ReferenceSceneBuilder:** NaN lights/transforms rejected at the public
  seams (a NaN position was silently accepted into the reference world); the
  spot-direction sanitizer is now one shared function (`Renderer/LightCommon.h`)
  used by Scene.cpp and the builder so the mirrors cannot drift.
- **Probe bake parallelized** (ParallelFor over probes, per-probe seeds keep
  it bit-identical) and its sphere sampler deduped onto `PathTracing`'s
  (verified bit-identical before deleting the copy).
- Smaller: lightmap UBO upload dirty-guarded; entries addressing atlas pages
  > 0 rejected at resolve (only page 0 is uploaded); the CommandBucket
  lightmap-region overflow warning got its own once-per-frame flag.

Dismissed with reasons (not silently dropped): the shared-struct refactor for
bake settings (individual-field hashing is pinned by tests; memcmp-hashing a
padded struct is its own trap), removing the `Page` field (kept for format
stability; guarded instead), a synchronous probe-bake button rework and a
bake-orchestrator extraction (follow-ups — house precedent).

## Deliberately deferred — all five now tracked as issues

Each is filed with the mechanism, the design decision it needs, and acceptance
criteria, so none of this lives only in a PR body:

- **#865 — deferred-path sampling.** The G-Buffer carries no UV2 and all five
  render targets are packed, so a deferred scene resolves its bake and never
  samples it. This is the only deferral that could surprise someone, so this PR
  also makes it *loud*: `SceneLightmapRuntime::WarnIfActivePathCannotSample`
  warns once per resolved bake when the deferred path is active, naming the fix.
  The issue tracks the real feature (which G-Buffer option to take, and what the
  always-on cost is for scenes with no lightmap).
- **#866 — Vulkan sampling.** Needs a third vertex-pull stream (the reserved
  pair is vertex + bone) plus UBO 1 / TEX 16 publication; that's an ADR 0011 §5
  contract change. Compiled out today rather than reading unbound state.
- **#867 — receivers beyond classic `MeshComponent`.** Instanced / Model /
  VirtualGeometry each break the bake's `UUID → one region` model differently;
  Instanced is the highest-value and the biggest change. Includes an explicit
  "do not half-fix the VirtualGeometry fallback path" note — wiring only the
  toggle-off side would make baked GI appear and disappear with the VG master
  switch and destroy that toggle's A/B property.
- **#868 — multi-page atlas.** The format is paging-ready (`PageCount`, per-entry
  `Page`) but bake and runtime are single-page, so oversized scenes degrade
  region sizes instead. Includes a cheap measurement to run *before* building it,
  since the degrade path may well be adequate.
- **#869 — albedo textures and sky/HDRI in the bake.** Inherited from the
  reference path tracer's deliberate factor-only / uniform-environment design.
  Filed as one issue because it's one architectural decision: growing the oracle
  weakens every parity test that depends on it sitting outside the raster's
  sampling conventions, so the choice (grow it / split the roles / bound the gap)
  must be made before any code.

Two MCP tooling gaps found during verification were logged to #607.

Closes #439

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Renderer/Commands/CommandBucket.cpp` (batch region
   stream) + `CommandDispatch.cpp` — the per-draw lightmap region riding the generic
   vec4 stream through auto-batching; a mistake here shares one entity's region
   across a batch and only shows on same-mesh-same-material scenes.
2. `OloEngine/src/OloEngine/Scene/SceneLightmap.cpp::ComputeBakeKey` — the staleness
   gate's completeness. The geometry proxy (counts+bounds+handle) is deliberately not
   a full vertex hash; if you can name an input the bake captures that the key
   misses, that's a real hole (both failure directions are silent).
3. `OloEngine/src/OloEngine/Renderer/Baking/LightmapUnwrap.cpp` — the all-or-nothing
   vertex/index rebuild after xatlas seam-splitting (submesh rebasing, shadow-index
   regeneration, the refusal set).

**What I verified, and how** — every new suite green (LightmapBakeParity,
LightProbePathTracedBake, LightmapUnwrap, ReferenceSceneBuilder,
LightmapAssetSerialization, SceneLightmapStaleness, InstanceDataLayout,
LightmapVisualEvidence on GPU); full suite green three times — twice during the
feature build and once after the self-review batch (59 targeted tests across the
9 lightmap/probe/serialization/governance suites re-run green first, then the
full run); glslc validation of both edited shaders after the ladder hoist; a
clang-cl `-fsyntax-only` preflight over every edited TU (which caught three real
compile errors before the build); live editor bakes with before/after
screenshots and the log trail — re-run after the review batch since bake-key v2
stales pre-review bakes; the perf numbers above from olo_perf_pass_timings.

**Least confident about** — the bake key's geometry proxy (point 2 above), and the
one-atlas-page ceiling: a scene whose static surface area exceeds the atlas degrades
region sizes (logged, never silent) rather than paging.

**Deliberately not tested** — Vulkan rendering of lightmapped scenes (the branch is
compiled out); asset-pack round-trip of `.olmap` beyond the serializer's own pack
tests; editor bake cancellation mid-flight (the token is checked per job, but no
automated test drives it).

---


drsnuggles8:
**Self-review round, and how the result was re-verified**

The self-review ran as 8 independent finder angles (efficiency, reuse, simplification, conventions, altitude, removed-behavior audit, line-by-line, cross-file tracing); every confirmed finding is folded into the six `fix(...)` commits ending at 7cae9ac3 — the severity-ranked list and the dismissals are in the PR body's self-review section.

Post-review verification, since review fixes are code changes:

- 59 targeted tests across the 9 lightmap/probe/serialization/governance suites, then the **full suite, all green** (exit 0; GPU visual tests ran, not skipped, on the RTX 4090).
- Both edited shaders glslc-validated after the `AmbientLadder.glsl` hoist.
- A clang-cl `-fsyntax-only` preflight over every edited TU caught three real compile errors (a duplicate `case AssetType::Lightmap` and a `Ref<T>` const-propagation) before the build queue was touched.
- **Live editor re-bake** (bake-key v2 stales pre-review bakes, so this matters): opened `LightmapTest.olo` over MCP, clicked Build ▸ Bake Lightmaps via injected input, and watched the full trail —

  ```
  Lightmap bake started: 6 entities, 59802 texels, 64 spp
  LightmapSerializer: wrote 'LightmapTest.olo-344d0c45.olmap' (126747 bytes, 512x512 x1 pages, 6 entries)
  Lightmap bake complete: 6 entities baked, 0 skipped
  SceneLightmapRuntime: resolved lightmap 512x512 with 6 entity regions
  ✨ Auto-imported new asset: ...\lightmaps\LightmapTest.olo-344d0c45.olmap
  ```

  The hash-suffixed file name and the immediate valid resolve at key v2 are both the new code paths working live; the viewport screenshot shows the room bounce-lit with correctly-signed bleed. No hot-reload deadlock (the ride-along `ReloadData` fix holding).

drsnuggles8:
**Follow-up (9d5eb3b8): the NVIDIA shader-recompile warning from the live-bake log is fixed**

The flagged `131218 — Vertex shader in program 15 ['PBR_MultiLight'] is being recompiled based on GL state` warning was this branch's doing: `a_TexCoord2` (location 3) was buffer-backed only on meshes with a UV2 stream, so the same program drew both enabled- and disabled-attribute VAO layouts, and NVIDIA specializes a vertex-shader variant per layout permutation. The repeat-recompile warning (plus its expensive draw-site stack capture) fired exactly when the first post-bake UV2 mesh was drawn — and any scene mixing lightmapped and non-lightmapped static meshes would pay the same first-frame hitch.

Fix: `VertexArray::AddConstantVertexBuffer` — a stride-0 DSA binding is literal (every vertex of every instance fetches offset 0), so an 8-byte `(0,0)` stub backs the attribute on every static mesh that lacks the stream and all static VAOs expose one identical layout. Vulkan overrides it as a no-op that records nothing (appending it would shadow the bone-pull stream). Verified by re-running the identical live repro — scene open → menu-click bake → resolve — which now logs **zero** 131218 lines where the pre-fix session logged the recompile-#2 warning at the bake transition; frame unchanged; 17 targeted tests incl. the GPU evidence room green.

drsnuggles8:
**Deferred work is now tracked, not just documented (ce4a9fd2)**

Fair challenge that the five "deliberately deferred" bullets would evaporate the moment this PR merged. They are now five filed issues, each with the mechanism, the design decision it needs, and acceptance criteria:

- **#865** — deferred-path sampling (G-Buffer carries no UV2; all five RTs packed)
- **#866** — Vulkan sampling (needs a third vertex-pull stream; ADR 0011 §5 contract change)
- **#867** — receivers beyond classic `MeshComponent` (Instanced / Model / VirtualGeometry)
- **#868** — multi-page atlas (format is paging-ready; bake and runtime are not)
- **#869** — albedo textures + sky/HDRI in the bake (one architectural decision about the oracle, not two tasks)

**And one of them got fixed here rather than filed-and-forgotten.** Four of the five fail per-draw and visibly — an unreached receiver falls back to probes/IBL exactly as an unbaked mesh does. The deferred path fails *scene-wide and silently*: it resolves the bake, logs `resolved lightmap NxN with M entity regions`, and then never samples it, which reads as a failed bake. `SceneLightmapRuntime::WarnIfActivePathCannotSample` now says so once per resolved bake, naming the path, the workaround, and #865. The latch clears on `Invalidate()` so a re-bake re-warns, and it stays silent when no valid bake is resolved.

I did **not** attempt the other four in this PR, deliberately: each is architecture-scale (a sixth G-Buffer target, a third pull stream, three separate bake data-model changes, a texture-sampling model for the oracle), and #866 additionally can't be verified live while the Vulkan deferred editor frame is frozen under #823.

Two traps are recorded in the issues and the companion guide so nobody re-derives them: half-wiring the VirtualGeometry *fallback* path would make baked GI appear and disappear with the VG master switch, destroying that toggle's A/B value; and growing the reference tracer's material/environment model weakens every parity suite that depends on it sitting outside the raster's sampling conventions.

Build green, 15 lightmap/parity/GPU tests re-run green.

drsnuggles8:
@coderabbitai review

Issue #862, opened 2026-08-21
Renderer/Vulkan: virtual geometry's SOFTWARE raster path device-faults (READ of invalid address) — hardware paths stable


drsnuggles8:
Found during the #813 mesh-shader bring-up (PR #859) — the first time the full `VirtualGeometryPass` ever ran a complete frame on Vulkan. The mesh-shader work is **exonerated**: the fault reproduces bit-for-bit with the mesh path forced off.

## Symptom

Opening a virtual-geometry scene (`VirtualGeometryTest.olo`) on the Vulkan backend at **default settings** kills the device on the first VG frame:

```
[Vulkan] DEVICE FAULT REPORT: '' (2 address record(s), 0 vendor record(s))
[Vulkan]   fault address: 0x2000c9860 (precision ±0x10) — instruction pointer (faulting)
[Vulkan]   fault address: 0x4cca53000 (precision ±0x1000) — READ of invalid address
=== CRASH: C++ Exception (caught) — Vulkan bring-up: vkWaitForFences failed (VkResult -4) ===
```

`VK_ERROR_DEVICE_LOST`, with `VK_EXT_device_fault` naming a GPU **read** of an invalid address (per docs/agent-rules/rhi-abstraction-boundary.md §12, without the fault report this would be unattributable on Windows).

## The bisect (live over MCP, `olo_virtual_geometry_set` — the #813 levers)

| `hwRasterMode` | `swRasterMode` | result |
|---|---|---|
| `auto` (mesh shaders) | `auto` | **device lost** |
| `forcemdi` | `auto` | **device lost** — identical fault, mesh path never engaged |
| `forcemdi` | `disabled` | stable (15 instances, 24 300 clusters tested, 4 143 MDI-drawn, editor runs) |
| `auto` (mesh shaders) | `disabled` | stable (same numbers through `vkCmdDrawMeshTasksEXT`, zero-pixel parity vs MDI) |

⇒ the faulting ingredient is the **software-raster path**: present in both crashing configs, absent in both stable ones.

## What is already ruled out / fixed

- **Not the mesh path** (row 2 above).
- **Not the resolve pipeline**: `VirtualVisibilityResolve.glsl` had no `OLO_VULKAN` vertex-pull branch — pipeline creation failed VUID-07904 and the resolve silently never drew. Fixed in PR #859 (own commit). The device fault is what remains *after* that fix.
- The cull compute is common to all four rows and stable.

## Suspect set

[...]

Issue #858, opened 2026-08-20
build: a ccache direct hit leaves ninja with no header dependencies, so that TU never rebuilds


drsnuggles8:
## Summary

In the `dev-cached` (Ninja + ccache) tree, **an object served by a ccache *direct* hit is recorded in
`.ninja_deps` with zero header dependencies**, so that translation unit is never rebuilt again when a
header it includes changes. The object silently stays at whatever the headers said when it was last
actually compiled.

Measured on a fresh `build-cached` in this worktree, right after one cold build:

```
$ ninja -C build-cached -t deps | grep -c '#deps'      # objects with a record
1408
$ ninja -C build-cached -t deps | grep -c '#deps 0,'   # ...with NO header deps
506
```

36% of the tree. `ccache --show-stats` for the same period reports **Direct hits: 2989**,
Preprocessed hits: 902, Misses: 2232 — i.e. exactly the calls where ccache never invokes the
compiler, so `/showIncludes` never runs and Ninja has nothing to parse.

## How it surfaced

Adding a row to `OloEngine/src/OloEngine/Core/DebugLevers.inl` and rebuilding:

```
lld-link: error: undefined symbol: bool __cdecl OloEngine::Levers::TerrainVtFullRebuild(void)
```

`DebugLevers.cpp` — which generates every lever's definition from that `.inl` — was **not
recompiled**. Its record:

```
OloEngine/src/CMakeFiles/OloEngine.dir/Debug/OloEngine/Core/DebugLevers.cpp.obj: #deps 0, deps mtime ... (VALID)
```

`.inl` mtime 22:14, object mtime 21:33. Ninja considered it up to date because "valid record, zero
dependencies" means nothing it depends on can have changed.

## Why it is worse than a link error

A link error is the *lucky* outcome — it only happened because the change added a new symbol. The
same mechanism silently produces **objects compiled against different versions of the same header**,
which is an ODR violation with no diagnostic. In this session, a change to
`TerrainVirtualTexture.h`'s class layout rebuilt only the TUs that happened to have a dependency
record; `Scene/Components.h` includes that header, so a large fraction of the engine was potentially
reading `GetStats()` at the wrong offsets. The resulting editor ran and rendered.

Note this does **not** mean ccache returns stale objects — a ccache hit is content-hashed and is
correct for the inputs it was compiled from. The defect is that Ninja never *asks*.

## Scope

[...]

Issue #712, opened 2026-07-31
Renderer: triangle-level culling in the virtual-geometry raster (backface + sub-sample-miss)


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

Virtual-geometry culling stops at the cluster. `VirtualClusterCull.comp` does frustum + normal-cone + Hi-Z per cluster, and everything that survives is rasterized whole — by the compute software rasterizer (`VirtualClusterRaster.comp`) or hardware MDI. Backfacing triangles inside a surviving cluster are only removed by fixed-function culling on the hardware path, and on the *software* path there is no fixed-function stage to do it at all. Micropoly triangles that land between rasterization sample points are shaded and discarded.

**Describe the solution you'd like**

Two per-triangle rejects before rasterization:

1. **Backface** — sign of the screen-space triangle area. On hardware this avoids a state switch and is often faster than letting fixed function do it; on the software rasterizer it is the only way to do it.
2. **Sub-sample miss** — reject triangles whose screen-space bounding box misses every rasterization sample point. At micropoly density this is a meaningful fraction.

**Describe alternatives you've considered**

Relying on fixed-function backface culling — not available on the software raster path, which is the one that handles the small clusters where this matters most.

**Additional context**

Reference: `D:\repos\Timberdoodle\src\shader_lib\cull_util.hlsl` and the per-triangle rejects inside `src/rendering/rasterize_visbuffer/draw_visbuffer.hlsl`; rasterization helpers in `src/shader_lib/rasterization.hlsl`. Apache-2.0.

Related: #654 (virtualized-geometry roadmap — this is a Tier-1-shaped item not currently listed there), #629, #651.

**Acceptance criteria**

- Measured reduction in software-rasterizer time on the #629 stress scene at identical output.
- Overdraw debug view shows lower counts; visibility buffer is pixel-identical.

## Score
```olo-score
capability: 3
craft: 1
stability: 1
decay: 1
effort: 3
confidence: 0.8
learning: 5
fun: 3
kano: performance
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

drsnuggles8:
[...]

PR #871, opened 2026-08-21
Make a virtual-geometry scene that draws nothing say so (#864)

### LINKED ISSUE #864: Scenes: VirtualGeometryStress/Sponza submit 0 instances (unresolvable VirtualMeshComponent handles) — the silent-zero needs a loud check
Found during #813 verification (PR #859). Two of the three virtual-geometry test scenes silently exercise nothing:

- **`VirtualGeometryStress.olo`** opens fine (28 entities) but submits **0 instances / 0 clusters** — `olo_virtual_geometry_stats` reads zero everywhere while the scene claims to be the stress case. Its `VirtualMeshComponent` handles do not resolve.
- **`VirtualGeometrySponza.olo`** has the same known problem (docs/agent-rules/cluster-lod-simplification.md §5: "the `VirtualMeshComponent` handle does not resolve in a clean checkout — the editor loads the scene and settles without registering a virtual mesh").
- `VirtualGeometryTest.olo` is the one scene that works (15 instances, 24 300 clusters) — every live VG measurement currently funnels through it.

This is the `terrain-tessellation-had-no-scene` genre: a feature's showcase scene that silently renders nothing gives every future verification session a false negative ("VG doesn't work on X") or, worse, a vacuous pass ("no diff between modes" because nothing draws). The #813 A/B had to prove non-vacuity via the VG on/off pixel toggle precisely because of this trap.

## What fixing it needs

Re-point the components at resolvable mesh assets (or commit the missing sources), and ideally an assertion in the scene-open path or a stats-tool warning when a scene contains `VirtualMeshComponent`s but the registry ends the frame with 0 registered parts — the silent-zero is the actual bug; the stale handles are just today's cause.

Also worth folding in: the Stress scene predates streaming budgets — once its handles resolve, verify it actually stresses (cluster count, instance count) rather than duplicating `VirtualGeometryTest`.

## Score
```olo-score
capability: 3
craft: 1
stability: 5
decay: 3
effort: 1
confidence: 0.8
learning: 1
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #864.

## What this turned out to be

The issue's premise was that `VirtualGeometryStress.olo` and `VirtualGeometrySponza.olo`
have stale/unresolvable `VirtualMeshComponent` handles. **Neither scene is broken, and
neither handle is stale.** I decoded `AssetRegistry.oar` properly (it is a binary v2
archive: `u32 version`, `u32 count`, then `{u64 handle, u32 type, u32 status, u32 pathLen,
path}` per entry — a text grep over it proves nothing, which is the trap the handover
warned about) and all three handles resolve to real entries:

| scene | handle | registry entry | on disk |
|---|---|---|---|
| `VirtualGeometryTest` | `…228` | `DamagedHelmet.gltf` | committed |
| `VirtualGeometrySponza` | `…234` | `Sponza.gltf` | committed |
| `VirtualGeometryStress` | `…304` | `xyzrgb_dragon.ply` | **fetch-on-demand** |

The dragon is a deliberately-uncommitted 137 MB asset declared in
`scripts/assets/asset-manifest.json`, absent until someone runs `scripts/Fetch-Assets.ps1`.
Its registry entry is intentional — `AssetContentValidityTest` allowlists it on purpose,
because deleting it would silently unregister the stress scene for everyone.

So **repointing the handles would have been the wrong fix**: it would have thrown away the
173M-triangle stress case to "fix" a machine that had merely never run a documented setup
step.

The actual zero in the original report has a simpler cause that nobody had written down:
**virtual geometry only submits on the Deferred path, and the editor boots on Forward.**
My very first live reading reproduced it exactly — `renderingPath: "Forward"`, every
counter zero. That is indistinguishable from a broken scene, which is the whole problem.

The manifest already states the contract that was being violated:

> Every consumer (scene, test) must degrade gracefully when an asset is absent: SKIP with a
> message naming the fetch command, never fail.

`VirtualGeometryStress.olo` documents its fetch step in a YAML comment (line 48) that
nothing reads at runtime. **The silent zero is the bug; the asset state was only ever
today's trigger.**

## The fix — one seam, three consumers

`VirtualMeshRegistry::SubmissionDiagnostics` (reset in `BeginFrame`, filled by the Scene
submission loop) records why a frame drew what it drew: enabled components, how many failed
to resolve, how many failed to build a cluster DAG, how many were submitted, and whether the
master switch routed them to the classic path. One predicate, `SilentlyDrewNothing()`, is
the trigger for all three reporters:

- **`Scene.cpp`** — a warning naming `Fetch-Assets.ps1`, logged on state *change* rather
  than per frame. The existing per-asset warning also now names the fetch command.
- **`StatisticsPanel.cpp`** — the panel previously appeared only when something was
  *already* drawing, so a broken scene hid the very panel its own header tells you to read.
  It now shows whenever the scene asked for virtual geometry, and explains the zero in red.
- **`McpToolsRender.cpp`** — the old note said *"no VirtualMeshComponent in the scene, or
  all of them are disabled"*, which was actively misleading on exactly this bug. It now
  emits a `diagnostics` object plus a note that warns an A/B taken here passes vacuously.

Deliberately silent for: a scene with no `VirtualMeshComponent`, components with no mesh
assigned, and the classic-path fallback (the intended A/B baseline). A spurious fire would
warn on nearly every scene in the project and get muted, which lands us back at square one.

## Live verification

All readings from `olo_virtual_geometry_stats` against a running editor on Deferred.

| scene | components | submitted | instances | clusters tested | `silentlyDrewNothing` |
|---|---|---|---|---|---|
| `VirtualGeometryTest` | 15 | 15 | 15 | 4 860 | `false` |
| `VirtualGeometrySponza` | 1 | 1 | 25 | 5 018 | `false` |
| `VirtualGeometryStress` | 24 | 24 | 24 | **2 834 448** | `false` |

Each was screenshotted over MCP and the pixels inspected, not just the counters: the Test
scene shows 15 PBR helmets with shadows and IBL; Sponza shows textured stonework (its 25
submeshes shading with their imported materials); the Stress scene shows two receding rows
of Stanford dragons. Counters alone would not have distinguished "drawing" from "drawing
something wrong", which is the failure mode this issue is about.

**Does the stress scene actually stress?** Yes, decisively — 2 834 448 clusters tested vs
4 860 for `VirtualGeometryTest` (**583×**), with a 3 165 SW / 51 HW raster split, exactly
the software-raster regime the scene header claims. Cook, from the log:

```
VirtualMeshBuilder: submesh 0 -> 118102 clusters in 7206 groups across 17 levels
                    from 7219045 triangles
Model::CookVirtualMesh: cooked 7219045 triangles into 1 part(s) / 118102 clusters
                        (200486 KB) in 107499.5 ms
VirtualMeshRegistry: mesh asset 4544736348022301304 registered as 1 part(s)
```

**Proof the loud check fires.** I renamed the dragon away and relaunched, reproducing the
original report verbatim (28 entities, 0 instances, 0 clusters) — it now explains itself:

```json
"diagnostics": { "enabledComponents": 24, "unresolvedAssets": 24,
                 "submitted": 0, "silentlyDrewNothing": true }
```
```
Scene: 24 enabled VirtualMeshComponent(s) submitted ZERO virtual-geometry instances this
frame (24 with an unresolvable mesh-source asset, 0 whose cluster DAG failed to build) —
… run 'scripts/Fetch-Assets.ps1' … Do NOT treat a VG measurement taken on this scene as
meaningful.
```

Logged **once**, not per frame (verified: 1 occurrence). And on `VehiclesTest.olo` (no
virtual geometry) it stays silent — `enabledComponents: 0`, `silentlyDrewNothing: false`.
The asset was restored afterwards.

## Tests

New `VirtualGeometrySceneCoverageTest.cpp` (`// OLO_TEST_LAYER: unit`), deliberately
GL-free so it runs in CI — the suite was green for this bug's entire life:

1. every `VirtualMeshComponent` handle in every sandbox scene is registered (the failure the
   issue assumed), without asserting on-disk presence, which is legitimately absent for a
   fetch-on-demand asset;
2. `SilentlyDrewNothing()` fires on both real failure shapes and on **none** of the four
   legitimate zeros;
3. the outcome buckets partition the enabled components, so a future Scene-loop path that
   leaves a component uncounted can't silently weaken the predicate.

`225/225` pass across 29 suites — `VirtualGeometrySceneCoverage`, `AssetContentValidity`,
`VirtualMesh*`, `Scene*`, `SystemScheduler*` and the component-codegen coverage suites.

Run after a **clean** `build-cached` rebuild, deliberately: this branch adds a member to
`Scene.h`, and per issue #858 **342 of 1426 objects in this tree carry no dep record**
(a ccache direct hit never runs `/showIncludes`, so ninja records nothing and never
rebuilds that TU for a header change). An incremental build is therefore unsound after a
layout change — it can leave TUs compiled against two different versions of one header with
no diagnostic. The first pass of these tests was green on the incremental build too, which
is exactly why it could not be trusted.

## Docs

`docs/agent-rules/cluster-lod-simplification.md` §5 carried two claims this work disproved,
both corrected in place rather than deleted, since they are what seeded the wrong premise:

- "Sponza's handle does not resolve in a clean checkout" — it does; 25 instances, measured.
- "Stale registry entries lie" about the dragon — that entry is fetch-on-demand, not stale,
  and must not be cleaned or repointed.

Both now also record the Deferred-path gate that explains the zeros.

## Not in scope

No evidence turned up for #862, #712 or #654. `AssetRegistry.oar` is untouched — an editor
run rewrites it, but the rewrite is pure byte-churn from hash-map iteration order (verified
semantically identical: same 119 entries, types, statuses and paths), so it was reverted.

## Summary by CodeRabbit

- **New Features**
  - Virtual geometry statistics now include clear submission diagnostics, including enabled components, unresolved assets, registration failures, fallback rendering, and silent zero-draw states.
  - The Statistics panel displays virtual geometry status and explains missing submissions or asset issues.
  - Added warnings and guidance for unresolved assets and fetch-on-demand content.

- **Bug Fixes**
  - Improved distinction between successful rendering, fallback behavior, missing assets, and disabled components.

- **Tests**
  - Added coverage for virtual geometry diagnostics, asset registration, fallback, recovery, and per-frame resets.

- **Documentation**
  - Updated virtual geometry measurement and troubleshooting guidance.

### DIFF
--- a/OloEditor/src/MCP/McpToolsRender.cpp
+++ b/OloEditor/src/MCP/McpToolsRender.cpp
@@ -4618,12 +4618,46 @@ namespace OloEngine::MCP
                 j["cull"] = std::move(cullJson);
                 j["residency"] = std::move(residencyJson);
                 j["settings"] = VirtualGeometrySettingsJson();
+                // Explain a zero rather than guess at it (issue #864). The old
+                // "no VirtualMeshComponent in the scene, or all of them are
+                // disabled" note was actively MISLEADING on the one case that
+                // matters: a scene full of VirtualMeshComponents whose mesh
+                // assets did not load reads exactly zero here, and that note
+                // sent the reader looking for a scene-authoring mistake that
+                // does not exist. The submission loop now records WHY, so say so.
+                const auto& diagnostics = registry.GetSubmissionDiagnostics();
+                j["diagnostics"] = Json{
+                    { "enabledComponents", diagnostics.EnabledComponents },
+                    { "unresolvedAssets", diagnostics.UnresolvedAssets },
+                    { "registrationFailures", diagnostics.RegistrationFailures },
+                    { "submitted", diagnostics.Submitted },
+                    { "fellBackToClassic", diagnostics.FellBackToClassic },
+                    { "silentlyDrewNothing", diagnostics.SilentlyDrewNothing() },
+                };
+
                 if (Renderer3D::GetRendererSettings().Path != RenderingPath::Deferred)
                     j["note"] = "Virtual geometry only renders on the Deferred path; the scene does not submit "
                                 "VirtualMeshComponents on Forward/Forward+, so every counter reads zero.";
+                else if (diagnostics.SilentlyDrewNothing())
+                    j["note"] = "WARNING: this scene contains " + std::to_string(diagnostics.EnabledComponents) +
+                                " enabled VirtualMeshComponent(s) but submitted ZERO virtual-geometry instances (" +
+                                std::to_string(diagnostics.UnresolvedAssets) +
+                                " mesh-source asset(s) did not load, " +
+                                std::to_string(diagnostics.RegistrationFailures) +
+                                " failed to build a cluster DAG). Every counter below reads zero for that reason, "
+                                "NOT because virtual geometry is working and finding nothing to draw. A missing "
+                                "asset is usually a fetch-on-demand one — run scripts/Fetch-Assets.ps1 (see "
+                                "scripts/assets/asset-manifest.json) — otherwise check OloEngine.log. Do NOT treat "
+                                "an A/B or performance measurement taken on this scene as meaningful: it will pass "
+                                "vacuously because nothing draws in either mode.";
+                else if (diagnostics.FellBackToClassic && diagnostics.EnabledComponents > 0)
+                    j["note"] = "The virtual-geometry master switch is OFF, so this scene's " +
+                                std::to_string(diagnostics.EnabledComponents) +
+                                " VirtualMeshComponent(s) were drawn through the CLASSIC mesh path instead. Zero "
+                                "counters are expected here — this is the intended A/B baseline, not a fault.";
                 else if (registry.GetFrameInstances().empty())
-                    j["note"] = "No virtual-mesh instances were submitted this frame (no VirtualMeshComponent in "
-                                "the scene, or all of them are disabled).";
+                    j["note"] = "No virtual-mesh instances were submitted this frame: the scene contains no "
+                                "VirtualMeshComponent, or every one of them is disabled or has no mesh assigned.";
                 return j; });
             return ToolResult::Structured(result);
         }
@@ -6502,8 +6536,12 @@ namespace OloEngine::MCP
                 "page budget, and cumulative page uploads + evictions). Use it to verify the cull is "
                 "actually culling (tested >> drawn), to check the HW/SW split after "
                 "olo_virtual_geometry_set { swRasterMode }, and to catch streaming thrash (uploads and "
-                "evictions climbing every frame under a tight budget). Zero everywhere means no "
-                "VirtualMeshComponent was submitted — virtual geometry renders on the DEFERRED path only.";
+                "evictions climbing every frame under a tight budget). When everything reads zero, "
+                "'diagnostics' says WHY — crucially it distinguishes a scene with no virtual meshes from "
+                "a scene whose VirtualMeshComponents all failed to load their mesh asset, which looks "
+                "identical in the counters and makes any A/B measured on that scene pass VACUOUSLY. "
+                "Check diagnostics.silentlyDrewNothing before trusting a zero. Virtual geometry renders "
+                "on the DEFERRED path only.";
             tool.InputSchema = Schema::EmptyObject();
             tool.OutputSchema = Schema::Object()
                                     .Prop("renderingPath", Schema::String())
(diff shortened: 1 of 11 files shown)


drsnuggles8:
Review round applied in `1a0fbffa`. Three of four findings fixed, one declined with evidence.

**Fixed — the panel had a hole the MCP tool did not.** The submission diagnostics are filled by
the Scene loop, which runs on the **Deferred path only**, so on Forward every field stays 0 and
the whole Virtual Geometry section vanished — for a scene full of virtual meshes. That is the
same silent zero in a different disguise, and it is the likeliest one to hit, because the editor
boots on Forward. The panel now asks the scene directly when the path cannot answer, and explains
three cases in the same order the MCP tool does (wrong path → classic fallback → submitted
nothing). Guarded to `!deferred && frameInstances == 0`, so a drawing frame never pays for the scan.

Verified live — `VirtualGeometryTest.olo` on Forward, `enabledComponents: 0` in the tool, and the
panel now reads:

> **15 VirtualMeshComponent(s) present, but the render path is not Deferred**
> Virtual geometry submits on the Deferred path only, so every counter below reads 0. Switch to
> Deferred in Renderer Settings to exercise it.

Before this commit that section did not render at all.

**Fixed — `SubmitVirtualMesh` now returns `[[nodiscard]] bool`** rather than the caller inferring
the outcome from a `GetSubmissions()` size delta. To be clear, the delta was *correct* (there is
exactly one `Submit()` call) — but it depended on an invariant nothing enforced. `RenderPathDrift`
still passes; its regex matches the new call shape and Scene.cpp still routes through the shared
material resolver.

**Fixed — doc triangle figure.** `cluster-lod-simplification.md` uses "7.2 M" in four places; my
new paragraph gave only the byte size. It now reads "the 7.2 M-triangle, 137 MB asset", and the
173 M figure shows its work as `24 × 7.2 M`.

**Declined — excluding probe-baker capture passes**, with the mechanism traced rather than assumed:

- There is **no accumulation**. Each cubemap face calls `RenderScene3D` → `BeginScene` →
  `PrepareFrame` → `BeginFrame()`, which *resets* the diagnostics.
- Every field is **view-independent** — the submission loop walks all entities and never
  frustum-culls (the GPU cull does that later) — so a capture pass produces identical counts.
- Decisively: `m_FrameInstances` and `m_TotalFrameClusterCount`, printed immediately beside the
  diagnostics, share the exact same `BeginFrame` lifetime. Scoping only the diagnostics to the
  primary render would make the panel internally inconsistent — diagnostics from one render, the
  instance and cluster counts beside them from another.

Documented in the struct instead, so the next reader does not re-derive it. (Minor: the finding
referred to `residency.FellBackToClassic`; that field is on `diagnostics`.)

`227/227` pass across 30 suites on a clean rebuild — clean again because this round touched two
more headers, and per #858 an incremental build here is unsound after a layout change.

**One unrelated anomaly, flagged not buried:** a single cold-start launch after a clean rebuild hit
`OLO_CORE_ASSERT(passRef, "BuildRenderPipelineGraph requires a valid render node")`
(`RenderPipelineBuilderInternal.h:265`) and died before the first frame. It did not reproduce on
four subsequent launches, it is in render-graph construction this branch does not touch, and it is
absent from every earlier run's log. Recording it here in case it rings a bell — not investigated
further, since chasing it would widen this PR.

PR #874, opened 2026-08-21
feat(renderer): structured GPU readback-stats channel for GPU-driven passes

### LINKED ISSUE #721: Renderer: structured GPU readback-stats channel for GPU-driven passes

GPU-driven passes report almost nothing about what they actually did. `VirtualMeshRegistry::VirtualResidencyStats` exists, and `GPUPassTimerPool` gives timings, but there is no general way for a compute pass to publish counters (clusters culled by which test, pages allocated/evicted, lights per cluster, work-item overflow) and have them arrive on the CPU cheaply and without stalling. Meanwhile #551 records that we currently do per-frame synchronous readbacks and log them unthrottled — the opposite of what is wanted.

One small, general channel: a per-frame stats SSBO with a typed layout that any pass can `atomicAdd` into, copied into a ring of persistent-mapped staging buffers and read N frames later (never synchronously). Surfaced in the debug overlay and over MCP, and — importantly — carrying **capacity-overflow flags**, so a pass that silently truncated its output says so instead of just rendering slightly wrong.

Per-pass ad-hoc readbacks — what we have; each one is a stall and a potential source of the #551 log flood.

**Additional context**

Reference: `D:\repos\Timberdoodle\src\shader_shared\readback.inl` and `readback_stats.inl` (Apache-2.0); their `ShaderDebugOutput` in `shader_shared/debug.inl:84-96` shows the overflow-flag pattern (`exceeded_*_capacity` per channel).

Related: #551, #607 (MCP capability gaps), #519, #463.

**Acceptance criteria**

- At least two GPU-driven passes publish counters through the channel with no synchronous readback.
- An intentionally overflowed buffer surfaces as a visible overflow flag rather than silent truncation.

## Score
```olo-score
capability: 5
craft: 1
stability: 5
decay: 1
effort: 2
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
One small, general channel any GPU-driven pass can `atomicAdd` into, read back N frames
late behind a fence — never synchronously — and carrying **capacity-overflow flags** so a
pass that silently truncated its output says so.

## What landed

**The channel** — `Renderer/Debug/GPUReadbackStats.{h,cpp}`, a shared name→slot registry
(`GPUReadbackStatsRegistry.h`) and its GLSL twin (`include/GPUReadbackStats.glsl`). One
144-byte SSBO at binding 64, cleared at `BeginFrame` (in `PrepareFrame`, i.e. `BeginScene`)
and copied into a 3-slot device-to-host ring at `EndFrame` (in `EndScene`, after the graph).
Retirement is a fence **poll** — there is no `ClientWaitFence` anywhere in the class, and a
full ring skips the capture rather than blocking on the oldest slot.

**The contract is names, not offsets.** Mirroring a *struct* between C++ and GLSL makes drift a
byte-offset problem, and this repo's history says what that costs: every counter after an
inserted field reads its neighbour's value, with no error and no wrong pixel. A flat `uint[32]`
plus a shared name→index registry makes drift a *name* problem, detectable by reading the two
files as text — which is what `GPUReadbackStatsLayoutTest` does. It is built so it cannot pass
vacuously (the #847 `("",0) == ("",0)` trap): every parse is asserted non-empty against an
independently known count before anything is compared.

**Two adopters, no synchronous readback** (acceptance criterion 1):

- **GPU instance cull** (`InstanceFrustumCull.comp` / `InstanceOcclusionCull.comp`) — input,
  frustum rejects, occlusion rejects, drawn, dropped, two-phase re-test/recovery.
- **Virtual shadow map page cache** (`VSM_MarkRequiredPages.comp` / `VSM_AllocatePages.comp`) —
  requests, requests dropped by a full ring, allocations, evictions, pool-exhaustion failures.

**Overflow flags are real, and forceable** (acceptance criterion 2). Both cull appends were
unbounded writes trusting a capacity nothing verified; they are now bound-checked and report the
refusal. `GPUFrustumCuller::SetDebugOutputCapacity(n)` shrinks the bound so the buffer
*genuinely* truncates — the draw really renders fewer instances and the dropped counter really
counts them. Nothing fakes a flag.

**Surfaced** on the F3 debug overlay (overflows unconditional, counters behind a fold) and over
MCP as `olo_gpu_readback_stats` (read; optional `enabled` write).

## Two things worth reading closely

**A bound check and its "rollback" are not the obvious pair.** The natural
`atomicAdd(count, +1)` / `atomicAdd(count, -1)` shape is correct single-threaded and **corrupts**
under concurrency: giving a reserved slot back lets a later invocation be handed an index another
one already wrote — a clean truncation becomes a lost-and-duplicated instance. Both appends now
reserve from a monotonic cursor and commit the count only after the write lands. Written up as
`gpu-readback-stats-channel.md` §5, including why `VSM_MarkRequiredPages` must do *neither*.

**The SSBO namespace is now full.** 64 was the only number never used as an SSBO (it is
`TEX_DDGI_VISIBILITY` in the sampler namespace). Cross-namespace reuse only breaks *within one
shader* on Vulkan's single-set model, so the constraint is checked rather than assumed:
`NoStatsConsumerAlsoSamplesBinding64` asserts no shader includes both `GPUReadbackStats.glsl` and
`DDGICommon.glsl`. The next feature wanting a buffer binding has to ride an existing block or
renumber — worth saying in its issue before starting.

## One ride-along fix, in its own commit

`fix(tests): set u_CullViewProjection in the raw-GL cull harness` is **not part of
#721**. Merging master in surfaced it: #726 made both cull shaders read
`u_CullViewProjection` unconditionally for their frustum planes, but
`GPUOcclusionCullGPUTest` hand-fills that UBO with raw GL and was never updated, so
the field is the zero matrix. `extractPlane` then normalises a zero vector, every
plane is NaN, and `signedDistance < -worldRadius` is false for NaN — **the frustum
stage is silently inert.**

Measured, not reasoned: all three tests pass on this RTX 4090 with the matrix left
at zero. They do not currently depend on the frustum stage — none of them places an
instance outside the frustum — so it was latent rather than active. That is exactly
what makes it a trap for whoever adds the first out-of-frustum case. CI could not
have caught it either way: the harness is `OLO_ENSURE_GPU_OR_SKIP` and no runner has
a GL 4.6 context.

It is the same shape as the `u_OutputCapacity` omission #721 introduced into the same
file and the same block, which is how I noticed it.

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Renderer/RenderPipeline.cpp:228` — the frame bracket. Self-review
   caught this one: `BeginFrame` was in `UploadExecutionState` (where the sibling
   `ShaderDebugDraw` resets), which runs inside `EndScene` — but the GPU instance cull dispatches
   at *submission* time, before it. The clear wiped the cull's counters after they were written,
   so adopter #1 reported a permanent zero in the real engine while every hand-ordered test
   passed. If you change where either hook lives, it must still enclose submission.
2. `OloEditor/assets/shaders/compute/InstanceOcclusionCull.comp:321-360` — the reserve-then-commit
   split on two different buffers (output and reject list), and the two-word
   `InstanceCullRejectedCount` block whose C++ twin must be zeroed in both words every dispatch.
   `rejectedCount` bounds phase 2's read of `InstanceData`, so an overshoot there is garbage
   transforms rather than missing geometry.
3. `OloEngine/src/OloEngine/Renderer/ShaderBindingLayout.h:2276` — taking binding 64, and the
   argument for why the sampler-namespace collision is safe *given the test that enforces it*.

**What I verified, and how**

- `GPUReadbackStatsEvidence` (3 tests, real compute dispatches on an RTX 4090): the shader's own
  input count == the 512 instances the CPU submitted; `Drawn + FrustumRejected == Input`;
  `Drawn == indirect.instanceCount` (different buffer, different atomic, different readback);
  `Latency > 0`, proving the value came through the ring and not a synchronous read. Forced
  overflow: flag raised, `drawn == 32`, `dropped == 224`, draw count stopped at the capacity.
- `GPUReadbackStatsFullFrame.RealFrameFeedsTheInstanceCullCounters` — a real frame through
  `Scene::OnUpdateRuntime` with a 1156-instance field. **Validated by reintroducing the bug**:
  with `BeginFrame` moved back to `UploadExecutionState` it fails with its own diagnostic, and
  passes again when restored.
- `GPUReadbackStatsLayout` (4) — the C++/GLSL contract and the binding-64 constraint.
  `McpGpuReadbackStats` (6) — the JSON shaping, in particular that an overflow is reachable
  without reading the counters.
- `GPUOcclusionCullGPUTest` (3, pre-existing) — these would have silently read zero survivors
  once the bound check landed; the harness now sets `u_OutputCapacity`, sizes its reject counter
  to two words, and binds a zeroed stand-in at SSBO 64 (it deliberately runs without
  `Renderer::Init()`).
- **No-stall, measured rather than inspected.** The naive A/B was useless — two consecutive
  same-config runs disagreed by 0.70 ms and then 0.05 ms. A paired interleaved design gives
  +0.010 ± 0.072 ms (n=12) and −0.156 ± 0.092 ms (n=14), both within 2 SE of zero and opposite
  in sign. Honest statement: **not resolvable above a ±0.1 ms noise floor at a 5.0 ms frame** —
  not "free", because this instrument cannot see 10 µs.
- Live over MCP: `olo_gpu_readback_stats` against a running editor returns
  `valid: true, latencyFrames: 1, slotsInFlight: 0`.

**Least confident about** — the Vulkan path. Everything above was measured on GL, and the Vulkan
deferred editor frame is frozen (#823) independently of this change, so I did not execute it.

I did read it rather than assume it, and every primitive the channel needs is implemented, not a
`Phase6Stub`: `AllocateBufferStorage(DeviceToHost)` → `VulkanRawBufferRegistry::Allocate`;
`CreateFence` queues a timeline signal on the **next submit**, which is the semantics the ring
wants (`TerrainVirtualTexture` and `GPUBufferLock` already rely on it); `ReadBufferSubData` names
this exact "readback-ring shape (A7)" as its production consumer. The one that could have bitten
is `CopyBufferSubData`/`MemoryBarrier`, which stub out when called *outside a recording bracket* —
but `VulkanContext` opens that bracket around the whole frame-render callback, so `EndScene` (and
`EndFrame` with it) is inside it. **Read, not run** — worth an actual Vulkan frame before anyone
depends on the counters there.

**Deliberately not tested** — a *sustained* VSM request-ring overflow. The flag and dropped counter
are wired and the shader path is straightforward, but forcing the ring full needs a scene dense
enough to exceed `VSM_MAX_REQUESTS`, and the sandbox has none. The instance-cull overflow is the
one exercised end to end. Also untested: two GPU-culled batches in one frame contending on the
counters — correct by construction (atomics into one block) but not measured.

Closes #721

## Summary by CodeRabbit

* **New Features**
  * Added non-blocking GPU readback statistics for instance culling and virtual shadow map rendering.
  * Added debug overlay visibility for counters, latency, ring status, and overflow flags.
  * Added the `olo_gpu_readback_stats` diagnostic tool with channel controls and structured reports.
  * Added capacity safeguards to prevent out-of-bounds culling output and report dropped instances.

* **Documentation**
  * Documented GPU statistics usage, counters, overflow behavior, synchronization, and diagnostics.

* **Tests**
  * Added coverage for layout, reporting, overflow handling, disabled states, and real-frame GPU results.


drsnuggles8:
## 🤖 Self-review @ `f53ed4dcf6d5ce49a634a677e22cb1240c5655a3`

Reviewed the full diff at **high** effort before the PR existed. **5 findings, all real, all fixed** — and the first one is the reason this comment is worth reading:

1. **The frame bracket was in the wrong place.** `GPUReadbackStats::BeginFrame()` sat in `UploadExecutionState` (where the sibling `ShaderDebugDraw` resets), which runs inside `EndScene` — but the GPU instance cull dispatches at *submission* time, before it. The per-frame clear wiped the cull's counters **after** they were written, so adopter #1 would have reported a permanent zero in the real engine. Every GPU test passed anyway, because they all drive `BeginFrame → cull → EndFrame` by hand and so tested the order they assumed. Moved to `PrepareFrame`; added `GPUReadbackStatsFullFrame.RealFrameFeedsTheInstanceCullCounters`, which renders a real frame through `Scene::OnUpdateRuntime`.
2. **`GPUOcclusionCullGPUTest` never set the new `u_OutputCapacity`** → capacity 0 → the new bound check refuses every append → all three tests would have read `instanceCount == 0`. Fixed.
3. **Same harness: its reject counter was still one word** while `InstanceCullRejectedCount` gained `rejectedReserve`, putting an atomic outside the bound range (UB per the GL spec). Fixed.
4. **Same harness runs without `Renderer::Init()`**, so nothing bound SSBO 64 — and the cull shaders now read `b_StatsEnabled` out of it. Reading an unbound SSBO is undefined in GL, which the new GLSL header itself documents. It now binds a zeroed stand-in.
5. **The disabled path still uploaded and re-bound 144 bytes every frame**, contradicting the "off: no clear, no copy, no fence" claim repeated in the header, `RendererSettings` and the MCP schema — the exact claim the tool's `enabled` argument exists to let someone A/B. Now a one-shot collapse.

**Nothing dismissed.**

One thing I did on top of the review: the new full-frame guard was **validated by reintroducing the bug**. With `BeginFrame` moved back, it fails with its own diagnostic (*"the GPU instance cull published nothing during a real frame"*) and passes again when restored. A guard nobody has watched fail is a guard you are guessing about.

Two mistakes of my own, corrected before the first commit, in case they matter to a reviewer:

- my first bound-check used the natural `atomicAdd(count,+1)` / `atomicAdd(count,-1)` rollback, which is correct single-threaded and **corrupts under concurrency** (a returned slot gets handed to an invocation that overwrites an instance already written there). Replaced by reserve-then-commit.
- I initially wrote — in shader comments *and* the agent-rules doc — that a dropped VSM page request stayed dropped for the session. It does not: `VSM_EndFrame` strips the bit every frame. Corrected everywhere; the real defect is a one-frame coarse-level fallback that recurs under sustained pressure.

Issue #887, opened 2026-08-22
DecalComponent texture references are lost on a scene round-trip (empty path via YAML, dropped entirely via .scenebin)


drsnuggles8:
A `DecalComponent`'s texture references do not survive a scene round-trip. Found while verifying #853 (PR #886); it is not that bug and is filed rather than folded in, because the root cause is not pinned and guessing at it would be worse than reporting it precisely.

## Reproduction

`OloEditor/SandboxProject/Assets/Scenes/DecalModeMatrixTest.olo` (added in PR #886) has an Emissive decal with:

```yaml
    DecalComponent:
      EmissiveTexturePath: Assets/Textures/Checkerboard.png
```

Open the scene in the editor and read the component back through the real serializer (`olo_scene_get_entity` calls `SceneSerializer::SerializeEntity`, so this is exactly what a re-save would write):

| load path | what comes back |
|---|---|
| YAML (sidecar deleted) | `EmissiveTexturePath: ""` — the `Ref<Texture2D>` is **non-null** (the serializer only emits the key `if (dc.m_EmissiveTexture)`), but its `GetPath()` is empty |
| `.scenebin` sidecar | the key is **absent entirely** — the texture is dropped outright |
| **control:** `SpriteRendererComponent` in `PinkCubeWithTextures.olo` | `TexturePath: Assets/Textures/Otter.png` — round-trips correctly, and is even self-healingly rewritten from the older working-directory spelling to the asset-relative one |

The control matters: `GetPath()` works, and `OpenGLTexture2D::InvalidateImpl` does set `m_Path` on a successful load. So this is **decal-specific**, not a general texture-path defect.

## Consequence

Two independent ways to lose the reference, both silent:

1. **Any re-save of a scene containing a textured decal writes an empty path**, permanently dropping the reference from the `.olo`. The next load has no texture.
2. **The binary sidecar drops decal textures**, so even without a re-save the fast path and the YAML path disagree about what the scene contains — the `scene-binary-sidecar` hybrid-coverage contract does not hold for `DecalComponent`.

Neither warns. Combined with the fact that a decal whose texture is missing renders as **nothing at all** (its alpha is zeroed and every fragment is discarded), the visible symptom is "my decal disappeared after saving", several steps removed from the cause.

## What is *not* the cause

Ruled out during investigation:

[...]

PR #915, opened 2026-08-22
fix(renderer): stop DecalRenderPass's GLStateGuard error spam (#895)

### LINKED ISSUE #895: DecalRenderPass leaks 8-9 GL state fields every frame, reported at error level (2723 lines in minutes)
`DecalRenderPass::Execute` leaks 8–9 GL state fields every frame it draws, and reports each one at **error** level. Found while verifying #853 (PR #886); present before and after that PR, so it is not a regression from it.

## What it looks like

Measured just now, `DecalModeMatrixTest.olo` loaded, a couple of minutes of runtime: **2723 occurrences**, once per frame that drains a decal, on **both** rendering paths (63 on Deferred, then +70 in the first seconds after switching to Forward).

```
[error] OloEngine: GLStateGuard[DecalRenderPass]: 9 state mutation(s) escaped the pass:
[error] OloEngine:     DepthTest: false -> true
[error] OloEngine:     BlendSrcRgb: 1 -> 770        (GL_SRC_ALPHA)
[error] OloEngine:     BlendDstRgb: 0 -> 771        (GL_ONE_MINUS_SRC_ALPHA)
[error] OloEngine:     BlendSrcAlpha: 1 -> 770
[error] OloEngine:     BlendDstAlpha: 0 -> 771
[error] OloEngine:     ActiveProgram: 0 -> 53
[error] OloEngine:     VAO: 25 -> 28
[error] OloEngine:     Texture2D[10]: 0 -> 35       (TEX_USER_0, decal albedo)
[error] OloEngine:     Texture2D[19]: 0 -> 19       (TEX_POSTPROCESS_DEPTH)
```

(8 or 9 depending on whether the albedo slot changed that frame.) `DecalRenderPass` is the only pass producing these — nothing else in the session logged a guard diff at all.

## Why it is not self-correcting

`DecalRenderPass.cpp:144` constructs the guard with the **default policy**:

```cpp
GLStateGuard guard("DecalRenderPass");
```

`GLStateGuard`'s default is `Policy::Log`, which prints each diff via `OLO_CORE_ERROR` and **does not roll anything back**. `Policy::Restore` exists and does exactly that (`GLStateSnapshot::ApplyCore()`), and logs the residue at trace level instead — but it is not what this call site asks for.

The comment immediately above that line is **stale** and is probably why:

```cpp
// Detector-only guard: ... The explicit restore calls further down still perform the
// actual restoration (the current GLStateGuard only detects leaks, it does not roll back).
```

`Policy::Restore` does roll back. And the "explicit restore calls further down" demonstrably do not cover these fields — the diff is taken at destruction, after them.

## Two separate problems

1. **Signal hygiene, and it is the worse one.** Thousands of `[error]` lines per session for a condition nobody is acting on trains everyone — human and agent — to skim past `[error]` in `OloEngine.log`. This directly cost time on #853: the log was the thing that eventually explained the bug (`ActiveProgram`/`VAO` in this very diff proved the decal draw *was* being issued), and it was buried in its own noise.

2. **The state genuinely is not restored.** Whether each leaked field causes a downstream defect is **not established** and is part of the work here. Command-driven draws re-apply their own `PODRenderState`, and shader/VAO binds are tracked in `CommandDispatch`'s cache, so the likely exposure is non-command GL work (clears, blits, immediate `RGCommandContext` draws) that runs after this pass and inherits `DepthTest`, the blend factors or a bound texture slot.

## What the fix probably is

Either switch the call site to `Policy::Restore` (which is what the policy was written for — "use this when a pass does enough state flipping that a manual restore is error-prone"), or extend the pass's explicit restores to cover the fields above and keep the loud policy so a future leak is caught.

Prefer whichever leaves the guard **loud for new leaks**: dropping to `Policy::Restore` silences the report to trace, so it should come with a check that no *other* field starts escaping. Note also `docs/agent-rules/render-pass-published-state.md` — a pass whose outputs are engine-global bindings must not be wrapped in a restoring guard at all, so confirm `DecalRenderPass` is not in that category before choosing `Restore`.

## Reproduce

1. Open `OloEditor/SandboxProject/Assets/Scenes/DecalModeMatrixTest.olo` (added in PR #886).
2. Either rendering path.
3. `grep -c "escaped the pass" OloEditor/OloEngine.log`.

## Score
```olo-score
capability: 1
craft: 4
stability: 3
decay: 4
effort: 2
confidence: 0.8
learning: 3
fun: 2
kano: table-stakes
blocked_by: []
blocks: []
```
<sub>score = confidence × (capability + craft + stability + decay) / effort. First-pass estimate from the filer. `decay` is the high axis on purpose: error-level log spam that nobody acts on degrades every future diagnosis, and this one sits in the exact file an agent reads when a decal misbehaves. `effort` is low if `Policy::Restore` is the right answer and higher if the pass needs hand-written restores plus a check that the guard stays loud.</sub>

drsnuggles8:
Fixes #895 — `DecalRenderPass` construted its `GLStateGuard` with the default `Policy::Log`, which reports every escaped GL state field at **ERROR** level and rolls nothing back. The pass's manual restore only covered `DepthMask`/`BlendState(enable)`/`DepthFunc`/`Cull`, so `DepthTest`, the four blend func factors, `ActiveProgram` and `VAO` escaped on every frame that drained a decal — 2723 ERROR lines in a couple of minutes on `DecalModeMatrixTest.olo`, on both rendering paths.

`Closes #895`

## The fix

Switch to `Policy::Restore`. This is safe here — unlike the DDGI/fog pattern `docs/agent-rules/render-pass-published-state.md` warns about — because `GLStateSnapshot::ApplyCore()` restores only the **core** GL subset (depth/blend/stencil/cull/polygon-mode/viewport/scissor/FBO/program/VAO) and *deliberately never touches per-slot texture bindings*. So it fixes exactly the fields that were escaping without undoing this pass's texture publishes at `TEX_POSTPROCESS_DEPTH` / `TEX_USER_0`:

- `TEX_POSTPROCESS_DEPTH` is re-published fresh by every downstream consumer (SSAO, Fog, DOF, MotionBlur, TAA, ToneMap all rebind it themselves before sampling) — nothing depends on `DecalRenderPass` leaving it bound.
- `TEX_USER_0-2` are documented pass-local-reuse slots (`ShaderBindingLayout.h`) — same contract.

Residual texture diffs still surface, just at TRACE instead of ERROR — quiet in steady state, discoverable for a future leak hunt. Same shape as `PlanarReflectionRenderPass` / `OverdrawRenderPass` / `ShaderDebugDrawPass`, which also replay geometry with their own programs/VAOs under `Policy::Restore`.

Also added the `CommandDispatch::InvalidateRenderStateCache()` call the non-OIT restore path was missing (the OIT branch already had it) — without it the render-state cache would go stale after the guard's raw restore GL calls bypass the tracked `RenderCommand` layer, risking a downstream pass eliding a call it actually needs to reissue.

**Audit of other `GLStateGuard(` construction sites** (in scope per the issue): every other site already explicitly chooses `Policy::Ignore` or `Policy::Restore` — `DecalRenderPass` was the only one left on the noisy default. No broader sweep needed.

## Verification

- `GLStateGuardTest.*` (37 tests) pass, including `RestorePolicy_RestoresCoreStateOnDtor` and `RestorePolicy_DoesNotUnbindTexturesOrUbosButLogsLeak`, which pin exactly the ApplyCore behaviour this fix relies on.
- Live editor verification on `DecalModeMatrixTest.olo` (per `CLAUDE.md` — headless tests are not sufficient for a rendering change): played the scene on both **Forward** and **Deferred** paths.
  - **Zero** `GLStateGuard[DecalRenderPass]` ERROR lines across ~8000 logged frames on either path (down from 2723 in a couple of minutes).
  - Decals still render their authored colours correctly on both paths — screenshots attached below, consistent with PR #886's recorded baseline (Albedo decal `1.0, 0.251, 0.149`; the checkerboard arm is the known missing-texture placeholder, unrelated to this change).
  - The three remaining `[error]` lines in the session log are a pre-existing missing `Checkerboard.png` asset, unrelated to this fix.

## Out of scope

- **#887** (`DecalComponent` texture references lost on scene round-trip) — same area, deliberately unclaimed this round per the issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved decal rendering state management to ensure graphics settings are restored correctly after decal passes.
  - Updated non-order-independent transparency rendering to keep subsequent rendering operations synchronized with the restored state.
  - Helps prevent visual inconsistencies and rendering artifacts that could occur after decals are processed.

### DIFF
--- a/OloEngine/src/OloEngine/Renderer/Passes/DecalRenderPass.cpp
+++ b/OloEngine/src/OloEngine/Renderer/Passes/DecalRenderPass.cpp
@@ -136,22 +136,30 @@ namespace OloEngine
         if (m_SelectedSceneDepthTexture.IsValid())
             depthTextureID = context.ResolveTextureHandle(m_SelectedSceneDepthTexture);
 
-        // Detector-only guard: captures GL state at entry and on destruction
-        // diffs against exit state, logging any field this pass failed to
-        // restore. The explicit restore calls further down are what actually
-        // restore -- the default Policy::Log does not roll anything back.
+        // ISSUE #895 (fixed): the manual restore below only ever covered
+        // DepthMask/BlendState(enable)/DepthFunc/Cull, so DepthTest, the
+        // blend func factors, ActiveProgram and VAO escaped every frame that
+        // drained a decal -- 2723 GLStateGuard ERROR lines in a couple of
+        // minutes under Policy::Log, which rolls nothing back.
         //
-        // ISSUE #895: it does not restore enough. This reports 8-9 escaped
-        // fields at ERROR level on EVERY frame that drains a decal, on both
-        // rendering paths (DepthTest, the four blend factors, ActiveProgram,
-        // VAO and the two bound texture slots). Note GLStateGuard::Policy
-        // ::Restore exists and DOES roll back via GLStateSnapshot::ApplyCore()
-        // -- an earlier version of this comment said rollback was impossible,
-        // which is probably why this call site never asked for it. Fixing the
-        // leak is #895; check render-pass-published-state.md before switching
-        // policy, since a pass whose outputs are engine-global bindings must
-        // not be wrapped in a restoring guard at all.
-        GLStateGuard guard("DecalRenderPass");
+        // Policy::Restore is safe here, unlike the DDGI/fog pattern
+        // render-pass-published-state.md warns about: GLStateSnapshot::
+        // ApplyCore() restores only the *core* GL subset (depth/blend/
+        // stencil/cull/polygon-mode/viewport/scissor/FBO/program/VAO) and
+        // deliberately never touches per-slot texture bindings. So it fixes
+        // exactly the fields above without undoing this pass's texture
+        // publishes at TEX_POSTPROCESS_DEPTH / TEX_USER_0 -- both of which
+        // are re-published fresh by whichever pass needs them next (every
+        // other TEX_POSTPROCESS_DEPTH consumer -- SSAO, Fog, DOF, MotionBlur,
+        // TAA, ToneMap -- rebinds it before sampling; TEX_USER_0-2 are
+        // documented pass-local-reuse slots), so nothing downstream depends
+        // on this pass's texture bindings surviving. Residual texture diffs
+        // still surface, just at TRACE instead of ERROR, keeping the log
+        // quiet in steady state while staying discoverable for a leak hunt.
+        // Same shape as PlanarReflectionRenderPass / OverdrawRenderPass /
+        // ShaderDebugDrawPass, which replay geometry with their own
+        // programs/VAOs under Policy::Restore.
+        GLStateGuard guard("DecalRenderPass", GLStateGuard::Policy::Restore);
 
         // Helper: decide whether a packet should be drained by *this*
         // (the graph-scheduled) Execute. In the Deferred path the opaque
@@ -304,6 +312,7 @@ namespace OloEngine
         context.SetBlendState(false);
         RenderCommand::SetDepthFunc(RHI::CompareOp::Less);
         RenderCommand::BackCull();
+        CommandDispatch::InvalidateRenderStateCache();
 
         m_SceneFramebuffer->Unbind();
 


drsnuggles8:
Live verification screenshots (Forward and Deferred paths, DecalModeMatrixTest.olo) — sent to the user directly in this session's chat, since gh comments can't attach local files without an upload step. Both show all four decal modes rendering correctly with zero GLStateGuard[DecalRenderPass] ERROR lines in OloEngine.log.

Issue #800, opened 2026-08-15
Vulkan: one VUID-vkCmdDraw-None-09600 per window resize (sampled image UNDEFINED, self-heals next frame)


drsnuggles8:
Found during #691 Phase 9 live verification, on a real editor session log.

## Symptom

Every window resize under `--rhi=vulkan` produces exactly one validation error:

```
vkQueueSubmit2(): pSubmits[0] command buffer ... expects VkImage 0x…
(subresource: aspectMask = VK_IMAGE_ASPECT_COLOR_BIT, mipLevel = 0, arrayLayer = 0)
to be in layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL
-- instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
```

It self-heals on the following frame (the recreated targets get their normal
per-frame transitions), so there is no visible artifact — but it violates the
zero-validation-errors bar Phase 8 set, and an image read in `UNDEFINED` is
formally undefined content, not merely "stale".

## Reproduction (deterministic, 4-for-4 and 9-for-9)

1. Launch the editor with `--rhi=vulkan` (any sandbox scene; VehiclesTest used).
2. Restore / move / maximize the window a few times while the scene renders.
3. `grep VUID-vkCmdDraw-None-09600 OloEditor/OloEngine.log` — one hit per resize
   (capped at 10 by the validation layer's duplicate-message limit).

## What is known

- **Pre-existing on master**, not introduced by #691 Phase 9: first observed
  from a binary built from pristine master before any Phase 9 commit existed.
- The error always follows the mid-frame resize storm in the log
  (`GTAORenderPass: Resized …` → `UICompositeRenderPass: Resized …` →
  `UICompositePass: scene inputHandle=… colorTex=(idx=…, gen=N)` → error), and
  the named `VkImage` differs every time — i.e. it is a freshly created
  attachment from that resize.
- **Not** the heap-descriptor route: Phase 9 closed that gap (the heap path had
  no bind-time transition at all — the debt the Phase 8 issue text recorded as
  "amendment (63) covers the slot path only"), and the symptom is unchanged, so
  the failing sample comes from somewhere else.
- Remaining candidates, in order of suspicion:
  1. **The ImGui viewport image.** `VulkanImGuiBackend::GetTextureID` mints a
     descriptor set baked at `SHADER_READ_ONLY` during the ImGui *build* phase;
[...]

PR #934, opened 2026-08-24
feat(rhi/vulkan): adopt the Vulkan 1.4 host image copy and maintenance5

### LINKED ISSUE #809: RHI/Vulkan: adopt the Vulkan 1.4 conveniences — host image copy, maintenance5/6
Follow-up to #691. The device targets Vulkan 1.4 (ADR 0010's floor) but the
backend only uses what the port needed. Two core-1.4 features would simplify
code this epic just spent significant effort on.

## Host image copy (`VK_EXT_host_image_copy`, core in 1.4)

Upload texture data straight from host memory to an optimally-tiled image — no
staging buffer, no queue submit, no layout dance.

This lands directly on the machinery #801 just split into per-class files: the
staging path, the one-shot submit outside a recording bracket, and the
mid-frame flush rules that several deferred findings in #803 are about. A
load-time texture upload would stop needing the command stream at all, which
removes a whole class of ordering hazard (amendment (72)'s "one-shot submits
are ordered BEFORE the still-recording frame") rather than managing it.

Not a blanket replacement: mid-frame uploads of GPU-written images still need
the command path. Scope it to the load-time/asset path first.

## Maintenance5/6 (core in 1.4)

Smaller, mostly ergonomics: `vkCmdBindIndexBuffer2` with an explicit size (the
facade's `DrawIndexed(va, 0)` "whole buffer" sentinel — amendment (9b) — maps
onto it honestly), plus relaxed pipeline-creation rules.

## Worth checking while here

`VkPhysicalDeviceVulkan14Features` is a single struct; amendment (51)'s rule
applies — once it exists, promoted feature structs must fold into it rather
than being chained separately, or validation rejects the device creation.

## Score
```olo-score
capability: 3
craft: 1
stability: 5
decay: 1
effort: 3
confidence: 0.8
learning: 5
fun: 3
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
The device has targeted Vulkan 1.4 since the port (ADR 0010's floor), but the backend only used what #691 needed. This adopts the two core-1.4 features the issue names, and records the decisions as ADR 0011 amendment (91).

## Host image copy — the load-time upload leaves the command stream

`VulkanTexture2D::UploadPixels` now prefers `vkCopyMemoryToImage`. For a texture with **no mip chain** the whole upload happens on the host: no staging buffer, no command buffer, no submit — so there is nothing left for amendment (72)'s *"one-shot submits are ordered BEFORE the still-recording frame"* to order. That hazard is removed for those uploads rather than managed.

Mip generation is a **blit**, which has no host form, so a mipped upload still submits — it just submits a blit chain whose base level arrived from the host. Both routes now share one `RecordMipChain` with a stated precondition, so their barrier ladders cannot drift.

**The route is confined to load time by two checks, not by intent.** The host commands execute on the CPU the instant they are called — no queue ordering, no fence — so this is a correctness question:

- **no frame may be recording** (the guard `SubImage`'s mid-frame arm already had), and
- **it must be the image's first upload** — because "no frame is recording" says nothing about frames already *submitted*. `VulkanImageInfoRegistry::InitialLayout` is already exactly that record, so nothing new had to be invented. This is what keeps `VideoTexture::UpdateFrame` (per video frame) and `OceanFFTField::Upload` (per tick) on the command path.

**Three capability gates, only one of them device-wide:** the `hostImageCopy` feature bit, the per-format `VK_FORMAT_FEATURE_2_HOST_IMAGE_TRANSFER_BIT` (queried through `VkFormatProperties3` — the bit is above bit 31, so the 32-bit `VkFormatProperties` cannot report it), and the driver's `pCopySrcLayouts` / `pCopyDstLayouts`, of which only `VK_IMAGE_LAYOUT_GENERAL` is spec-guaranteed. The route **declines entirely** if it cannot finish in `SHADER_READ_ONLY_OPTIMAL`, because `GetData` and `SubImage` both name that as their barrier's `oldLayout` — a fast path that quietly changes a resource's resting state is a second contract, not a fast path.

`VK_IMAGE_USAGE_HOST_TRANSFER_BIT` is kept **off framebuffer attachments** via a backend-internal `renderTargetOnly` constructor flag. This matters: `identicalMemoryTypeRequirements` is **`VK_FALSE` on NVIDIA** (measured on this RTX 4090), so the bit may change memory type requirements — and every render target would have paid it, on every resize, for a route it can never take. Gating the *feature* on that property was tried and is wrong: it disables host image copy outright on exactly the hardware it was built for.

## maintenance5 — the index bind states a size it already knew

`vkCmdBindIndexBuffer2` with the buffer's real byte extent. Amendment (9b)'s `DrawIndexed(va, 0)` "whole buffer" sentinel was already resolved to `GetCount()` at the draw sites; the *bind* still said "whole buffer" implicitly. Now an overrunning draw is a validation error naming the bind instead of an out-of-bounds index fetch that renders garbage.

**maintenance6 is deliberately not enabled** — nothing in the backend consumes any of its relaxations, and an unused feature bit is dead weight the validation layer still has to reason about (the rule the EDS3 and mesh-shader bits already follow). The issue named the pair; only the half with a consumer shipped.

## Amendment (51) sweep

`VkPhysicalDeviceVulkan14Features` joins the device chain, so every standalone struct already in it was checked against what 1.4 promoted. Nothing folded — `VK_EXT_descriptor_heap`, `VK_KHR_shader_untyped_pointers`, `VK_EXT_extended_dynamic_state3`, `VK_EXT_device_fault` and `VK_EXT_mesh_shader` are all post-1.4 or never-promoted. Recorded so the *next* promoted feature is known to belong inside the struct.

## A bisect lever

`OLO_VULKAN_NO_HOST_IMAGE_COPY=1` forces every upload back to staging — same genre as `OLO_GAMEPLAY_SCHEDULER_SEQUENTIAL`. Kept permanently because the host route changes *when* an upload happens relative to the queue, and that class of difference is otherwise only answerable by rebuilding the backend. It is what attributed the pre-existing VUIDs below.

---

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/Platform/Vulkan/VulkanTexture.cpp` — the `firstUpload && !frameRecording` gate in `UploadPixels`. This is the correctness core: it is what makes "load-time only" true rather than aspirational. Its first version had only the recording check, which left `VideoTexture`/`OceanFFTField` doing unsynchronised CPU writes to live images. If either condition is too weak, the failure is a race that no test on an idle machine will show.
2. `OloEngine/src/Platform/Vulkan/VulkanTexture.cpp` — `RecordMipChain`'s extraction. It moved mip 0's `TRANSFER_DST→TRANSFER_SRC` transition out to each caller and moved the per-mip transition to *after* the blit. I believe it is barrier-for-barrier equivalent to the old inline ladder, but it is the change most likely to be subtly wrong in a way only sync validation catches.
3. `OloEngine/src/Platform/Vulkan/VulkanDevice.cpp` — the two-call `VkPhysicalDeviceHostImageCopyProperties` query and the `VK_IMAGE_LAYOUT_MAX_ENUM` wildcard handling. A wrong answer here silently changes which layouts the host path believes it may use.

**What I verified, and how**

- **Full suite green: 6584 passed, exit 0** (`OloEngine-Tests.exe`, `build-cached` Debug), run after the review fixes.
- **All 108 Vulkan tests pass** (`--gtest_filter='Vulkan*:*Vulkan*'`), including the 50-test `VulkanPassSuite` and `VulkanDrawPath` which exercises the new index bind. The `VulkanResourceFactory` fixture asserts **zero validation errors in TearDown**, sync validation included — so "it round-tripped" means "it round-tripped legally".
- **The host route is proven *taken*, not merely permitted.** `VulkanTexture2D::GetHostImageCopyUploadCount()` exists because both routes produce byte-identical pixels in the same layout, so no pixel assertion can tell them apart. `HostImageCopyRouteIsActuallyTakenNotJustPermitted` asserts the counter advances on a first upload **and does not advance on a re-upload** — the latter is the assertion that would catch the race in finding 1 coming back.
- **New tests:** `HostImageCopyGateIsSelfConsistent` (asserts both the enabled and disabled directions off one branch, since every machine that can run these has the feature), `HostUploadOfAMippedWidenedTextureRoundTrips` (RGB8→RGBA8 widening *plus* a mip chain — the combination the host arm has to get right), `HostUploadedTextureStillAcceptsASubImageUpdate`, `Maintenance5IsEnabledOnAContractSatisfyingDevice`.
- **Live editor on `--rhi=vulkan`** (RTX 4090, driver 610.88, device API 1.4.341): log confirms `host image copy enabled … maintenance5 enabled`. `MaterialSpheres.olo` renders correctly — full PBR sphere grid with IBL reflections and a legible environment map (picture frames, sofa, patterned floor visible in the reflections), which is textured content sampling correctly through the host-upload path. `olo_shader_errors` returns 0.

**Least confident about**

The `VK_PIPELINE_STAGE_2_HOST_BIT` / `VK_ACCESS_2_HOST_WRITE_BIT` source scope on the mip-0 barrier in the mipped host path. `vkQueueSubmit`'s implicit host-write ordering already covers the host copy, so naming it explicitly should be redundant-but-correct rather than load-bearing — sync validation is clean and the mips read back byte-exact, but I would welcome a second read on whether that is the right spelling.

Second: the `renderTargetOnly` flag catches `VulkanFramebuffer`'s attachments, which is the large population, but any other site that creates a `VulkanTexture2D` purely as a render target still gets the usage bit. I did not audit for such sites beyond the framebuffer.

**Deliberately not tested**

- **`SubImage` was left on the staging path** — it is a partial update with a mid-frame arm of its own, and #809 scopes the host route to the load-time upload. Extending it would need the host-side layout lists to include `SHADER_READ_ONLY_OPTIMAL` as a *source*, which is not spec-guaranteed.
- **The disabled arm of every gate is untested on real hardware** — this box supports host image copy, so the "declines cleanly" paths are only covered by the gate test's assertions, not by execution.
- **Texture2DArray / Cubemap upload paths are untouched.** They are engine-internal GPU-written images, not the asset path.

## Pre-existing issue found, not introduced here

The IBL bake emits **2 image-layout validation errors** (`VUID-vkCmdDraw-None-09600`, `VkImage … expects SHADER_READ_ONLY_OPTIMAL — instead, current layout is TRANSFER_DST_OPTIMAL`, and the converse), alongside `IBLCache: Failed to read face 0 mip 0 for caching`. **These are not from this change**: an A/B on the identical binary with `OLO_VULKAN_NO_HOST_IMAGE_COPY=1` reproduces both, identically, with the host route fully off. Worth its own issue.

Closes #809

## Summary by CodeRabbit

* **New Features**
  * Vulkan 1.4 devices can use host-image-copy for faster initial texture uploads when supported.
  * Texture uploads automatically fall back to the staging-buffer path when host copying is unavailable or unsuitable.
  * Mipmap generation and supported texture formats are handled automatically across both upload paths.
  * Added an option to disable host-image-copy behavior for troubleshooting or compatibility testing.

* **Documentation**
  * Documented Vulkan feature requirements, upload limitations, fallback behavior, and validation coverage.

### DIFF
--- a/OloEngine/src/OloEngine/Core/DebugLevers.inl
+++ b/OloEngine/src/OloEngine/Core/DebugLevers.inl
@@ -45,6 +45,11 @@ OLO_LEVER_TOGGLE(BindlessDescriptorHeap, "OLO_RHI_BINDLESS",
 OLO_LEVER_TOGGLE(VulkanTraceBuffers, "OLO_VK_TRACE_BUFFERS",
                  "Log every Vulkan vertex/index buffer's device-address range at create time — the currency "
                  "for pairing a GPU fault address back to its buffer.")
+OLO_LEVER_TOGGLE(VulkanNoHostImageCopy, "OLO_VULKAN_NO_HOST_IMAGE_COPY",
+                 "Force every Vulkan texture upload back onto the staging buffer + one-shot submit path, "
+                 "disabling the Vulkan 1.4 host-image-copy route (#809). The host route changes WHEN an "
+                 "upload happens relative to the queue, so this is the A/B for attributing a frame or "
+                 "validation difference to it without rebuilding the backend.")
 
 // --- Assets and bakes -------------------------------------------------------
 // TEXT, not a toggle, and that is a correction: the variable NAMES THE OUTPUT
--- a/OloEngine/src/Platform/Vulkan/VulkanDevice.cpp
+++ b/OloEngine/src/Platform/Vulkan/VulkanDevice.cpp
@@ -6,9 +6,10 @@
 #include "Platform/Vulkan/VulkanCapabilities.h"
 #include "Platform/Vulkan/VulkanShader.h"
 
+#include "OloEngine/Core/DebugLevers.h"
+
 #include <algorithm>
 #include <atomic>
-#include <cstdlib>
 #include <mutex>
 #include <optional>
 #include <stdexcept>
@@ -742,8 +743,12 @@ namespace OloEngine
         // permanently because the host route changes WHEN an upload happens
         // relative to the queue, and that class of difference is exactly what
         // a bisect lever is for.
-        const char* const hostCopyOverride = std::getenv("OLO_VULKAN_NO_HOST_IMAGE_COPY");
-        const bool hostCopyForcedOff = hostCopyOverride != nullptr && hostCopyOverride[0] == '1';
+        // Declared in DebugLevers.inl and read through Levers, not through a
+        // raw getenv or even a bare Env::IsTruthy: the registry is what makes
+        // a lever discoverable (it carries the help text and shows up in the
+        // snapshot), and DebugLeversTest fails the build for any engine code
+        // that reads an OLO_* variable around it.
+        const bool hostCopyForcedOff = Levers::VulkanNoHostImageCopy();
         m_HostImageCopyEnabled = wantHostImageCopy && !hostCopyForcedOff && vkCopyMemoryToImage != nullptr &&
                                  vkTransitionImageLayout != nullptr;
         if (hostCopyForcedOff)
--- a/OloEngine/src/Platform/Vulkan/VulkanDevice.cpp
+++ b/OloEngine/src/Platform/Vulkan/VulkanDevice.cpp
@@ -773,10 +773,6 @@ namespace OloEngine
             hostCopyProps.pCopySrcLayouts = m_HostCopySrcLayouts.empty() ? nullptr : m_HostCopySrcLayouts.data();
             hostCopyProps.pCopyDstLayouts = m_HostCopyDstLayouts.empty() ? nullptr : m_HostCopyDstLayouts.data();
             vkGetPhysicalDeviceProperties2(m_PhysicalDevice, &hostCopyProps2);
-            const auto isWildcard = [](VkImageLayout layout)
-            { return layout == VK_IMAGE_LAYOUT_MAX_ENUM; };
-            m_HostCopyAllSrcLayouts = std::ranges::any_of(m_HostCopySrcLayouts, isWildcard);
-            m_HostCopyAllDstLayouts = std::ranges::any_of(m_HostCopyDstLayouts, isWildcard);
             m_HostCopyMemoryTypeNeutral = hostCopyProps.identicalMemoryTypeRequirements == VK_TRUE;
         }
         OLO_CORE_INFO("[Vulkan] 1.4 conveniences: host image copy {}{}, maintenance5 {}",
@@ -915,8 +911,6 @@ namespace OloEngine
         m_Maintenance5Enabled = false;
         m_HostCopySrcLayouts.clear();
         m_HostCopyDstLayouts.clear();
-        m_HostCopyAllSrcLayouts = false;
-        m_HostCopyAllDstLayouts = false;
         m_HostCopyMemoryTypeNeutral = false;
         {
             std::lock_guard<std::mutex> lock(m_HostImageCopyFormatMutex);
@@ -961,7 +955,7 @@ namespace OloEngine
         {
             return false;
         }
-        return m_HostCopyAllSrcLayouts || std::ranges::find(m_HostCopySrcLayouts, layout) != m_HostCopySrcLayouts.end();
+        return std::ranges::find(m_HostCopySrcLayouts, layout) != m_HostCopySrcLayouts.end();
     }
 
     bool VulkanDevice::IsHostCopyDstLayoutSupported(VkImageLayout layout) const
@@ -970,7 +964,7 @@ namespace OloEngine
         {
             return false;
         }
-        return m_HostCopyAllDstLayouts || std::ranges::find(m_HostCopyDstLayouts, layout) != m_HostCopyDstLayouts.end();
+        return std::ranges::find(m_HostCopyDstLayouts, layout) != m_HostCopyDstLayouts.end();
     }
 
     void VulkanDevice::LogDeviceFaultInfo() const
--- a/OloEngine/src/Platform/Vulkan/VulkanDevice.cpp
+++ b/OloEngine/src/Platform/Vulkan/VulkanDevice.cpp
@@ -8,6 +8,8 @@
 
 #include <algorithm>
 #include <atomic>
+#include <cstdlib>
+#include <mutex>
 #include <optional>
 #include <stdexcept>
 #include <string>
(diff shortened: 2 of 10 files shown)


drsnuggles8:
## 🤖 Self-review @ `e719cdb095f83b086003dfbd15b860c26f5f262e`
Reviewed the PR diff at **high** effort before opening.

- **Findings: 3 · all fixed.**
  1. *Correctness (real race).* The host route was taken for **every** `SetData`, not just load-time — `VideoTexture::UpdateFrame` and `OceanFFTField::Upload` re-upload live, currently-sampled textures, and the host commands run on the CPU with no queue ordering or fence. Fixed with two gates (no frame recording **and** first upload, read off `VulkanImageInfoRegistry::InitialLayout`), plus a regression test asserting the route counter does **not** advance on a re-upload.
  2. *Duplication.* `HostUploadTexelBytes` was a functional twin of the existing `VkFormatTexelBytes` in the same anonymous namespace — and one of them guarded a raw host-heap read. Deleted; the existing helper is used.
  3. *Cost on resources that cannot benefit.* `VK_IMAGE_USAGE_HOST_TRANSFER_BIT` landed on every single-sampled colour texture, framebuffer attachments included. Fixed with a backend-internal `renderTargetOnly` constructor flag. **First attempt at this fix was wrong and is worth flagging:** gating on `identicalMemoryTypeRequirements` disabled host image copy outright on this RTX 4090 (the property is `VK_FALSE` on NVIDIA) — caught because the test suite went red, not by reasoning.
- **Dismissed:** none.
- **Re-verified after the fixes:** full suite 6584 passed (exit 0); all 108 Vulkan tests pass with the fixture's zero-validation-errors assertion; live `--rhi=vulkan` editor re-checked.

drsnuggles8:
## 🤖 Self-review @ `74b62a71596082142c35ccd37870e068a03b784a` (round 2)

Round 1 CI reached a complete picture before I pushed: **every job green except SonarCloud.**
`build` ×2, `ASan (Windows/clang-cl)`, `ASan + LSan (Linux)`, `UBSan (Linux)`, `TSan (Linux)`, `Configure + build OLO_WITH_VULKAN=OFF`, pre-commit and the Steamworks-TU gate all **SUCCESS**. Notably TSan is clean, which is the CI-side check on the upload-path synchronisation this PR changes.

**The one real failure, and the fix.** SonarCloud's quality gate failed on `new_reliability_rating` = 4 (threshold 1), from a single `cpp:S990` — `std::getenv` in the new `OLO_VULKAN_NO_HOST_IMAGE_COPY` bisect switch. This repo funnels env reads through `Core/Environment` exactly so there is one S990 suppression and one definition of truthiness.

Worth recording: `Env::IsTruthy` would have satisfied Sonar but **not** the repo's own rule — `DebugLeversTest.NoEngineCodeReadsAnOloVariableOutsideTheRegistry` fails any engine code reading an `OLO_*` variable outside the lever registry, and it caught that locally before this push. The switch is now declared in `DebugLevers.inl` and read via `Levers::VulkanNoHostImageCopy()`, which is its right home anyway — it carries help text and appears in the lever snapshot beside the other RHI switches.

**Re-verified locally at this head:**
- Full suite **6584 passed, exit 0**.
- All **108** Vulkan tests pass; `DebugLevers` + `VulkanResourceFactory` = 20 passed.
- The lever works both ways: default logs `host image copy enabled …`, and with it set logs `host image copy disabled by OLO_VULKAN_NO_HOST_IMAGE_COPY=1` with the route-taken test skipping cleanly (12 → 11 passed).

**One correction to the PR body's "Deliberately not tested".** It said the disabled arm of the gate was untested on real hardware. That is no longer true: the lever exercises exactly that arm on this machine, and I ran it.

**Still outstanding:** CodeRabbit has not reviewed yet — its first pass hit the OSS rate limit ("Review limit reached"), so the current unresolved-thread count of 0 reflects *no review*, not a clean one. I am not treating the review gate as passed.

drsnuggles8:
@coderabbitai review

drsnuggles8:
## 🤖 Self-review @ `a00e3a92f23661bcafd3d08b884b0364cb945090` (round 3 — external review findings)

Three findings, all verified against current code before acting. **Two were mine to have got wrong, and one of those is a correctness bug** the reported finding didn't mention — I found it while checking the reported one.

**1. `IsHostTransitionTargetSupported` accepted the wrong list — FIXED (not in the original report).**
It returned `src || dst`. `VUID-VkHostImageLayoutTransitionInfo-newLayout-09057`, read out of the SDK's `validusage.json` rather than recalled, says `newLayout` must be in **`pCopyDstLayouts`**. A layout present only in `pCopySrcLayouts` is a legal copy *source* and an illegal transition *target*. Every driver at hand reports identical lists — which is precisely why this would have shipped and failed on someone else's.
*Verified the narrowing didn't silently disable the feature*: `HostImageCopyRouteIsActuallyTakenNotJustPermitted` still passes, so `SHADER_READ_ONLY_OPTIMAL` really is in the destination list here.

**2. `VK_IMAGE_LAYOUT_MAX_ENUM` wildcard — VALID, FIXED.** I checked `vk.xml` and `validusage.json`: no such wildcard exists for these lists. I invented it. It was worse than dead code — had a driver ever placed that value in a list, the branch would have reported *every* layout as supported, i.e. fail-open on a safety check. Removed along with `m_HostCopyAllSrc/DstLayouts` and the doc claims built on it.

**3. Duplicate `§16` — VALID, FIXED.** The existing §16 is #800's layout tracker; my section is now §17, and the amendment's cross-reference follows.

**Nothing skipped.**

**Re-verified at this head:** full suite **6584 passed, exit 0**; Vulkan + DebugLevers **116 passed**; the `OLO_VULKAN_NO_HOST_IMAGE_COPY` lever still flips the route both ways.

Both spec facts are now cited to `validusage.json` in the code comments and the amendment, so the next reader can check them the way I should have the first time.

PR #936, opened 2026-08-24
Renderer: sample baked lightmaps on the deferred path (#865)

### LINKED ISSUE #865: Renderer: sample baked lightmaps on the deferred path (no UV2 in the G-Buffer)
Baked lightmaps (issue #439, PR #861) ship on the forward PBR path only. A scene rendering through `RenderingPath::Deferred` resolves its bake normally and then never samples it — **the whole feature is a no-op for that scene**, not a per-draw fallback.

## Why it can't sample today

Deferred shading reads from the G-Buffer, and by the time `DeferredLightingShared.glsl` runs the fragment has no UV2 and no instance identity. `PBR_GBuffer.glsl` passes 5 varyings (worldPos, normal, texCoord, clipPosCurr, clipPosPrev) — no `a_TexCoord2` — and all five render targets are occupied:

| target | format | contents |
|---|---|---|
| 0 | RGBA8 | albedo + metallic |
| 1 | RGBA16F | octahedral normal + roughness + AO |
| 2 | RGBA16F | emissive + flags |
| 3 | RG16F | velocity |
| 4 | RED_INTEGER | entity ID |

Routing is in `Renderer3DMeshSubmission.cpp` (~line 357): a default-PBR material under `RenderingPath::Deferred` gets `s_Data.PBRGBufferShader`, which has no lightmap branch. Note a *custom-shader* material that isn't deferred-capable is rerouted to `ForwardOverlayPass` instead, so it may still sample — the silently-dropped case is the common default-PBR static mesh.

## Current mitigation (already in PR #861)

`SceneLightmapRuntime::WarnIfActivePathCannotSample` logs once per resolved bake when the deferred path is active, so the no-op is visible in `OloEngine.log` instead of looking like a failed bake. That is a signpost, not a fix.

## Options, and the decision this issue needs

1. **A sixth G-Buffer target carrying atlas UV** (RG16F). Simple and orthogonal, but costs bandwidth on *every* deferred frame whether or not any lightmap exists in the scene.
2. **Fold the atlas sample into the G-Buffer pass** and store resulting irradiance. No new UV plumbing downstream, but every channel is taken so it still needs a new target — and the geometry pass would have to pull in the ambient-ladder helpers, which it currently has no business knowing about.
3. **A packed/conditional target** enabled only when the scene has a resolved bake. Avoids the always-on cost, but a runtime-conditional G-Buffer layout interacts with the frame-graph fingerprint cache (see `virtual-shadow-map-page-cache.md` §5 — a `Setup()` that branches on a runtime toggle gets frozen).

Option 3's trap is real and documented, so this is a deliberate architecture call, not a mechanical port.

## Acceptance

- A lightmapped static scene renders identical baked GI on Forward and Deferred (region-mean comparison, not per-pixel).
- The `WarnIfActivePathCannotSample` call and its `#865` reference are removed.
- Measured cost of the chosen option on a scene with **no** lightmap, so the always-on tax is known rather than assumed.

Context: `docs/agent-rules/baked-lightmap-pipeline.md` §6.

## Score
```olo-score
capability: 8
craft: 5
stability: 3
decay: 3
effort: 5
confidence: 0.8
learning: 5
fun: 5
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Baked lightmaps (#439, PR #861) shipped forward-only. A scene rendering through `RenderingPath::Deferred` resolved its bake and then **never sampled it** — the whole feature was a scene-wide no-op, not a per-draw fallback, because the deferred lighting pass shades from the G-Buffer and by then the fragment has neither UV2 nor instance identity.

## The design decision the issue asked for

**A sixth G-Buffer target (RGBA16F) carrying the baked irradiance E in `.rgb` and coverage in `.a`, fetched in the G-Buffer pass** — the last stage that still holds both the UV2 stream and the per-draw `InstanceData::LightmapScaleOffset`. `DeferredLightingShared.glsl` consumes that vec4 at the same ambient-ladder rung, with the same coverage gate and the same helpers, that `include/AmbientLadder.glsl` uses on the forward path.

This is the issue's **option 2, refined**: it stores raw irradiance rather than a shaded ambient term, so the geometry pass needs only one texture fetch (`LightmapSampling.glsl`) and never pulls in the ambient-ladder helpers the issue worried about.

Why the alternatives lose:

- **Not the atlas UV in RT5 (option 1),** despite looking cheaper at RG16F. An MSAA resolve averages colour attachments, and averaging two charts' UVs across a silhouette addresses an *unrelated* atlas texel — a wrong-address bug that renders as a plausible patch of light rather than as anything obviously broken. Averaging irradiance and coverage together is exactly the alpha-weighted blend `sampleLightmapIrradiance` already documents for a bilinear tap straddling a coverage edge. 16F UV precision near 1.0 is also ~1/1024 — several texels on a 2048/4096 atlas — so option 1 is really RG32F and its bandwidth edge mostly evaporates.
- **Not the shaded ambient term (option 2 as literally phrased).** Decals run *after* the G-Buffer pass and rewrite albedo, normal and RMA. An ambient term pre-shaded against the pre-decal albedo would be silently stale under every decal. Storing raw E keeps the ladder — and the albedo it multiplies — in the lighting pass, where the decals have already landed.
- **Not a conditional target (option 3).** The render graph declares the G-Buffer's attachment list as the transient-pool aliasing key (`RenderPipeline.cpp`'s `buildGBufferFramebufferDesc`), so a runtime-conditional layout is the frozen-`Setup()` trap in `virtual-shadow-map-page-cache.md` §5.

## Measured always-on cost

A/B on this branch against **its own base tree** (WIP-commit, check out the base, rebuild, measure), on `VirtualGeometryPerf.VirtualVsClassicFrameBudget`: 15.7M triangles at 1280x720, deferred path, **no lightmap in the scene**, min of 20 `GL_TIME_ELAPSED` samples, three runs each.

| | with RT5 | without RT5 |
|---|---|---|
| classic geometry frame | 6.487 / 6.488 / 6.493 ms | 6.491 / 6.491 / 6.493 ms |
| virtual geometry frame | 2.309 / 2.312 / 2.313 ms | 2.302 / 2.309 / 2.396 ms |

**No measurable frame-time cost** — every delta is smaller than the benchmark's own run-to-run spread, on a scene heavy enough to cover the frame many times over.

The committed `perf_baselines.txt` was deliberately **not** used as the "before": its `virtual_geometry` entry reads 4.63 ms against a measured 2.31 ms, so it has drifted from master through other merged work and is not a clean A/B.

The cost that **is** certain is memory: 8 bytes/pixel — ~16.6 MB at 1920x1080, and again on the single-sample resolve target when MSAA is on (it mirrors the layout).

## Three contracts this adds, each of which fails silently

- **Every G-Buffer writer must write RT5.** An MRT output a fragment shader never assigns is UNDEFINED in that attachment — not zero — and RT5's `.a` is a coverage flag, so undefined there reads as "this pixel has baked GI". A new G-Buffer shader that forgets the line renders correctly on every scene with no bake and wrong on the ones with one. All ten writers now write `vec4(0)`, and `GBufferBakedGIContractTest` scans the shader sources for it.
- **`DecalRenderPass`'s draw-attachment arrays are sized to `GBuffer::Count`.** A short initializer list value-initializes the tail to `0`, which is not "writes nowhere" — it is "writes attachment 0". Every entry is spelled out, and `fullDrawBufs` (the RESTORE set handed to later passes) names RT5 rather than masking it.
- **`ClearAllAttachments` fills every float RT with one colour,** which would clear coverage to 1.0 everywhere the G-Buffer pass never ran. `SceneRenderPass` clears RT5 to `vec4(0)` explicitly — the same reason the velocity RT already has its own clear.

`SceneLightmapRuntime::WarnIfActivePathCannotSample` and its `#865` reference are removed; there is no path-dependent no-op left to warn about.

## Verification

**GPU parity test** — `LightmapVisualEvidenceTest.BakedBleedSurvivesTheDeferredPath` bakes the colour-bleed room once and captures it on Forward and Deferred, comparing **region means** (never per-pixel, per the issue's acceptance criterion). It asserts both halves because they fail differently: the deferred ON-vs-OFF gap catches "the chain is broken anywhere" (UV2 attribute, atlas fetch, target write, sampler bind, rung), and the forward-vs-deferred hue ratio catches "the chain works but addresses the wrong chart". The bake block is factored into a fixture helper so the deferred case cannot drift into baking something subtly different from the forward case it is compared against.

Evidence PNGs (`Lightmap_PathParity_*.png`): Deferred-ON is indistinguishable from Forward-ON — warm pink bleed on the floor and ceiling beside the red wall, green beside the green wall — while Deferred-OFF is flat neutral grey.

**Headless contract tests** — `GBufferBakedGIContractTest` (new, `shaderpipe`): every G-Buffer writer declares and writes RT5; the sampler slot mirrors `ShaderBindingLayout::TEX_GBUFFER_BAKEDGI` in both lighting shaders; `PBR_GBuffer.glsl` includes the sampler, reads UV2 at location 3, writes `sampleLightmapIrradiance(...)` into RT5, and guards its Vulkan pull on `LightmapScaleOffset.x > 0.0` (an unguarded pull on an unbaked static mesh is a device loss, ADR 0011 amendment (89)); and both paths' top ambient rung gates on **coverage**, never on the colour.

**Full suite: 6559 passed, 0 failed.** One pre-existing mirror guard needed updating — `ShaderStageContract.DecalGBufferOutputLocationsMatchTheirDrawAttachmentMaps` pins the decal draw-attachment map literals as source text, and they grow with the layout.

**Live editor** (deferred path, MCP): zero shader errors, clean log, frame renders correctly.

## Backend

No Vulkan mirror needed: `VulkanShader::Reflect`'s deferred-capability scan matches the `o_GBuffer` prefix, and the G-Buffer goes through the generic `Framebuffer` abstraction. Adding the texture slot did move `TEX_SHADER_GRAPH_0` (69 → 70) and therefore `HEAP_IMAGE_SLOT_BASE`, so `OLO_HEAP_IMAGE_BASE` in `BindlessHeap.glsl` and its copy in `BindlessHeapGpuTest.cpp` move with it — pinned by `BindlessShaderPipeline.HeapImageBaseMatchesTheBindingLayout`, which passes.

`docs/agent-rules/baked-lightmap-pipeline.md` §6 is updated (the deferral is now DONE) and a new §7 records the decision, the rejected options and the three contracts.

## Second commit (unrelated, droppable)

`fix(renderer): stop logging our own KHR_debug pass markers` — found while verifying this change in the live editor, kept as its own commit.

`OloEngine.log` was **97.3% one message**: 192,415 of 197,680 lines were `OpenGL debug message (source: APPLICATION, type: , id: 0): <PassName>`. Those are the engine's own `glPushDebugGroup`/`glPopDebugGroup` pass markers echoed back through the debug callback; `OpenGLDebug.cpp`'s `typeStr` switch has no case for `GL_DEBUG_TYPE_PUSH_GROUP`/`POP_GROUP` (hence the empty `type: `), so they fell through to the generic branch at INFO — two lines per pass per frame.

It was costing real frame time, which is why it is a fix rather than a tidy-up: same scene, same deferred path, same MCP-attached state, one sample each — **18 FPS / 55.86 ms before, 198.75 FPS / 5.03 ms after**. After: 4,811 log lines instead of 197,680, zero marker lines, real warnings still surfacing.

Closes #865

## Summary by CodeRabbit

* **New Features**
  * Deferred rendering now supports baked lightmaps, including irradiance and coverage data.
  * Lightmap sampling is supported across Forward, Forward+, and Deferred rendering paths.
  * Added baked-GI frame capture support for debugging and analysis.

* **Bug Fixes**
  * Removed obsolete warnings about unsupported deferred lightmaps.
  * Reduced OpenGL debug-log noise from internal render-pass markers.

* **Tests**
  * Added coverage for baked-GI integration, deferred lightmap visibility, and the expanded G-buffer layout.


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

PR #938, opened 2026-08-24
feat(asset): import OpenVDB volumetric grids as engine-native volumes

### LINKED ISSUE #724: Asset: OpenVDB volumetric import (authored smoke / fire / cloud volumes)

Every volumetric in the engine is procedural — `Renderer/CloudNoise.cpp` generates cloud noise, `FogVolumeComponent` uses analytic shapes, `Precipitation`/`Snow`/`Particle` are simulated. There is no way to import an *authored* volume, which is how smoke, fire and hero cloud assets are actually produced (Houdini/EmberGen/Blender all export `.vdb`). #437 (dynamic fire propagation) and #633 (volumetric clouds) both eventually want this.

An OpenVDB importer producing an engine 3D texture / sparse volume asset: read the grid, resolve its transform and voxel size, resample the sparse tree into a dense or brick-tiled 3D texture, and expose it as an asset a fog-volume or cloud component can reference. Animated sequences (a `.vdb` per frame) as a follow-up slice.

Baking VDBs to 3D texture slices offline in a DCC and importing them as flipbooks — works, but loses the transform/voxel-size metadata and makes iteration painful.

**Additional context**

Reference: `D:\repos\Timberdoodle\src\scene\openvdb_loader.cpp` / `.hpp` (Apache-2.0) — a compact, non-invasive integration worth reading before deciding how much of OpenVDB to vendor. Their `src/scene/tex_compression/sdf_bc1_compressor.cpp` is a related trick for compressing SDF/volume data on the GPU.

Fits the interchange work in #655 (USD/Alembic importer registry) — the `MeshImporterRegistry` pattern there is the natural place to hang a volume importer.

**Acceptance criteria**

- A `.vdb` smoke grid imports and renders through the existing volumetric path with correct world-space scale and orientation.
- Import is asset-cook-friendly (no runtime OpenVDB dependency in the shipped runtime if avoidable).

## Score
```olo-score
capability: 3
craft: 5
stability: 1
decay: 1
effort: 3
confidence: 0.8
learning: 5
fun: 5
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary

- Adds a cook-time **OpenVDB (.vdb) importer**, confined to a new editor-only static lib (`OloEngine-VolumeCook`) so OpenVDB and its dependency chain (blosc/boost/imath/openexr/tbb) never reach the shipped runtime (`OloRuntime`/`OloServer`) — only `OloEditor` and `OloEngine-Tests` link it. The importer resamples a grid's active region into a dense f32 buffer and derives the grid→world transform directly from OpenVDB's own `indexToWorld()`, preserving translation/rotation/voxel-size exactly as authored.
- Adds a new engine-native `.olovol` binary format (versioned, zlib-compressed, CRC32-checked, mirrors `.olmap`) plus `AssetType::Volume` / `VolumeAsset` / `VolumeSerializer` — the reader has **zero OpenVDB dependency** and supports two-phase async loading like textures.
- `FogVolumeComponent` gains a `Texture3D` shape that samples an imported volume within the entity's extents, reusing the existing local-fog-volume UBO/shader plumbing (`VolumetricFogPass` / `FroxelFogScatter.comp` / `FogVolumeCommon.glsl`) rather than a new render pass.
- `Texture3D::SetData` is a new RHI primitive (OpenGL + Vulkan) to upload the density buffer.
- The editor's Content Browser cooks a dropped `.vdb` to `.olovol` via a dedicated "Import as Volume" action (OpenVDB is not a recognized extension in `OloEngine` itself).

Closes #724

## Test plan

- [x] `VolumeImportTest.cpp` — `.olovol` binary format round-trip + corruption rejection (bad magic/checksum/truncation/non-finite/size-mismatch), always compiled; plus (under `OLO_WITH_OPENVDB`) OpenVDB import tests: dimensions/background-value preservation, **grid-transform reproduces `openvdb::math::Transform::indexToWorld()` exactly** at several sample points, downsampling above the resolution cap, missing-grid error handling, grid-name listing.
- [x] `VolumeVDBVisualEvidenceTest.cpp` — full-pipeline visual evidence: a synthetic dense sphere is written through the real `.olovol` codec, loaded as a memory-only `VolumeAsset`, referenced by a `FogVolumeComponent`, and rendered from **side, three-quarter, and top-down** angles, once with the volume off and once on. Differential/golden-free assertions: no black frames, the volume's screen footprint changes measurably when enabled, and a region outside the volume stays stable. See the screenshots below.
- [x] Full `OloEngine-Tests` suite: **6593/6593 passing** (Debug, `dev-cached`, MSVC-triplet + clang-cl toolchain).
- [x] Confirmed via a real build that `OloEngine-VolumeCook` (and OpenVDB) is linked into `OloEditor` and `OloEngine-Tests` only, not `OloRuntime`/`OloServer` — the CMake wiring gates it the same way `httplib::httplib` is confined to `OloEditor`.

### Screenshot evidence

Side view (off → on):

| Off | On |
|---|---|
| off | on |

Top-down (on), showing the bounding-box gizmo the editor now draws for the `Texture3D` shape:

topdown

## Review guide

**Where I'd look hardest**
1. `OloEngine-VolumeCook/src/OpenVDBVolumeCook.cpp` — `DeriveGridTransform()`. It sidesteps OpenVDB's row-vector matrix convention entirely by reading world-space corner positions straight from `indexToWorld()`, rather than transposing OpenVDB's `Mat4d` by hand. `VolumeImportTest.cpp`'s `GridTransformReproducesIndexToWorldExactly` verifies this against OpenVDB's own transform at 5 sample points — I'd want a second pair of eyes on whether that test is actually exercising a non-trivial transform (it uses a uniform scale + translation; it does not cover rotation, since `openvdb::math::Transform::createLinearTransform` doesn't take one directly).
2. `CMakeLists.txt` / `OloEngine-VolumeCook/CMakeLists.txt` / `OloEditor/CMakeLists.txt` / `OloEngine/tests/CMakeLists.txt` — the "no OpenVDB in shipped runtime" wiring. I'd double check the `target_link_libraries` scoping (PUBLIC on the cook target so `blosc_static` and OpenVDB headers reach the test target for fixture authoring, but the cook target itself is never linked onto `OloEngine`/`OloRuntime`/`OloServer`).
3. `OloEditor/assets/shaders/include/FogVolumeCommon.glsl` — two evaluator functions now exist (`evaluateFogVolumesAtPoint` and `evaluateFogVolumesAtPointVDB`), and only the VDB-aware one can see the density texture. A self-review pass caught the plain evaluator falling through to the `BOX` SDF default for a `Texture3D`-shaped volume (rendering it as a solid opaque box) before this was fixed to skip it explicitly — worth checking I didn't miss a third call site.

**What I verified, and how** — named above: the transform math against OpenVDB's own function (unit test), the full render path against real screenshots (visual evidence test, images embedded above), and the build topology by actually building and checking `OloEditor`/`OloEngine-Tests` link `OloEngine-VolumeCook` while `OloRuntime`/`OloServer` do not (their `CMakeLists.txt` has no such line).

**Least confident about** — the froxel fog pass's temporal history: while building the visual evidence test I found that toggling `FogVolumeComponent::m_Enabled` off doesn't immediately clear `VolumetricFogPass`'s scatter-volume history (I had to bump the test's settle-frame count from 2 to 8 to get a clean "off" capture). That's very likely fine in practice — a one-or-two-frame fade when a volume is disabled is normal for a temporally-accumulated effect — but I haven't verified there's no user-visible "ghost" persistence beyond a few frames in the live editor, only in the headless test harness.

**Deliberately not tested**
- `VolumeAsset::GetGridTransform()`'s rotation/off-center-translation component is computed and round-tripped through the `.olovol` format, but **not yet consumed at render time** — `FogVolumeComponent` places the sampled volume using its own entity transform + `m_Extents` (a symmetric half-size), not the source `.vdb`'s full affine map. The editor now seeds `m_Extents` from the imported grid's real dimensions/voxel-size when a volume is dropped onto the component (so the box at least sizes correctly), but a rotated or off-center source grid will still need manual entity-transform tuning. Full transform consumption is a reasonable follow-up rather than in scope here.
- Multiple simultaneous `Texture3D`-shaped fog volumes: only the single highest-priority enabled one binds a texture this frame (documented in `Components.h`); a second one falls back to zero density rather than silently reusing the first's texture.
- Live-editor MCP screenshot verification (`olo_screenshot` / `olo_render_capture_target`) — I don't have an interactive editor session in this environment; the visual evidence above comes from the same full `Renderer3D` pipeline the editor uses (`RendererAttachedTest`'s `RunEditorFrames`), not a live MCP session.


drsnuggles8:
## 🤖 Self-review @ `0b4e5804e`

Reviewed the PR diff at high effort (`/code-review high`) before the first push, plus a focused manual review of the two build-fix commits pushed afterward to drive CI green.

- **Findings from the initial review:** 3, all fixed before the PR was opened — a `Texture3D`-shaped fog volume silently rendering as a solid opaque box through the non-VDB-aware GLSL evaluator (now explicitly skipped there), a zero-extent `FogVolumeComponent::m_Extents` producing NaN/Inf sampling UVW with no CPU or shader-side guard (now clamped + floored, both), and the imported grid's dimensions/voxel-size being computed but never consumed anywhere (now seeds the editor's `m_Extents` on drag-drop).
- **Findings from CI:** 2 real build-topology bugs, both fixed and verified:
  1. `OpenVDB::openvdb`'s vcpkg-shipped `FindOpenVDB.cmake` isn't multi-config aware and silently bound the Debug `.lib` into a Release `OloEditor` build (`_ITERATOR_DEBUG_LEVEL` mismatch) — fixed by setting explicit per-config `IMPORTED_LOCATION_DEBUG/RELEASE` on `OpenVDB::openvdb` and the transitively-pulled `TBB::tbb`.
  2. That fix's first version hardcoded Windows-shaped vcpkg paths and broke every Linux CI job at configure time — gated it to `WIN32 AND CMAKE_CONFIGURATION_TYPES`, the actual surface where the bug exists.
- **Dismissed:** none — every finding from both passes was fixed, not rebutted.
- **Not independently reviewed by CodeRabbit:** it hit its OSS rate limit on the very first push and never posted an actual review (see its comment above) — the zero unresolved review threads reflects "never ran," not "reviewed clean."

CI is now fully green (14/14 checks passing, including a rerun of `TSan (Linux/Clang)` which hit a job timeout mid-suite on its first attempt with every test passing up to that point — infra, not a regression).

drsnuggles8:
@coderabbitai review

PR #940, opened 2026-08-24
fix(renderer): bound the virtual-geometry SW-raster work-list append (#862)

### LINKED ISSUE #862: Renderer/Vulkan: virtual geometry's SOFTWARE raster path device-faults (READ of invalid address) — hardware paths stable

drsnuggles8:
## Summary

Fixes #862 — virtual geometry's software-raster path device-faulted (READ of invalid address) on Vulkan at default settings, while hardware-only configs stayed stable.

**Root cause:** `VirtualClusterCull.comp`'s software-raster work-list append —

```glsl
uint swSlot = atomicAdd(swList.Count, 1);
swList.Records[swSlot] = swRecord;
```

— had **no bounds check**, unlike the reject-list append a few lines below it in the same shader, which has always been guarded with `if (slot < u_RejectCapacity)`. An unbounded SSBO write like this is silently clamped/undefined on GL (never fatal there), but on Vulkan without `robustBufferAccess` it can genuinely device-fault the moment a later shader indexes off a corrupted `InstanceIndex`/`ClusterIndex` record — exactly the `VK_EXT_device_fault` "READ of invalid address" this issue reports.

**Fix:**
- Repurposed the cull UBO's unused padding field (`Pad0`/`_vccPad0` at offset 140) into `SwCapacity`/`u_SwCapacity`, mirroring the existing `RejectCapacity`/`u_RejectCapacity` plumbing.
- `VirtualGeometryPass.cpp` now uploads `cullParams.SwCapacity = frameClusterCount` at both cull-dispatch sites (phase 1 and phase 2) — the same bound the SW-list buffer is already sized to.
- `VirtualClusterCull.comp` now bounds-checks the SW-list append. On the should-never-happen overflow case it falls through to the hardware draw-command path instead of silently dropping the cluster (matching the reject list's own "never lose a cluster" contract).
- Added `VirtualClusterTwoPhaseOcclusion.SwListAllocationHoldsTheWorstCaseSwRecordCount`, mirroring the existing `RejectListAllocationHoldsTheWorstCaseRejectCount` test, proving the buffer's allocation actually holds the bound the shader is now told to trust.

**What this does NOT fix (separate, pre-existing, harmless):** live verification surfaced a one-time `'VirtualClusterRaster_Int64'/'VirtualVisibilityResolve' buffer binding 42 has no published occupant — null block` warning on the very first frame after enabling the SW path. This is `VulkanRendererAPI.cpp`'s existing, deliberate safety net (issue #691) — an unfed binding substitutes a real zero-filled block instead of a null GPU address, specifically to prevent this exact fault class. It's logged once (dedup'd) and never recurs; the same warning fires for the classic mesh-shader path too, which was already proven stable. Not touched here since it's cosmetic and orthogonal to the device fault.

## Verification

Live over MCP against a running Vulkan-backend editor (`VirtualGeometryTest.olo`, deferred renderpath — see `docs/agent-rules/rhi-abstraction-boundary.md` §9-14 and the issue's own bisect table):

| config | before | after |
|---|---|---|
| `swRasterMode: auto` (default), `hwRasterMode: auto` (mesh shaders) | **VK_ERROR_DEVICE_LOST** | stable — 4143 clusters drawn (3539 HW + 604 SW), matches the old `disabled`-mode baseline |
| `swRasterMode: auto`, `forcePortableSwRaster: true` (the portable two-pass variant, the other suspect the issue named) | not tested before (issue's own "next bisect step") | stable — 4152 clusters drawn (3619 HW + 533 SW) |

Ran thousands of frames across both configs with no device loss, no new Vulkan validation errors beyond pre-existing/known ones (the IBL-cache layout VUIDs — see `docs/agent-rules/ibl-bake-preexisting-vulkan-layout-vuids.md`), and `olo_virtual_geometry_stats` reporting healthy, expected numbers throughout.

Headless: `VirtualClusterTwoPhaseOcclusion.*` (7/7), `VulkanBringUp.*` / `VulkanPassSuite.*` / `VulkanDrawPath.*` / `VulkanRenderGraphExecution.*` (62/62) all pass on this box's live GPU context.

## Test plan

- [x] `OloEngine-Tests.exe --gtest_filter=VirtualClusterTwoPhaseOcclusion.*` — 7/7 pass, including the new regression test
- [x] `OloEngine-Tests.exe --gtest_filter=VulkanBringUp.*:VulkanPassSuite.*:VulkanDrawPath.*:VulkanRenderGraphExecution.*` — 62/62 pass
- [x] Live editor over MCP: previously-crashing `swRasterMode: auto` default now stable, both raster variants (int64 single-pass and portable two-pass)
- [x] Checked `OloEngine.log` / Vulkan validation output for new warnings — none beyond pre-existing/known ones

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved virtual geometry occlusion culling to prevent software-raster work-list overflows.
  * Automatically falls back to hardware command generation when the software work list reaches capacity.

* **Tests**
  * Added regression coverage for minimum, exact-fit, and worst-case work-list allocations across multiple frame sizes.

### DIFF
--- a/OloEditor/assets/shaders/compute/VirtualClusterCull.comp
+++ b/OloEditor/assets/shaders/compute/VirtualClusterCull.comp
@@ -196,7 +196,7 @@ layout(std140, binding = 69) uniform VirtualClusterCullParams
     uint  u_ArgsSlotBase;            // 128 — added to the instance index when indexing args[]
     int   u_DebugDrawClusters;       // 132 — cluster-bounds debug bit field, 0 = off (issue #725)
     uint  u_DebugDrawClusterStride;  // 136 — emit only every Nth cluster
-    int   _vccPad0;                  // 140
+    uint  u_SwCapacity;              // 140 — SW-raster work-list record capacity (this frame's cluster count)
     // ---- Culling-camera override (issue #726) ---------------------------
     // A FLAGGED override, not an unconditional replacement like the one in
     // InstanceCullParams: this program is bound by two different views, and the
@@ -455,15 +455,26 @@ void EmitSurvivor(uint instanceIndex, uint clusterIndex, VirtualInstance inst, V
         bool nearSafe = nearDistance > worldRadius * 1.05;
         if (projRadiusPixels < u_SwRasterThresholdPixels && nearSafe)
         {
+            // Bounds-checked exactly like the reject list below (u_RejectCapacity):
+            // an unchecked atomicAdd here overruns swList.Records[] on GL only benignly
+            // (driver-clamped/undefined, never fatal) but device-faults on Vulkan with a
+            // READ of invalid address downstream in VirtualClusterRaster.comp, once the
+            // corrupted InstanceIndex/ClusterIndex indexes far out of range (issue #862).
             uint swSlot = atomicAdd(swList.Count, 1);
-            VisibleCluster swRecord;
-            swRecord.InstanceIndex = instanceIndex;
-            swRecord.ClusterIndex = clusterIndex;
-            swRecord._p0 = 0u;
-            swRecord._p1 = 0u;
-            swList.Records[swSlot] = swRecord;
-            atomicAdd(args[argsIndex].SwCount, 1);
-            return;
+            if (swSlot < u_SwCapacity)
+            {
+                VisibleCluster swRecord;
+                swRecord.InstanceIndex = instanceIndex;
+                swRecord.ClusterIndex = clusterIndex;
+                swRecord._p0 = 0u;
+                swRecord._p1 = 0u;
+                swList.Records[swSlot] = swRecord;
+                atomicAdd(args[argsIndex].SwCount, 1);
+                return;
+            }
+            // Over capacity (should not happen — swListBytes is sized to
+            // this frame's total cluster count): fall through to the
+            // hardware compaction path below so the cluster still draws.
         }
     }
 
--- a/OloEngine/src/OloEngine/Renderer/ShaderBindingLayout.h
+++ b/OloEngine/src/OloEngine/Renderer/ShaderBindingLayout.h
@@ -932,7 +932,7 @@ namespace OloEngine
             u32 ArgsSlotBase;                  // 128
             i32 DebugDrawClusters;             // 132 — bit field, 0 = off (issue #725)
             u32 DebugDrawClusterStride;        // 136
-            i32 Pad0;                          // 140
+            u32 SwCapacity;                    // 140 — SW-raster work-list record capacity (this frame's cluster count)
             // ---- Culling-camera override (issue #726) ---------------------
             // Unlike InstanceCullUBO's unconditional field, this one is a
             // flagged override because THIS shader is bound by two views: the
--- a/OloEngine/src/OloEngine/Renderer/VirtualGeometry/VirtualGeometryPass.cpp
+++ b/OloEngine/src/OloEngine/Renderer/VirtualGeometry/VirtualGeometryPass.cpp
@@ -455,6 +455,7 @@ namespace OloEngine
         cullParams.DebugDrawClusters = static_cast<i32>(m_ClusterBoundsDebugMode);
         cullParams.DebugDrawClusterStride = std::max(m_ClusterBoundsDebugStride, 1u);
         cullParams.RejectCapacity = frameClusterCount;
+        cullParams.SwCapacity = frameClusterCount;
         cullParams.CommandSlotBase = 0u;
         cullParams.ArgsSlotBase = 0u;
         applyCullCameraOverride(cullParams);
@@ -678,6 +679,7 @@ namespace OloEngine
             // statement of intent, not as the mechanism.
             cullParams.DebugDrawClusters = 0;
             cullParams.RejectCapacity = frameClusterCount;
+            cullParams.SwCapacity = frameClusterCount;
             cullParams.CommandSlotBase = frameClusterCount;
             cullParams.ArgsSlotBase = instanceCount;
             applyCullCameraOverride(cullParams);
--- a/OloEngine/tests/Rendering/VirtualClusterCullParityTest.cpp
+++ b/OloEngine/tests/Rendering/VirtualClusterCullParityTest.cpp
@@ -314,3 +314,184 @@ TEST(VirtualClusterCullParity, GpuSelectionMatchesCpuReferenceAtEveryThreshold)
     glBindBufferBase(GL_UNIFORM_BUFFER, ShaderBindingLayout::UBO_CAMERA, 0);
     glUseProgram(0);
 }
+
+// -----------------------------------------------------------------------------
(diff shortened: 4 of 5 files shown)


drsnuggles8:
@coderabbitai review

drsnuggles8:
Addressed the review comments:

**`VirtualClusterCull.comp` — bounded CAS reservation:** verified but not applied. The concern is that the unconditional `atomicAdd(swList.Count, 1)` lets `Count` grow past `u_SwCapacity`. Traced every consumer of `swList.Count`:
- The write itself is already safe — it only happens inside `if (swSlot < u_SwCapacity)`, and `u_SwCapacity` exactly matches the buffer's allocated record count, so every write is in-bounds regardless of how high `Count` climbs.
- `VirtualClusterRaster.comp`'s only read of `Count` is the early-out `if (swRecordIndex >= swList.Count) return;`, and `swRecordIndex`'s range is bounded independently by the CPU-issued dispatch grid (`VirtualGeometryPass.cpp`'s `frameClusterCount`-derived `groupsX/groupsY`), not by `Count`. An inflated `Count` only makes that check *more* permissive for indices already in range — it can't cause an out-of-bounds read.
- Nothing else (no CPU readback, no indirect-dispatch argument) treats `Count` as an array-size driver.

So `Count` exceeding capacity is a purely cosmetic property with no path to memory unsafety — the actual defect class #862 was about (an OOB write) is already closed by the existing guard. I also checked the reject-list append a few lines below (`atomicAdd(rejectList.Count, 1u); if (slot < u_RejectCapacity) {...}`), which is the pattern the SW-list guard was deliberately modeled on — it's already shipped with the identical "unconditional add, guarded write" shape, not a CAS. Adding CAS only to the SW list would create an unjustified asymmetry with that already-reviewed sibling; doing both would widen this PR's scope well past the device-fault fix. Left as-is.

Did add real GPU coverage for the guard itself, since that was the valid part of the finding — see `VirtualClusterCullParity.SwListAppendNeverWritesPastAnUndersizedCapacity` (new commit): it calibrates the eligible-cluster count, then dispatches the actual compute shader against a `Records[]` buffer sized to half that, forcing the guard to reject more than half the appends, and asserts the accepted count matches capacity exactly, the excess falls through to the hardware path, no GL error is raised, and every written record is in-range.

Also checked: my fix touches **only** `VirtualClusterCull.comp`, which has no `#ifdef VIRTUAL_RASTER_INT64` branching at all (that split exists solely in `VirtualClusterRaster.comp`'s downstream *raster* variants, not the cull shader). So there's no "portable vs int64" distinction to cover for this particular change — the two raster shaders are unaffected by it. The PR description's live-verification table does already cover both raster variants for completeness (`forcePortableSwRaster: true/false`), since that's the bisect the original issue called for.

**`ShaderBindingLayout.h` — rename `SwCapacity` to `m_SwCapacity`:** not applied — this field is `UBOStructures::VirtualClusterCullUBO`, a GPU-mirror struct that byte-mirrors a GLSL `uniform` block (`VirtualClusterCullParams` in `VirtualClusterCull.comp`). Per `CLAUDE.md`'s documented convention, GPU-mirror structs deliberately use bare `PascalCase` so each field reads one-for-one against its GLSL twin — `m_`-prefixing is for engine member variables, not this struct family. Every neighboring field in this exact struct (`RejectCapacity`, `CommandSlotBase`, `ArgsSlotBase`, `DebugDrawClusterStride`, ...) is unprefixed; `m_SwCapacity` would be the only `m_`-prefixed field in the entire file and would break the one-to-one naming match with the GLSL side (`u_SwCapacity`). Left as-is.

Issue #944, opened 2026-08-25
Docked-panel hit-test rect desyncs from its paint area after a new dock split (root cause of #921)


drsnuggles8:
Split from #921 per that issue's guidance ("it's fine to fix one and file the other as a scored follow-up issue"). #921 shipped the diagnostic this issue's investigation needed (`olo_input_inject`'s `after.hoveredWindow` / `hoveredId` / `activeId`, read straight off ImGui's `g.HoveredWindow` / `g.HoveredId` / `g.ActiveId`); this issue is the root cause that diagnostic exposed.

## Symptom

A panel docked into a **newly created split** (dragged there mid-session — confirmed reproducible via `olo_input_inject`'s synthetic drag, not yet confirmed with a real mouse drag) ends up with an ImGui hit-test rect **narrower than the area it visually paints**. A click on the visually-correct position for a control near the far edge of the panel silently lands on whatever sibling panel's rect still (incorrectly) extends into that space, instead of the panel you're looking at.

## Repro (via MCP, in this worktree's environment)

1. Attach to a live editor (`run-oloengine` skill, `driver.ps1 -Action attach -AllowWrites`).
2. Open the Terrain Editor panel (`Window ▸ Terrain Editor`) while it's floating (undocked). Confirm its radio buttons (`Edit Mode: None / Generate / Sculpt / Paint / Erosion`) respond correctly to `olo_input_inject` clicks — they do.
3. Drag the panel's title bar onto an existing dock tab bar (or the top of the dockspace) to dock it into a **new** split — a single `olo_input_inject` `drag` action from the panel's title bar to the target works for this.
4. Re-test the same radio buttons. Reproducible finding: `olo_input_inject` `move` probes across the row at the panel's own Y coordinate show `after.hoveredWindow` correctly naming the panel for the LEFT portion of its width, then **silently switching to a completely different, sibling panel's name** (e.g. `"Scene Hierarchy"`) partway across — at a screen X that, per `imgui.ini`, matches almost exactly the sibling's own column width. The switch boundary does not move with the newly-docked panel's actual (wider) visual paint area.
5. This state also **survives a full editor restart** (imgui.ini persists the corrupted layout) — it is not an artifact of the live drag alone.

## What this means concretely

[...]

Issue #854, opened 2026-08-20
MCP: olo_input_inject drag reports ok, does nothing, and latches input state for the rest of the session


drsnuggles8:
Split out of #607, where it was logged as a comment during #634's visual verification. It is a **bug with a repro**, not a missing capability, so it does not belong in that tracker's capability-gap log.

## Symptom

Driving the editor over MCP-on-HTTP, `olo_input_inject` with a `drag` action between two small targets inside a floating ImGui panel:

- returns **`ok: true`** with 22 frames injected,
- produces **no effect** — in the observed case, dragging between two ~9 px node pins in the Visual Script Editor created no wire,
- and, worse, **every subsequent `click` in the same editor session also stops having any effect**, while continuing to return `ok: true`. A restart cleared it.

Single `click` actions on menus and list rows worked reliably both before the `drag` and in sessions where no `drag` was issued.

## Why this is worse than it looks

The tool reports success for work it did not do, and then **silently invalidates every later injection in the session**. An agent driving the editor has no way to tell "the click landed and the UI did not respond" from "the click never happened" — so any verification performed after a `drag` is silently untrustworthy, including verifications of something completely unrelated. That is the failure class this repo treats as most expensive: not a crash, but a confident wrong answer.

It also cost real time once already: the #634 panel check degraded into a pixel-hunt partly because of this.

## Repro

Easy, per the original report:

1. Launch the editor with the MCP diagnostics server (`run-oloengine` skill → `attach`).
2. Open the **Visual Script Editor** panel.
3. `olo_input_inject` a `drag` from one node pin to another (~9 px targets).
4. Observe `ok: true`, no wire created.
5. Issue any `click` that previously worked — it now also has no effect, and still returns `ok: true`.

## Suspected area

Some input state is left latched after a `drag` plan drains. `SyntheticInput` carries both an absolute cursor override and (since the 2026-08-09 batch on #607) a persistent relative offset that deliberately outlives the plan — the absolute override wins outright while set. Worth checking whether a `drag` leaves the absolute override latched, which would explain both halves: the drag itself not registering as motion ImGui accepts, and every later click landing at a stale overridden position.

`SyntheticInputTest` already pins the absolute path's "sums to zero however long it is held" property, so it is a natural place to add the regression case once the mechanism is confirmed.

## Acceptance criteria

- A `drag` between two small targets inside a floating panel either performs the drag, or fails loudly — never `ok: true` with no effect.
[...]

PR #946, opened 2026-08-25
feat(mcp): olo_input_inject exposes ImGui hover/active state (#921)

### LINKED ISSUE #921: MCP: olo_input_inject clicks land but do not activate widgets inside docked panels (menu bar works, panels do not)
Split out of #607 per that tracker's policy: this is a **bug**, not a missing capability, so it does not belong in the capability-gap log. Sibling of #854 (`drag` reports ok / latches state), which was **fixed 2026-08-22 in PR #876** — both observations below are from sessions *after* that fix, so this is a separate defect, not a #854 regression.

## Symptom

Driving the editor over MCP-on-HTTP, `olo_input_inject` `click` actions **land on the menu bar but not on widgets inside docked panels**:

- **#717 (GPU terrain picking), 2026-08-23** — repeated clicks on the Terrain Editor panel's **Edit Mode** radio buttons, and on the panel's own **dock tab**, never took. The reply reported `landed: true` and the coordinates were confirmed inside the 1280x720 client rect. The widget simply did not react, so `TerrainEditorPanel::IsActive()` stayed `false` and the editor never entered the picking path at all. Menu-bar items at the same DPI, in the same call shape, in the same session, *did* react.
- **#866 (Vulkan lightmap sampling), 2026-08-22** — landing a reliable click on the **main-menu-bar `BeginMenu` popup items** (Build > Bake Lightmaps) did not pan out either, after a non-trivial amount of investigation. Coordinate space and/or click timing against `BeginMenu` suspected; root cause unconfirmed. The bake was never triggered, and that verification fell back to the headless GL `LightmapBleedRoom` test.

So the reliable surface today is *menu-bar items*; docked-panel widgets are not, and menu **popups** are at best inconsistent.

## Why it matters

`olo_input_inject` exists so a session can prove an editor-facing code path actually fires. Today that only demonstrably works for one widget class, which undercuts the tool's purpose — and the failure is the expensive shape this repo keeps hitting: **`ok: true`, `landed: true`, no effect.** The session cannot tell "the click landed and the widget ignored it" from "the click never reached ImGui", so it burns rounds on coordinate derivation for a problem that is not coordinates.

Concrete cost already paid: #717's editor-facing acceptance half ("the brush cursor lands under the mouse in a live session") could not be verified over MCP at all, and #866's bake verification was downgraded to a delivery-path check on an unbaked scene.

## Repro sketch

1. Launch the editor with the MCP diagnostics server (`run-oloengine` skill, `driver.ps1 -Action attach`).
2. Open the **Terrain Editor** panel (it docks on the right).
3. `olo_input_inject` a `click` on one of its **Edit Mode** radio buttons, coordinates taken from a `driver.ps1 -Action shot` PrintWindow capture and converted to window-logical pixels.
4. Reply comes back `ok: true`, `landed: true`, `after.cursor` at the requested position — and `TerrainEditorPanel::IsActive()` is still `false`.
5. Same session, click a **menu-bar** item: it reacts.

## Suspected area — unconfirmed, and #854's lesson applies

#854's stated suspicion turned out **not** to be the cause, so treat the following as leads to eliminate rather than a diagnosis:

- The plan builder already does the right-looking thing: `BuildPlan` emits `MousePos` then `AppendIdleFrames(s_MoveSettleFrames - 1)` before the press (`OloEditor/src/MCP/McpInputInject.h`), so ImGui gets hover frames ahead of the button-down. That rules out the most obvious "pressed before hovered" explanation.
- Cursor **position** is verifiably correct — `SampleMcpCursorLanding` (added by #876) reads it back after `NewFrame` and reported `landed: true`. So the break is *between a correct `io.MousePos` and ImGui activating the item*, not in the coordinates.
- Worth checking next: what ImGui thinks is the **hovered window / hovered item** at that moment (dockspace hit-testing, a `ImGuiConfigFlags_ViewportsEnable` viewport-routing mismatch when the editor window is unfocused, or an overlay/dockspace host window swallowing the hover). The menu bar living in the host window while the panel lives in a dock node is the most obvious axis separating the working case from the broken one.
- The `BeginMenu` popup case may be a different defect again (popups are their own ImGui window, created the frame *after* the parent item activates), so do not assume one fix covers both.

## Acceptance criteria

- A `click` on a widget inside a docked panel (a radio button, a dock tab) activates it, or the tool **fails loudly** — never `ok: true` / `landed: true` with no effect.
- `olo_input_inject` reports enough to tell the two apart: alongside `landed`, expose what ImGui had hovered/active at the injected position (hovered window name and/or hovered item id) in the `after` block, so a session can distinguish "reached the widget, widget ignored it" from "never reached the widget".
- A two-step menu interaction (`BeginMenu` parent then a popup item) either works, or is documented as unsupported with the reason.
- A regression test pins whichever behaviour the fix establishes — `SyntheticInputTest` is the existing home for this family.

## Score

```olo-score
capability: 3
craft: 1
stability: 3
decay: 1
effort: 3
confidence: 0.6
learning: 3
fun: 2
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary

`olo_input_inject`'s `cursorLanding` check (#854) only proves the injected cursor *position* took — it says nothing about whether ImGui's hit-test then routed that position to a widget. That's exactly the gap #921 reports: a docked-panel click reports `ok: true` / `landed: true` and the widget never reacts, with no way to distinguish "reached the widget, widget ignored it" from "never reached anything ImGui knows about".

- `after.hoveredWindow` / `hoveredId` / `activeId` close it, read straight off ImGui's own `g.HoveredWindow` / `g.HoveredId` / `g.ActiveId`.
- Latched at the end of `EditorLayer::OnImGuiRender()` — after every panel has drawn for the frame, the same point `ButtonBehavior` itself resolves against — while an injected plan is active or in its post-plan settle window. A live read taken when the tool reports back (after the plan has drained, possibly after the cursor has already been handed back to "nowhere") answers a different, useless question.
- Getting the capture point right took two iterations, both empirically verified live: capturing inside `DrainMcpInputQueue` (before `NewFrame`) always read the *previous* frame's resolution; capturing after the plan and countdown drain read stale/absent hover. The current point — end of `OnImGuiRender`, guarded on the plan still being active or settling — was confirmed against a known-good control (a menu-bar click correctly reports `hoveredWindow: "##MainMenuBar"`) before trusting it against the actual bug.

## What the live diagnostic actually found

Using the now-working diagnostic to drive the investigation (not just add it), I traced #921's mechanism: a panel docked into a **newly created split** ends up with an ImGui hit-test rect **narrower than the area it visually paints**. Concretely, on the Terrain Editor panel docked into a new top-level split:

```jsonc
// x well inside the panel's visually-full-width strip, y on the "Edit Mode" row
// olo_input_inject { "action": "move", "x": 246, "y": 90, "space": "window" }
{ "after": { "hoveredWindow": "Terrain Editor", "hoveredId": 0 } }   // reaches the panel, misses every control

// x further right, still visually well inside the same panel
// olo_input_inject { "action": "move", "x": 280, "y": 90, "space": "window" }
{ "after": { "hoveredWindow": "Scene Hierarchy", "hoveredId": 1603940588 } }  // a SIBLING panel's rect wins instead
```

The switch boundary (~x=270) matches the sibling panel's own dock-column width almost exactly, and survives a full editor restart (the corrupted layout persists in `imgui.ini`) — so this is a genuine ImGui docking-layout bug, not an artifact of the injection path itself. Filed separately as #944, since it needs a live-GL-context editor test fixture to pin properly (out of scope for the headless `McpInputInjectTest`/`SyntheticInputTest` home this issue named), and it isn't yet confirmed whether a real (non-synthetic) mouse drag reproduces the identical corruption.

## Review guide

**Where I'd look hardest**
1. `OloEditor/src/EditorLayer.cpp` (the `OnImGuiRender()` capture site, ~line 2829) — the guard condition (`!m_McpInputQueue.empty() || m_McpCursorRestoreCountdown > 0`) is the one place this PR's whole value depends on; get the timing wrong and the diagnostic silently goes back to reporting nothing useful, the way two earlier iterations did before I found this one.
2. `OloEditor/src/MCP/McpToolsInput.cpp` — the intermediate `stateJson` round-trip. The first version of this fix set the fields correctly in `EditorLayer::GetMcpInputState()` but never wired them through this JSON hop, so they silently reset to null/0 every time — caught only by adding a temporary `OLO_CORE_WARN` and checking `OloEngine.log` against a live editor. Worth double-checking the key names (`imguiHoveredWindow`/`imguiHoveredId`/`imguiActiveId`) don't collide with the existing entity-hover `hoveredId` key in the same object.

**What I verified, and how**
- Unit tests: `McpInputInjectTest.*Hovered*` (3 new tests) plus the full `McpInputInject*`/`SyntheticInputTest.*` suite (56 tests) — `OloEngine-Tests.exe --gtest_filter=...`, all pass.
- Live, over MCP, against a running editor (`run-oloengine` skill `attach`): confirmed `hoveredWindow: "##MainMenuBar"` on a menu-bar click (known-good control), then used the diagnostic to trace the actual docked-panel failure mechanism described above — screenshots and raw MCP responses captured throughout the investigation.
- Confirmed the fix required BOTH the `EditorLayer.cpp` capture-point relocation AND the `McpToolsInput.cpp` JSON round-trip fix — verified each was broken independently (first attempt: correct capture point, but always-null in the reply; traced to the missing JSON keys) and both together produce the correct live result.

**Least confident about** — whether the "narrower hit-test rect after a new split" root cause (#944) is specific to synthetic/interpolated drag sequences or reproduces identically with a real mouse. I could not test a real drag from this session; #944's acceptance criteria calls that out explicitly.

**Deliberately not tested** — a live-GL-context regression test for the #944 root cause itself; that needs an editor-level fixture (`RendererAttachedTest`-shaped, not the headless `McpInputInjectTest`/`SyntheticInputTest` home this issue named), tracked as part of #944 rather than blocking this PR.

Closes #921

## Summary by CodeRabbit

* **New Features**
  * Enhanced input-injection diagnostics with hovered window, hovered item, and active item details.
  * Responses now distinguish between missed windows, missed widgets, and successfully targeted widgets.

* **Documentation**
  * Added guidance and examples for interpreting input hit-test results, including dock-layout mismatches.

* **Tests**
  * Added coverage for window-level, widget-level, and successful input targeting scenarios.

### DIFF
--- a/OloEditor/src/EditorLayer.cpp
+++ b/OloEditor/src/EditorLayer.cpp
@@ -13,6 +13,7 @@
 #include "OloEngine/Core/SyntheticInput.h"
 
 #include <backends/imgui_impl_glfw.h>
+#include <imgui_internal.h> // GImGui->HoveredWindow/HoveredId/ActiveId for GetMcpInputState (issue #921)
 #include <GLFW/glfw3.h>
 #include <stb_image/stb_image_write.h>
 
@@ -1214,6 +1215,15 @@ namespace OloEngine
         state.CursorAskedY = m_McpCursorAskedY;
         state.CursorLandedX = m_McpCursorLandedX;
         state.CursorLandedY = m_McpCursorLandedY;
+
+        // What ImGui's hit-test resolved on the last frame the plan actually ran
+        // (issue #921) — latched by DrainMcpInputQueue every tick, not read live
+        // here. A live read at THIS point (after the plan drained, and possibly
+        // after the cursor has already been handed back to "nowhere") answers a
+        // different question than the one this field exists to answer.
+        state.HoveredWindowName = m_McpHoveredWindowName;
+        state.HoveredId = m_McpHoveredId;
+        state.ActiveId = m_McpActiveId;
         return state;
     }
 
@@ -2802,6 +2812,25 @@ namespace OloEngine
         if (m_EditorPreferencesPanel.OnImGuiRender(m_Prefs))
         {
             ApplyPreferences();
+        }
+
+        // Latch what ImGui's hit-test resolved THIS frame (issue #921), while an
+        // injected plan is in flight (or just drained — one extra frame past empty
+        // so the plan's LAST tick, where a press/release actually lands, is never
+        // missed). g.HoveredWindow/HoveredId/ActiveId are computed once per frame in
+        // NewFrame and then USED by every widget's ButtonBehavior during THIS same
+        // frame's Begin/End calls above — so reading them here, after every panel
+        // has run, is what makes the value match what ImGui itself actually decided.
+        // Reading them from DrainMcpInputQueue (tried first) reads the PREVIOUS
+        // frame's resolution, since that runs before NewFrame; reading them from the
+        // tool's post-hoc GetInputState() call reads whatever is hovered at report
+        // time, which can be long after the plan (and its cursor) are gone. Neither
+        // answers "what did ImGui do with the click" — this does.
+        if (::GImGui != nullptr && (!m_McpInputQueue.empty() || m_McpCursorRestoreCountdown > 0))
+        {
+            m_McpHoveredWindowName = ::GImGui->HoveredWindow != nullptr ? ::GImGui->HoveredWindow->Name : std::string{};
+            m_McpHoveredId = ::GImGui->HoveredId;
+            m_McpActiveId = ::GImGui->ActiveId;
         }
     }
 
--- a/OloEditor/src/EditorLayer.h
+++ b/OloEditor/src/EditorLayer.h
@@ -510,6 +510,17 @@ namespace OloEngine
         f32 m_McpCursorAskedY = 0.0f;
         f32 m_McpCursorLandedX = 0.0f;
         f32 m_McpCursorLandedY = 0.0f;
+        // What ImGui's hit-test resolved on the LAST frame an injected plan actually
+        // ran (issue #921), captured at the end of OnImGuiRender (after every panel
+        // has run its Begin/widgets for the frame) while a plan is in flight. A live
+        // read taken when the tool reports back — after the plan has drained and the
+        // cursor may already have been handed back to "nowhere" — is stale by
+        // construction: g.HoveredWindow answers "what's under the cursor NOW", and by
+        // then there may BE no cursor. Latching it during the plan is what makes the
+        // diagnostic answer the question it's meant to answer.
+        std::string m_McpHoveredWindowName;
+        u32 m_McpHoveredId = 0;
+        u32 m_McpActiveId = 0;
 
         // Undo/Redo
         CommandHistory m_CommandHistory;
--- a/OloEditor/src/MCP/McpInputInject.h
+++ b/OloEditor/src/MCP/McpInputInject.h
@@ -1113,6 +1113,14 @@ namespace OloEngine::MCP::InputInject
                                            { "landedY", state.CursorLandedY },
                                            { "landed", landedRaw } };
         }
+        // What ImGui's hit-test resolved at the moment of this snapshot (issue #921):
+        // the difference between "reached the widget, widget ignored it" (a hovered
+        // window/id that names the widget you clicked) and "never reached the widget
+        // at all" (empty/unrelated hoveredWindow), which the cursor-landing check
+        // above cannot tell apart on its own.
+        after["hoveredWindow"] = state.HoveredWindowName.empty() ? Json(nullptr) : Json(state.HoveredWindowName);
+        after["hoveredId"] = state.HoveredId;
+        after["activeId"] = state.ActiveId;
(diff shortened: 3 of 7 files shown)


drsnuggles8:
Update: #944 (the docked-panel hit-test-rect root cause this PR's diagnostic traced) has been closed as not reproducible under normal conditions — re-verification on an idle machine could not reproduce the corruption across three attempts; the original repro almost certainly hit a load-induced race while 10-17 concurrent compiler processes from sibling worktrees had the editor's FPS crashing to 5-15 during the multi-frame synthetic drag. See #944 for the full write-up. This PR's actual deliverable — the `after.hoveredWindow`/`hoveredId`/`activeId` diagnostic — is unaffected and remains verified working.

drsnuggles8:
@coderabbitai review

PR #961, opened 2026-08-25
test(renderer): cover RGBA16F lightmap arrays across backends

### LINKED ISSUE #951: Lightmap atlas GPU upload: RGBA16F texture arrays and the paged sampling path are unverified on Vulkan (and untested on both backends)
#868 (PR #947) made the baked lightmap atlas a `Texture2DArray` with one layer per page, and moved the f32 -> half conversion to the CPU (`glm::packHalf1x16`) because `Texture2DArray::SetLayerData`'s contract is native-data-per-format, unlike `Texture2D::SetData` which accepts `GL_FLOAT` for an RGBA16F target. That upload and the `sampler2DArray` sampling path are verified **only on OpenGL**.

Two gaps, and the first is wider than "the Vulkan arm is untested".

## 1. RGBA16F texture arrays are first-of-kind, on BOTH backends

Formats actually instantiated as `Texture2DArray` in the engine today:

| format | users |
|---|---|
| `DEPTH_COMPONENT32F` | 3 (shadow atlas / CSM) |
| `RGBA8` | 2 (terrain layer arrays) |
| `RGBA32F` | 1 (ocean FFT ping-pong) |
| `RGBA32UI`, `BC7` | 1 each (terrain VT cache) |
| **`RGBA16F`** | **2 — both added by #947** (the lightmap atlas and its 1x1x1 placeholder) |

And the only `SetLayerData` call in the whole test tree is `TerrainVTCompressBC7Test.cpp:178-179`, which uploads `RGBA32UI` staging data. So **`SetLayerData` with `RGBA16F` has no test coverage on either backend** — the GL path is exercised in production only through #947's own GPU evidence test, and the Vulkan path is exercised by nothing at all.

The half pack is the specific risk: it is a CPU-side loop the engine did not previously need, and a wrong rounding mode, a wrong element stride, or an endianness assumption is **silent** — the frame renders, with subtly or badly wrong baked GI.

## 2. The paged sampling path is GL-only verified

`LightmapVisualEvidenceTest.BakedBleedSurvivesAMultiPageAtlas` (#947) proves per-entity GI is correct across four pages, but it runs on the GL `RendererAttachedTest` harness. Untested on Vulkan:

- the descriptor minted for the array at `TEX_LIGHTMAP` (slot 16) via `HeapBinding::PublishTextureOffsetAndBind`, now that the shader declares `sampler2DArray`;
- `NullSamplerKind::Texture2DArray` for the unpublished-slot fallback (#947 changed this from the `Texture2D` default — an unpublished slot previously minted a 2D null view, which is a type mismatch against the shader);
- `VulkanTexture2DArray::SetLayerData`'s mid-frame vs one-shot upload split for a layer count > 1.

Note the shape of the #866 precedent: Vulkan lightmap sampling looked like a pure no-op-needed change and still device-lost live, on an unbounded vertex-pull read that no test covered. The pattern is that the Vulkan lightmap path keeps failing on the thing nobody thought needed a test.

## Not blocked on new infrastructure

This is deliberately **not** a "we cannot test Vulkan" issue. `VulkanPassSuiteTest.cpp` already runs 20+ real passes through the Vulkan backend and reads results back, and `VulkanResourceFactoryTest` / `VulkanDrawPathTest` exist alongside it. The work is wiring cases into harnesses that are already there, not building one.

(The editor `driver.ps1 -Action attach` path genuinely cannot launch a Vulkan editor, so live MCP inspection is not the route — but it is also not needed.)

## Acceptance

- A backend-agnostic round-trip test for `Texture2DArray` `RGBA16F`: upload known half-encoded layers, read back, assert the values survive per layer. Runs on whichever backend the binary was built for, skipping cleanly with no device.
- A Vulkan case covering the paged lightmap atlas end to end — at minimum that the array is created, all layers upload, and slot 16 samples the layer the encoded region names (the wrong-page failure is a plausible-looking frame, not an error).
- If the CPU half pack turns out to differ from the driver conversion the pre-#947 `Texture2D` path used, say so with numbers and decide deliberately whether it matters — the lightmap contracts are region means, so a 1-ULP difference is invisible to them by construction.

## Worth checking first

Whether the Vulkan test jobs on CI actually have a device, or whether every `VulkanPassSuite` case skips there. If they skip, this is a dev-workstation-verified task like the perf suite, and the issue should say so rather than implying CI coverage it will not get.

Context: `docs/agent-rules/baked-lightmap-pipeline.md` §8 (the encoding and the GPU-side traps), `docs/agent-rules/rhi-abstraction-boundary.md` §9-§14.

## Score
```olo-score
capability: 1
craft: 1
stability: 5
decay: 1
effort: 2
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
## What and why

- Adds one backend-neutral RGBA16F `Texture2DArray` upload/readback contract and invokes it on OpenGL and Vulkan through the public renderer facade.
- Adds a real Vulkan paged baked-lightmap shader sample through `TEX_LIGHTMAP` (slot 16), covering pages 0/1/2, one-shot uploads, a live-recording layer upload, and the typed `Texture2DArray` null fallback.
- Measures CPU `glm::packHalf1x16` output against the legacy OpenGL driver float-to-RGBA16F conversion across edge, midpoint, signed, and representative lightmap values.
- No production fix was required: the new coverage demonstrated that the current transport, sampling, live upload, and typed-null paths satisfy the contract.

Closes #951

## Validation

Local workstation: NVIDIA GeForce RTX 4090, driver 610.88.

- Locked Debug build of `OloEngine-Tests`: passed.
- Focused feature tests: 5/5 passed, 0 skipped.
- Exact Vulkan-to-OpenGL order regression after lifecycle hardening: 28/28 passed.
- Affected contract batch: 98/98 passed.
- Full suite: 6,693 passed, 26 unrelated/environmental skips, 0 failed (6,719 total).
- `pre-commit run --all-files`: all hooks passed.
- Final independent Standards review: clean.
- Final independent Spec review: clean.

Conversion measurement on NVIDIA OpenGL 4.6.0 / driver 610.88:

- compared half words: 104
- differing half words: 0
- maximum difference: 0 ULP

Decision: retain CPU half conversion. It is bit-identical on the measured driver; the test permits at most 1 ULP, which is immaterial to the region-mean lightmap contract.

Vulkan was exercised locally on the same RTX 4090 (API 1.4.341, driver 610.88) with no skips. Ordinary GitHub-hosted PR runners do not guarantee usable Vulkan hardware. The scheduled/manual AMD GPU workflow includes the full suite on RX 5600 XT hardware, but currently does not explicitly fail on Vulkan-specific skips, so cross-vendor Vulkan execution remains a CI visibility limitation.

## Review guide

Please focus on:

1. Whole-array RGBA16F readback extent and exact layer/texel/channel attribution in the shared contract.
2. Vulkan `SetLayerData` coverage split between the one-shot path and the live command-buffer path.
3. Descriptor-heap ownership during GL/Vulkan facade swaps, slot-16 typed-array publication, and transparent-black null sampling.

Least certain area: CI enforcement of real Vulkan execution on AMD, not the locally exercised NVIDIA behavior.

Deliberately not tested locally: a live Vulkan editor/MCP session and non-NVIDIA hardware.

## Summary by CodeRabbit

* **Tests**
  * Added cross-backend validation for RGBA16F texture-array uploads and readbacks.
  * Added coverage for layer-specific data, edge cases, midpoint values, and representative inputs.
  * Added lightmap sampling tests, including mid-frame uploads and transparent-black fallback behavior.
  * Improved diagnostics with detailed mismatch reporting and graphics hardware metadata.

### DIFF
--- a/OloEditor/assets/shaders/tests/LightmapPagedSampling.glsl
+++ b/OloEditor/assets/shaders/tests/LightmapPagedSampling.glsl
@@ -1,0 +1,53 @@
+#type vertex
+#version 460 core
+
+#ifdef OLO_VULKAN
+layout(std430, binding = 57) readonly buffer OloVertexPull
+{
+    float v[];
+} b_Vertices;
+
+layout(location = 0) out vec2 v_TexCoord;
+
+void main()
+{
+    int base = gl_VertexIndex * 5;
+    vec3 position = vec3(b_Vertices.v[base + 0], b_Vertices.v[base + 1], b_Vertices.v[base + 2]);
+    v_TexCoord = vec2(b_Vertices.v[base + 3], b_Vertices.v[base + 4]);
+    gl_Position = vec4(position, 1.0);
+}
+#else
+layout(location = 0) in vec3 a_Position;
+layout(location = 1) in vec2 a_TexCoord;
+
+layout(location = 0) out vec2 v_TexCoord;
+
+void main()
+{
+    v_TexCoord = a_TexCoord;
+    gl_Position = vec4(a_Position, 1.0);
+}
+#endif
+
+#type fragment
+#version 460 core
+
+layout(location = 0) out vec4 o_Color;
+layout(location = 0) in vec2 v_TexCoord;
+
+#include "include/LightmapSampling.glsl"
+
+layout(std140, binding = 2) uniform TestLightmapRegions
+{
+    vec4 u_FirstRegion;
+    vec4 u_SecondRegion;
+    vec4 u_ThirdRegion;
+};
+
+void main()
+{
+    vec4 region = v_TexCoord.x < (1.0 / 3.0)
+        ? u_FirstRegion
+        : (v_TexCoord.x < (2.0 / 3.0) ? u_SecondRegion : u_ThirdRegion);
+    o_Color = sampleLightmapIrradiance(vec2(0.5), region);
+}
--- a/OloEngine/tests/CMakeLists.txt
+++ b/OloEngine/tests/CMakeLists.txt
@@ -445,6 +445,7 @@ add_executable(OloEngine-Tests
 		Rendering/PropertyTests/ModelLoadDeterminismTest.cpp
 		Rendering/PropertyTests/DrawIndexedRawOffsetTest.cpp
 		Rendering/PropertyTests/TextureSaveRoundTripTest.cpp
+		Rendering/PropertyTests/Texture2DArrayRoundTripTest.cpp
 		Rendering/PropertyTests/RHIHandleNativeIdentityTest.cpp
 		Rendering/PropertyTests/TextureInPlaceReloadTest.cpp
 		Streaming/SceneStreamingTest.cpp
--- a/OloEngine/tests/Rendering/PropertyTests/Texture2DArrayRoundTripTest.cpp
+++ b/OloEngine/tests/Rendering/PropertyTests/Texture2DArrayRoundTripTest.cpp
@@ -1,0 +1,122 @@
+#include "OloEnginePCH.h"
+
+#include "PropertyTests/RenderPropertyTest.h"
+#include "Rendering/Texture2DArrayRoundTrip.h"
+
+#include "OloEngine/Renderer/RendererAPI.h"
+#include "OloEngine/Renderer/Texture.h"
+
+#include <gtest/gtest.h>
+#include <glad/gl.h>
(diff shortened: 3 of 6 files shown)


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

PR #966, opened 2026-08-26
fix(renderer2d): match Vulkan UI composite MRT contract

### LINKED ISSUE #957: Vulkan: Renderer2D shaders write unused MRT outputs in single-attachment passes
## What

Vulkan validation reports that the Renderer2D_Quad and Renderer2D_Text fragment stages write MRT outputs that the active dynamic-rendering pass does not provide:

- location 1, o_EntityID
- location 2, o_ViewNormal

The draw has no VkRenderingInfo::pColorAttachments[1] or [2], so those writes are unused. The messages repeat until Vulkan's duplicate-message limit is reached.

This was observed while validating #955 and is deliberately tracked separately because it is unrelated to the Drift foliage/G-buffer contracts fixed there.

## Reproduction

1. Launch OloEditor with --rhi=vulkan and validation enabled.
2. Open Scenes/Drift.olo (the warning also appears during ordinary editor/Renderer2D drawing).
3. Inspect the Vulkan validation output for draws bound to Renderer2D_Quad and Renderer2D_Text.

## Expected

Each Renderer2D pipeline variant exposes fragment outputs that match the color attachments supplied by its pass, without unused-output validation messages.

## Evidence

A clean rebuilt-binary run on 2026-08-25 reproduced both location-1 and location-2 messages for both shaders. The separate #955 target messages were absent in the same run, confirming this is an independent contract.

## Scope

Fix the owning Renderer2D shader/pipeline variant contract and add reflection or pipeline-contract coverage where practical. Do not suppress Vulkan validation.

## Score
```olo-score
capability: 3
craft: 1
stability: 5
decay: 1
effort: 1
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary
- keep MRT locations 0–2 active while UIComposite invokes its Renderer2D callback
- mask the entity-ID and normal attachments during UI drawing so UI stays non-pickable and does not modify normal data
- add Vulkan SPIR-V reflection coverage for every shared Renderer2D batch shader and the production UIComposite attachment contract

Closes #957

## Verification
- `ShaderStageContract.Renderer2DOutputsMatchUiCompositeCallbackDrawAttachments` — passed
- `VulkanPassSuite.UiCompositeClearsBlitsAndBlendsTheOverlayCallback` — fails before the callback at an existing producer-to-UIComposite background-sampling synchronization barrier (`WRITE_AFTER_WRITE`; source is incorrectly `TRANSFER_WRITE` after a color attachment store). Reproduced twice; not changed in this scoped fix.

## Review guide
- The risky seam is the UIComposite callback: it must expose all three Renderer2D fragment outputs to Vulkan dynamic rendering while masking only the auxiliary attachments.
- Review `UICompositeRenderPass.cpp` for map/mask restoration after the callback.
- Review `ShaderStageContractTest.cpp` for reflected output locations/types and the production framebuffer contract.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved UI overlay rendering so color output remains correct while entity and normal data are excluded from the composite.
  - Ensured render targets are restored properly after UI rendering.

- **Tests**
  - Added coverage verifying that Renderer2D shaders and UI composite attachments use compatible output mappings and data formats.

### DIFF
--- a/OloEngine/src/OloEngine/Renderer/Passes/UICompositeRenderPass.cpp
+++ b/OloEngine/src/OloEngine/Renderer/Passes/UICompositeRenderPass.cpp
@@ -10,6 +10,7 @@
 #include "OloEngine/Renderer/ResourceHandle.h"
 #include "OloEngine/Renderer/RenderPipelineBuilderInternal.h"
 
+#include <array>
 #include <span>
 
 namespace OloEngine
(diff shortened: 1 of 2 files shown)


drsnuggles8:
@coderabbitai review

Issue #953, opened 2026-08-25
Drift: scattered single-pixel black specks in the sky above the horizon


drsnuggles8:
## What

Drift's sky carries **scattered single-pixel black specks** in the band just above the horizon — dozens of them, isolated, pure black against a smooth sky gradient. From a low camera near the water (which is where the game's chase camera sits) they read as dirt on the lens or dead pixels.

## Repro

1. Open `Scenes/Drift.olo`.
2. Put the camera low and near the water looking at the horizon — e.g. position `(0, 4, -262)`, target `(0, 3, -235)`.
3. Look at the sky band immediately above the horizon.

Seen first in Play at the dusk squall leg, then reproduced in edit mode. Counted programmatically: **31–52 isolated dark specks** per frame depending on view.

## What is established

- They are **already in `SceneColor`**, i.e. present before any post-processing, and are near-black there (RGB ≈ `1,3,0`).
- At those pixels the depth buffer reads the **far plane** (`device = 1.0`) and `SceneEntityID` reads **-1** — so no geometry rasterised there.
- They are **world/view-locked, not screen-locked**: move the camera and they appear at different screen coordinates. So not dead texels in a render target.
- They occupy the **near-horizon band only**. Looking up at open sky from a higher camera shows none.
- At 8× magnification they are 1–2 px, pure black, with no falloff into the surrounding gradient — the signature of isolated bad values rather than a shading gradient.

## What was ruled out (each verified to actually take effect)

The frame-level RMSE is quoted because a toggle that silently does nothing would otherwise look like a successful elimination.

| suspect | test | frame changed? | specks |
|---|---|---|---|
| Precipitation | weather `Storm` → `Clear` (only Storm enables it) | yes | unchanged |
| Volumetric clouds | `CloudscapeComponent.Enabled = false` | yes, RMSE 1.64 (max 129) | unchanged, identical positions |
| Sky cubemap texels | `CubemapResolution` 256 → 1024 (forces a re-bake) | yes, RMSE 2.86 (max 128) | unchanged, identical positions and values |

The cubemap one is the informative negative: a 4× change to the bake resolution visibly changed the sky and did **not** move a single speck, so they are not bad texels in the sky cubemap.

Rain was also independently unlikely — `PostProcess_Precipitation.glsl` colours every particle type bright (`0.85,0.9,1.0` and similar), so it cannot produce black.

## Not established

[...]

PR #986, opened 2026-08-31
feat(renderer): band-limited three-cascade FFT ocean (#969)

### LINKED ISSUE #969: Renderer: add band-limited multi-cascade FFT ocean fields

## Problem

Olo's FFT ocean owns a single patch size and one displacement/normal field. That constrains broad swell, mid-scale wave structure, and close chop to one repeating frequency domain; it is a likely source of the weak near-to-horizon coherence visible in Drift water captures.

## Proposal

Extend the existing GPU FFT field to a fixed three-band preset before exposing artist controls:

- use disjoint, band-limited frequency ranges for broad, mid, and fine waves;
- choose non-commensurate tile sizes and rotate one sampling domain to avoid phase-aligned repetition;
- derive cascade resolutions from each band's shortest wavelength rather than running every chain at a global maximum;
- sum the fields consistently in vertex/tessellation shading, foam, the CPU wave sampler, and buoyancy;
- retain the current one-cascade path as a tested fallback during rollout.

Threepp's `Ocean` is a concrete reference for the band partition, non-commensurate tile selection, and resolution derivation. Its Vulkan/BLAS architecture is not a porting target.

## Evidence

- Threepp: `include/threepp/objects/Ocean.hpp`, `src/threepp/objects/Ocean.cpp`, and `src/threepp/renderers/vulkan/shaders/ocean_cascade.glsl`.
- Olo: `OloEngine/src/OloEngine/Renderer/Ocean/OceanFFTField.*`, `OceanFFTGpu.*`, `OloEditor/assets/shaders/include/WaterVertexStage.glsl`, and `WaterSurface.*`.
- Olo's existing future-water design already identifies cascaded FFT as the natural follow-on: `docs/design/WATER_FUTURE_IMPROVEMENTS.md` §1.3.

## Acceptance

- A controlled Drift capture sequence at near, mid, and horizon distances demonstrates stable broad swell plus close chop without obvious tile repetition.
- Adjacent frequency bands have no discontinuity or double-counted energy at their handoff.
- CPU/GPU summed height parity and buoyancy parity are covered for all enabled cascades.
- Generated-field resource lifetime, render-graph transitions, and OpenGL/Vulkan bindings are tested.
- Existing one-cascade scenes retain their current authored behavior unless they opt into the preset.

## Non-goals

- Do not expose a large set of cascade knobs before the fixed preset is visually proven.
- Do not introduce ray-tracing-specific water code.

## Score
```olo-score
capability: 5
craft: 8
stability: 3
decay: 1
effort: 8
confidence: 0.5
learning: 13
fun: 13
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #969.

Extends the Tessendorf FFT ocean from one tile carrying the whole spectrum to a
fixed **three-band, band-limited preset** — broad swell for the horizon, the
authored tile in the middle, close chop under the bow — opt-in per water surface
and off by default, so no existing scene changes.

## The preset

`m_FFTCascades` is 1 (the pre-#969 field) or 3 (the preset). Everything else
derives from `m_FFTPatchSize` / `m_FFTResolution`, which the scene already
authors: the issue's non-goal is a knob surface before the preset is visually
proven. The full design and the reasoning behind each constant is at the top of
[`Ocean/OceanCascades.h`](OloEngine/src/OloEngine/Renderer/Ocean/OceanCascades.h);
the four load-bearing decisions:

| decision | what it buys |
|---|---|
| The authored patch size becomes the **mid** band, with broad and fine tiles derived either side | an author keeps the wave scale they tuned; the preset adds an octave in each direction |
| Band boundaries sit at the **next tile's fundamental** (2π/L\_{i+1}), as half-open ranges | no gap and no double-counted energy at the handoff — every wave vector belongs to exactly one band |
| Tile ratios (6.29, 4.43) are deliberately **non-commensurate**, and the mid band's sampling domain is **rotated** (with its wind rotated into that frame to match) | the three lattices share neither a period nor an axis. At Drift's 140 m patch that is an 881 m / 140 m / 31.6 m set whose lattices first re-align at **6.2 km** — past any sea this engine draws |
| Per-band resolutions derive from each band's **shortest wavelength** | the bounded bands need 64 where the fine band needs the authored resolution |

The three fields are **layers of one `Texture2DArray` pair**, so the preset costs
the same two engine texture slots the single-cascade field did, and the
single-cascade path is a one-layer array through the same code. One binding pair,
one shader loop, one CPU sampler — the alternative is two mirrors of the same
contract.

## Two things worth a reviewer's attention

**Every chain runs at the array resolution, not at its own derived one**, because
a texture array's layers must share a size. That costs GPU time and nothing else:
a band-limited spectrum on a larger grid is the same spectrum with the extra bins
zero, and its inverse FFT is the exact band-limited reconstruction of the smaller
one. That is a claim, so it is pinned rather than asserted in a comment —
`OceanCascadeTest.DerivedResolutionReproducesTheArrayResolutionField`. The
alternative (one sampler pair per cascade) would spend four engine texture slots
and shift the shader-graph user base to save a fraction of a millisecond.

**The `Texture2D` → `Texture2DArray` move silently dropped the REPEAT wrap.**
Every cascade layer is periodic over its own patch size and the shader addresses
it with an unbounded `worldXZ * (1/L)`; `Texture2DArray` defaults to
`CLAMP_TO_EDGE` (right for a shadow cascade, wrong here), which would have
rendered one tile of sea and smeared its border across the rest of the frame. The
old `Texture2D` got REPEAT for free from the GL texture default — an inherited
invariant a container swap drops with nothing to notice it. Fixed by a
`RepeatWrap` flag honoured on **both** backends, and pinned on the texture object
(`OceanFFTGpuContractTest.CascadeArraysTileRatherThanClamp`).

## One sum per side, not five

The summation contract — including which SPACE its argument is in, which is the
half a same-point parity test cannot check — is written once in
`OceanCascades.h`, and evaluated in exactly two production places:

- `OceanFFTField::SampleCascades` (CPU — what buoyancy floats on)
- `include/OceanCascadeCommon.glsl::sampleOceanCascades` (GPU), **included** by
  the water vertex, tess-eval and fragment stages rather than copied into each,
  with the texture fetches behind a hook so one walk serves three stages'
  sampler declarations.

`OceanCascadeTest.EveryWaterStageSumsTheCascadesThroughTheSharedFunction` is a
text test that keeps it that way: a C++ test cannot see a shader stage growing
its own second copy of the walk.

Slopes are summed, never normals — normals do not add, and a shading normal built
from an average of three unit normals describes none of the three surfaces the
geometry actually has.

## A third thing, which the tests could not have told me

**A spectrum is a DENSITY, and a second grid makes its bin spacing
load-bearing.** Tessendorf sets `|h0(k)| = sqrt(Phi(k))` with no bin-area factor
— sound with one grid, because the missing constant is absorbed into the
amplitude scale. Across grids it is not: the fine band's tile is 4.43x smaller,
so its k-lattice is 4.43x *coarser* and each of its bins stands for 19.6x more of
the spectrum, while carrying the same amplitude.

The signature is what makes it dangerous:

| quantity | single cascade | three cascades (unfixed) |
|---|---|---|
| summed height RMS | 0.167 m | 0.170 m — **preserved** |
| summed **slope** RMS | 0.0425 | 0.0262 — **38% lost** |

The amplitude normalisation is defined on height, so it held height constant
while the energy drained out of the short waves. Every assertion passed and the
captures showed a mirror-smooth sea you could see the sky reflected in. Found by
looking at the pixels, then measuring which statistic had moved; written up as
[water-shading-nyquist.md §7](docs/agent-rules/water-shading-nyquist.md) and
pinned by `OceanCascadeTest.DiagBandEnergyAndSlopeAtDriftSettings`, which fails
if the preset costs more than 30% of the slope the author tuned.

The general lesson, past water: **when a change redistributes a signal across
scales, write the acceptance on the DERIVATIVE — the value is what the
normalisation is holding constant.**

## Verification

- **Full suite: 6833 tests, 6806 passed, 0 failures.** `OloEngine.log` clean of
  shader compile/link errors.
- `ShaderCompilation.AllProductionShadersCompileUnderVulkanTarget` and
  `VulkanPassSuite.WaterTessellatedPipelineBuildsAndRasterizesAPatchDraw` both
  pass, so the new `sampler2DArray` declarations are SPIR-V-clean and build a
  real Vulkan pipeline — the issue's "OpenGL/Vulkan bindings are tested" bar.
- `ShaderUBOSizeConsistency.CrossShaderUBOMemberOffsetsAgree` keeps the five
  identical GLSL `WaterParams` declarations honest about the added `vec4`.
- **CPU/GPU summed parity is exactly 0** on height, horizontal displacement and
  foam (`OceanFFTGpuContractTest.SummedHeightMatchesBetweenTheCpuSamplerAndTheCascadeTextures`),
  evaluated on the read-back texels the shader actually samples rather than on
  the CPU field. The foam channel was 0.006 out at first — the CPU interpolated
  the Jacobian then clamped, while the texture stores clamped foam and
  interpolates that, and clamping does not commute with interpolation. Fixed,
  not tolerated.
- New tests: `OceanCascadeTest` (19), `OceanCascadeVisualEvidenceTest` (2),
  `+4` GPU contract, `+2` buoyancy, `+1` round-trip.
- Captures at near/mid/horizon for both modes are committed under
  `OloEditor/assets/tests/visual/OceanCascade_*.png`. I looked at them, twice —
  the first set is what caught the density bug.

## Notes / follow-ups

- **Drift still ships with `UseFFT: false`.** That is a measured decision recorded
  in the scene (the plateau A/B), and flipping it is a feel change this PR does
  not make on its own. The acceptance captures use Drift's authored water
  settings so the evidence answers "what would Drift look like"; the flip is a
  separate call once someone has looked at them.
- The CPU physics proxy is now evaluated **per band** (three ≤64² IFFTs per tick
  instead of one) — the price of buoyancy following the summed surface rather
  than one band of it.
- `OceanFFTGpu`'s three compute programs are now shared across instances (a
  three-band field owns three of them, and they all run the same code).

---

## Follow-up commits: running it in Drift

The three-band preset was reviewed in the real Drift scene over MCP, not just in
the synthetic evidence scene. That turned up a performance problem that mostly
**pre-dates this branch**, and a cloud bug that has nothing to do with water.

### The FFT ocean rebuilt its whole spectrum every frame

`DriftWeatherDirector.lua` eases the sea state every tick, writing
`fftWindSpeed` and `fftAmplitude` with a continuously-changing value. Both were
in the h0 cache key, which is a bit-exact compare — so "regenerate when the key
changes" meant "regenerate always", and each regeneration ran a
full-resolution CPU inverse FFT per band purely to measure an RMS.

Gameplay tick, Play mode, chase camera behind the boat (Debug):

| mode | before | after |
|---|---|---|
| Gerstner (what Drift ships) | 5.15 ms | 5.07 ms |
| FFT, 1 cascade | 43.40 ms | **8.82 ms** |
| FFT, 3 cascades | 110.8 ms | **14.44 ms** |

GPU cost was flat at ~10.1 ms across all three the whole time — the three FFT
chains really are free, as designed. Single-cascade FFT being 8x Gerstner is
almost certainly why Drift ships with `UseFFT: false`.

Three exact fixes, no behaviour change: amplitude out of the regeneration key
(it is a linear scale, so it is a multiply); the reference RMS from **Parseval**
instead of eight inverse FFTs; and the remaining continuous params quantised in
the *key only*. Plus the per-band physics proxy grid, which is this branch's own
contribution to the cost. Each is pinned by a new test.

### Clouds composited in front of terrain (#987)

Reported while reviewing the captures; unrelated to water (identical across all
three water modes). Measured against the depth buffer: **0** cloud texels over
terrain in the raymarch output, **6** in the resolved buffer — the temporal
resolve was adding them, because its 3x3 neighbourhood clamp legitimately
contains sky cloud at a ridge, so history was never clamped away. Fixed by
treating "the ray hit geometry" as an answer rather than a missing sample.
After: resolved equals raw exactly.

I tried two other cloud fixes first (a conservative depth reduction for the
half-res march, and a rework of the composite's bilateral weights) and
**reverted both** — neither showed a clean measured benefit, and a half-understood
shader edit is worse than none. #987 stays open for the residual silhouette-width
effect.

### Not fixed: the black sky specks (#953)

Confirmed still present on master and independent of the water path (24 isolated
specks with `UseFFT: false`, 24 with the three-band preset). #953 is closed as
completed but both its closing PRs are in this branch's history. Evidence posted
to the issue. I could not get a stable repro in the current session — the count
is strongly viewport/projection dependent (1 speck at the issue's documented
pose here, against 31-52 originally) — so I have not attempted a fix.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in three-band FFT ocean preset for broader, finer, and mid-scale wave detail.
  * Added a Cascades setting to the water inspector and Lua scripting interface.
  * Improved ocean surface sampling for rendering and buoyancy across all enabled cascades.
  * Added texture-array support for efficient layered ocean data and repeating wave patterns.
* **Bug Fixes**
  * Preserved legacy single-cascade behavior by default.
  * Added backward-compatible saving and loading for cascade settings.
* **Tests**
  * Added coverage for cascade rendering, sampling, serialization, and visual output.


drsnuggles8:
@coderabbitai review

Issue #433, opened 2026-06-28
Foliage: impostor/billboard LOD for distant trees & dense vegetation


drsnuggles8:
## Problem
Foliage LOD is distance **fade-only** (`FadeStartDistance` in `Terrain/Foliage/FoliageLayer`). There are no impostors or billboard swaps, and the general mesh `LODGroup` (`Renderer/LOD`) isn't applied to instanced foliage. Distant forests render full-triangle meshes.

## Why it matters (Tier 2)
Dense vegetation at long view distances (the Far Cry look) blows up triangle count without impostors. Standard high-impact LOD technique the engine is missing.

## Proposed approach (compute-friendly bake)
- Octahedral / hemi-octahedral impostor baking: render each tree mesh to an atlas of view-angle slices (albedo + normal + depth), baked offline or in a compute/render pre-pass.
- At runtime, swap instanced mesh LODs for a camera-facing impostor card beyond a distance threshold; cross-fade into the existing alpha fade.
- Hook into the `FoliageRenderer` instance path + the existing `Renderer/LOD` machinery.

## Acceptance criteria
- A dense forest holds its triangle budget at distance via impostors; smooth transition (no pop) into mesh LOD up close.
- Impostors receive sun/shadow and wind consistently enough not to read as billboards at gameplay distances.

---
**Tier 2 — performance ceiling.**

## Score
```olo-score
capability: 3
craft: 5
stability: 1
decay: 1
effort: 5
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
Implemented in #666 (branch `feature/foliage-impostor-lod-433`).

**Octahedral impostor LOD for distant foliage** — each foliage layer's mesh is baked into an octahedral atlas (albedo+coverage / object-normal+card-depth) and instances render as view-dependent camera-facing impostor cards: 3-tile blended (no slice pop), depth-parallaxed, re-lit from the baked normal (dynamic sun), wind-swayed, with a distance detail ramp. Enable per layer with `UseImpostor` + a `MeshPath`.

**Acceptance criteria:**
[...]

PR #988, opened 2026-08-31
fix(renderer): foliage impostor cards render at terrain-local coordinates (#953)

### LINKED ISSUE #953: Drift: scattered single-pixel black specks in the sky above the horizon

drsnuggles8:
Fixes #953.

## What it actually was

The "black speckles in the sky" are **pine impostor cards rendered at terrain-local coordinates**.

`FoliageRenderer::GenerateInstances` derives instance positions straight from the heightfield — x/z from tile coordinates starting at 0, y from `GetHeightAt() * heightScale` with no base offset — so they live in the same space the terrain mesh is authored in, and become world positions only through the owning terrain's transform. `CommandDispatch::DrawFoliageLayer` uploads that transform as the shaders' single `u_Model` entry, and `Foliage_Instance.glsl` has always multiplied through it.

`Foliage_Impostor.glsl` did not:

```glsl
vec3 instWorld = a_PositionScale.xyz - u_RenderOrigin;
```

No island in Drift sits at the origin, so all six islands' pines rendered in one heap over open water near (0,0,0), hanging above anything the terrain can reach. From the boat that is a swarm of dark dots in the sky — which is why this was reported as a sky artefact and chased through the atmosphere, cloud and water paths first.

The shader's comment justified skipping `u_Model` by the out-of-bounds hazard from #433. That hazard is about *indexing* `instances[gl_InstanceIndex]` when foliage uploads only one entry; `OLO_INSTANCE_SINGLE` — already defined in this shader — pins the read to `instances[0]` and makes `u_Model` safe. The two got conflated, and the transform was dropped with them.

**Second site, same cause:** `FoliageRenderer::RenderShadows` uploaded plain identity. It is live (`ShadowRenderPass` drives it straight off the renderer) and has no command packet to carry the transform, so foliage shadows were cast from that same origin heap.

## Evidence

Measured over MCP on `Scenes/Drift.olo` at render 5248x2517, counting isolated dark pixels in the sky band. Same cameras, same resolution, only the pivot expression toggled between runs:

| camera | before | after |
|---|---|---|
| horizon-island | 34 | **0** |
| origin-look | 18 | **0** |
| boat-chase | 62 | **1** |
| high-wide | 0 | 0 |

The remaining one is foliage (0 with every `FoliageComponent` disabled, 1 restored) sitting 1–9 px above the island's ridge line — a tree on the skyline, which is scenery rather than a defect.

## Hypotheses falsified first

The issue had already ruled geometry out because depth reads the far plane and `SceneEntityID` reads -1 — sound for opaque geometry, but foliage writes neither, so that inference pointed away from the real owner. Working at native resolution with controls (`olo_screenshot` downscales to 1024 wide and smears a one-pixel speck into its neighbours, so `olo_render_probe_pixel` was the only usable instrument), these were each tested and **falsified**:

- **bad mip / undefined derivatives** from the `texture()` calls inside the non-quad-uniform `continue` — `textureQueryLod` = 0.00, all three taps run
- **UV clamped to a tile border** — mid-tile at (0.45, 0.72)
- **corrupt impostor atlas** — contents match `grass.png` x the authored `BaseColor` exactly
- **viewport parity / resolution** — resolution only decides whether the sub-pixel card clears the alpha cutoff at all, which is why a small window shows zero specks and "cannot reproduce" there means "the window was too small"

The pass and shader were then identified positively rather than by elimination: `SceneColor` after `ScenePass` = [1.31, 0.83, 0.09] (sky) vs after `FoliagePass` = [0.006, 0.014, 0.001], and marking the two foliage shaders in distinct colours made the speck read pure impostor with terrain and sky controls untouched.

## Second commit: the card was also ~8x too small

Independent defect found on the way. The card never read the per-instance height:

```glsl
float radius = u_ImpostorParams1.y * scale;
```

`u_ImpostorParams1.y` is the object-space radius `ImpostorBaker` framed the mesh with, and the foliage meshes are authored unit-height (`pine.obj` spans y in [0,1], radius 0.560). The near path supplies the world size itself (`localPos.y *= height * scale`, height 9–16 m for the Pines layer); the impostor left that factor out. The arithmetic closes on the live frame — a probed card reported radius 0.745 m, and 0.745 / 0.560 = 1.33, inside the layer's MinScale 0.9 / MaxScale 1.4.

**Uniform rather than anisotropic, decided by looking at the frame.** Matching the near quad's footprint exactly reproduces its own 1:16 stretch and renders the baked pine as a needle — the skyline grew in 83 of 1024 columns, as isolated one-column spikes. The near quad is deliberately a different thing (a tufted billboard; the impostor is what "gives the tree line a 3D silhouette from any azimuth at range", #433), so the card scales uniformly by `height * scale`: the drawn tree is exactly that tall, nothing pops vertically across the transition, and it keeps its proportions. The same skyline that had no trees on it before now shows recognisable conifer silhouettes.

## Tests

New `FoliageImpostorPlacementTest` (3 tests, `unit`) asserts both foliage vertex stages place the shared instance stream the same way — the two-mirrors-drift shape, since they consume one stream and must agree on what it means. Source-text rather than a render: a GPU test only catches this in a scene whose terrain is off the origin, and the visual result is easy to read as art. **All three fail on the pre-fix shader and pass on this one.**

Targeted run (`*Foliage*:*Impostor*:*Terrain*:*Shadow*`): 365 passed, 1 pre-existing skip, 0 failures.

## Not verified

- The shadow-path change rests on it using the same matrix as the main pass — `olo_render_capture_target` returns "no GPU backing this frame" for the shadow map, so I could not capture one.
- `TransformComponent.Rotation` is not MCP-editable, so the dusk scene could not be relit for a daylight check; logged against #607.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected foliage placement across terrain islands.
  * Fixed foliage shadows appearing at the world origin instead of beneath their plants.
  * Improved render-relative positioning and impostor sizing for more accurate visuals.

* **Tests**
  * Added coverage to verify foliage impostor placement, transforms, and height-based sizing.

### DIFF
--- a/OloEditor/assets/shaders/Foliage_Impostor.glsl
+++ b/OloEditor/assets/shaders/Foliage_Impostor.glsl
@@ -105,7 +105,24 @@ void main()
 
     float scale = a_PositionScale.w;
     float rotation = a_RotationHeight.x;
-    float radius = u_ImpostorParams1.y * scale;  // world-space card radius
+    float height = a_RotationHeight.y;
+
+    // World-space card radius. u_ImpostorParams1.y is the OBJECT-space radius the
+    // bake framed the mesh with, and the meshes are authored unit-height
+    // (pine.obj spans y in [0,1], radius 0.560) — so the per-instance world
+    // height has to come back in here, exactly as the near path applies it
+    // (`localPos.y *= height * scale` in Foliage_Instance.glsl). This used to
+    // read only `scale`, drawing a 9-16 m pine as a ~1.5 m card: an ~8x shrink
+    // the moment an instance crossed ImpostorStartDistance (issue #953).
+    //
+    // UNIFORM, not anisotropic. Matching the near quad's aspect exactly would
+    // mean stretching the card 1:16, which renders the baked pine as a needle —
+    // and the near quad is a deliberately different thing anyway (a tufted
+    // billboard; the impostor is what "gives the tree line a 3D silhouette from
+    // any azimuth at range", issue #433). Scaling uniformly puts the drawn tree
+    // at exactly `height * scale` tall, so nothing pops vertically across the
+    // transition, and leaves it its own proportions.
+    float radius = u_ImpostorParams1.y * height * scale;
 
     // Instance pivot. Foliage's per-instance positions are TERRAIN-LOCAL (x/z in
     // [0, WorldSize], y the raw sampled height), so they only become world
--- a/OloEditor/assets/shaders/Foliage_Impostor.glsl
+++ b/OloEditor/assets/shaders/Foliage_Impostor.glsl
@@ -141,13 +141,22 @@ void main()
     // which is the out-of-bounds hazard the old comment was really about (issue
     // #433). The two got conflated, and the transform was dropped with them.
     vec3 instWorld = (u_Model * vec4(a_PositionScale.xyz, 1.0)).xyz;
-    // Anchor the impostor so its bounding sphere RESTS ON the ground: centre one
-    // world radius above the instance ground point. This is independent of how
-    // the source mesh is authored (centred-on-origin vs base-at-origin) — using
-    // the baked centre.y directly would half-bury a centred mesh, and a
-    // camera-looking-down view then flattens that half-buried card into the
-    // terrain and depth-culls it (issue #433).
-    vec3 cardCenter = instWorld + vec3(0.0, radius, 0.0);
+    // Anchor the card on the MESH CENTRE, because that is what the bake framed:
+    // ImpostorBaker centres each tile on the source mesh's bounding-box centre
+    // and spans +-u_ImpostorParams1.y around it, so the card's centre has to
+    // land on that same point in world space or the tree floats inside its own
+    // card. The meshes are authored base-at-origin (pine.obj spans y in [0,1])
+    // and the near path already relies on that (`localPos.y *= height * scale`
+    // in Foliage_Instance.glsl), so the centre is half the drawn height up.
+    //
+    // This offset by `radius` instead, which was indistinguishable while radius
+    // was the UNIT-mesh radius (~0.5 m). The moment radius became
+    // R0 * height * scale (6-12 m for Drift's pines) the same line lifted the
+    // card centre — and the tree drawn around it — metres into the air: trees in
+    // the sky. The card's BOTTOM stayed on the terrain, so the card extent
+    // looked right and only its contents floated, which is what made this read
+    // as a placement bug rather than an anchoring one.
+    vec3 cardCenter = instWorld + vec3(0.0, 0.5 * height * scale, 0.0);
 
     // Subtle whole-card wind sway (legacy sine model, matching the near
     // billboard's fallback branch) so a distant tree still moves with the wind.
--- a/OloEditor/assets/shaders/Foliage_Impostor.glsl
+++ b/OloEditor/assets/shaders/Foliage_Impostor.glsl
@@ -107,13 +107,23 @@ void main()
     float rotation = a_RotationHeight.x;
     float radius = u_ImpostorParams1.y * scale;  // world-space card radius
 
-    // Instance pivot. Foliage's per-instance positions are absolute world in
-    // a_PositionScale; the foliage command uploads only a SINGLE model-matrix
-    // entry to the InstanceData SSBO, so indexing it by gl_InstanceIndex for
-    // instance > 0 reads out of bounds (zero matrix on this driver) and would
-    // collapse every card to the origin. Shift to render-relative space directly
-    // (subtract the render origin) instead of going through u_Model (issue #433).
-    vec3 instWorld = a_PositionScale.xyz - u_RenderOrigin;
+    // Instance pivot. Foliage's per-instance positions are TERRAIN-LOCAL (x/z in
+    // [0, WorldSize], y the raw sampled height), so they only become world
+    // positions after the owning terrain's transform — which DrawFoliageLayer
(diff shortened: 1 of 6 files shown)


drsnuggles8:
@coderabbitai review

drsnuggles8:
Rebased onto current master (`95f34a30a`, which includes #986) and added a third commit.

## The second commit had a regression, reported and now fixed

Sizing the card from the plant height (commit 2) exposed a latent line in the anchor:

```glsl
vec3 cardCenter = instWorld + vec3(0.0, radius, 0.0);
```

That was written when `radius` was the **unit-mesh** radius (~0.75 m), where the offset is invisible. Once `radius` became `R0 * height * scale` — 6–12 m for Drift's pines — the same line lifted the card's centre, and the tree drawn around it, into the air. Trees rendered floating in the sky.

The card's **bottom** stayed on the terrain the whole time, which is why this read as a placement bug rather than an anchoring one. Drawing the whole card quad opaque is what separated the two: the card rectangle sat correctly on the ground with the tree's coverage floating near its centre.

The bake settles the correct anchor — `ImpostorBaker` centres each tile on the source mesh's bounding-box centre and spans ±`u_ImpostorParams1.y` around it, so the card centre must land on that same world point. The meshes are authored base-at-origin (`pine.obj` spans y ∈ [0,1]) and the near path already relies on that (`localPos.y *= height * scale`), so the anchor is half the drawn height. Card bottom is then `instWorld.y - 0.06·s` and the tree's base inside the tile is `+0.06·s` from it — exactly on the ground.

Confirmed on a live instance: ground → card-centre offset **+6.54 m**, which is `0.5 × 13 × 1.0` for that instance exactly. Trees verified grounded from a high view, a sea-level view, and from inside the island.

Guarded by `ImpostorCardIsAnchoredOnTheMeshCentreNotItsRadius`, which fails on the pre-fix expression.

## Two things visible in the diagnostics, not addressed here

- The near **billboards** were never affected. Marking the two foliage paths in different colours shows every billboard correctly on the terrain; only the impostor cards floated.
- One near pine still reads as a detached canopy: its base does meet the ridge, but the trunk is too sparse to survive the alpha test at that range. Separate and much milder, deliberately left alone rather than fixed speculatively.

PR #989, opened 2026-08-31
fix(renderer): occlude cloudscape against terrain

### LINKED ISSUE #987: Renderer: volumetric clouds composite in front of terrain in Drift
## What

Volumetric cloud puffs are composited **in front of** an island in Drift — small, sharp, white clouds drawn over the terrain silhouette, at screen positions where clouds sit in the surrounding open sky. Seen in Play from the boat's chase camera at low sun.

Reported by the user while reviewing #969 captures; not caused by that work (see below).

## Why it is a bug, not intent

The raymarch already clips against scene geometry — `PostProcess_Cloudscape.glsl`:

```glsl
// Geometry distance (absolute world) — clouds render behind geometry.
float geomT = 1.0e12;
if (depth < 0.9999) { ...reconstruct... geomT = length(...); }
vec2 slab = cloudLayerIntersect(cameraPos, rayDir);
float tEnd = min(min(slab.y, geomT), kMaxMarchDistance);
```

and the pass reads the real scene depth (`blackboard.Scene.SceneDepth`, or
`Post.UpscaledSceneDepthTexture` when upscaling is on —
`CloudscapeRenderPass.cpp:46-49`). So clouds are *meant* to be occluded by
terrain, and here they are not.

## Not the water path

Established while chasing an unrelated question: the frames are otherwise
identical across `UseFFT: false` / 1 cascade / 3 cascades, so nothing about the
ocean is involved.

## Two candidate mechanisms, neither confirmed

I did not isolate this — filing so it is not lost. Both are worth checking before
looking further afield:

1. **Half-res raymarch + bilateral upsample.** The march runs at half
   resolution, so a half-res texel straddling the silhouette reads a depth that
   may land on sky (far plane), marches the full slab, and produces cloud. The
   full-res composite
   (`PostProcess_CloudscapeComposite.glsl`) then bilateral-upsamples it, and
   every tap carries a `+1.0e-3` weight floor:

   ```glsl
   float w = exp(-depthDelta * 400.0) + 1.0e-3;
   ```

   With all four taps mismatched the weights collapse to the floor and the
   result is a plain average of four wrong taps rather than a rejection. That
   explains an edge fringe cleanly; it explains puffs well *inside* the
   silhouette less well.

2. **Temporal resolve without depth rejection.** `PostProcess_CloudscapeResolve.glsl`
   reprojects a history buffer. If history is not rejected where geometry has
   since become occluding, cloud colour persists over the island as the boat
   moves — which matches the observation being from a moving chase camera, and
   matches puffs appearing away from the silhouette edge.

(2) fits the reported picture better and is the one I would test first, e.g. by
holding the camera still and seeing whether the puffs decay.

## Repro

1. Open `Scenes/Drift.olo`, enter Play.
2. Chase camera behind the boat, looking at an island at low sun (the dusk squall
   leg shows it well).
3. Look at cloud puffs over the island silhouette.

## Score
```olo-score
capability: 1
craft: 5
stability: 3
decay: 1
effort: 2
confidence: 0.5
learning: 3
fun: 5
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary

- remove residual half-resolution cloud radiance at terrain silhouettes by reducing each cloud pixel's full-resolution 2x2 depth footprint to its nearest conventional-depth sample
- treat only the exact `1.0` clear sentinel as sky so valid far-ridge depth is retained
- exercise the production GLSL helper through three single-property L2 shader-unit tests
- add inspected native and FSR1 Quality visual evidence and record the diagnostic signature

## Confirmed mechanism

PR #986 already fixed the originally reported temporal-history ghost: that case existed only in `CloudsResolved`. The remaining edge-width leak was present in `CloudsRaw`, so it was neither the temporal resolve nor the composite weight floor.

Two pass-A classification errors combined:

1. one half-resolution cloud invocation covers four full-resolution depth texels, but the previous filtered sample could select sky while another texel contained terrain;
2. the `depth < 0.9999` cutoff classified Drift's valid far ridge depth (about `0.999904`) as sky.

The fix takes the minimum of the exact 2x2 footprint and uses `depth < 1.0` for the conventional-depth geometry predicate. Odd extents clamp to the terminal row/column.

## Verification

- build-lock Debug build: `OloEngine-Tests`
- broad shader/reflection/stage/Vulkan/cloud-depth gate: 23/23 passed
- cloud/atmosphere gate: 49/49 passed
- full engine suite: 6,883 tests; 6,845 passed, 38 skipped, 0 failed
- live Drift, native path: front/left/right/sky-only inspected; 6,907 cloud pixels, 0 over terrain; cloud animation remained live; 0 shader errors
- live Drift, FSR1 Quality path: selected upscaled scene depth was 1411x942 and `CloudsRaw` was 706x471; 6,535 cloud pixels, 0 over corresponding conservative terrain footprints; 0 shader errors
- independent standards and issue-spec reviews: clean after findings were addressed

## Visual evidence

- `OloEditor/assets/tests/visual/CloudscapeTerrainOcclusion_Front.png`
- `OloEditor/assets/tests/visual/CloudscapeTerrainOcclusion_Left.png`
- `OloEditor/assets/tests/visual/CloudscapeTerrainOcclusion_Right.png`
- `OloEditor/assets/tests/visual/CloudscapeTerrainOcclusion_SkyOnly.png`
- `OloEditor/assets/tests/visual/CloudscapeTerrainOcclusion_UpscaledQuality.png`

## Review guide

The behavior change is concentrated in `PostProcess_Cloudscape.glsl` and `include/CloudscapeDepth.glsl`. The probe shader includes that same production helper, so the mixed-footprint, odd-extent, and exact-sentinel tests cannot drift onto a duplicate implementation. The intended conservative tradeoff is that any terrain texel in a 2x2 footprint occludes that half-resolution cloud sample, avoiding a one-pixel foreground leak at silhouettes. FSR1 Quality exercises the alternate upscaled-depth publication route; other FSR1 quality modes select the same route and were not captured separately.

Closes #987

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volumetric cloud depth classification to prevent residual cloud artifacts along terrain silhouettes.
  * Correctly handles mixed terrain-and-sky pixels, odd viewport dimensions, and far-plane depth values.

* **Tests**
  * Added GPU validation coverage for cloud depth selection and sky classification scenarios.

* **Documentation**
  * Added troubleshooting guidance for diagnosing cloud artifacts using depth and cloud rendering captures.


drsnuggles8:
Current-head self-review summary (`6e46f82e3`):

- Standards review initially found that the depth rule needed an L2 shader-unit contract and that one expanded Vulkan test covered multiple properties. Fixed by extracting `CloudscapeDepth.glsl`, including it from both production and the probe shader, splitting coverage into three single-property tests, and removing the Vulkan test's net diff.
- Standards also caught one explicit-type declaration where repository style requires `auto` after a `static_cast`; fixed in `6e46f82e3` and reverified by build plus the 23-test shader gate.
- Spec review initially requested proof of the upscaled-depth route and clearer separation from PR #986. Fixed with a live FSR1 Quality capture/correlation and revised diagnostic chronology.
- The suggestion to omit PNGs was rejected because this task's handover explicitly requires inspected multi-angle evidence under `OloEditor/assets/tests/visual/`.
- Final standards and spec re-reviews are clean; no unresolved local findings.

Issue #974, opened 2026-08-30
Renderer tooling: deterministic AAA benchmark scenes, AOV captures, and publication-quality hero frames


drsnuggles8:
## Problem

OloEngine already has a strong renderer-test pyramid: CPU/contract tests, visual-evidence scenes, golden images, MCP screenshots, intermediate-target inspection, and OpenGL/Vulkan parity checks. Those tests answer **"did the renderer regress?"**, but the repository still lacks a small, deterministic benchmark suite that also answers:

* Is the renderer producing a representative high-quality frame?
* Which lighting term or post-process is responsible for a visible problem?
* How does a change affect image quality, temporal stability, and GPU cost together?
* Are public screenshots showing the renderer at its real quality ceiling rather than a minimal technical demo?

The present screenshots under-sell the renderer because scene composition, assets, exposure, temporal warm-up, camera placement, and capture settings are not standardized.

Note this is distinct from the existing perf-baseline machinery (`--olo-perf-rebase`, `PerfRegressionTests`), which measures cost but not image quality, and from the golden-image suite, which measures change but not quality.

## Proposed approach

Add three deterministic renderer benchmark scenes with checked-in capture manifests:

1. **Material laboratory**
   * Neutral studio environment and 18% gray reference.
   * Dielectric/metallic and roughness sweeps.
   * Known linear color patches, normal-map strength cases, emissive values, and a white-furnace mode.
   * Fixed manual-exposure capture plus a separate auto-exposure convergence capture.

2. **Dense architectural / foliage scene**
   * Opaque and masked geometry, decals, many local lights, shadows, DDGI/SSGI, reflections, volumetric fog, and distant geometry.
   * Camera paths that expose disocclusion, thin geometry, foliage motion, and temporal-history failures.
   * Use redistributable assets only; record asset origin and license in the scene manifest.

3. **Ocean / coast hero scene**
   * Ocean, coastline or island geometry, a vessel or other scale reference, useful reflected geometry, and a controlled sun/sky setup.
   * Fixed spectrum seed, wind, simulation time, camera, exposure, and weather state.
   * Capture both a diagnostic composition and a publication-quality hero composition.

Each scene gets a versioned YAML/JSON capture manifest containing at least:

* scene and camera identifier;
* output resolution and render scale;
* RHI and quality preset;
* deterministic seeds and simulation time;
* manual/automatic exposure mode and parameters;
[...]

Issue #706, opened 2026-07-31
Renderer: shared blue-noise sampler + temporal resolve utility for stochastic passes


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

There is no blue-noise sampler in the engine (grep for `BlueNoise` / `Sobol` across `OloEngine/src` and the shader tree returns nothing) and no reusable temporal accumulation helper — `PostProcess_TAA.glsl` is TAA-specific. Every stochastic effect we have or want (SSR, SSGI, SSS, DDGI ray directions, distance-impostor reflections, soft shadows) therefore either avoids stochastic sampling entirely or invents its own noise, which is exactly how you get correlated artifacts and per-pass reinvention.

**Describe the solution you'd like**

One shared utility, three parts:

1. **Blue-noise sample sequence** — a scrambling/ranking tile (Heitz et al. 2019, *A Low-Discrepancy Sampler that Distributes Monte Carlo Errors as a Blue Noise in Screen Space*) as an SSBO or texture, with a `SampleRandomVector2D(pixel, frameIndex)` GLSL helper that golden-ratio-advances per frame. Blue-noise-distributed error is far cheaper to denoise than white noise at the same sample count.
2. **GGX VNDF importance sampling** (Heitz, JCGT 2018) in `include/PBRCommon.glsl` — sampling the *visible* normal distribution, with the matching `G2/G1` weight so the estimator stays unbiased.
3. **Reusable temporal resolve** — history reprojection + neighbourhood clamp + disocclusion rejection, parameterised so SSR / SSGI / reflections can each instantiate it instead of hand-rolling.

**Describe alternatives you've considered**

Per-pass ad-hoc noise (`Core/FastRandom.h` / hash-based) — what would happen by default; produces visibly worse convergence for the same sample budget and cannot be denoised as aggressively.

**Additional context**

Both sampling techniques are published and freely implementable. A worked integration of all three together (blue-noise sampler feeding VNDF feeding a temporal resolve) can be seen in `D:\repos\tachyonfire\th1\shaders\cubemap.comp:290-372` and `th1/shaders/temporal.comp` / `median.comp` — **GPLv3, read for approach only.**

Enables the distance-impostor reflection issue and improves SSR/SSGI/DDGI quality at no extra ray cost.

**Acceptance criteria**

- A single `include/` header exposes the sampler + VNDF + temporal resolve; at least two existing passes adopt it.
[...]

PR #1001, opened 2026-09-01
refactor(renderer): move the cloudscape resolve onto TemporalResolve.glsl

### LINKED ISSUE #903: Renderer: move PostProcess_CloudscapeResolve onto the shared TemporalResolve.glsl

#706 added `include/TemporalResolve.glsl` to stop every stochastic pass
hand-rolling its own history blend, and refactored `PostProcess_TAA.glsl` onto
it. `PostProcess_CloudscapeResolve.glsl` is the remaining hand-rolled
implementation: world-point reprojection, an RGB min/max clamp, a fixed
feedback. It was left alone deliberately, so the header could land without
touching the cloud goldens.

That leaves the engine with two temporal resolves again, which is the exact
situation the shared header exists to end — and the cloudscape one is the weaker
of the two: a componentwise **clamp** rather than a clip (so a rejected history
can land on a colour the neighbourhood never contained), no variance box, and no
disocclusion test.

Move `PostProcess_CloudscapeResolve.glsl` onto `TemporalResolve.glsl`.

Two things make this more than a mechanical swap, and are why it is its own
issue:

- **The signal is RGBA.** Alpha carries transmittance, so it is not a colour
  channel and must not go through YCoCg. Either the header grows an
  alpha-aware entry point or the pass resolves RGB through the shared kernel and
  handles alpha separately — decide deliberately rather than by accident.
- **The cloud goldens are unusually sensitive.** See
  [`docs/agent-rules/volumetric-cloud-debugging.md`](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/volumetric-cloud-debugging.md):
  a *correct* change can move the frame by 0.3%, there are two suite-level
  golden killers, and "darker" versus "directionally darker" is the only thing
  that separates several look-alike causes. Budget the verification, not just
  the edit.

Leaving it — which is what #706 did, and is defensible while there are only two
implementations. This becomes worth doing when a third stochastic pass wants a
resolve, or when a fix has to be applied to both copies.

**Additional context**

Deferred from #706 (PR #889).
[`docs/agent-rules/stochastic-sampling-and-temporal-resolve.md`](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/stochastic-sampling-and-temporal-resolve.md)
§5 documents what the shared kernel gets right that the cloudscape one does not,
and §6 names this pass as the obvious next adopter.

**Acceptance criteria**

- `PostProcess_CloudscapeResolve.glsl` uses `TemporalResolve.glsl`; no
  hand-rolled reprojection/clamp/blend left in it.
- The alpha/transmittance decision is stated in a comment where the code makes
  it.
- Cloud goldens either unchanged or rebaked in the same PR with the visual
  reasoning written down (per `procedural-generator-golden-coupling.md`).

## Score
```olo-score
capability: 1
craft: 3
stability: 3
decay: 3
effort: 3
confidence: 0.5
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Moves `PostProcess_CloudscapeResolve.glsl` onto `include/TemporalResolve.glsl`, removing the engine's last hand-rolled temporal resolve. Deferred from #706 (PR #889).

## What changed

**The pass** no longer answers any of the resolve's four questions itself. Reprojection, neighbourhood clip and feedback blend all come from the shared header; what remains pass-specific is the cloud-layer-midpoint anchor (clouds have no depth buffer) and the #987 occlusion gate.

**The header** grew the pieces that were missing, rather than the pass growing special cases:

- `OloTemporalReprojectWorldPoint` — question 1 had no helper at all. Owns the `w <= 0` guard, which needs its own answer rather than the caller's memory: a divide by negative `w` mirrors the point through the origin and lands somewhere perfectly plausible on screen, so the off-screen test downstream would pass it. Returns `(-1,-1)` so a caller who ignores the flag still fails safe.
- `OloTemporalScalarStats` / `OloTemporalClipHistoryScalar` / `OloTemporalBlendScalar` / `OLO_TEMPORAL_GATHER_3X3_RGBA` / `OloTemporalResolveRGBA` — the alpha-aware entry point.
- `OloTemporalFeedbackWeight` — one owner of the 0.98 cap, so no channel can drift from another.

### The alpha decision (acceptance criterion 2)

Stated in the header where the code makes it, under *"Channels that are NOT colour"*.

Alpha is **transmittance**, consumed downstream as `scene * a + rgb`. It is an occupancy, not a hue, so it must not enter YCoCg: that transform exists to give luma and chroma separate variance, and a fourth non-colour quantity pushed through it acquires a chroma that means nothing — the history would then be rejected for having the wrong colour in a channel that has no colour. So alpha gets its own 1-D box, gathered in the same nine-tap loop.

The part worth reading twice, because it otherwise looks like a leftover: **in one dimension the segment clip and the componentwise clamp are the same operation.** The failure clamping causes — landing on a value the neighbourhood never contained — needs at least two coupled axes to exist, and one axis has no second axis to leave. So `OloTemporalClipHistoryScalar` clamps, and that *is* `OloTemporalClipToAABB` degenerate. Pinned by `StochasticSampler.ScalarClipIsTheOneDimensionalCaseOfTheClip` against `ClipToAABB` itself, rather than asserted in prose.

### One knob deliberately not adopted

**No motion-scaled feedback.** `OloTemporalMotionFeedback` exists because TAA's reprojection is least trustworthy exactly where motion is largest — at the silhouettes of moving objects, which velocity dilation only half-fixes. This pass has no moving objects and no velocity buffer: a camera pan reprojects the layer analytically and is *correct*, while the signal it accumulates is a jittered raymarch that needs its history most while the camera moves. Cutting feedback on motion would trade a ghost this pass does not have for noise it does. The reason is in a comment so the omission does not read as an oversight.

## Goldens: unchanged, not rebaked (acceptance criterion 3)

All 30 cloud/atmosphere tests pass against the **committed** `Atmosphere_*` / `Drift_*` baselines with no rebase. Nothing to rebake, so **the cloud baseline does not move under #974's benchmark-scene work** — flagging that here as the handover asked.

## Verification

Noise floor measured first, per `live-verification-noise-floor.md`:

| measurement | worst RMSE | note |
|---|---|---|
| noise floor — same shaders, two runs | **0.361** | most captures byte-identical |
| signal — master vs shared kernel | **0.867** | 3–13× the floor where it moves |
| pre-existing driver delta vs committed goldens | **2.27** | test threshold is 8.0 |

The signal is **structured**, which is what separates it from the uniform veil in `volumetric-cloud-debugging.md`: overcast cells (smooth deck) barely move at 0.006–0.05 RMSE, storm cells (high coverage variance) move most at 0.17–0.87, and a 20×-amplified diff of `Drift_DawnHero` puts the change **only on the cloud puffs**, tracing their exact positions with the water and clear sky untouched. Signed band means are ~0, so history is being re-decided, not dimmed — not "darker", and not "directionally darker" either, which is the correct answer for a resolve change.

Live over MCP (Drift, overcast, noon): `PostProcess_CloudscapeResolve` link status 1 / Ready, no shader errors in `OloEngine.log`. `CloudsResolved` shows a dense deck with a sharp black occlusion wedge and no cloud bleeding onto terrain (#987 intact); after a drastic pose change it comes back clean with **no ghost of the previous pose's wedge**, which is the history-rejection path in situ.

## Review guide

**Where I'd look hardest**

1. `include/TemporalResolve.glsl` — `OLO_TEMPORAL_GATHER_3X3` now **delegates** to the RGBA gather with a discarded alpha stats struct. This changes the code path for TAA, SSR and SSGI, not just the cloudscape. My claim is that it is bit-identical (same accumulation order; `texture()` returns a `vec4` regardless; the alpha accumulators are local dead stores) and the 130-test sweep agrees — but this is the one edit that reaches passes this issue is not about.
2. `PostProcess_CloudscapeResolve.glsl` — `kClipGamma = 1.25`. The variance box is strictly *tighter* than the raw min/max it replaces. My argument that this is safe is that the box is built from **this** frame's neighbourhood, so it opens where the raymarch jitter is noisy and closes only where the deck is smooth (and there the history is near the mean anyway). That is reasoning, not a measurement of every regime.
3. `include/TemporalResolve.glsl`, the hard min/max intersection — `sqrt(E[x²] − mean²)` does **not** return 0 on a flat channel: one ULP of cancellation (5.96e-8 at x = 0.6) becomes 2.4e-4 after the sqrt, measured on GPU at 3.2e-4. So the intersection is load-bearing, and tidying it away as redundant would silently start admitting a sliver of history on flat sky.

**What I verified, and how** — the numbers above; `ShaderUnitTemporalResolveTest` (4 GPU tests) drives the real GLSL, because the risk here is the gather filling two stats structs from one loop, and a crossed accumulator passes every CPU mirror of the functions it feeds; `StochasticSampler` CPU mirrors for the clip/reprojection algebra; a ratchet test that the pass cannot regrow its own resolve; the 130-test sweep over every shared-header consumer; all six consumers compile under `glslc` on `opengl4.5` and `vulkan1.2`.

**Least confident about** — the γ = 1.25 choice, honestly. The evidence tests cannot interrogate it: with a static camera `prevUV == uv`, so once converged `history == current`, which is trivially inside the box, and **both implementations are provably the same no-op**. That is exactly why the 20×-amplified sky reads as pure black in the converged cells. It is a correctness result, not a null one, but it means the offline captures under-exercise the clip and the rejection path rests on the GPU probe plus the live pose-change capture.

Two things worth a second opinion beyond that: whether the `AtmosphereVisualEvidenceTest` scene should be showing a flat grey sky at every weather setting (it does, on master too — look-alike cause 4, feature scale vs the 30° FOV — so those 13 captures contribute almost nothing to a cloud change, and it may deserve its own issue); and whether the ratchet test's substring matching earns its keep.

**Deliberately not tested** — a live A/B of magnitude. I ran one and it is **confounded, not merely noisy**: the cloud wind and ocean sim advance independently of the paused sun (live floor 1.25 RMSE with the sun frozen), so two editor sessions show genuinely different cloud phase. I kept the live run for what it does prove — clean link, #987 intact, no ghost after a pose change — and took magnitudes from the deterministic mock-clock captures instead. Vulkan backend not exercised; the shader compiles on the `vulkan1.2` target but I did not run the Vulkan pass suite on a GPU.

Closes #903

## Summary by CodeRabbit

- **Rendering Improvements**
  - Improved temporal rendering reliability for RGBA data, including independent alpha handling and accurate color-channel results.
  - Updated cloudscape temporal resolving to use the shared rendering behavior for more consistent visual output.

- **Bug Fixes**
  - Improved handling of history reprojection, screen-edge coordinates, invalid points, and low-confidence history.

- **Tests**
  - Added GPU and CPU coverage for temporal resolve, alpha statistics, history clipping, and reprojection scenarios.

- **Documentation**
  - Documented RGBA temporal-resolve behavior and troubleshooting guidance for related rendering issues.


drsnuggles8:
## 🤖 Self-review @ `9bcd8d3478a11276f447d51d27b408a95f8ad07f`

Reviewed the full diff at **high** effort before opening the PR (Phase 3), then re-ran the 130-test sweep after the fixes.

- **Findings:** 5 · **Fixed:** 4
  1. The new probe shader `ShaderUnit_TemporalResolveRGBA.glsl` was **untracked** — a `git commit -a` would have shipped four tests that hard-fail on every GPU run while headless CI stayed green. Staged explicitly.
  2. `OLO_TEMPORAL_GATHER_3X3_RGBA` duplicated the entire colour half of `OLO_TEMPORAL_GATHER_3X3`. The colour-only macro now **delegates** to the RGBA one (discarded alpha stats, dead stores the compiler drops), so the colour statistics have exactly one implementation. A header that exists to stop two passes carrying two resolves must not itself carry two gathers.
  3. `OloTemporalResolveRGBA` inlined the feedback blend, so a change to the 0.98 cap would have reached TAA/SSR/SSGI but not the cloudscape. Extracted `OloTemporalFeedbackWeight`; every blend goes through it.
  4. The ratchet test's bare substrings would have fired on innocent future code (`nMin` matches `nMinSteps`, `mix(current` matches `mix(currentA,`) and its comment stripper handled only `//`, so a `/* */` description of the removed clamp would have tripped it. Needles now carry their delimiter and block comments are stripped.
- **Dismissed:** 1 — "ships with no captured frame or evidence, and residual shimmer on a converged deck is what a CPU mirror plus a synthetic probe cannot see." The first half is addressed (see the PR body's measurement table, the 20× cloud-shaped diff, and the live `CloudsResolved` captures). The shimmer half is answered by data already in hand rather than by a new test: the converged evidence cells are **byte-identical between the two arms**, with a same-shaders noise floor of 0.000 on those images — so a converged deck is stable under both implementations, which is precisely the shimmer question.

drsnuggles8:
## Known-red check: `GoldenBaselineAudit.VendorSetsAgreeWithTheSharedSet` → #1008

Flagging this so the red is not read as unexplained. **It is the guard working, not the PR being wrong.**

The atmosphere commit caps `FogMaxOpacity` so the cloud deck is visible in the Overcast/Storm cells, which had been a flat fog fill. Both vendor baseline sets are rebaked here — the AMD one on the AMD runner itself, per §7 of the crosscheck doc. The committed AMD set turned out to carry the **identical** defect (sky std `0.000`, mean exactly `165.00`/`149.59` at every time of day), so the audit had been green on two sets that agreed about nothing.

With both sets rebaked the sky band now agrees cross-vendor (0.39 mean-abs) and the ground agrees (0.10). What remains is a systematic **−21.7 luma across 89.5% of the horizon band**, where AMD renders the *uncapped* fog colour while NVIDIA shows sky through the cap — a real divergence the flat-fill baselines were masking. Measurements and two candidate mechanisms are in **#1008**.

Widening the audit's thresholds to go green would disable the check that found this, so it stays red until #1008 is fixed.

Issue #1002, opened 2026-09-01
Renderer/Vulkan: Deferred path renders no lit output — DeferredLightingPass's fragment output reaches no render target


drsnuggles8:
## Problem

On the **Vulkan** backend, the **Deferred** rendering path produces no lit output:
`DeferredLightingPass`'s fragment shader output reaches **no render target at
all**. The viewport shows a flat, colourless scene. OpenGL is unaffected —
Deferred renders correctly there, on both the slot-based and the bindless route.

What the user sees is not "nothing", which is why it reads as a shading bug: the
grey, dimly-shaded spheres visible in the viewport are the **AO buffer applied
over the flat clear colour**, not lit geometry.

## Evidence (measured 2026-09-01, NVIDIA RTX 4090, this box)

Scene: `Assets/Scenes/PBRClosureV2Test.olo`, editor launched `--rhi=vulkan`.

**1. The lit chain is greyscale from `SceneColor` onward.** Sweeping all 69
render-graph targets and measuring per-pixel colour spread (`max(rgb)-min(rgb)`):

| target | colour spread |
|---|---|
| `GBufferAlbedo` / `GBufferResolved` | 39.8 |
| `GBufferNormal` / `SceneNormals` | 140.2 |
| `SceneColor`, `ToneMapColor`, `FXAAColor`, `UIComposite` | **0.0** |

The G-Buffer is correct and everything downstream of lighting is pure greyscale.

**2. The G-Buffer inputs are correct.** Probing a gold ClosureV2 sphere:
albedo `(0.949, 0.749, 0.400)`, flags lane `2.0`, depth `0.992`
(linear 12.4), entityID 9. All as expected.

**3. The fragment output reaches nothing.** Forcing the shader to
`o_Color = vec4(1,0,0,1)` unconditionally:

| backend | sphere final colour |
|---|---|
| OpenGL | `(0.659, 0.0, 0.0)` — red, as expected |
| Vulkan | `(0.31, 0.31, 0.31)` — unchanged |

Scanning **all 69 Vulkan targets** for red pixels: **none**. The output is not
landing in the wrong target; it is not landing anywhere.

**4. Forward works on the same session.** `SceneColor` min 0.006 / **max 9.22**
on Forward vs min 0.10 / **max 1.00** on Deferred — so the device, the scene,
the lights and the post chain are all fine.

**5. The pass executes.** `DeferredLightingPass` is present in the 49-pass
[...]

Issue #998, opened 2026-09-01
Tooling: check in the GGX energy-table generator (emit both language twins)


drsnuggles8:
## Problem

The #975 GGX energy tables (`GgxEnergyTables.h` /
`include/PBRClosureV2Energy.glsl`) are marked **GENERATED, DO NOT HAND-EDIT**,
but no generator is checked in — regeneration exists only as a prose recipe in
the header comment (4096-sample Hammersley VNDF estimator, 64-point E_avg
midpoint quadrature, IEEE-half round-to-nearest-even packing, identical hex
words emitted into two languages). The actual script that produced the tables
lived in a session scratchpad and is gone.

The recompute pin (`ClosureV2.EnergyTablesMatchTheirOwnEstimator`) keeps the
tables from *rotting*, but it cannot *produce* them: any future grid widening,
`MIN_ROUGHNESS` change, or a V3 table means re-implementing the whole pipeline
from a comment and hand-splicing hex into two files that must stay word-for-word
identical — the rebake-without-the-baker trap, and the one place ADR 0016 §6's
"regenerating the energy tables is a generator run emitting both files" promise
is currently false.

## Proposed approach

Check the generator in as a small tool — `OloEngine/tests/scripts/` next to
`generate_test_catalogue.py` is the natural home (Python precedent, no build
wiring needed), or `tools/` if a C++ tool reusing `ReferenceBRDF.h` directly is
preferred (which would make generator-vs-engine estimator drift structurally
impossible). Either way it must:

* recompute Ess/E_avg with exactly the documented estimator (the recompute
  test is the referee — a checked-in generator whose output fails the pin is
  wrong by definition);
* emit **both** files in full (header + GLSL include), never a paste-me
  fragment;
* take the grid size / sample counts as parameters with today's values as
  defaults, so a widening is a flag rather than an edit;
* update the two files' provenance comments to name the tool instead of the
  prose recipe.

## Acceptance criteria

* Running the tool with default parameters reproduces the currently committed
  files byte-for-byte.
* `GgxEnergyTables.h`, `PBRClosureV2Energy.glsl`, and ADR 0016 §6 all point at
  the tool.
* The recompute + GLSL-twin drift pins in `ClosureV2Test.cpp` pass against the
  regenerated output.

## References

* #975 / PR #990 — the tables and their pins; ADR 0016 §4/§6; the regeneration
  recipe currently living in `GgxEnergyTables.h`'s header comment.

## Score
```olo-score
[...]

Issue #975, opened 2026-08-30
Renderer: PBR closure v2 — versioned Evaluate/Sample/Pdf contract, energy compensation, near-mirror fix


drsnuggles8:
## Problem

OloEngine's CPU reference path tracer deliberately mirrors `OloEditor/assets/shaders/include/PBRCommon.glsl`, which is the right policy for renderer-parity testing. Issue #706 already landed a large part of a modern specular stack there — GGX VNDF importance sampling (`sampleGGXVNDF`), the Heitz G2/G1 sample weight (`ggxVNDFWeight`), height-correlated Smith visibility, a shared blue-noise sampler, plus white-furnace tests (`PathTracerFurnaceTest.cpp`) and CPU/GPU parity tests (`ReferenceBRDFGpuParityTest.cpp`).

What is **not** yet in place is the part a GPU path tracer and a ReSTIR stack actually depend on:

* **No versioned material model.** Changing the closure today silently changes every existing scene and golden. There is no way to introduce a corrected closure without a migration cliff.
* **No `Evaluate` / `Sample` / `Pdf` triple as one contract.** BRDF evaluation exists as renderer code and VNDF sampling exists as shader code, but nothing pins them to a single agreeing density. Path tracing needs all three to be mathematically consistent; MIS needs the density they agree on.
* **Near-mirror lobe collapse.** `distributionGGX` divides by `max(denom, EPSILON)` with `EPSILON = 1e-4` (`PBRCommon.glsl:151`). As roughness approaches 0 the numerator goes to zero while the clamped denominator does not, so the lobe collapses toward black instead of approaching a mirror.
* **Two geometry terms coexist.** The direct-light path still uses the Schlick-GGX `k` approximation (`PBRCommon.glsl:222`) while the VNDF path uses height-correlated Smith. Both are defensible; having both under one "the BRDF" label is not.
* **No energy compensation.** There is no multiple-scattering / energy-loss correction anywhere in the tree, so rough metals lose energy that a path tracer will make obvious.
* The current Fresnel/diffuse split is common in real-time renderers but mildly non-reciprocal.

Full path tracing amplifies inconsistencies between `Evaluate`, `Sample`, and `Pdf`. ReSTIR then reuses those samples across pixels and frames, making a superficially plausible but biased material implementation much harder to diagnose.

## Goal

[...]

PR #1004, opened 2026-09-01
Renderer: make the G-Buffer flags lane resolve-safe and multi-bit (#996)

### LINKED ISSUE #996: Renderer: make the G-Buffer flags lane resolve-safe and multi-bit (deferred PBRModel transport)
## Problem

The G-Buffer RT2 alpha "MaterialFlags" lane (bit 0 = unlit, bit 1 = the #975
`PBRModel` closure selector, encoded by `oloEncodeGBufferPbrFlags` in
`PBRCommon.glsl`, decoded in `DeferredLightingShared.glsl`) is **not
average-safe**, and it is transported as a **single bit** against an enum whose
contract is append-only.

Two concrete defects, both documented as known limitations in ADR 0016 §5 and
at the encode helper, found during #975's self-review (mechanism code-verified
end to end):

1. **The resolved-MSAA deferred mode averages the bitfield.** With
   `MSAASampleCount > 1` and *Per-sample Deferred Lighting* unchecked (both
   live, non-default, session-local toggles in the Renderer Settings panel),
   `GBuffer::Resolve()` average-blits every colour attachment and the non-MSAA
   `DeferredLighting.glsl` reads the result. A silhouette pixel half-covered by
   a ClosureV2 surface (a=2.0) over Legacy geometry (a=0.0) resolves to exactly
   **1.0 → decodes as UNLIT → raw-emissive black fringe** around every v2
   object; a v2-over-skybox edge (2.0 vs 1.0) resolves to 1.5, where GLSL
   `round()` half-way behaviour is implementation-defined. The pre-#975
   `{0,1}` encoding had the milder variant (unlit silhouettes decoding lit);
   the widened co-domain made the failure mode worse.

2. **A third `PBRModel` silently truncates on Deferred only.** The lane
   carries one model bit; `PBRModel.h` mandates append-never-renumber. When
   `PBRModel(2)` is appended, Forward (which reads the full `u_PBRModel` int
   from the material UBO) shades it correctly while the deferred decode reads
   bit 1 = 0 → Legacy — a silent Forward/Deferred photometric divergence, the
   exact class `ClosureV2DeferredParityEvidenceTest` pins for the current
   model. The `min()` clamp in the encoder makes this deliberate-and-documented
   today, but the lane must be widened the day a third model lands.

## Proposed approach

Make the flags lane resolve-safe and multi-bit in one change, since both fixes
touch the same seam:

* Exclude the flag bits from the averaged resolve — candidates: move the flags
  to an integer RT (MSAA integer attachments resolve via sample 0, not
  averaging), a small custom resolve shader that carries RT2's alpha through
  `texelFetch` of sample 0, or decode-before-resolve in the per-sample path.
  Whichever lands, `DeferredLighting.glsl`'s non-MSAA read should move from
  filterable `texture()` to `texelFetch` for symmetry with the MSAA variant
  (today it is only safe by exact 1:1 texel alignment).
* Widen the model field to 2–3 bits with the layout still owned solely by
  `oloEncodeGBufferPbrFlags` + the one decode site, and extend
  `ClosureV2DeferredParityEvidenceTest` to sweep every encodable model value.

## Acceptance criteria

* With MSAA > 1 and per-sample lighting off, a ClosureV2-over-Legacy and a
  ClosureV2-over-sky silhouette show no unlit misclassification (pinned by an
  evidence test that sets exactly those two toggles).
* A hypothetical model index above the current maximum round-trips the
  G-Buffer un-truncated (or is rejected loudly at the encoder — but never
  silently remapped to Legacy).
* Legacy-only scenes remain byte-identical on all paths and MSAA modes.

## References

* #975 / PR #990 — where the lane was introduced; ADR 0016 §5 records the
  limitation; `oloEncodeGBufferPbrFlags` (PBRCommon.glsl) is the layout's one
  executable home.
* Verifier evidence: `GBuffer.cpp` resolve blit, `DeferredLightingPass.cpp`
  resolved-vs-MS attachment selection, `RendererSettingsPanel.cpp` toggles.

## Score
```olo-score
capability: 2
craft: 1
stability: 5
decay: 3
effort: 2
confidence: 0.8
learning: 2
fun: 2
kano: table-stakes
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md). Stability 5: a user-reachable rendering-correctness footgun (two live toggles) plus a future silent Forward/Deferred divergence in a core lighting path; Decay 3: every flag bit added to the lane inherits the defect, and the ceiling becomes a live bug the day a third model lands; Effort 2 includes the MSAA-mode visual-verification tax. `blocked_by` clears when #975 merges.</sub>

drsnuggles8:
Closes #996.

The G-Buffer RT2 alpha "MaterialFlags" lane was neither average-safe nor wide
enough for a third `PBRModel`. Both defects share one seam, so both are fixed
here, as the issue asks.

## 1. Resolve-safety

Three of RT2's channels are radiometry and want the hardware's averaging
resolve; the fourth is a **bitfield**, and an averaged bitfield decodes to a
material nobody wrote. `GBuffer::Resolve()` keeps the average blit — that is
what leaves the emissive channels bit-identical — and then runs
`GBufferFlagsResolve.glsl`, which overwrites RT2's **alpha only** with the flags
one real sample wrote.

**Sample selection is "the first LIT sample, else sample 0"**, which is
deliberately stronger than the plain sample-0 the issue suggests: with plain
sample 0 a background sample can still speak for a pixel a real surface covers,
which is the fringe wearing a different hat. A `texelFetch` at a fixed sample
index is exactly defined on every backend and vendor, which no averaged encoding
of a bitfield can be.

**Why this and not an integer RT** (the issue offered three candidates and
mandated none): it costs no attachment, so there is no G-Buffer seam change
while #974 is in flight, no new binding constant, and no new "every writer must
write this RT" obligation. And "MSAA integer attachments resolve via sample 0"
is not actually guaranteed by the GL spec — that option would have traded one
implementation-defined case for another, which is the specific thing this issue
is about.

## 2. Multi-bit model

The model field is now the whole rest of the lane (encode `model * 2`, decode a
plain `>> 1`, **no mask**), so appending to `PBRModel.h` can no longer truncate
to Legacy on the deferred path while Forward shades it correctly. Encode and
decode now live **together** in `PBRCommon.glsl`, so the layout has one
executable home in both directions. The only real ceiling is exact integer
representation in RGBA16F, asserted at `kPBRModelGBufferLaneMax` — rejected
**loudly at compile time**, never silently remapped.

`DeferredLighting.glsl`'s flags read moves to `texelFetch`. The texel comes from
`v_TexCoord * textureSize` rather than `gl_FragCoord`, so it stays a nearest
fetch of the right texel even if the lighting viewport ever stops being 1:1 with
the G-Buffer — the invariant the issue notes nothing enforces.

## Evidence

Measured via the new `OLO_GBUFFER_NO_FLAGS_RESOLVE` lever, which switches off
exactly this fix and nothing else (same binary, same scene):

| dark-fringe pixels | control (MSAA 1) | MSAA 4, fix OFF | MSAA 4, fix ON |
|---|---|---|---|
| ClosureV2 subject | 14 | **62** | **3** |
| Legacy subject | 14 | 3 | 3 |

The control is the scene's noise floor — a lit sphere has a real terminator, so
the assertion is a comparison, not a constant. Frame-diffing the two ClosureV2
MSAA frames: **86 pixels along the silhouette, mean luma 27 (min 0 — literally
black) becoming mean 151.**

**On "Legacy-only scenes remain byte-identical on all paths and MSAA modes"** —
the one acceptance criterion that cannot hold literally, so stating it exactly:
both **non-MSAA** frames are byte-identical with the lever on and off (matching
MD5s, zero differing pixels), ClosureV2 and Legacy alike. Under **MSAA** the
Legacy frame does move, on **23 pixels, every one brighter** (luma 76 → ~102),
all on the lower silhouette. Those are the milder variant of the same defect
that ADR 0016 §5 already recorded against the pre-#975 encoding — a Legacy
silhouette averaging with the unlit background to a half-way value. They were
already wrong and are now right.

### Live verification, both backends

Probing the lane in the running editor on `PBRClosureV2Test.olo`:

| probe | RT2 alpha | meaning |
|---|---|---|
| Legacy spheres | 0.0 | model 0, lit — unchanged |
| ClosureV2 sphere | 2.0 | `model * 2` = 1, lit |
| background | 1.0 | unlit bit |

**Identical on OpenGL and Vulkan**, and `OloEngine.log` is clean on both — no
shader compile/link errors, no validation VUIDs.

### Tests

* `GBufferFlagsLaneTest` — sweeps all **1024** encodable model indices through
  the production encode/decode helpers and the real fp16 quantisation. It
  includes those helpers rather than transcribing them, so it tests the layout
  and not a copy of it.
* `GBufferFlagsResolveEvidenceTest` — sets exactly the two toggles the issue
  names and pins both silhouette cases (ClosureV2-over-Legacy and
  ClosureV2-over-background) against the non-MSAA control, plus a Legacy-only
  regression guard.
* Both are non-vacuous: with the lever flipped the ClosureV2 case fails at
  62 vs 14.
* Full suite: **6908 run / 6868 passed / 0 failed** (rest skipped).

ADR 0016 **§5 only** updated to record the limitation as gone — §6 left
untouched for #998.

## Coordination

* **No G-Buffer attachment was added**, so the G-Buffer seam #974 was told to
  coordinate on is unchanged.
* ADR 0016: **§5 only**. `ClosureV2Test.cpp` untouched (#998 owns it); this
  branch's test is `ClosureV2DeferredParityEvidenceTest.cpp`.

## Found while verifying, filed separately

* **#1002** — on **Vulkan**, the Deferred path renders no lit output at all:
  `DeferredLightingPass`'s fragment output reaches no render target. Confirmed
  **pre-existing on master** (A/B with this branch's shader edits stashed and
  the shader cache cleared), so it is not caused by this PR. Filed with the full
  evidence table and five refuted hypotheses.
* **#607** — `olo_renderer_settings_set` cannot reach `MSAASampleCount` /
  `PerSampleLighting`, so this issue's own toggle combination is not reachable
  in the live editor over MCP.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dark fringes and incorrect material shading around silhouettes in MSAA deferred rendering.
  * Improved reliability of material and lighting flags during G-Buffer resolution.
  * Expanded supported PBR model index handling without truncation or unintended fallback to Legacy shading.

* **Documentation**
  * Updated rendering documentation to reflect the corrected G-Buffer flags behavior and supported model range.

* **Tests**
  * Added coverage for MSAA resolves, material-flag round trips, and silhouette rendering parity.


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

PR #1005, opened 2026-09-01
feat(terrain): GPU-resident painting, erosion and undo (#716)

### LINKED ISSUE #716: Terrain: GPU-resident painting and erosion — remove the CPU splatmap and the per-op readback

Terrain authoring round-trips through the CPU on every operation:

- `Terrain/Editor/TerrainPaintBrush.h` says it plainly — *"Paints on CPU splatmap data and marks dirty regions"* — so brush cost scales with brush area on the main thread, and every stroke re-uploads dirty rects.
- `Terrain/TerrainMaterial.cpp:280` pulls the splatmaps back with `GetData(readback)` to keep `m_CPUSplatmaps` in sync.
- `Terrain/Editor/TerrainErosion.cpp:80` runs the erosion compute shader and then **reads the entire heightmap back** unless `skipReadback` is passed — a full GPU->CPU sync in the middle of an interactive operation.

The result is that large brushes stutter and erosion is not interactive.

Keep the authoring textures GPU-resident and never read them back except on save:

- Brush strokes become a compute dispatch / blit into the height and splat render targets, with the brush shape, falloff, opacity and rotation as uniforms. Brush size stops affecting responsiveness.
- Erosion iterates on the GPU with no readback; the CPU heightmap becomes a save-time export rather than a live mirror.
- **Undo/redo for GPU textures** — a bounded ring of region snapshots taken before each stroke, restored by blit. This is the piece that normally blocks the move, and there is a working reference for it.
- Physics heightfield and CPU-side height queries need a defined sync point (relates to #469/#428).

Keeping the CPU as the source of truth and optimising the upload — does not help erosion, and leaves brush cost on the main thread.

**Additional context**

Reference: `D:\repos\Hollow-TerrainSystem\Packages\PhotoTerrain\Shaders\Editor\BrushUpdate.compute`, `BrushRegionCopy.compute`, `Painting/Brush.hlsl`, `Tools/{PaintHeight,SmoothHeight,SetHeight,PaintOMPVPSplat}.shader`; editor-side tools in `Editor/TerrainTools/Painting/`; **undo stack in `Runtime/Layers/TextureUndoStack.cs`**; brush presets and the alt-drag size/opacity/rotation overlay in `Editor/TerrainTools/Painting/TerrainBrushPreset.cs` + `Editor/TerrainTools/Overlays/`. MIT. Their README notes the payoff directly: *"Painting remains responsive no matter the brush size"*.

Related: #551 (per-frame GPU->CPU readback logging — this removes one of the offenders), #469, #428.

**Acceptance criteria**

- Brush stroke frame time is flat with respect to brush radius.
- No `glGetTexImage`/`GetData` on the terrain authoring path outside save/export.
- Undo/redo works across paint, sculpt and erosion operations.
- Erosion is interactive (visible convergence while dragging a slider).

## Score
```olo-score
capability: 5
craft: 3
stability: 3
decay: 1
effort: 5
confidence: 0.8
learning: 5
fun: 5
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
Terrain authoring round-tripped through the CPU on every operation. The paint brush mutated a CPU splatmap, the sculpt brush a CPU height array, and `TerrainErosion::Apply` read the **entire heightmap back** after every compute dispatch. Brush cost scaled with brush area on the main thread, and erosion was not interactive.

## What changed

**The GPU textures are now the source of truth while authoring.** `m_Heights` and `m_CPUSplatmaps` become *mirrors*, refreshed lazily at one named sync point each — `TerrainData::SyncFromGPU` and `TerrainMaterial::SyncSplatmapsFromGPU` — which the read accessors call.

Inverting ownership is what makes the rest tractable. The issue asks to enumerate every remaining CPU consumer, and the list is long: physics height field, the per-edit collision update, quadtree height pyramid, chunk-mesh rebuild, tile stitching, splatmap generation, foliage density masks, the exporters, the editor undo commands. Rather than patch twelve call sites, the sync lives in the accessor, so each of them keeps working unchanged and pays one readback *per sync* instead of one *per operation*. That placement is deliberate: a consumer reading a stale mirror is a silent failure — the terrain looks right and gameplay disagrees with it — so it must not be something a call site can forget.

**Brushes are compute dispatches** (`Terrain_SculptBrush.comp`, `Terrain_PaintBrush.comp`, `TerrainGPUBrush`) over the brush rect. The sculpt kernel reads a rect-sized scratch copy so `Smooth`'s neighbour taps stay order-independent; the CPU brush copied the whole heightmap per apply for that. `TerrainBrushUtils::ComputeBrushRect` is now shared by the CPU and GPU paths so they cannot disagree about which texels a stroke covers.

**Erosion dispatches with no readback at all**, and the panel gains a continuous mode that iterates every frame — affordable only because the dispatch no longer stalls the main thread.

**Undo/redo is a bounded ring of GPU region snapshots restored by blit** (`TerrainTextureUndoStack`), bounded by *both* an entry count and a byte budget: a hundred small dabs and three full-map strokes are the same VRAM problem from opposite ends. Entries hold `Ref<Texture2D>`, not raw handles, so eviction and teardown release through the refcount — the trap in `shared-atlas-allocator.md`.

## Acceptance criteria

| Criterion | Evidence |
|---|---|
| Brush frame time flat vs radius | measured, below |
| No `GetData`/`glGetTexImage` on the authoring path outside save/export | `TerrainAuthoringReadbackTest` — a source scan, runs headless in CI |
| Undo/redo across paint, sculpt and erosion | `TerrainTextureUndoStack` tests + `TerrainGPUBrushVisualEvidenceTest` (undo frame is byte-identical to the pre-stroke frame) |
| Erosion interactive | measured, below; **and confirmed by hand by the repo owner in the editor** |

### Measurements

129² terrain, 200 calls averaged, **GPU work included** (batch drained inside the timed region — a main-thread-only timing cannot tell a cheap dispatch from a cheap dispatch that queues expensive work):

| radius | rect | GPU brush | CPU brush |
|---|---|---|---|
| 4 | 8×8 (64 texels) | 7.96 µs | 2.06 µs |
| 120 | 129×129 (16,641) | 8.10 µs | 974 µs |

**260× the area → 1.02× the cost.** The CPU brush is 473× over the same span.

Erosion, 20 iterations at the panel default of 70k droplets:

- throughput from dropping the readback: **1.16–1.29×** — a real but modest win;
- main-thread cost per continuous-erosion frame: **2.7 µs**.

Being precise about which number carries the "interactive" claim: it is the second one. What made the old slider unusable was a full-map GPU→CPU stall *per iteration*, not throughput. At 70k droplets an iteration is ~24 ms of GPU erosion work regardless of this change, and the droplet count — not this PR — sets how fast it converges per frame.

## The bug this nearly shipped with

`SetData` rebuilds the mip chain as a **side effect**, so no producer had ever needed to ask for it; `TerrainData::UploadRegionToGPU` even re-uploads the whole image when a chain exists, for exactly this reason. A compute kernel writing level 0 bypasses all of that.

That is not cosmetic for the heightmap: the tessellation stage picks a mip from the triangle footprint and `textureLod`s it to **position the vertex**. A stale chain means terrain near the camera is correct, terrain in the distance keeps its pre-edit shape, and it snaps as you approach — correct everywhere the artist is looking and wrong everywhere they are not. Every non-visual test in this PR passed with it present, and a manual editor session reported the feature working.

This PR adds `Texture2D::RegenerateMips()` (GL + Vulkan) and calls it from every writer of level 0. There turned out to be **four**, and the fourth — `TerrainTextureUndoStack::Restore` — was only found *after* fixing the other three, because until then every level was uniformly stale and the visual test passed for the wrong reason. Fixing one instance of this bug exposes the others. Written up as `docs/agent-rules/compute-written-texture-mip-chain.md`.

## Scope note worth reviewing

**The GPU sculpt path is gated on GPU-driven LOD being active** (`m_TessellationEnabled` + `TerrainChunkManager::IsGpuDrivenLODEnabled()`), because it defers the chunk-mesh rebuild to stroke settle — that rebuild reads the CPU mirror, so doing it per drag frame would reinstate the very readback this removes — and deferring is only invisible when the terrain is drawn from the heightmap texture. With GPU LOD off, the CPU brush still runs: no better than before, but never a silent regression.

The component default is `true`, but `ProceduralTerrainTest.olo` and `ScriptedTerrainDemo.olo` both set it `false` and therefore keep the CPU path. That is uncomfortably close to the "a gating flag no scene sets is a feature with zero coverage" lesson in `terrain-gpu-lod-quadtree.md`, so it is called out rather than buried. `TerrainGpuLodTest.olo` and the new visual evidence test both exercise the GPU path.

## Verification

- Full suite: **6923 ran, 6885 passed, 38 skipped, 0 failed**.
- New: `TerrainGPUBrushParityTest` (5 tools × CPU/GPU parity on a deliberately **non-square** terrain — a texel-space radius would look correct on a square one, which is the space mismatch `cpu-gpu-surface-parity.md` is about), paint parity across both splatmaps at 6 layers, sync-point tests, 4 undo-ring tests, `TerrainAuthoringReadbackTest`, `TerrainGPUBrushVisualEvidenceTest`.
- std140 layout verified by reflecting the compiled SPIR-V (`block_size = 96`, offsets matching the C++ twin) rather than by reasoning — an earlier `ivec3` pad silently made the block 112 bytes. `TerrainBrushParams` is now registered in `ShaderUBOSizeConsistencyTest`, the guard that would have caught it.
- Live editor: both kernels compile, `TerrainGpuLodTest.olo` renders via the GPU LOD path, the panel activates Sculpt mode and reports terrain hits. **A full mouse-driven stroke was verified by the repo owner, not by me** — I could not drive the ImGui panel reliably from the diagnostics server.

## Known, not fixed

The brush falloff steps from **0.5 → 0** at the rim for the default falloff of 0.5 (`ComputeFalloff` returns `lerp(1, smooth, falloff)`, which is `1 - falloff` at `t = 1`), so an accumulated stroke leaves a visible crease at the brush boundary. This is **pre-existing CPU brush behaviour** faithfully reproduced by the kernel — the parity tests pin that — and changing it would alter shipped brush results, so it is out of scope here. Visible in the evidence PNGs at 300 accumulated applies.

Takes UBO binding **83**, the last index below the GL 4.6 minimum guarantee of 84, shared by both brush kernels. The next binding addition has to reclaim a retired one or move to an SSBO; the `static_assert` makes that a compile error rather than a silently unbound block.

Closes #716

## Summary by CodeRabbit

* **New Features**
  * Added GPU-accelerated terrain sculpting and painting.
  * Added Raise, Lower, Smooth, Flatten, Level, and continuous erosion tools.
  * Added GPU-based terrain undo and redo support.
  * Added automatic synchronization between GPU terrain edits and CPU data.
  * Improved texture mipmap regeneration after terrain updates.

* **Bug Fixes**
  * Fixed stale distant terrain detail after GPU edits.
  * Ensured terrain changes remain visible and undo restores the previous result.

* **Documentation**
  * Added guidance for maintaining mipmaps after GPU texture writes.


drsnuggles8:
@coderabbitai review

PR #1014, opened 2026-09-02
fix(build): make Linux desktop launchers relocatable

### LINKED ISSUE #913: Linux .desktop launcher: not relocatable, icon format mismatch
`BuildPipelinePlatform::WriteDesktopEntry` (added in #891) writes a freedesktop `.desktop` launcher entry for Linux game builds, but it has two known gaps flagged by CodeRabbit on that PR:

1. **Not relocatable.** `Exec=` embeds the build-time absolute path to the copied runtime executable (`<OutputDirectory>/<GameName>/<GameName>`). Moving the finished game folder — to a different directory, a different machine, or packaging it into an installer/AppImage/Flatpak — leaves the launcher pointing at a path that no longer exists, so double-clicking the `.desktop` file does nothing.
2. **Icon format mismatch.** `GameBuildSettings::IconPath` is a single field shared with the Windows icon-embedding step, which requires a `.ico` file. Linux's `Icon=` field expects a PNG/SVG/XPM file or an icon-theme name — a Windows-authored `.ico` selected in the editor won't render on Linux, silently falling back to a generic icon.

- Resolve the executable path at install/first-run time instead of baking in the build-time absolute path — e.g. write `Exec=` relative to the `.desktop` file's own location (there's no portable field-code for "directory of this .desktop file" in the Exec key itself, so this likely means a small wrapper script, or generating/repairing the `.desktop` file when the game is first launched from a moved location) — or document a required "regenerate launcher" step as part of a real install process.
- Either convert/re-encode a selected `.ico` into a PNG for the Linux target, or give the Build Game panel a platform-aware icon picker/field so a Linux build can point at its own PNG/SVG/XPM directly (the panel's file-open filter was already made platform-aware in #891; this is the deeper "produce a working icon file for the target" half).

Leaving it as-is: the `.desktop` entry works today for a build run in place (the scenario #891's acceptance criteria asked for), and is a strict improvement over having nothing. This issue tracks the harder "ship it somewhere else" case as a separate, explicitly scoped follow-up rather than silently expanding #891.

**Additional context**
Surfaced by CodeRabbit on PR #909 (issue #891): https://github.com/drsnuggles8/OloEngineBase/pull/909

## Score

```olo-score
capability: 1
craft: 1
stability: 3
decay: 1
effort: 3
confidence: 0.7
learning: 1
fun: 1
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
## Summary
- stage a portable Linux desktop launcher, wrapper, and target-native packaged icon
- resolve launches from the moved desktop file through the standard %k field code
- refresh the icon path from the moved package without blocking a read-only package from launching
- cover relocation, icon validation/regeneration, and Linux wrapper execution

## Verification
- GameBuildPipelineTest.LinuxLauncher* — 3 passed on Windows; Linux execution case skips there and runs on Linux
- pre-commit run --all-files

Closes #913

## Review guide

**Where I'd look hardest**
1. OloEngine/src/OloEngine/Build/LinuxDesktopLauncher.cpp — Desktop Entry escaping and the explicit /bin/sh bootstrap must preserve %k as a standalone field-code argument.
2. OloEngine/src/OloEngine/Build/LinuxDesktopLauncher.cpp — wrapper icon refresh is deliberately best-effort so a read-only moved package still launches.
3. OloEngine/tests/Build/GameBuildPipelineTest.cpp — Linux-only execution test exercises the generated path after relocation.

**What I verified, and how** — focused launcher tests passed on Windows, pre-commit passed, and Linux CI will run the execution/desktop-file-validation branch.

**Least confident about** — desktop-environment icon-cache refresh timing after the launcher rewrites its own entry; the entry is standards-compliant after launch, but each desktop may refresh visual metadata at a different cadence.

**Deliberately not tested** — a real Linux desktop session was not available locally; the Linux-only test is left for CI.

## Summary by CodeRabbit

* **New Features**
  * Linux game packages now include a relocatable desktop launcher and startup script.
  * Launchers continue working when the packaged game directory is moved.
  * Optional PNG, SVG, and XPM icons are packaged with the launcher and resolved automatically.
  * Invalid, unsupported, or corrupted icons are rejected with clear build errors.
* **Documentation**
  * Linux shipping guidance now explains the additional launcher files, package distribution requirements, and which files should be excluded from Steam depots.

### DIFF
--- a/OloEngine/src/CMakeLists.txt
+++ b/OloEngine/src/CMakeLists.txt
@@ -207,6 +207,8 @@
 		"OloEngine/Build/GameBuildSettings.cpp"
 		"OloEngine/Build/GameBuildPipeline.h"
 		"OloEngine/Build/GameBuildPipeline.cpp"
+		"OloEngine/Build/LinuxDesktopLauncher.h"
+		"OloEngine/Build/LinuxDesktopLauncher.cpp"
 		"OloEngine/Build/BuildPipelinePlatform.h"
 		"OloEngine/Serialization/AssetPackFile.h"
 		"OloEngine/Serialization/FileStream.h"
--- a/OloEngine/src/OloEngine/Build/BuildPipelinePlatform.h
+++ b/OloEngine/src/OloEngine/Build/BuildPipelinePlatform.h
@@ -19,8 +19,8 @@ namespace OloEngine::BuildPipelinePlatform
     /// Write a freedesktop.org `.desktop` launcher entry next to `exePath`
     /// (#891) — the Linux arm of icon handling. Linux has no PE-resource slot
     /// to embed an icon into, so instead of the Windows-only nicety silently
-    /// vanishing on Linux, the pipeline emits a standard launcher entry that
-    /// points at the (optional) icon file and marks the executable runnable.
+    /// vanishing on Linux, the pipeline emits a relocatable launcher and stages
+    /// the optional Linux-native icon inside the package.
     /// Returns true on success, populates `outError` on failure.
     /// On platforms without an implementation, returns false and sets an error.
     bool WriteDesktopEntry(const std::filesystem::path& exePath,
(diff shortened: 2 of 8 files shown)


drsnuggles8:
@coderabbitai review

PR #1025, opened 2026-09-03
ci(linux): clang-23 on hosted runners, drop the dead clang-19 pin on the box

drsnuggles8:
Option 2 of the three we discussed: hosted Linux to clang-23, and the self-hosted arm fixed to describe reality instead of a pin that was never there.

Follows #1024 (Windows LLVM 23.1.0). Kept separate on purpose — a four-major compiler jump can red the sanitizer jobs on its own findings, and that must not be able to sink the Windows fix.

## Hosted: clang-19 → clang-23

`setup-llvm-apt`'s default and the apt lists that name the compiler directly (`setup-linux-build`, `video-ffmpeg`).

Verified the suite exists **before** bumping — `llvm-toolchain-noble-23` publishes `clang-23`, `lld-23` and `libc++-23-dev` in its `binary-amd64` index. A version with no suite fails as an apt 404 several steps after the step that chose it, so this is worth checking rather than assuming.

The reason hosted needs apt.llvm.org at all is unchanged: ubuntu-24.04's default libstdc++ has no `std::forward_like`, and `Core/Reflection` uses it.

## Self-hosted: the clang-19 pin was never provisioned

Two findings, either one sufficient:

1. **The pin didn't exist on the box.** `setup-linux-build` selected `/usr/lib64/llvm19/bin` when present — on the box only `llvm21` exists. So every self-hosted sanitizer run had already been taking the documented warn-and-fall-back path, silently, for as long as the pin stood. The comment described a machine state that was not true.
2. **The box cannot follow 23.** EPEL on Rocky 10.2 tops out at `clang20` (20.1.8); no `clang22`, no `clang23`. Pinning a version the distro doesn't ship just recreates the same unprovisioned pin one number higher.

So the box uses what it has — system clang 21.1.8 with `compiler-rt-21` and `lld-21` — and the action now verifies the **sanitizer runtimes** rather than a version. That's the check that protects something real: a clang without compiler-rt configures fine and fails every sanitizer *link* on `cannot find libclang_rt.asan.a`. The runtime dir is asked of the compiler itself, so it follows whatever layout the distro uses. Still a warning, never a failure — `asan.yml` routes every same-repo PR's sanitizer jobs here.

Confirmed on the box over ssh before writing any of it:

```
clang version 21.1.8 (RESF 21.1.8-1.el10)     LLD 21.1.8
asan: present   tsan: present   ubsan: present
compiler-rt-21.1.8-1.el10.x86_64   lld-21.1.8-1.el10.x86_64
```

The resulting skew between the arms is coverage, not a defect — the #1010 argument for that is unchanged and quoted where it was.

## Provisioning + docs

- `setup-olo-ci-host.sh` installs `compiler-rt` + `lld` for the system clang instead of EPEL's clang19 trio, and checks **all three** runtimes rather than asan alone (a compiler-rt missing tsan fails only the TSan job's link, hours later, in a job nobody was watching).
- `setup-olo-ci-runners.sh` drops the three `clang19` packages.
- Docs updated in six places, including the two that stated the pin as current fact (`docs/README.md`, `self-hosted-gpu-runner.md`).

## Checks

Every touched YAML parses; every shell step in `setup-linux-build` passes `bash -n`; both provisioning scripts pass `bash -n`.

**What CI has to prove:** that the engine and its sanitizer flags are clean under clang-23. That's the part I can't verify locally, and the reason this is its own PR.

## Not touched

`setup-linux-build`'s line about "the HOSTED SDK build (1.4.321.0 + clang-19)" — that records what a past measurement was made against, not what the job runs now.

## Summary by CodeRabbit

- **Build & Tooling**
  - Updated hosted Linux build and video workflows to Clang/LLD 23.
  - Self-hosted runners now use system Clang and verify sanitizer runtimes.
  - Added protection against concurrent CMake configuration conflicts and improved toolchain mismatch warnings.
  - CMake now avoids repeating expensive LTO capability checks.

- **Bug Fixes**
  - GameNetworkingSockets initialization and networking tests now run consistently under sanitizers.
  - Updated the bundled networking dependency to include an upstream stack-overflow fix.

- **Documentation**
  - Added guidance for concurrent CMake configuration and refreshed toolchain setup instructions.

### DIFF
--- a/.claude/hooks/claude-build-lock-guard.py
+++ b/.claude/hooks/claude-build-lock-guard.py
@@ -72,7 +72,14 @@ _STATEMENT = r"(?:^|[;|(){}\n]\s*|&+\s*[\"']?\s*|(?:-Command|-c)\s+[\"']?\s*)"
 _PATH = r"(?:[A-Za-z]:)?(?:[\w.\-]*[\\/])*"
 
 # `cmake` alone is not enough -- `cmake --preset msvc` only configures, and
-# configuring is cheap.
+# configuring is cheap. That stays true: a configure will not OOM the box, which
+# is what this mutex exists to prevent. It is NOT the same as saying concurrent
+# configures are safe -- two configures against one binary directory corrupt each
+# other's try_compile scratch projects. That is guarded on the CMake side, by the
+# file(LOCK) at the top of CMakeLists.txt, because it also has to cover configures
+# this hook never sees (VS Code's CMake Tools, Visual Studio, a plain shell). Do
+# not "fix" it by adding bare `cmake` here; see
+# docs/agent-rules/concurrent-cmake-configure.md.
 BUILD_PATTERNS = (
     (
         "cmake --build",
--- a/.github/actions/setup-linux-build/action.yml
+++ b/.github/actions/setup-linux-build/action.yml
@@ -139,7 +139,7 @@ runs:
         # list is worth far more than those seconds, because a per-job list is how
         # one job quietly acquires a dependency the others do not have.
         sudo apt-get install -y \
-          clang-19 lld-19 \
+          clang-23 lld-23 \
           libvulkan-dev vulkan-validationlayers \
           libgl-dev libglu1-mesa-dev \
           libx11-dev libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev libxext-dev \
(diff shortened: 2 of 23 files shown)


drsnuggles8:
@coderabbitai review

Issue #993, opened 2026-08-31
Renderer: canonical GPU Scene light and environment records (#977)


drsnuggles8:
## Parent

Child of #977. Was blocked by the GPU Scene substrate in #991 (landed in PR #1003, 2026-09-01).

## Goal

Add one canonical GPU Scene light/environment record shared by clustered raster lighting, next-event estimation, and future ReSTIR candidate generation, preserving stable IDs and photometric semantics.

## Scope

- Define explicitly laid-out analytic light records covering type, position/direction, color/intensity units, radius/range, shape dimensions, shadow/visibility flags, stable ID, and generation.
- Define the ownership/home for environment and future emissive-distribution records without prematurely implementing those distributions.
- Extract and incrementally upload canonical records once, with deterministic ordering, dirty ranges, safe removal/reuse, and lifecycle reset.
- Route the existing raster light consumer through an adapter over this canonical record or prove field-for-field parity until the raster-migration child removes the adapter.
- Enforce the invariants in `docs/agent-rules/light-path-photometric-parity.md` so raster, CPU/reference, and future GPU transport do not silently disagree about units or shape.

## Acceptance criteria

- Light/environment layouts have C++/shader static/reflection parity tests.
- Stable IDs survive movement and compatible edits; removal/reuse changes generation.
- Analytic type, radius/range, shape, intensity units, and shadow flags are not dropped between ECS extraction and the canonical record.
- One-light edits produce incremental dirty uploads.
- Existing raster lighting remains threshold-equivalent on OpenGL and Vulkan.
- CPU tests cover deterministic extraction, allocation/reuse, generation rollover, dirty ranges, and lifecycle reset.

## Out of scope

- Building environment/emissive sampling distributions.
- ReSTIR candidate generation.
- BLAS/TLAS or ray queries (#978).
- Migrating all specialized lighting paths.

## Score
```olo-score
capability: 8
craft: 3
stability: 8
decay: 5
effort: 3
confidence: 0.8
learning: 5
fun: 8
kano: table-stakes
blocked_by: []
blocks: [994]
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
PR #1011 implements this. Evidence per acceptance criterion:

[...]

Issue #992, opened 2026-08-31
Renderer: canonical GPU Scene material records and shader-layout parity (#977)


drsnuggles8:
## Parent

Child of #977. Was blocked by the instance/geometry substrate in #991 (landed in PR #1003, 2026-09-01).

## Goal

Add the canonical GPU Scene material record and its stable identity/versioning contract so raster, ray-query, and path-tracing consumers share one material representation rather than translating independently.

## Scope

- Define a compact, explicitly laid-out `GPUSceneMaterial` record for texture/resource indices, base-color/emissive factors, roughness/metallic/alpha data, flags, and closure version.
- Allocate stable material IDs and increment a generation/version after incompatible edits so temporal consumers reject stale material history.
- Resolve assets through the existing RHI resource-handle/descriptor-heap model without assuming full runtime-indexed bindless #805.
- Add incremental dirty material updates and lifecycle reset on scene reload/backend switch.
- Establish one C++/GLSL layout source or a strict static/reflection parity check that fails loudly when the two mirrors drift.

## Acceptance criteria

- C++ and shader material layouts agree on size, alignment, offsets, field semantics, and closure version on both backends.
- Ordinary movement does not change material identity; incompatible material edits change its version/generation.
- Editing one material uploads only that material's dirty range.
- Texture/resource handles remain valid across normal updates and are cleared/re-resolved across scene reload and backend switch.
- CPU tests cover stable allocation, reuse/version rollover, dirty ranges, and deterministic extraction.
- Shader-layout/reflection tests fail if either side changes alone.

## Out of scope

- Reworking closure math or `Evaluate`/`Sample`/`Pdf` semantics (#975).
- Canonical light/environment records (separate #977 child).
- Raster-path migration (separate #977 child).

## Score
```olo-score
capability: 8
craft: 3
stability: 8
decay: 5
effort: 2
confidence: 0.8
learning: 5
fun: 8
kano: table-stakes
blocked_by: []
blocks: [994]
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
PR #1011 implements this. Evidence per acceptance criterion:

[...]

Issue #991, opened 2026-08-31
Renderer: GPU Scene instance/geometry records, stable IDs, and incremental uploads (#977)


drsnuggles8:
## Parent

Child of #977. This is the first implementation seam required by the parent issue's explicit split.

## Goal

Establish the RHI-neutral GPU Scene foundation for ordinary mesh instances and geometry: compact CPU/GPU mirror records, stable identity across frames, deterministic extraction, incremental dirty-range upload, safe removal/reuse, and renderer-profiler telemetry.

This child owns the identity/allocation/upload substrate. It must not introduce a second Vulkan-specific scene model.

## Scope

- Define explicitly laid-out `GPUSceneInstance` and `GPUSceneGeometry` records with current/previous transforms, geometry/material references, stable instance identity, visibility/feature flags, generation, resource handles, optional device addresses, vertex/index format and primitive range.
- Define the GPU Scene ownership/lifetime boundary and build/extraction entry point so scene/ECS state is produced once per view/frame.
- Add stable slot allocation with generation-based removal/reuse, deterministic extraction order, dirty-range merging, frame-safe retirement, and capacity growth.
- Preserve previous-frame transform truth at the same ownership boundary as current transforms.
- Define explicit support-policy hooks and diagnostics for virtualized/software-raster geometry, procedural geometry, terrain, foliage, particles, fluids, and skinned meshes. Unsupported categories are counted, never silently omitted.
- Upload through existing RHI-neutral buffer/resource APIs on both OpenGL and Vulkan.
- Publish extraction time, upload bytes, live/capacity counts, fragmentation/growth, and unsupported-category counts through `RendererProfiler`.

## Acceptance criteria

- Instance and geometry layouts have compile-time C++ assertions plus shader-layout/reflection parity coverage.
- Stable IDs survive ordinary movement; removal/reuse increments generation so stale temporal identity is rejected.
- Current and previous transforms are updated deterministically and covered by CPU tests.
- A one-instance transform change uploads only the merged dirty range, not the full scene buffer.
- Scene reload, renderer shutdown, and backend switch cannot retain stale handles, slots, or generations.
- Capacity growth and retirement are frame-safe and tested without requiring a live GPU where a CPU contract test suffices.
- Profiler metrics expose extraction time, uploaded bytes, capacity/utilization, fragmentation/growth, and unsupported geometry counts.
[...]

Issue #977, opened 2026-08-30
Renderer architecture: one GPU scene for raster, ray queries, and path tracing


drsnuggles8:
## Problem

OloEngine now has OpenGL and Vulkan render backends, a render graph, deferred/forward paths, virtualized geometry, heap-backed Vulkan resources, and a CPU reference path tracer. Adding ray queries, a GPU path tracer, and ReSTIR directly to today's per-pass scene submission would create a second scene representation with different material IDs, transforms, lifetime rules, and visibility semantics.

That is the failure mode to avoid: raster, hybrid RT, and path tracing must not become separate renderers that happen to read the same ECS.

## Goal

Create one canonical, frame-versioned **GPU Scene** produced once from scene/ECS state and consumed by:

* raster/deferred/forward rendering;
* GPU culling and virtualized-geometry integration;
* Vulkan BLAS/TLAS construction;
* ray-query effects;
* the GPU path tracer;
* temporal validation and ReSTIR.

The data model is RHI-neutral. Backend-specific code decides how buffers, device addresses, descriptor-heap entries, and OpenGL bindings are exposed.

## Effort note — this is expected to split

Scored effort **13**, which per `docs/process/issue-scoring.md` §4 means **split it** (the anchor is #435, "three rendering sub-features"). This issue defines the contract and the acceptance bar; the implementation should land as sub-issues along roughly these seams:

1. instance + geometry records, stable identity, dirty-range upload, profiler counters;
2. material records and the shader-layout parity mechanism;
3. the canonical light/environment record;
4. the first raster-path migration and the anti-duplication ratchet.

File those as children rather than attempting one PR.

## Children

- [x] #991 — instance/geometry records, stable IDs, dirty-range upload, profiler counters
- [x] #992 — canonical material records and C++/shader layout parity
- [x] #993 — canonical light/environment records and photometric parity
- [x] #994 — first raster-path migration and anti-duplication ratchet

## Canonical records

Define compact, explicitly laid-out records along these lines:

```cpp
struct GPUSceneInstance
{
    Matrix3x4 CurrentTransform;
    Matrix3x4 PreviousTransform;
    u32 GeometryIndex;
    u32 MaterialIndex;
    u32 InstanceId;
    u32 VisibilityMask;
    u32 Flags;
    u32 Generation;
};

struct GPUSceneGeometry
{
    ResourceHandle VertexBuffer;
    ResourceHandle IndexBuffer;
[...]

PR #1026, opened 2026-09-03
feat(renderer): render the classic mesh path through GPU Scene records

### LINKED ISSUE #994: Renderer: migrate one raster path to GPU Scene and add an anti-duplication ratchet (#977)
## Parent

Final integration child of #977. Blocked by #992 and #993 (#991 landed in PR #1003).

## Goal

Convert one existing raster path to consume canonical GPU Scene instance, geometry, material, and light records with no visual regression, while adding the adapter and ratchet that prevent new direct scene-layout duplication.

## Scope

- Select one representative ordinary-mesh raster path shared by OpenGL and Vulkan.
- Build the GPU Scene once per view/frame at a documented ownership boundary and bind its semantic buffers through the existing RHI facade.
- Consume current/previous transforms and canonical material/light identity from GPU Scene records.
- Keep explicit legacy adapters for unmigrated specialized renderers; unsupported categories remain visible in diagnostics.
- Add a structural ratchet test that rejects new direct scene/ECS layout duplication outside named legacy adapters.
- Record threshold-equivalent output and performance/capacity telemetry.

## Acceptance criteria

- One production raster path renders through GPU Scene on OpenGL and Vulkan.
- Multi-angle visual evidence and live editor/MCP verification show no threshold-significant regression.
- Motion vectors use the same current/previous transform truth future RT consumers will use.
- Scene reload/backend switch cannot leave stale bindings or generations.
- Legacy paths have explicit adapters and diagnostics; unsupported geometry never disappears silently.
- A ratchet test prevents new direct scene-layout duplication and names the remaining approved legacy adapters.
- Extraction/upload timings and GPU Scene memory/capacity remain visible in `RendererProfiler`.

## Out of scope

- Migrating every specialized renderer.
- BLAS/TLAS or ray queries (#978).
- GPU path tracing or ReSTIR.

## Testing / visual verification

Follow the renderer verification contract: CPU/contract coverage, both RHI backends, multi-angle screenshots, a live editor run with shader/log inspection, and MCP capture of relevant intermediate targets.

## Score
```olo-score
capability: 8
craft: 5
stability: 8
decay: 5
effort: 3
confidence: 0.5
learning: 8
fun: 8
kano: table-stakes
blocked_by: []
blocks: [978]
```

drsnuggles8:
The GPU Scene epic (#977) built canonical instance, geometry, material, light and environment
records in #991 / #992 / #993, and **nothing consumed them**. This makes one production raster path
read its transforms and its material from the records, on both backends, and adds the ratchet that
stops the next path growing a second scene representation.

## The path, and why this one

**`Scene::SubmitMeshSourceClassic` → `Renderer3D::DrawMesh` → `DrawMeshCommand` → `CommandDispatch`**
— the classic ordinary-mesh (`MeshComponent`) path, plus the virtual-geometry fallback that shares
it. It is the right first migration because it is the only mesh path that is *both* shared by
OpenGL and Vulkan and used by all three rendering paths (Forward, Forward+, Deferred), so one
change is exercised by every existing golden. Virtual geometry, foliage, terrain and the
explicitly-instanced path are out of scope by the issue and stay named legacy adapters.

## How it works: a draw link

Extraction and submission are interleaved — `Scene` walks the registry once and, per submesh,
stages a record *and* builds the draw packet. Slots are only assigned at commit, in key order, so a
draw cannot know its slot at submit time. It carries a **link index** into a per-frame table
instead, and the table is resolved in one pass right after the commit:

| step | what happens |
|---|---|
| `BeginGPUSceneExtraction` | the frame's link table is cleared |
| `ExtractGPUSceneMesh` | stages a record, appends its key, returns a link index |
| `DrawMesh` | carries that index in the command packet |
| `EndScene` | `EndExtraction` → `Upload` → **`ResolveGPUSceneDrawLinks`** |
| command dispatch | reads resolved links only; never resolves a key itself |

That ordering **is** the "built once per view/frame at a documented ownership boundary" criterion,
written as code in `GPUScene/GPUSceneDrawLink.h`. It is also what makes the transforms usable as
they are: the record encoded them with this frame's render origin, so the dispatcher must *not*
shift them again.

A link that does not resolve still draws. Both generation lanes of `InstanceData::GPUSceneRef` read
zero — the invalid generation on both sides of the mirror — and the draw falls back to exactly what
it used before.

## Motion vectors: a bug fixed, not just a new source

The per-entity previous-transform cache is keyed on entity id and written on **every** `DrawMesh`
call, so a multi-submesh mesh overwrote its own history between submeshes: submesh 0 reported the
real motion and submeshes 1..N-1 reported zero velocity. The instance record is keyed per
`(entity, geometry, submesh)` — the same identity a future TLAS / ReSTIR consumer resolves — so
raster motion and the RT contract now read one truth. Pinned by
`GPUSceneRasterMigration.EverySubmeshKeepsItsOwnPreviousTransform`.

## GPU side: one kind at a time

`include/GPUScene.glsl` stops declaring storage blocks; each record kind gets its own one-line
declaration file. A consumer takes only what it reads, which is what lets `PBR_GBuffer.glsl` take
the canonical **material** table on slot 17 while it still reads per-draw `InstanceData` on 15.
`oloGPUSceneMaterial` validates the record's **generation** before reading the slot — an index
without its generation is a stale-slot read, since a retired slot is reused after two frames.

Slot 17 is shared with the instance cull's draw-indirect buffer, so `GPUFrustumCuller` now calls
`CommandDispatch::InvalidateGPUSceneMaterialBinding()` when it takes the slot, and a draw only
presents its material link when the bind actually happened.

## The ratchet

`GPUScene/GPUSceneLegacyAdapters.h` names all 12 source files that still assemble previous-frame
transform truth outside the records, each with what it is and **what has to happen before it can
go**. `GPUSceneAntiDuplicationRatchetTest` enforces the list in both directions: an unnamed
duplicator fails, *and* a named file that stopped duplicating fails — so the list shrinks as paths
migrate instead of turning into decoration.

## Acceptance criteria

| criterion | evidence |
|---|---|
| One production raster path renders through GPU Scene on OpenGL **and** Vulkan | Live editor, `MaterialSpheres.olo` on Deferred, both backends — see below |
| Multi-angle visual evidence, no threshold-significant regression | 4 committed goldens `OloEditor/assets/tests/visual/GPUSceneRaster_{Front,Left,Right,Above}.png`; **full suite green twice**, every pre-existing golden unchanged |
| Motion vectors use the same current/previous transform truth as future RT consumers | `GPUSceneRasterMigration.EverySubmeshKeepsItsOwnPreviousTransform`, `.DecodedTransformIsExactlyTheRenderRelativeMatrix` |
| Scene reload / backend switch cannot leave stale bindings or generations | `.SceneReloadLeavesNoResolvableLink`, `.BackendSwitchInvalidatesLinksTakenOnTheOldDevice`; `ResetGPUScene` drops the link table |
| Legacy paths have explicit adapters and diagnostics; unsupported geometry never disappears | `GPUSceneLegacyAdapters.h` (12 named, 2 carrying an unsupported category); unresolved links fall back and are counted |
| A ratchet prevents new direct scene-layout duplication and names the approved adapters | `GPUSceneAntiDuplicationRatchetTest` (3 cases, both directions + entry quality) |
| Extraction/upload timings and memory/capacity remain visible in `RendererProfiler` | unchanged, plus two new rows: links resolved/unresolved and draws consumed/fell back |

## Verification

- **Full suite, twice** (before and after the self-review fixes): 7086 → **7087 tests, 0 failures**
  (42 / 40 skipped, all environment-gated). Every existing golden image test passed unchanged,
  which is the real "no visual regression" evidence — the migration is expected to be
  pixel-identical because the record and the per-draw UBO are both built from the same `Material`.
- **Live editor, OpenGL**, `MaterialSpheres.olo`: correct on Deferred from two angles and on
  Forward; `olo_shader_errors` = 0; no shader/pipeline errors in `OloEngine.log`.
- **Live editor, Vulkan** (`--rhi=vulkan`), same scene on Deferred: correct, `olo_shader_errors` = 0,
  no validation errors (only pre-existing missing-font warnings).
- **A found-and-fixed Vulkan crash**, worth calling out because it is the reason the live run
  mattered: the first version bounds-checked the material slot with `g_GPUSceneMaterials.length()`.
  That compiles to `OpArrayLength`, which the Vulkan RHI's
  `VK_DESCRIPTOR_MAPPING_SOURCE_INDIRECT_ADDRESS_EXT` mapping rejects at `vkCreateGraphicsPipelines`
  — the editor then died on the null pipeline. **Every OpenGL test and the whole suite stayed
  green.** The generation is the bound instead; written up as `glsl-shaders.md` §6b and pinned by
  `GPUSceneLayoutTest.NoGPUSceneIncludeAsksAStorageBufferForItsLength`.

## Deliberately not done, and why

**The canonical light table is not read by a shader.** `GPUSceneLight` lives on SSBO slot 9, which
is `SSBO_FPLUS_POINT_LIGHTS`; every raster light evaluator — `DeferredLighting.glsl` and
`PBR_MultiLight.glsl` alike — reaches `ForwardPlusCommon.glsl`, which declares that slot. Taking the
canonical table therefore *requires* retiring the Forward+ per-type light buffers and rewriting the
clustered evaluator first, which is its own change and its own risk. Light identity is still
consumed through the canonical record on the CPU: `Scene` reads each authored light exactly once
into a `GPUSceneLight` and derives every raster struct from it through `GPUSceneLightAdapter`
(#993, pinned field-for-field by `GPUSceneLightAdapterTest`). I've left the aliasing comments
accurate rather than claiming the retirement happened. This is the obvious next child of #977.

Also unmigrated by design: `DrawMeshInstanced` and `CommandBucket` auto-batching, both of which need
a per-instance link lane in `InstanceData` before N sources can each name their own record. Named in
`GPUSceneLegacyAdapters.h` with that exit condition.

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Renderer/Commands/CommandDispatch.cpp` (`UploadModelInstance`) — the
   whole migration lands in one branch. The claim that the record's transform is *already*
   render-relative and must not be shifted again is the single thing that would silently move every
   migrated mesh; it rests on `s_Data.RenderOrigin` being the same value at extraction and at
   dispatch. Pinned bitwise by `DecodedTransformIsExactlyTheRenderRelativeMatrix`, but that test
   assumes the origin, it does not prove it.
2. `OloEditor/assets/shaders/include/GPUSceneMaterialResolve.glsl` — no bounds check by design (see
   above). If the generation is ever set without the record being in the uploaded range, this reads
   past the table. I argue that cannot happen because `Upload` grows the buffer before any draw
   runs, but it is an invariant held by convention rather than by a check.
3. `OloEngine/src/OloEngine/Renderer/GPUScene/GPUSceneDrawLink.h` + `Renderer3DGPUScene.cpp` — link
   indices are positions in a per-frame vector handed to command packets. `BeginGPUSceneExtraction`
   and `ResetGPUScene` are the only two places that clear it; a third clear, or a missed one, hands
   this frame's draw last frame's record.

**What I verified, and how** — the full suite twice (7087 tests, 0 failures, every pre-existing
golden unchanged); four committed multi-angle goldens I looked at, which show the three
differently-coloured cubes in the correct relative positions from front, both sides and above;
live editor runs on OpenGL *and* Vulkan with `olo_shader_errors` = 0 and clean logs. The check that
would have failed if this were wrong: `ClassicMeshPathRendersThroughTheRecordsFromEveryAngle`
asserts on `CommandDispatch::GetGPUSceneConsumedDrawCount()`, not on extraction — a total fallback
would produce byte-identical PNGs and this is what refuses to pass anyway.

**Least confident about** — the material-table bind on slot 17. Only `GPUFrustumCuller` binds that
slot today and both its sites now invalidate the cache, but the cache is frame-scoped
(`ResetState`), not pass-scoped, so a *new* consumer of slot 17 that forgets to invalidate would
make a linked draw read its buffer as material records. The generation check would reject almost
any such read, but "almost" is doing work there. A pass-scoped bind would be sturdier and I did not
build one.

**Deliberately not tested** — the migrated shader's *Vulkan pipeline* is never created by a test.
This is the sharp one, and it is the gap that let the `OpArrayLength` bug above survive a green
suite, so it is worth stating precisely rather than as "Vulkan is untested":

- There **is** substantial device-backed Vulkan coverage — 126 tests across `VulkanPassSuite` (52),
  `VulkanDrawPath`, `VulkanResourceFactory`, `VulkanShaderPipeline` and friends — and it ran as part
  of the full suite on this branch, on a real device.
- None of it creates a pipeline for `PBR_GBuffer.glsl`. `VulkanPipelineBuilder` returns
  `VK_NULL_HANDLE` and logs on failure, so a test that asks it for a pipeline for each mesh/G-Buffer
  shader and asserts non-null would have caught the crash. Filed as **#1029** for the whole
  G-Buffer shader family rather than folded in here.
- It cannot be closed on the AMD self-hosted runner. That box already has the Vulkan SDK, RADV and
  lavapipe, and `gpu-conformance-amd.yml` already builds and runs the suite with Vulkan enabled —
  but every device-backed Vulkan test **skips** there with *"No device satisfies the ADR 0010
  capability contract (`VK_EXT_descriptor_heap` et al.)"*. Mesa 25.2.7's RADV exports 136
  `VK_EXT_*` strings including `descriptor_buffer` and `descriptor_indexing`, and does **not**
  export `descriptor_heap`. So Vulkan CI coverage is blocked on driver support, not on runner
  configuration — nothing to install.

Meanwhile the cheap structural guard for this specific class does run everywhere:
`GPUSceneLayoutTest.NoGPUSceneIncludeAsksAStorageBufferForItsLength`.

The auto-batched path (a linked draw collapsed into an instanced call) is also not exercised by any
test scene I added; it falls back by construction and the profiler counts it, but no test pins that.

Closes #994

## Summary by CodeRabbit

* **New Features**
  * Classic mesh rendering now uses resolved GPU Scene records for transforms and materials.
  * Added per-frame renderer metrics showing linked, unlinked, GPU Scene-consumed, and fallback draws.
  * Added generation tracking to help validate scene records across reloads and rendering backend changes.

* **Bug Fixes**
  * Improved rendering reliability by falling back safely when GPU Scene records cannot be resolved or materials are unavailable.
  * Preserved per-submesh previous transforms for more accurate motion-vector rendering.


drsnuggles8:
@coderabbitai review

Issue #722, opened 2026-07-31
Renderer: depth-aware cluster culling — active-cluster compaction + 2.5D depth-cell mask


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

`OloEditor/assets/shaders/compute/LightCulling.comp` states the design at :14-15 — *"there is NO depth-texture input: every cluster of the fixed grid is culled, so the pass is depth-prepass independent."* That independence was a deliberate and defensible choice, but it has two costs:

1. Every cluster in the fixed grid is culled and gets a light list, including clusters containing no geometry at all — pure waste, and it scales with grid resolution rather than with scene content.
2. Within a cluster that *does* contain geometry, a light sitting in an empty depth gap (in front of or behind the actual surfaces) still passes the frustum test and is shaded against.

We do have a depth prepass (`DepthPrepass.glsl`, `DepthPrepass_Mask`, `_Skinned` variants), so the information is available — it is simply not consumed.

**Describe the solution you'd like**

An **optional** depth-aware path, gated on the depth prepass being enabled, leaving the current depth-independent path as the fallback:

- Per screen tile, a shared-memory min/max depth reduction, plus a **32-bit depth-cell occupancy mask**: quantise each pixel's depth into one of 32 cells between the tile's min and max and `atomicOr` the bit. A light whose depth extent intersects only zero bits is rejected — this is the classic 2.5D culling optimization and it is very effective in scenes with large depth discontinuities.
- **Active-cluster compaction**: mark clusters actually containing geometry, compact them into a list, and dispatch culling indirectly over just those.

**Describe alternatives you've considered**

Raising cluster Z-slice count — narrows the depth gaps but multiplies the number of clusters, which is the cost this is trying to reduce. #683 (VRCS) attacks the *shading* cost of the same passes and is complementary, not an alternative.

**Additional context**

Technique: Harada's 2.5D culling. A compact reference implementation combining tile min/max, the 32-bit depth mask and the tile cluster AABB in one pass is `D:\repos\tachyonfire\th1\shaders\cheap_ao_tile_prepare.comp` (shared-memory reduction at :150-160, depth mask at :181-187) — **GPLv3, read for approach only.**

Related: #435 (the clustered lighting this extends), #683, #684.

**Acceptance criteria**

- Light-culling GPU time drops measurably in a scene with many local lights and large depth discontinuities.
- Lit output is identical to the depth-independent path (this is a culling optimization, not a shading change) — verified by image diff.
[...]

PR #1027, opened 2026-09-03
feat(renderer): analytic sphere-proxy ambient occlusion for off-screen occluders

### LINKED ISSUE #710: Renderer: analytic sphere-proxy ambient occlusion for off-screen occluders

All our ambient occlusion is screen-space — `GTAORenderPass` (+ `GTAO.comp` / `GTAO_Denoise.comp`) and `SSAORenderPass`. Screen-space AO is structurally blind to any occluder that is not in the depth buffer: a large object just off the edge of the frame, or behind the camera, casts no occlusion at all, so contact darkening pops as the camera pans.

A complementary analytic AO term from coarse sphere proxies:

- Each significant occluder gets one or more bounding spheres (authored, or auto-fitted from the existing bounding volumes).
- Rasterize each proxy's screen extent and accumulate the closed-form sphere-occlusion integral per pixel — no sampling, no noise, no ray budget.
- Bin proxies per screen tile (reuse the froxel grid from #435) so a pixel only evaluates spheres that can affect it.
- Combine multiplicatively with GTAO: GTAO keeps the fine detail, proxies supply the large-scale term GTAO cannot see.

Widening the GTAO radius — costs more samples and still cannot see outside the depth buffer. Baked AO — static only, and we do not have a bake path yet (#439).

**Additional context**

The closed-form integral is Inigo Quilez's, published at <https://iquilezles.org/articles/sphereao> — implement from there. An integrated example (proxy insert pass + tiled prepare + shade) exists at `D:\repos\tachyonfire\th1\shaders\cheap_ao.frag`, `cheap_ao_tile_insert.{vert,frag}`, `cheap_ao_tile_prepare.comp`, `cheap_ao_tile_shade.frag`, CPU side `src/fn_engine/th_occlusion.h` — **GPLv3, read for approach only.** That repo also has `directionalocc.frag` / `directionalocc_binary.frag` if a directional (bent-normal) variant is wanted later.

**Acceptance criteria**

- A large object moving off-screen no longer causes its contact shadow to pop away; verified with a scripted camera pan.
- Cost is a small fraction of GTAO's; no added temporal noise (the term is analytic).

## Score
```olo-score
capability: 3
craft: 5
stability: 1
decay: 1
effort: 3
confidence: 0.5
learning: 5
fun: 5
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
Screen-space AO is structurally blind to anything outside the depth buffer: a large object just off the edge of the frame — or behind the camera — casts no occlusion at all. This adds the complementary analytic term, and combines it multiplicatively with GTAO so GTAO keeps the fine contact detail while the proxies supply the large-scale occlusion it cannot see.

## What landed

**The integral.** Inigo Quilez's closed-form sphere occlusion ([sphereao](https://iquilezles.org/articles/sphereao)), implemented from the article in `SphereProxyAO.h` with a textual mirror in `SphereProxyAO.comp`. Two regimes — the cone entirely above the receiver's tangent plane (`nl / h²`) and the horizon-clipped case — which agree exactly at the boundary, so the term is continuous. The GPLv3 reference the issue names was not consulted.

**Proxy source.** Auto-fitted from the world AABBs the frame's DDGI caster funnel already builds. That source is chosen over the two obvious alternatives for reasons that are load-bearing rather than incidental: the lit submit path is view-frustum culled, so harvesting from it would drop exactly the occluders this feature exists to keep, and the shadow-caster list is gated on a shadow-casting light, so an artist turning a shadow off would silently turn the occlusion off with it. No edit to `Scene.cpp` or to the ordinary-mesh submit path.

**The fit.** Volume-matched *and contained* — the radius is capped at the box's smallest half-extent — with an anisotropy-driven split of up to four spheres along the long axis, so a corridor gets a line of proxies rather than one sphere that is far too fat at the ends.

**Binning.** A dedicated per-16×16-tile pass, not the #722 cluster grid. The cluster grid is bounded by the view frustum, so a proxy beside or behind the camera has no cluster to live in — it is structurally the wrong container for this feature. Instead each workgroup builds its tile's view-space AABB from the tile's own depth range and tests proxies against it in 3D. Nothing projects, so nothing clips, and off-screen proxies survive.

**Combine.** A read-modify-write of `AOBuffer` in place, after the AO producer and before `DeferredLightingPass`. Every existing consumer of that resource picks the combined value up with no change of its own, and "disabled" means the node contributes nothing rather than a branch inside it.

Bound to the GTAO technique: SSAO's AO buffer is half-resolution RG16F while GTAO's is full-resolution R8UNorm, and a storage image's format is part of its binding contract. Under SSAO the pass declines and says so once.

The params block takes `UBO_SPHERE_PROXY_AO = 64`, one of the last free numbers in a namespace `ShaderBindingLayout.h` documents as full; the 128-entry proxy array rides the same block rather than claiming an SSBO the namespace cannot give. The within-shader rule holds and is asserted by a test.

## Acceptance criteria

**"A large object moving off-screen no longer causes its contact shadow to pop away; verified with a scripted camera pan."** — `SphereProxyAOVisualEvidenceTest.ContactDarkeningSurvivesTheOccluderLeavingTheFrame`. The pan is derived, not hard-coded: the test searches for a yaw that puts every corner of the occluder's AABB off screen while keeping the receiver inside, and fails if none exists. It then measures the proxy term's *contribution* at the receiver (a within-pose difference, so the camera move and the tone curve cancel) on both sides of the pan and asserts it survives. `SphereProxyAO_ProxyTerm_OffFrame.png` is the picture worth looking at: the occluder is entirely out of frame and its ambient contact darkening is still on the floor.

**⚠️ One thing I could not reproduce, and said so in the test rather than papering over it.** The issue frames the criterion as the *screen-space* term's contact shadow popping. It does not pop on this engine, because GTAO here barely produces that contact shadow to begin with — measured 88% of its (already small) receiver darkening retained across the pan, i.e. what it loses is noise. `GTAOVisualEvidenceTest` already records the same gap in its own words ("GTAO's horizon search here does not yet produce a strongly visually distinct darker band at this cube/floor junction"). Asserting on that difference would have been asserting on GTAO's noise floor, so the test measures the proxy term's own contribution and its invariance across the pan instead. The header says all of this at the top.

**"Cost is a small fraction of GTAO's; no added temporal noise."** — No noise: `ProxyTermIsTemporallyStable` asserts two captures of an identical proxy-on frame are *byte-identical*, which is why the tile bin result is a bitmask rather than an atomicAdd-compacted list (the shading loop is an order-dependent float product). On cost I am reporting structure, not a number: one 16×16 compute dispatch reading depth + normals with a shared-memory bin over at most 128 spheres, against GTAO's HZB build + 9 slices × 3 samples + four bilateral denoise passes. It has its own `GPUPassTimerPool` sub-pass bracket ("SphereProxyAO") so the number is separable, but I did not measure it — see *Deliberately not tested*.

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Renderer/SphereProxyAO.h:182` — `SelfOcclusionFade`. This is the one modelling decision with no clean answer. A proxy sits inside its own object, where the closed form's honest answer is "fully enclosed" (occlusion 1) — which painted a visible dark disc across the occluder's front face in the first build. The fix is containment *plus* a narrow ramp, and the ramp width trades self-occlusion artefacts against contact occlusion it deletes: at 0.35 with an uncontained fit it hollowed out the whole contact ring around an object's base. 0.15 with a contained fit is where I landed, and I would like a second opinion on it.
2. `OloEditor/assets/shaders/compute/SphereProxyAO.comp:250-285` — the tile AABB. The four corner rays are swept between the tile's own min/max depth, and the min/max over all four `{lo,hi} × {minZ,maxZ}` products is what makes it correct when `lo` is negative. An off-by-one or a dropped sign here does not crash, it just quietly stops binning some proxies, and the symptom is a term that is subtly too weak.
3. `OloEngine/src/OloEngine/Renderer/RenderPipeline.cpp:221` — arming the collection at `PrepareFrame`. It must be at BeginScene (the bounds are gathered during the traversal that follows) and it must NOT be in the pass's `Setup()`, which is fingerprint-cached and would freeze the list at whatever the first frame saw. Gated on all three of the pass's own conditions because arming it makes `Scene.cpp` build a full `DDGIMeshCaster` per static opaque submesh.

**What I verified, and how**

- `SphereProxyAOMathTest` — 21 CPU contract tests pinning the integral against configurations whose answers geometry forces, not the implementation: the form factor on the normal axis, a tangent sphere overhead (1.0), a tangent sphere centred in the tangent plane (exactly 0.5), a sphere below the horizon (0), continuity across the branch boundary, monotonicity, rigid-transform invariance, and **a sphere behind the camera producing the same occlusion as its mirror image in front of it** — the property no screen-space term can have. Plus the fit's containment invariant over five box shapes, and CPU↔GLSL mirror checks on the proxy-array length, the fade width, the UBO binding and the binding-64 namespace rule.
- `SphereProxyAOVisualEvidenceTest` — 2 GPU tests, 7 PNGs committed under `OloEditor/assets/tests/visual/SphereProxyAO_*.png`. **I looked at every one of them**, and the first build's images are what found the self-occlusion disc — no assertion caught it.
- Full suite: **7099 tests, 7059 passed, 0 failures**, exit 0.
- Live editor (OpenGL): `SphereProxyAO` compiles at startup (`OloEngine.log`: "Compiled compute shader 'SphereProxyAO'", `Success: true`), the pass initialises at 1920×1080, `olo_shader_errors` returns 0 with the pass enabled and dispatching, all six settings are reachable over MCP, and toggling the pass changes the frame in a loaded scene.
- `/code-review` at high effort before the PR existed: 8 findings, all 8 fixed (see the self-review comment).

**Corrected after the first commit** — see the second review comment. A live check on Sponza showed the term saturating to near-black across the atrium: the influence cutoff admitted dozens of ~1% contributors and the visibility product compounded them. Fixed at source (a tighter cutoff, plus an influence window that makes the hard binning threshold invisible). The lesson is now in `docs/agent-rules/notes-renderer.md`: **a one-occluder test scene cannot see an accumulation bug** — nothing in the 23 tests green at the first commit could have caught it.

**Least confident about**

The fade width, as above. Second: the `MaxRadius = 25` and `InfluenceScale = 4` defaults. Both are now measured against Sponza rather than reasoned about, which is better than where they started, but they remain two numbers tuned on one dense scene — a project at a different world scale will want different ones. (An earlier draft of this section claimed `MaxRadius` rejects all of Sponza's architecture. That was speculation from a bad camera angle and it is wrong: the term does plenty of work there.)

**Deliberately not tested**

- **The cost number.** The issue asks for "a small fraction of GTAO's" and I have not measured it. Debug builds cannot answer this (see `docs/agent-rules/` on `/MDd` checked iterators), and I did not want to report a Debug figure as if it meant something. The sub-pass bracket is in place so the measurement is a Release capture away.
- ~~**Vulkan.**~~ **Now verified — the original claim here was wrong.** I wrote that the driver could not launch the Vulkan backend; in fact only `driver.ps1` lacks an `--rhi` passthrough, and a ~30-line launcher covers it. Both backends were checked at the same pose and agree. See the second review comment.
- **Authored proxies.** The issue allows "authored, or auto-fitted"; only the auto path is here, which the plan called for.

Closes #710

## Summary by CodeRabbit

- **New Features**
  - Added analytic sphere-proxy ambient occlusion for more consistent contact darkening, including debug visualization.
  - Added controls for enabling the effect and tuning strength, proxy limits, radius, influence, occlusion, and debug view.
  - Added scene persistence and runtime configuration for the new ambient-occlusion settings.

- **Bug Fixes**
  - Corrected documentation links so referenced resources resolve correctly.

- **Tests**
  - Added coverage for AO mathematics, rendering behavior, temporal stability, and visual evidence.


drsnuggles8:
## 🤖 Self-review @ `05881018ffd7067b0ff0f73fda57a0727785f3f0`
Reviewed the PR diff at **high** effort with `/code-review` before the PR existed, then fixed every finding and re-ran the full suite (7099 tests, 0 failures).

- **Findings:** 8 · **Fixed:** 8
  1. `SelfOcclusionFade` was distance-to-centre only, so it hollowed out the contact ring around an object's base, not just the object's own surface — proxies are now fitted **contained** (radius capped at the box's smallest half-extent) and the ramp narrowed 0.35 → 0.15, which is a thin guard around the tangent point instead of a hole. New `EveryFittedSphereIsContainedInItsBox` pins the invariant.
  2. The CPU↔GLSL mirror guard opened one hard-coded relative path and `GTEST_SKIP`ped on a miss — but two fixtures in this binary permanently chdir into `OloEditor/`, so in a full-suite run the guard could silently stop guarding depending on test order. Now tries three roots.
  3. Proxy collection was armed on `SphereProxyAOEnabled` alone, so a full `DDGIMeshCaster` was built per static opaque submesh even when the GTAO gate would discard the list. Now gated on all three of the pass's own conditions, all knowable at BeginScene.
  4. The tile proxy list was `atomicAdd`-compacted while the accumulation is an order-dependent float product — a latent flake against the byte-identical stability assertion. Replaced with a **bitmask** iterated in index order, deterministic by construction.
  5. A negative `SphereProxyAOMaxProxies` (reachable via the panel's CTRL+Click type-in) cast to `u32` and selected the full 128 budget instead of none. `std::max` before the cast.
  6. `SelectProxies` reserved for the budget while the loop can push `bounds.size() * 4`; reserve now matches the loop, capped.
  7. Panel drag ranges (500 / 64) were narrower than the MCP and sanitizer ranges (100000 / 1000), so a valid saved value was truncated on first touch. All three surfaces now carry the same 1000 / 64.
  8. `retained` divided by a denominator only checked with non-fatal `EXPECT_GT`, producing NaN failures that would bury the real one. Promoted to `ASSERT_GT`.

- **Dismissed:** none.

The review also flagged untracked `SupportedTargetTestGame/`, `OceanFFT_*.png` and `fog1008/` artifacts in the tree — other tests' outputs, deliberately not staged.

drsnuggles8:
Second commit, from a live-editor check on **Sponza** that the synthetic evidence scene structurally could not make — with one occluder there is nothing to compound; in a real scene there is.

**At the first commit's defaults the term saturated to near-black across the whole atrium.**

The cause was the influence cutoff, not the combine. At scale 12 a proxy stayed binned while it could still contribute as little as `1/144`, so every tile in dense architecture carried dozens of negligible contributors — and combining them as independent occluders (the product of visibilities, correct only for occluders that do not overlap) compounded them to zero. A coarse sphere also over-covers the object it stands for, so the error only ever goes one way.

Tightening the cutoff alone trades one artefact for another: the binning test is hard, so a proxy just past the threshold is in one tile's list and absent from its neighbour's — **clear 16x16 tile seams at scale 2.5**. `InfluenceWindow` removes that class of artefact instead of hiding it: the occlusion goes smoothly to zero at exactly the radius the binning tests, so whatever the binning drops was already worth nothing. That is what makes the tighter cutoff safe.

`SphereProxyAOMaxOcclusion` now defaults to 1.0 (the raw product). My first attempt at this capped it at 0.6, which just moved the saturation — enclosed spaces went flat at the ceiling instead of shading.

**Verified on both backends.** I had claimed in the PR body that Vulkan was out of reach; that was wrong — only `driver.ps1` lacks an `--rhi` passthrough, and a ~30-line launcher covers it. OpenGL and Vulkan produce the same term at the same pose, so the saturation was never backend-specific, and the Vulkan gap in *Deliberately not tested* no longer applies.

- Full suite after the fix: **7101 tests, 7061 passed, 0 failures**, exit 0.
- New contracts: `InfluenceWindowReachesZeroAtTheBinningCutoff` and `ProxyContributionVanishesAtTheCutoffForEveryScale` — the second asserts through `ProxyOcclusion` for four influence scales, because it is the accumulated term the binning drops, not the window in isolation.
- `olo_shader_errors` = 0 on both backends with the pass dispatching.

The lesson generalises past this PR and is why the live check is worth its cost: **a one-occluder test scene cannot see an accumulation bug.** Nothing in the 23 tests that were green at the first commit could have caught this.

drsnuggles8:
@coderabbitai review

Issue #943, opened 2026-08-24
Water: boat wake/foam trail has hard-edged, aliased boundary


drsnuggles8:
## What

The foam/wake trail a moving boat leaves on the water surface has a visibly hard-edged, sawtooth/stair-stepped boundary instead of a smooth or anti-aliased one. Zoomed screenshot evidence from Drift's boat under sail (issue #878/#899): a jagged, dark trailing patch behind the hull with a distinct pixel-staircase edge, not a soft foam feather.

Found during live-editor verification for #881 (unrelated discovery-loop feature) — initially misread as a shadow artifact (Drift's `Sun.CastShadows` is authored `false` as a workaround for #901, so it can't be a CSM shadow); on a second look it's the water's wake/foam interaction mask around the hull.

## Repro

1. `olo_scene_open Scenes/Drift.olo`, `olo_scene_play` (or just open normally and press Play).
2. Let the boat sail a few seconds under wind (auto-trim; no input needed).
3. Screenshot from behind/above the boat and zoom into the trailing wake edge.

## Where I looked (not root-caused — just narrowed the area)

Water/foam rendering lives under `OloEngine/src/OloEngine/Renderer/Passes/WaterRenderPass.cpp`; didn't trace the exact foam-mask shader given time budget. Likely candidates: a hard `step()`/threshold comparison in the foam-mask shader where a `smoothstep()` (or a soft falloff against a signed distance / depth-difference field) would read cleanly, or a low-resolution interaction/foam texture being point-sampled without bilinear/soft blending near the boat.

## Acceptance

- Wake/foam edge reads as soft/anti-aliased at typical camera distances (the follow-camera's usual 9-13 m boom, matching #901's repro distance), not a visible pixel-staircase.

## Score
```olo-score
capability: 1
craft: 3
stability: 1
decay: 1
effort: 1
confidence: 0.8
learning: 1
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
Root-caused and largely fixed in #950 — but **please don't close this yet**, there's a residual (bottom).

## The premise was wrong, in a useful way

[...]

PR #1041, opened 2026-09-03
feat(water): shoal, refract and break waves against a seabed depth field (#1033)

### LINKED ISSUE #1033: Water: shore wave deformation — shoaling, refraction and breaking against a seabed depth field
Extracted from `docs/design/WATER_FUTURE_IMPROVEMENTS.md` §5.3 so that doc can stop being a backlog. Verified against the code 2026-09-03: nothing implements shoaling, refraction or breaking — there is no seabed depth input to the wave evaluation at all (`grep -i "shoal\|seabed\|waterDepthMap"` finds only unrelated caustics and AI-flocking hits).

## What's missing

Waves currently have the same shape everywhere. Approaching a shallow shore they should:

- **Shoal** — phase speed falls with depth, so the wave slows and bunches up.
- **Grow and steepen** — amplitude rises as the wave compresses.
- **Refract** — wavefronts turn toward the shore (Snell's law for water waves), so waves arrive parallel to the beach regardless of offshore heading.
- **Break** — detected via the Jacobian (already computed for FFT foam) or a steepness threshold.

## Why this one matters more than the rest of the water backlog

Drift (#878) is an island-sailing game. Every island shoreline in it currently meets deep-water waves that neither slow, turn, nor break — which is the single most visible way the water reads as "a shader" rather than a sea. The rest of the deferred water work is polish; this is a silhouette-level behaviour.

## Prerequisite

A seabed depth field the wave evaluation can sample per-vertex. That is the real cost of this issue and should be designed first — options range from a per-island authored depth texture to sampling the terrain heightfield the islands already use. The wave math is cheap once depth is available.

## Notes

- The Jacobian-based fold detection already exists in `OceanFFTField` for foam; breaking can reuse that signal rather than inventing a second one.
- Whatever is added must be evaluated in *both* the vertex/tessellation displacement chain and the surface-depth capture, which deliberately replay the same displacement (`WaterVertexStage.glsl` header comment) — a drift between them shows up as depth-test artefacts, not as wrong waves.
- Foam already has a `shorelinePower` term for the visual band; that is a shading effect and is not the same thing as geometric shoaling.

## Acceptance criteria

- Waves visibly slow, steepen and turn toward shore as depth decreases, verified on a Drift island from multiple angles per the CLAUDE.md rendering rule.
- The colour pass and the surface-depth capture agree (no new depth artefacts at the waterline).
- A CPU contract test pins the shoaling/refraction math against known configurations (deep-water limit reduces to current behaviour).

## Score
```olo-score
capability: 5
craft: 8
stability: 1
decay: 1
effort: 5
confidence: 0.5
learning: 8
fun: 8
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #1033.

Every shoreline in Drift met deep-water waves that neither slowed, turned nor broke — the single most visible way the sea reads as "a shader" rather than as water. All four behaviours the issue asks for fall out of one input the engine did not have: **the water depth under each surface point**.

## The design decision, made first

The issue is explicit that the seabed depth field is the real cost and should be designed before any wave code. Two options were named; the reasoning is written up in full in [`Renderer/Water/WaterShoreDepth.h`](../blob/feature/water-shore-waves-1033/OloEngine/src/OloEngine/Renderer/Water/WaterShoreDepth.h) §1 and summarised here.

**The depth comes from the terrain the scene already has.** Drift's six islands are `TerrainComponent`s whose island-falloff mask drives every tile border to a base height 24 m *under* water (their own `ShapingIslandFalloff` notes say so) — so the terrain already IS the sea floor for the entire approach to every island. Nothing had to be authored.

**The per-island authored depth texture was rejected**, and it is worth recording why, because it looks like the safer option: it is a *second source of truth* for a shape that already exists, and the two would drift apart silently the first time anyone moved an island or re-rolled a procedural seed. The waves would then shoal against a coastline that is not the one drawn.

What the terrain route is *not*: a per-vertex sample of six separate height maps. That would need six transforms and six samplers in the vertex stage and would still have nothing to say about the water *between* the islands. `WaterShoreDepthSystem` resamples all of them into **one 512×512 RGBA16F window** over the water tile (`r` = depth in metres, `gb` = the depth gradient) — 3.1 m per texel against Drift's 18–27 m wavelengths, which is far finer than a wave can feel. It is baked when its inputs change and **not per frame** (the sea floor does not move), and retained on the CPU so buoyancy reads the same field with no readback.

## The wave transform

`WaterShoreDepth.h` is the contract and the whole of the maths; `include/WaterShoreCommon.glsl` is its GPU twin, deliberately only the arithmetic in the same order with the same constants.

The load-bearing fact: **frequency is what is conserved** as a wave train runs into shallow water. So the local wavenumber comes from inverting `ω² = g k tanh(k h)`, and the phase speed is `ω/k` with ω fixed at its deep-water value — which is why the *time* term of every wave's phase is depth-independent and a camera crossing a slope sees no shimmer. It is the *spatial* term that compresses.

- **Shoal** — k rises as h falls, so crests bunch up and the phase speed drops.
- **Grow and steepen** — amplitude by Green's law (`a · c_g` conserved along a ray); steepness `Q = a k` grows on both factors at once.
- **Refract** — Snell's law with `c` in place of the refractive index, against the normalised depth gradient. `c ≤ c₀`, so the ray always turns *toward* the shore normal and never away.
- **Break** — **the Jacobian the FFT foam already reads**, not a second signal. `gerstnerWaveNormal`'s accumulated tangent/binormal *are* the off-identity entries of the horizontal displacement map's Jacobian, so the fold detection costs nothing extra (`waterGerstnerJacobian`). What the depth adds is the **limit**: amplitude capped at `breakerIndex · h`, which is what makes the surf zone decay toward the waterline instead of growing into a wall of water.

**Both chains, structurally.** The transform lives inside `sumGerstnerWavesShore` in the shared `WaterCommon.glsl`, which the vertex and tess-eval stages share and which `Water.glsl` and `Water_Depth.glsl` both include. There is no way to shoal in the colour pass and not in the surface-depth capture. Buoyancy is the third chain and follows through `WaterSurface`, calling the same `WaterShore::` functions rather than a second port of them.

**Deep water is unchanged, structurally rather than by tolerance.** A disabled sample makes every relation the identity; the sentinel depth is chosen so `tanh(k h)` is *exactly* `1.0f` for every wavelength the component admits.

## Two defects the contract test caught

Both would have shipped looking entirely plausible, and both are why the test asserts the *residual* rather than the iteration:

1. **The obvious fixed point does not converge in shallow water.** `k ← k₀/tanh(kh)` has derivative exactly −1 at the root there, so it oscillates about the answer forever; a fixed iteration count returns whichever side it happened to land on — a wave of the wrong length, in exactly the water this feature is about. Replaced with Newton from an Eckart seed.
2. **The deep sentinel was not deep.** At 200 m a 500 m wave still has `tanh(kh) = 0.987`, so the "unchanged" open ocean was quietly drifting. It is now the depth at which `tanh` is exactly `1.0f` for the whole admissible range.

## Verification — all three legs

**1. CPU contracts (19 tests, run in CI).** `WaterShoreWaveTest.cpp`: the dispersion residual at 24 wavelength/depth pairs; Green's-law coefficient *dipping below 1 before it grows* (the signature that distinguishes the real relation from any monotone fudge); Snell's turn direction, that it never overshoots, that offshore-travelling waves stay offshore; the breaker limit in metres and the decay to the waterline; the bake against an analytic ramp; and a pin on the GLSL constants + iteration count, because a CPU/GPU divergence there is invisible to any C++ test.

**2. Frames, from angles that can actually show it.** `WaterShoreVisualEvidenceTest.cpp` → `assets/tests/visual/WaterShore_*.png`, six poses. `TopDown` is included **as the negative control** — it is the shot that looks reasonable whether or not any of this works. `ShoalingProfile` shows the crest spacing compressing up the approach; `Grazing` shows crests curving to run parallel to the beach at eye level; `OpenOcean` is the visual half of the deep-water limit.

**3. Live in the editor on a Drift island, A/B'd.** Drift's Atoll, same camera pose, `ShoreWavesEnabled` true vs false: the surf band along the whole waterline is present in one and absent in the other. That A/B is not optional here — a `shorelineFoam` term already existed, so a white band at a shoreline is not by itself evidence of this change. `olo_shader_errors` reports none, and `OloEngine.log` has no shader compile or link errors.

## Scope notes

- **The FFT path gets the depth limit only.** A tiled spectrum has no per-train heading to turn, so refraction is not representable there without a per-ray domain warp; the crest height is still clamped to the breaker limit, because a metre-tall crest in 20 cm of water is the artefact that actually shows. Stated in the docs rather than left to be discovered.
- **Known limitation, recorded so it is not rediscovered as a bug:** this is ray theory evaluated pointwise rather than integrated along a ray, so crest *spacing*, heading, height and breaking are right and individual crest *positions* drift from an eikonal solution by the accumulated phase.
- **Drift's breaker index is authored at 0.06, not the physical 0.39**, and that is a deliberate art decision with its reasoning in the scene file: Drift's sea is ~0.35 m crest-to-trough for mesh-Nyquist reasons (#943), and at the physical index its surf zone is the last ~15 cm of water. The *relation* is untouched; only the authored constant moved.
- Adding an engine texture slot shifted `TEX_SHADER_GRAPH_0` and therefore `OLO_HEAP_IMAGE_BASE` and the heap offset table's length, in both of their mirrors — the existing `BindlessShaderPipeline` pins caught that, as designed.
- `docs/design/WATER_FUTURE_IMPROVEMENTS.md` §5.3 was the origin of this issue; that file no longer exists (it was retired when the issue was extracted), so the entry is recorded instead as `water-ocean.md` §5.3, now marked shipped.

## Off by default

`ShoreWavesEnabled` is off unless a scene asks for it — a water tile with no terrain under it would pay a bake and a vertex fetch to change nothing. Drift is the only scene that turns it on, for the same reason it is the only one that turns the wake on: a feature no scene enables has no coverage in practice however green its tests are.

## Summary by CodeRabbit

* **New Features**
  * Added shore-wave deformation for water, including shoaling, refraction, breaking, and shoreline foam.
  * Added terrain-aware seabed depth sampling for more realistic wave behavior near islands and shores.
  * Added Shore Waves controls to the water inspector, with settings for breaker strength and foam behavior.
  * Shore-wave settings are preserved in scenes and save data.
* **Tests**
  * Added automated coverage and visual validation for shore-wave rendering and depth sampling.
* **Documentation**
  * Documented the shore-wave deformation system and its supported behavior.


drsnuggles8:
@coderabbitai review

PR #1044, opened 2026-09-04
test(vulkan): gate the mesh / G-Buffer shader family on real pipeline creation (#1029)

### LINKED ISSUE #1029: Testing: no Vulkan pipeline is ever created for the mesh / G-Buffer shader family
## Problem

No test ever creates a **Vulkan graphics pipeline** for the mesh / G-Buffer shader family. A shader
construct that is legal GLSL, valid SPIR-V, and correct on OpenGL — but that the Vulkan RHI's
descriptor-heap binding model rejects — is therefore invisible to the entire suite and only shows up
when a human launches `OloEditor --rhi=vulkan`.

That is not hypothetical. It happened in #994:

```
vkCreateGraphicsPipelines(): pStages[1].pNext<VkShaderDescriptorSetAndBindingMappingInfoEXT>
  .pMappings[5].source (VK_DESCRIPTOR_MAPPING_SOURCE_INDIRECT_ADDRESS_EXT) ...
  %1180 = OpArrayLength %80 %356 0
=== CRASH: EXCEPTION_ACCESS_VIOLATION at 0x0 (write) ===
```

A single `.length()` call in a fragment shader. It compiled, it validated, the golden images were
byte-identical, and the **full 7086-test suite was green** — including all 126 device-backed Vulkan
tests, which ran on a real device and passed. `VulkanPipelineBuilder` returns `VK_NULL_HANDLE` and
logs; the editor then dereferenced the null pipeline and died. Only the mandated live-editor run
(CLAUDE.md → *Rendering changes must be visually verified*) caught it, and only because the change
was a rendering change that triggered that step.

## Why the existing coverage does not reach it

- `VulkanPassSuite` (52 tests), `VulkanDrawPath`, `VulkanShaderPipeline`, `VulkanResourceFactory`
  and friends **do** run with a real device on the dev box, and they do build pipelines — but from
  `FullscreenBlit.glsl` and a handful of test shaders, never from the mesh/G-Buffer family.
- The AMD self-hosted runner cannot help, and this is worth recording so nobody re-investigates it:
  `gpu-conformance-amd.yml` already builds and runs the full suite with Vulkan enabled, and the box
  has RADV and lavapipe installed. Every device-backed Vulkan test **skips** there with
  *"No device satisfies the ADR 0010 capability contract (`VK_EXT_descriptor_heap` et al.)"*.
  Mesa 25.2.7's RADV exports 136 `VK_EXT_*` strings including `descriptor_buffer` and
  `descriptor_indexing`, and does **not** export `descriptor_heap`. Vulkan CI coverage is blocked on
  driver support, not on runner configuration. This issue is therefore about the **dev-box suite**
  being a real gate, not about CI.

## Goal

Make `vkCreateGraphicsPipelines` failure for any shipped mesh / G-Buffer shader a **test failure**
on the dev box, instead of a live-editor discovery.

## Scope

- A device-gated Vulkan test (skipping cleanly where the ADR 0010 contract is unmet, like the rest
  of the `VulkanShaderPipelineTest` ladder) that asks `VulkanPipelineBuilder` for a pipeline for
  each shipped mesh / G-Buffer shader and asserts the handle is non-null.
- Cover the family, not one file: there are **17** `*GBuffer*` shaders under
  `OloEditor/assets/shaders/`, plus `PBR_MultiLight*.glsl` and the depth-prepass variants that share
  the same include closure.
- Drive it from a list that is derived, not hand-maintained, so a new G-Buffer shader is covered the
  day it lands rather than the day someone remembers to add it.
- Assert on the *reason* where possible: a null pipeline should fail with the builder's logged
  `VkResult`, not a bare "handle was null".

## Acceptance criteria

- A new test fails when a mesh/G-Buffer shader cannot produce a Vulkan pipeline, and names the
  shader and the `VkResult`.
- The `.length()` / `OpArrayLength` regression from #994 is reproducible by the test: reintroduce
  the call in a scratch branch and the test goes red without a live editor run.
- Adding a new `*_GBuffer.glsl` requires no test edit to be covered.
- Skips cleanly (never fails, never silently passes) on a device that does not satisfy the ADR 0010
  contract, and the skip reason says which.

## Out of scope

- Making Vulkan run in CI — blocked on `VK_EXT_descriptor_heap` reaching Mesa, see above.
- Pixel/golden comparison of the Vulkan output. This is a pipeline-*creation* gate; visual parity
  stays with the live-editor step.
- A non-descriptor-heap fallback path in the RHI (that would be its own issue, and ADR 0010 exists
  precisely to avoid one).

## Notes

The cheap structural guard for the specific `.length()` case already exists and runs everywhere:
`GPUSceneLayoutTest.NoGPUSceneIncludeAsksAStorageBufferForItsLength`, plus the rule in
`docs/agent-rules/glsl-shaders.md` §6b. That covers one construct in one include family. This issue
is the general gate.

## Score
```olo-score
capability: 3
craft: 1
stability: 5
decay: 3
effort: 2
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
## What

`vkCreateGraphicsPipelines` failure for any shipped mesh / G-Buffer shader is now a **test
failure on the dev box** instead of a live-editor discovery.

`VulkanMeshPipelineFamily.EveryShippedMeshShaderProducesAPipeline` asks
`VulkanPipelineBuilder::GetOrCreateGraphics` — the draw path's own call, with the shader's own
cached root-data layout, so the baked mapping array is byte-for-byte what a real draw bakes —
for a pipeline for every family shader, and requires a non-null handle.

- **Family derived, not listed**: `*GBuffer*` + `PBR_MultiLight*` + `DepthPrepass*` off the
  shader directory. 23 shaders today, including the tessellated `Terrain_GBuffer` and the
  task/mesh `VirtualMeshletGBuffer`. A new `*_GBuffer.glsl` is covered the day it lands.
- **A glob that matched nothing cannot pass**: four anchor shaders are asserted present, and the
  family size is floored.
- **Attachment plan is derived**: colour count and per-attachment format come from each fragment
  stage's reflected outputs, so the suite never fails for a reason unrelated to the shader.
- **Skips cleanly** where the ADR 0010 contract is unmet, naming it. Never `DISABLED_`.

## The correction this turned up

The issue (and `glsl-shaders.md` §6b) said the builder returns `VK_NULL_HANDLE` and the editor
then dereferences it. **That is not what happens.** Measured under cdb on this box's NVIDIA
driver, `vkCreateGraphicsPipelines` **access-violates inside the driver** and never returns:

```
00 nvoglv64+0x1372f3                                          <- 0xC0000005
...
13 VkLayer_khronos_validation!vvl::DispatchDevice::CreateGraphicsPipelines+0x342
```

I found this the expensive way: my first version fed the construct to the builder and asserted on
the result. The process died, and because the SEH unwind skips the builder's `std::lock_guard`,
`TearDown`'s `ReleaseAll()` then deadlocked on the still-held mutex — the run hung for nine hours
rather than failing.

So the gate **pre-flights each shader's SPIR-V for `OpArrayLength` and refuses to hand a violating
module to the driver**, reporting the shader, the block and the VUID instead. §6b is corrected in
its own commit.

## The red run (acceptance criterion 2)

`.length()` reintroduced in `PBR_GBuffer.glsl`'s fragment stage on `InstanceBuffer`, then reverted:

```
..\OloEngine\tests\Rendering\VulkanMeshPipelineFamilyTest.cpp(629): error: Failed
PBR_GBuffer calls .length() on a buffer block (InstanceBuffer) in stage 0x10.
Every buffer block on this backend is mapped through
VK_DESCRIPTOR_MAPPING_SOURCE_INDIRECT_ADDRESS_EXT, and the spec forbids OpArrayLength on such a
resource (VUID-VkPipelineShaderStageCreateInfo-pNext-11378). This is #994: legal GLSL, valid
SPIR-V, correct on OpenGL, and an access violation inside the driver when
vkCreateGraphicsPipelines sees it. Pass the count in explicitly — see
docs/agent-rules/glsl-shaders.md §6b.
Google Test trace:
..\OloEngine\tests\Rendering\VulkanMeshPipelineFamilyTest.cpp(595): PBR_GBuffer

[  FAILED  ] VulkanMeshPipelineFamily.EveryShippedMeshShaderProducesAPipeline (2022 ms)
```

Red in 2 s, exit code 1, no crash and no hang. Green again after revert (`git diff` over
`OloEditor/assets/shaders/` is empty in this branch).

`MeshPipelineSpirvContract` proves the detector itself: it fires on a probe carrying `.length()`
and stays quiet on an otherwise identical twin that takes its count from a UBO. Both are
device-free, so this half of the gate runs everywhere — including the CI arms with no Vulkan
device. (Vulkan CI stays blocked on `VK_EXT_descriptor_heap` reaching Mesa; unchanged by this PR.)

## Also in this PR

`VulkanPipelineBuilder` records the last creation failure (shader name + `VkResult`) so a null
handle is reported with its reason rather than "the handle was null". Additive and diagnostic
only — nothing reads it to decide anything, and it is written under the lock the creation already
holds. It is *not* the path #994 takes, since that call never returns.

## Evidence

| | |
|---|---|
| `VulkanMeshPipelineFamily` + `MeshPipelineSpirvContract` | 3 passed |
| All `Vulkan*` + `Shader*` suites (regression check on the builder change) | **334 passed, 1 pre-existing skip, 0 failed** |
| Family gate, warm | 281–419 ms for 23 shaders |
| Family gate, cold (first run, shaderc compile) | ~72 s |
| Validation errors during pipeline creation | 0 |

Cheap enough to sit in the default run; no flag gating it.

## Not done

Rendering is unchanged, so the visual-verification rule does not attach — no pixel is produced
differently by this branch. The builder edit adds a recorded field and touches no creation
parameter.

Closes #1029

## Summary by CodeRabbit

* **Bug Fixes**
  * Pipeline creation failures now provide structured diagnostic details, including the affected shader and Vulkan result.

* **Tests**
  * Added Vulkan mesh and G-Buffer pipeline coverage across supported shader configurations.
  * Added validation for unsupported shader constructs in production SPIR-V.
  * Improved handling and reporting of per-shader pipeline failures.

* **Documentation**
  * Updated shader guidance to describe driver crashes caused by storage-buffer `.length()` usage and the corresponding validation checks.

### DIFF
--- a/OloEngine/src/Platform/Vulkan/VulkanPipelineBuilder.cpp
+++ b/OloEngine/src/Platform/Vulkan/VulkanPipelineBuilder.cpp
@@ -464,6 +464,7 @@ namespace OloEngine
         {
             OLO_CORE_ERROR("VulkanPipelineBuilder: compute pipeline creation failed (VkResult {})",
                            static_cast<int>(result));
+            m_LastCreationFailure = { true, "<compute>", result };
             return VK_NULL_HANDLE;
         }
 
@@ -794,6 +795,7 @@ namespace OloEngine
         {
             OLO_CORE_ERROR("VulkanPipelineBuilder: vkCreateGraphicsPipelines failed for '{}' (VkResult {})",
                            shader.GetName(), static_cast<int>(result));
+            m_LastCreationFailure = { true, shader.GetName(), result };
             return VK_NULL_HANDLE;
         }
 
--- a/OloEngine/src/Platform/Vulkan/VulkanPipelineBuilder.h
+++ b/OloEngine/src/Platform/Vulkan/VulkanPipelineBuilder.h
@@ -54,6 +54,7 @@
 #include <array>
 #include <mutex>
 #include <shared_mutex>
+#include <string>
 #include <unordered_map>
 #include <vector>
 
@@ -142,6 +143,33 @@ namespace OloEngine
         // the maps (device teardown).
         void ReleaseAll();
 
+        // The last creation failure, kept so a caller can report the REASON
+        // rather than a bare null handle (issue #1029). GetOrCreate* return
+        // VK_NULL_HANDLE and log on a vkCreate*Pipelines failure; the log is
+        // not reachable from a test assertion, and "the handle was null" does
+        // not say which shader or which VkResult. Recorded under the same
+        // lock the creation already holds, so it is the failure of the call
+        // that just returned null on this thread as long as no other thread
+        // failed in between — a diagnostic, never a control-flow input.
+        struct CreationFailure
+        {
+            bool Valid = false; ///< False until a creation has failed.
+            std::string ShaderName;
+            VkResult Result = VK_SUCCESS;
+        };
+
+        [[nodiscard]] CreationFailure GetLastCreationFailure() const
+        {
+            std::shared_lock lock(m_Mutex);
+            return m_LastCreationFailure;
+        }
+
+        void ClearLastCreationFailure()
+        {
+            std::lock_guard<std::shared_mutex> lock(m_Mutex);
+            m_LastCreationFailure = {};
+        }
+
         [[nodiscard]] sizet GetCachedPipelineCount() const
         {
             std::shared_lock lock(m_Mutex);
@@ -186,6 +214,7 @@ namespace OloEngine
         // contention the #806 measurement found.
         mutable std::shared_mutex m_Mutex;
         std::unordered_map<Key, VkPipeline, KeyHash> m_Pipelines;
+        CreationFailure m_LastCreationFailure;
     };
 } // namespace OloEngine
 
--- a/OloEngine/tests/CMakeLists.txt
+++ b/OloEngine/tests/CMakeLists.txt
@@ -268,6 +268,13 @@ add_executable(OloEngine-Tests
 		# root-data binding mappings, vertex pulling, and the FXAA golden-pass
 		# checkpoint (ADR 0011 §4/§5). Same SKIP ladder + zero-validation bar.
 		Rendering/VulkanShaderPipelineTest.cpp
+		# #1029: the mesh / G-Buffer shader family reaches a REAL VkPipeline. Every
+		# other device-backed Vulkan suite builds from FullscreenBlit.glsl and test
+		# shaders, so a construct the descriptor-heap binding model rejects (#994's
+		# `.length()`) was invisible to all 7086 tests and only showed up in the
+		# live editor. Derived family list + a probe pinning that the gate bites.
+		# Same SKIP ladder + zero-validation bar.
+		Rendering/VulkanMeshPipelineFamilyTest.cpp
 		# #691 Phase 7 Stage 1.2: the buffer/texture factory arms on a REAL
 		# device — arena-versioned UBO addresses, texture upload/readback with
 		# the mip blit chain and RGB->RGBA widening, the descriptor slot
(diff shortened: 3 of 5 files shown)


drsnuggles8:
@coderabbitai review

PR #1051, opened 2026-09-04
feat(renderer): Vulkan RT acceleration-structure scene manager (#978)

### LINKED ISSUE #978: Vulkan RT: acceleration-structure scene manager (BLAS/TLAS, updates, alpha masks, telemetry)
## Problem

The completed Vulkan RHI (#691 / #801) provides the graphics/compute/resource foundation, but there is currently **no** ray-tracing code anywhere in the tree — `VK_KHR_acceleration_structure`, `VK_KHR_ray_query` and `VK_KHR_ray_tracing_pipeline` are not referenced at all. Hardware ray tracing needs a first-class scene-management subsystem before any shadow, reflection, path-tracing, or ReSTIR pass can be correct.

Acceleration structures are not an implementation detail of one effect. If each pass builds its own BLAS/TLAS or invents different update/lifetime rules, memory, synchronization, alpha masking, and dynamic-geometry behavior will diverge immediately.

## Goal

Add a Vulkan-only `RayTracingScene` that consumes the canonical GPU Scene (#977), owns acceleration-structure lifecycle and synchronization, and exposes a stable query/pipeline binding contract to RenderGraph passes.

This issue deliberately produces **no new lighting effect**. Its output is a validated ray-intersection substrate.

## Effort note — this is expected to split

Scored effort **13**, which per `docs/process/issue-scoring.md` §4 means **split it** (the anchor is #435, "three rendering sub-features"). Suggested child seams:

1. capability detection/validation + device-property capture + the unsupported-hardware path;
2. BLAS lifecycle (build, compact, retire) for static and rigid-dynamic geometry;
3. TLAS lifecycle, instance masks, and previous/current transform semantics;
4. the alpha-tested candidate-confirmation helper;
5. RenderGraph hazard representation, telemetry, and the MCP ray-trace diagnostic.

File those as children rather than attempting one PR.

## Capability and initialization

Require and validate the relevant Vulkan capabilities, including the acceleration-structure and ray-query extensions; enable the ray-tracing-pipeline path when available/selected. Fail clearly or expose an unsupported capability state — never silently create an incomplete RT scene.

Record device properties used for:

* acceleration-structure alignment;
* scratch alignment;
* maximum geometry/instance counts;
* shader-group handle/SBT alignment when the pipeline extension is used;
* supported build/update flags.

Do not gate any of this on `identicalMemoryTypeRequirements` — it is reported FALSE on this project's development GPU and has already caused one feature to be silently disabled here.

## BLAS lifecycle

Implement explicit geometry classes:

* **Static:** build once, query compacted size, compact, retire source storage safely.
* **Rigid dynamic:** BLAS remains static; transform changes update the TLAS instance only.
* **Deformed/skinned:** refit when legal and quality remains acceptable; rebuild under a documented heuristic.
* **Masked:** triangle candidates retain enough geometry/material/UV identity for alpha testing.
* **Unsupported:** counted and reported, with the raster fallback remaining visible.

Batch compatible builds, pool/reuse scratch buffers, and perform frame-safe deferred destruction. Avoid a full GPU idle for routine scene mutation.

## TLAS lifecycle

* Build/update one TLAS per relevant view or a documented shared scene TLAS.
* Consume stable GPU Scene instance IDs, transforms, masks, and geometry indices.
* Support per-effect instance masks and ray flags.
* Retain previous/current transform semantics needed by motion-aware RT effects.
* Rebuild only when topology/capacity requires it; update/refit for ordinary transform changes.

## Alpha-tested geometry

Ray-query shaders do not get a traditional any-hit shader automatically. Provide a shared candidate-confirmation helper that:

1. obtains primitive barycentrics and material/texture indices;
2. reconstructs UVs;
3. samples the alpha mask with a documented LOD policy;
4. confirms or rejects the candidate.

The ray-tracing-pipeline path should use the same semantic helper through any-hit or equivalent code. Raster and RT alpha cutoffs must agree.

## RenderGraph and synchronization

* Represent AS build/update/read hazards explicitly.
* Route build commands through the Vulkan frame command ordering, not one-shot work that can execute ahead of producers. `docs/agent-rules/vulkan-command-ordered-buffer-writes.md` covers why a life-stable device address makes an out-of-order write silently last-write-wins.
* Track geometry buffer and descriptor-heap lifetimes through the frame.
* Add validation-layer coverage for scene creation, update, resize, destruction, and backend restart. Note the IBL bake has two pre-existing layout VUIDs on master — A/B before attributing a validation error to this work.

## Telemetry and diagnostics

Expose:

* BLAS count by class;
* build/refit/compaction counts and GPU time;
* TLAS build/update time;
* AS and scratch memory;
* compaction savings;
* rebuild reason;
* unsupported/missing geometry count;
* masked-candidate accept/reject counts in a diagnostic trace mode.

Add a small RT diagnostic view or MCP command that traces deterministic primary/visibility rays and returns hit instance, primitive, barycentrics, distance, and material ID. Log it against the MCP follow-up tracker (#607) if it lands as a native tool.

## Acceptance criteria

* Static triangle meshes build, compact, and remain traceable across frames.
* Rigid transform changes update TLAS state without rebuilding static BLASes.
* Deformed geometry follows the documented refit/rebuild policy.
* Masked geometry agrees with raster visibility in a cutout-foliage test.
* Removed/reused GPU Scene instances cannot be hit through stale TLAS records.
* AS build/read barriers are represented in RenderGraph/Vulkan synchronization and validation stays clean against the documented pre-existing baseline.
* No routine scene update calls `vkDeviceWaitIdle`.
* Memory, build time, compaction, refit/rebuild, and unsupported-geometry counters are visible.
* A deterministic ray-hit test covers miss, closest hit, instance mask, transformed geometry, and alpha reject/accept.
* Unsupported RT hardware keeps the raster renderer usable and reports the capability reason.

## Explicitly out of scope

* Lighting, shadowing, reflections, GI, denoising, path tracing, or ReSTIR.
* Opacity micromaps in the first implementation.
* Procedural/AABB geometry, curves/hair, particles, fluids, or virtualized-geometry clusters.
* Solving texture LOD for arbitrary secondary rays beyond the alpha-test MVP.

## Dependencies

Hard-blocked by the canonical GPU Scene #977.

## References

* Vulkan `VK_KHR_acceleration_structure`, `VK_KHR_ray_query`, `VK_KHR_ray_tracing_pipeline`.
* Completed Vulkan RHI epic #691 / #801.
* `docs/agent-rules/rhi-abstraction-boundary.md` §9-§14 — accumulated Vulkan-epic lessons.

## Score
```olo-score
capability: 13
craft: 5
stability: 8
decay: 3
effort: 13
confidence: 0.5
learning: 13
fun: 8
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Adds the first hardware ray-tracing code in the tree: a `RayTracingScene` that consumes the canonical GPU Scene (#977), owns BLAS/TLAS lifecycle and synchronisation, and exposes a stable query contract to RenderGraph passes. It deliberately produces **no new lighting effect** — its output is a validated ray-intersection substrate.

## Deliberate deviation from the issue text, stated up front

The issue says "File those as children rather than attempting one PR" and lists five child seams. This lands as **one PR**, per the handover's brief and the epics-stay-whole policy that postdates the issue. The reasoning held up in practice: unlike #977, whose four children were each independently shippable, these five seams are phases of one class's bring-up and none ships anything alone — capability detection with no BLAS is dead code, BLAS with no TLAS is untraceable, TLAS with no alpha helper fails the cutout criterion. The issue's own framing agrees: *"Its output is a validated ray-intersection substrate"* — singular.

## Design decisions worth reviewing

**The capability is one value, not a predicate plus a reason.** #978 requires the renderer to report *why* RT is unavailable, and only `VulkanDevice::Init` can distinguish "extension absent" / "feature bit off" / "volk exported no entry point" / "lever turned it off". So `GetRayTracingCapabilities()` returns verdict, reason and captured device properties together, and `SupportsRayTracing()` forwards rather than re-deriving (rhi-abstraction-boundary §13c). Enabled Tier-2 style — never an ADR 0010 gate row.

**An acceleration structure needs no descriptor.** Verified with `glslc` + `spirv-val`: `accelerationStructureEXT(uvec2)` lowers to `OpConvertUToAccelerationStructureKHR` under `OpCapability RayQueryKHR` alone. Combined with `GL_EXT_buffer_reference` for the ray/hit batches and the three GPU Scene tables, the whole subsystem costs **one** buffer binding — `UBO_RAY_TRACING` (65), **the last free number in the engine**. This is ADR 0011 §4's root-data pointer model reaching somewhere it wasn't designed for and fitting exactly. Registered in `IsKnownUBOBinding`.

**A BLAS is per GEOMETRY; opacity is per INSTANCE.** The natural per-instance loop hands `vkCmdBuildAccelerationStructuresKHR` two entries writing one destination the moment two entities share a mesh. And alpha-testing is a property of the *material*, so one mesh can be an opaque wall for one entity and a cutout for another — geometries build `OPAQUE` and the instance's `FORCE_OPAQUE`/`FORCE_NO_OPAQUE` overrides it. One structure, both uses.

**Builds ride the frame command buffer.** `VulkanOneShot` submits ahead of the still-recording frame (amendment (72)), so a BLAS built that way reads vertex data the frame hasn't uploaded. The corollary: compaction can't block, so the size query is stamped in the build's own command buffer and polled without a wait bit two generations later.

**Seam shape** is `RHI::DescriptorHeap`'s, not `RendererAPI`'s: `RayTracingScene` owns the policy (classification, refit heuristic, instance packing, stale-record guard, telemetry) and holds an `IRayTracingBackend` implemented once under `Platform/Vulkan/`. The neutral half compiles and is tested with no GPU, which is every CI runner. Two `RHI::Access` members carry the build→read hazard through the neutral lattice into `VulkanBarrierLowering`; a `RayTracingScenePass` owns it in the graph.

## Acceptance criteria

| Criterion | Evidence |
|---|---|
| Static meshes build, compact, remain traceable | `RayTracingDevice.AStaticTriangleBuildsCompactsAndStaysTraceable` + live: 54 BLASes, 2.36 MB reclaimed by compaction |
| Rigid transform changes don't rebuild static BLASes | `RayTracingSceneFixture.AMovedInstanceRebuildsNoBlas`; live steady state `blasBuilds: 0, tlasUpdates: 1` |
| Deformed follows the refit/rebuild policy | ⚠️ **partial** — see below |
| Masked agrees with raster visibility | ⚠️ **partial** — see below |
| Removed/reused instances not hittable via stale TLAS | `AnInstanceThatStopsBeingStagedLeavesTheTlasAndItsBlasRetires`, `AReusedGeometrySlotDoesNotInheritTheDeadMeshsBlas` (real `GPUScene`), `RayTracingDevice.RetiringAGeometryRemovesItFromEveryLaterTrace` |
| AS build/read barriers represented; validation clean | `VulkanBarrierLowering` lowering + **0 validation errors** across every device tenant and both live arms |
| No `vkDeviceWaitIdle` on routine mutation | `RoutineSceneMutationNeverIdlesTheDevice` — a blocking query read would hang, not fail |
| Counters visible | `StatisticsPanel` "Ray Tracing" section + `olo_rt_scene_stats`, both exercised live |
| Deterministic ray-hit test | `DeterministicRaysReportMissClosestHitTransformAndAlpha` — miss, closest hit, barycentrics, instance mask, transformed geometry, alpha reject **and** accept, on a real RTX 4090 |
| Unsupported RT keeps raster usable, reports reason | Live A/B with `OLO_VULKAN_NO_RAY_TRACING=1`: reports the reason, 0 validation errors, frame **byte-identical** to the RT-on arm |

### The two partials, stated rather than ticked

**Deformed geometry** has a documented refit/rebuild heuristic and tests (`DeformedRefitsUntilTheRunBudgetThenRebuilds`), but **no live producer**: skinned, cloth, virtualized-cluster and particle entities never reach the canonical GPU Scene at all — `Scene.cpp` skips them and counts them in `GPUSceneUnsupportedCategory`. The policy is exercised by tests, not by a scene. One line in `Classify` changes the day a deformed-vertex stream lands in GPU Scene.

**Masked geometry**: UV reconstruction, material resolve and the cutoff comparison all ship and are verified on-device including reject *and* accept. Cutoff parity with raster is **structural** — both read the same `GPUSceneMaterial::AlphaCutoff`, there is no second constant to drift. What is missing is the *texel fetch* for an arbitrary material: outside `OLO_BINDLESS` a shader reaches a material texture through a per-draw slot binding, and one ray-query dispatch has no per-draw scope. The fetch is a caller-supplied macro; closing it needs the shader-visible sampler heap ADR 0011 §1.2a already records, not anything in this issue.

## Review guide

**Where I'd look hardest**

1. `VulkanRayTracingBackend.cpp` — `RecordTlasBuild`'s rebuild-vs-refit promotion. A TLAS update must use the *same* `primitiveCount` the source was built with, so any count change forces a rebuild, and a compaction that moved a BLAS address forces one too. Both were bugs the self-review found; this is the densest correctness logic in the PR.
2. `VulkanRayTracingBackend.cpp` — the compaction handshake and the instance-buffer ring. Compaction spans frames by design (never idle), and the ring exists because one persistent buffer rewritten every frame is the cross-frame form of the last-write-wins archetype.
3. `RayTracingScene.cpp::Update` — the two-pass structure. Pass one walks instances, pass two decides builds per unique geometry; collapsing them reintroduces duplicate builds for a shared mesh.

**What I verified, and how**

- 37/37 RT tests, 1373/1374 across every suite my changes touch (one known headless skip: `McpHeadlessAttachTest.HostUntilDetached`), 164/164 on the RT + Vulkan re-run after the final fix.
- Device tenants run on a real RTX 4090 and assert **zero validation errors** in TearDown, sync validation included.
- Live editor, both arms, `--rhi=vulkan`, `MaterialSpheres.olo` (57 entities): RT on → 54 BLASes / 54 TLAS instances / 2.36 MB compaction savings / 0 validation errors; RT off → reports the reason, 0 validation errors, byte-identical frame.
- Shaders validated with `glslc` + `spirv-val` before every build, and the sweep now derives the Vulkan-only compute family from source so the next ray-query shader is covered automatically.

**Least confident about**

The **compaction handshake's frame separation**. Queries are held back two `RecordBlasBuilds` generations, which matches frames-in-flight and is what makes the read legal — but that counter advances per *call*, not per *fence*. It's correct for the frame loop (one call per frame) and for the device tests (each records and waits), and I could not construct a case where it isn't, but it is an assumption about the caller rather than something the type system enforces. A second opinion there would be worth having.

Also: the `PackInstanceMask` fold is a judgement call. GPU Scene's `VisibilityMask` is always `0xFFFFFFFF` today, so the folding behaviour is unobserved in practice — I chose folding over truncation so a future effect setting only a high bit isn't silently invisible, but nothing currently proves that's the convention we want.

**Deliberately not tested**

- The ray-tracing *pipeline* path (SBT). Ray query alone satisfies this issue; the pipeline extension is enabled and its alignment properties captured, but nothing exercises them.
- The alpha helper's texel fetch, for the reason above — the policy half is tested, the fetch has no consumer to test through.
- Multi-GPU / non-NVIDIA behaviour. `accelerationStructureHostCommands` and `accelerationStructureIndirectBuild` are both false on this box; the code records them and gates on neither, but the false arms are the only ones observed.

Closes #978

## Summary by CodeRabbit

* **New Features**
  * Added hardware ray-tracing support for Vulkan, including acceleration-structure construction and ray-query diagnostics.
  * Added a Renderer Statistics panel section showing ray-tracing availability, memory usage, scene counts, build activity, and GPU timings.
  * Added the `olo_rt_scene_stats` diagnostics tool with capability and scene telemetry.
  * Added clear unsupported-hardware reporting and a Vulkan option to disable ray tracing for comparison.
* **Documentation**
  * Documented ray-tracing architecture, constraints, and diagnostic tooling.
* **Tests**
  * Added coverage for ray-tracing devices, scenes, shader compilation, and statistics reporting.


drsnuggles8:
## 🤖 Self-review @ `40e404af4d9a9b612fbc668a22bb9edf3cf7a833`

Reviewed the full diff at **high** effort before the PR existed, then re-verified in the live editor.

**Findings: 7 · all real, all mine, all fixed.**

Six from the static review:
- **Compaction never ran in production** — `RecordBlasBuilds` polls the compacted-size queries, but the caller guarded it behind `if (!m_PendingBuilds.empty())`. A settled scene produces no builds, so it never polled. The device test hid it by calling the backend directly. Now unconditional; live editor confirms 2.36 MB reclaimed.
- **TLAS refit with a changed instance count is invalid usage** — `VUID-…-pInfos-03407` requires an update to use the same `primitiveCount` the source was built with. Promoting only on growth left 2→1 illegal. Any count change now rebuilds.
- **`VK_PIPELINE_STAGE_2_ACCELERATION_STRUCTURE_COPY_BIT_KHR`** belongs to `VK_KHR_ray_tracing_maintenance1`, which isn't enabled — illegal, not merely redundant. Dropped; the BUILD stage covers copies without it.
- **Cross-frame hazard on the TLAS instance buffer** — one persistent buffer rewritten every frame with only an intra-command-buffer barrier, which is the cross-frame form of the exact archetype quoted three lines above it. Now a 2-deep ring.
- **`BlasByClass` counted instances, not structures** — a mesh drawn 500× reported 500 BLASes, which is the number a memory budget gets sized against. Now per unique geometry; unsupported *instances* reported separately. Regression tenant added.
- **Build/refit counters credited positionally** from a bare return count while the backend skips arbitrary entries. The backend now counts what it recorded.

One the static review could not have found, and the suite could not either:
- **`VK_BUFFER_USAGE_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT_KHR` set unconditionally** on every vertex/index buffer. A usage bit from an optional extension is invalid unless that extension is *enabled*, so on hardware without RT every buffer creation was `VUID-VkBufferCreateInfo-None-09499` — the "unsupported RT keeps raster usable" criterion failing. All 37 RT tests passed either way, because the device-backed ones only run where RT exists. Caught on the first live launch with `OLO_VULKAN_NO_RAY_TRACING=1`: **10 validation errors before, 0 after**, frame byte-identical. Written up as §8b of the agent-rules doc, because the general form — *enumerate every flag and stage bit a new capability introduces and ask which the device must have ENABLED* — is not RT-specific.

**Dismissed:** none.

Two of my own verification steps were also wrong and are corrected: my `glslc` pre-check passed a `-I` flag the engine's includer doesn't use (false green on an include path), and the first build's failure was masked by a trailing `echo "EXIT=$?"` — the exact trap `task-loop.md` documents. `build-lock.ps1` propagates exit codes correctly; that was my command shape.

drsnuggles8:
@coderabbitai review

Issue #1058, opened 2026-09-05
RHI/Vulkan: virtual geometry's mesh-shader arm draws nothing — a per-draw snapshot on a GPU-produced buffer, masking an out-of-bounds SSBO read


drsnuggles8:
## Summary

Two coupled defects, found while finishing #1052. Neither is fixed there; both are
fully diagnosed here and the patch for the first is attached below.

1. **`VulkanStorageBuffer` snapshots a GPU-produced buffer**, so the virtual-geometry
   mesh-shader raster arm launches `EmitMeshTasksEXT(0)` and draws nothing.
2. **Suppressing that snapshot makes the mesh arm run — and loses the device**, because
   the snapshot was masking an out-of-bounds SSBO read somewhere in the VG mesh/cull path.

The first is why virtual geometry is still invisible on Vulkan with **default settings**
on a mesh-shader-capable device, even after #1052. The MDI and software-raster arms are
correct (`olo_virtual_geometry_set hwRasterMode=forcemdi` is a working lever today).

## Defect 1 — a per-draw snapshot on a buffer the GPU produces

`VulkanStorageBuffer::SetData` versions a mid-frame write into the frame arena so draws
recorded after it see the new bytes and earlier ones keep what they recorded — GL's
command-ordered `glNamedBufferSubData` semantics (#691,
[vulkan-command-ordered-buffer-writes.md](../blob/master/docs/agent-rules/vulkan-command-ordered-buffer-writes.md)).

The split that matters (`VulkanRendererAPI::AssembleRootData`, the
`commandOrderedBufferReads` flag):

- **draws** read `GetRootDataAddress()` → the snapshot, when one is live
- **compute dispatches** read `GetDeviceAddress()` → always the persistent buffer

`VirtualMeshRegistry::PrepareFrame` zeroes the VG draw-args buffer every frame *before*
the cull dispatches. The cull then writes the real `DrawCount` into the persistent
buffer, while the CPU's zero snapshot stays live for the rest of the frame. Three
consumers, one victim:

| consumer | how it reads `DrawCount` | result |
|---|---|---|
| hardware MDI | `vkCmdDrawIndexedIndirectCount` on the parameter `VkBuffer` | persistent — correct |
| software raster | it is a dispatch | persistent — correct |
| mesh-shader **task stage** | it is a **draw**, so root data | the zero snapshot → `EmitMeshTasksEXT(0)` |

[...]

Issue #1052, opened 2026-09-04
RHI/Vulkan: virtual geometry does not render — SetVertexArrayIndexBuffer + two raw-buffer upload paths are still #691 no-ops


drsnuggles8:
> **STATUS (updated after PR #1054).** The device loss is **fixed** — #1054 makes a raw buffer a
> first-class SSBO occupant, so `SSBO_VIRTUAL_INDICES` binds and the GPU no longer faults.
> **Virtual geometry still does not RENDER on Vulkan**, and that is what this issue now tracks:
> three RHI entry points on the same path are still `#691` no-ops. Details in
> [this comment](#issuecomment-5542405190); the remaining work and *why nobody caught it* are below.

## What is still broken

A wide capture of `VirtualGeometryTest.olo` on Vulkan after #1054 shows the ground plane and the
skybox and **no geometry**. The cull runs (4143 clusters drawn, `silentlyDrewNothing: false`), the
shaders compile, the device survives — and no VG pixels land, because:

| stub | consequence |
|---|---|
| `SetVertexArrayIndexBuffer` | `indexed draw without an index buffer — dropped`; the hardware MDI arm draws nothing |
| `AllocatePersistentUploadStorage` | raw-buffer persistent mapping unavailable |
| `UploadBufferSubData` | raw-buffer sub-range uploads silently discarded |

The first is the one that matters and it is a **design question, not a transcription**: the Vulkan
draw path *pulls* vertices and has no vertex-input state at all (ADR 0011 §5, and the `OLO_VULKAN`
branch in `VirtualVisibilityResolve.glsl`), so "bind the VAO's index buffer" has to be decided
before it can be lowered.

## Why this was never caught — the part worth reading

Virtual geometry **is** tested on Vulkan. Two tests, and both pass:

- `VulkanPassSuite.VirtualGeometryMdiCountDrawsHandAuthoredClusters`
- `VulkanPassSuite.VirtualGeometryMeshTasksMatchTheMdiPath`

They pass because they **construct the broken buffer a different way**. The suite's own header
comment is honest that it substitutes — "the FULL VirtualGeometryPass is disproportionate
headlessly … so this tenant pins the INDIRECT-DRAW ENTRY itself — with the REAL
VirtualMeshGBuffer.glsl and a hand-authored cluster set". What the substitution changed is precisely
the thing that fails:

```cpp
[...]

PR #1062, opened 2026-09-05
fix(vulkan): lower the raw-buffer upload and element-buffer no-ops so virtual geometry renders

drsnuggles8:
## What changed

Virtual geometry now renders on Vulkan. Four `#691` no-ops are lowered, and the test that should
have caught them now builds its buffers the way production does.

**The issue's diagnosis was wrong about which stub mattered, and that is the useful part.** #1052
and PR #1054 both call `SetVertexArrayIndexBuffer` "the one that matters". It is not. Two of the
three stubs sit on the **page-upload** path and compose:

| stub | what the caller made of it |
|---|---|
| `AllocatePersistentUploadStorage` returned `nullptr` | `GPUCircularBuffer::Create` read it — correctly, by its own contract — as *"mapping failed, use the other path"* and never created a ring |
| `UploadBufferSubData` discarded its payload | that other path, now taken for **every** page load, wrote nothing |

So `VirtualMeshRegistry::CopyThroughRing` wrote nothing to the vertex arena or the index arena, and
no raster arm had geometry to draw. Fixing the index bind alone would have changed nothing on
screen. `UnmapBuffer` was a fourth stub, dead only because the ring it belongs to was never created
— lowering the ring wakes it, so it is lowered here too.

The generalisable half, recorded in `no-silent-fallbacks.md`: **an unlowered function whose return
value is a legal in-band answer does not read as unlowered at the call site — it reads as the
condition that value encodes**, and the caller's graceful-degradation branch then runs as designed.

## The design question: no ADR change needed

ADR 0011 §5 removes the vertex **attribute** axis from the PSO ("vertex input layout is not baked at
all"). The index buffer was never one of those axes — it is not in `VkGraphicsPipelineCreateInfo` at
all, it is `vkCmdBindIndexBuffer`, plain command state contributing zero PSO permutations. So §5's
"remove the axis entirely" cannot apply to it, and `BindIndexBufferFor` has been doing exactly this
for object-backed VAOs since #691.

The lowering records the raw handle on the `VulkanVertexArray` aggregate and resolves it at draw
time (so a re-allocate under the same identity is picked up without a re-set). Raw buffers also
needed `VK_BUFFER_USAGE_INDEX_BUFFER_BIT` at create time — #1054's create-time-usage-bit lesson, one
role later.

**Rejected:** routing the hardware MDI arm through the index SSBO the software raster reads.
`vkCmdDrawIndexedIndirectCount` *requires* a bound index buffer, so it would mean rewriting the
registry's `DrawElementsIndirectCommand` records into a non-indexed encoding and forking VG's GL and
Vulkan draw paths — which is what the raw facade exists to avoid.

## The ride-along fix (separate commit)

`GPUBufferLockManager::Wait` carries a standing note asking that its loop be audited when Vulkan's
`AllocatePersistentUploadStorage` stops being a stub. It just did. GL's `ClientWaitFence` flushes
inside the wait so an unsubmitted fence still completes; Vulkan **stages** its signal for the next
submit, so an intra-frame ring wrap would spin the render thread forever —
`VirtualMeshRegistry::RebuildPools` stages every resident page in one call with no submit in
between, so any scene whose page geometry exceeds the 8 MiB ring would hang. `Reserve` now refuses
to wrap on such a backend and the caller takes its existing direct-upload path, gated on a new
`SupportsIntraFrameFenceCompletion()` that defaults to false. **OpenGL is unchanged** — it overrides
to true and keeps wrapping exactly as before.

## The second ride-along fix (separate commit) — a pre-existing full-suite failure

`VulkanDrawPath.ProductionFrustumCullWritesRootDataForIndirectDraw` fails in a full-suite run and
passes in isolation. It is **not a flake and not this branch**: deterministic across runs, and the
merge base fails identically (same line, same two stub hits) with none of these changes applied.

The tenant drives production `CommandDispatch::DrawMeshInstanced`, which binds `Renderer3D`'s shared
UBOs — process statics created by the **first** `Renderer::Init` in the process, which in this binary
is always OpenGL (`RendererAttachedTest` is the only fixture that calls it, and it creates a GL 4.6
context). Any GL renderer test registered ahead of this one leaves `CommandDispatch` binding
GL-owned UBO handles that cannot resolve on Vulkan, so `BindUniformBuffer` counts two
`PreconditionFailure` stubs and the zero-stub assertion fails for a reason unrelated to the draw
path. Bisecting the suite order pinned it to the nearest such neighbour in registration order.

A shipped app cannot reach this — `--rhi` picks one backend before `Renderer::Init` and never
switches — so it is a test-binary artefact, not an engine defect, and paying a registry lookup on
every UBO bind would be the wrong trade. The guard is a **precondition skip, not a weakened
assertion**: the zero-stub check keeps full strength wherever the tenant runs (any filtered Vulkan
invocation, where it still passes), and the skip reason names the precondition instead of leaving a
confusing red.

## Verification

**Vulkan, live editor**, `[RHI] Backend: Vulkan (source: --rhi flag)`, RTX 4090 / driver 104.256.0 /
API 1.4.351. `VirtualGeometryTest.olo`, Deferred, **identical camera poses** before and after:

| raster arm | before | after |
|---|---|---|
| hardware MDI (`hwRasterMode: forcemdi`) | ground plane + skybox only | **renders, matches the OpenGL reference** |
| software raster (`swRasterMode: forcesoftware`) | ground plane + skybox only | **renders, matches the OpenGL reference** |
| mesh shader (`hwRasterMode: auto`) | nothing | still nothing — **#1058**, see below |
| classic mesh path (`enabled: false`) | renders | renders, unchanged |

Three angles captured and looked at each time; the OpenGL run at the same poses is the reference.
Log after: **0** `#691` stub hits, **0** dropped draws, **0** unfed storage bindings, **0** shader
errors, **0** device faults.

**Tests.** Full suite: 7404 tests, 0 failed (the one pre-existing failure above is now a documented
skip). `Vulkan*:VirtualGeometry*:VirtualCluster*:VirtualMesh*:GPUCache*:GPUCircular*` — 214
passed, 1 skipped (`VirtualMeshRealAssetCook.StanfordDragon`, a fetch-on-demand asset absent in this
worktree), 0 failed. New `VulkanRawBufferUploadTest` (plumbing, device-gated) pins that the
persistent mapping is real host-coherent memory, that a sub-range upload lands at its offset without
touching its neighbours, that an over-range one is refused and counted, and that the element-buffer
record is not a stub while a junk handle still is.

**The test-seam repair is the real deliverable.**
`VulkanPassSuite.VirtualGeometryMdiCountDrawsHandAuthoredClusters` built its index data with
`IndexBuffer::Create` while production mints a **raw** handle and binds it a second way as
`SSBO_VIRTUAL_INDICES` — so the buffer it exercised was not the buffer that breaks. It now builds it
the production way (`CreateBufferHandle` → `AllocateBufferStorage` → `UploadBufferSubData` → raw VAO
→ bound at binding 42) and asserts `GetUnimplementedStubHitCount()` **and**
`GetUnfedStorageBindingCount()` are unchanged across the frame. That assertion is backend-shaped
rather than feature-shaped: it catches the next unlowered path in whatever pass reaches it first.

**OpenGL:** every Vulkan edit is under `Platform/Vulkan/`; the one shared-layer change is the new
capability whose GL override preserves today's behaviour exactly.

## What this does NOT do — read before merging

**Virtual geometry is still invisible on Vulkan at DEFAULT settings** on a mesh-shader-capable
device, because `hwRasterMode: auto` picks the mesh arm. That is **#1058** and it is not a stub:
`VulkanStorageBuffer` serves *draws* a CPU snapshot of the GPU-produced draw-args buffer, so the
mesh task stage reads `DrawCount == 0` and calls `EmitMeshTasksEXT(0)` — a legal launch, so nothing
warns. #1058 carries the patch, the regression test, and the reason it cannot ship here: applying it
makes the mesh arm run and then **loses the device**, reproducibly (3/3 with the guard, 0/3 without,
one build apart), on an out-of-bounds SSBO read the snapshot had been absorbing. Landing half of
that would turn "does not render" into "kills the editor".

So this PR uses `Refs #1052`, not `Closes`. The issue's own acceptance table — the three stubs — is
complete and evidenced; whether that closes it is your call.

## Review guide

**Where I'd look hardest**

1. `GPUCircularBuffer.h` `Reserve` — the refuse-to-wrap arm. It changes a shared-layer class for a
   backend property, and I am relying on "the caller's direct-upload path is complete on both
   backends" being true after this PR. If that reasoning is wrong the failure is a silent
   performance cliff, not a crash.
2. `VulkanRendererAPI::UploadBufferSubData` — the barrier bracket is copied from `CopyBufferSubData`
   minus the host-read half. I convinced myself the host half is unnecessary for an upload; a second
   opinion would be worth having.
3. `VulkanPassSuiteTest.cpp` — the MDI tenant's rewritten buffer construction. If I got the
   production shape subtly wrong, the test goes back to being green about the wrong object, which is
   the exact failure this PR is about.

**What I verified, and how** — named above: three camera angles on Vulkan against an OpenGL
reference at identical poses, the arm-by-arm A/B table, a clean log on four counters, and the two
new suites. The check that would have failed if this were wrong is the pass-suite stub-count
assertion, which every pre-fix build violated.

**Least confident about** — the `requireHostCoherentMap` allocation. Demanding
`HOST_VISIBLE|HOST_COHERENT` makes `vmaCreateBuffer` fail rather than hand back an unmapped buffer,
which is what I want, but it is a hard requirement on a memory type; on a device where no such type
exists the ring never forms and every page load takes the direct path. That is correct and now
loud, but it is untested on any GPU but this one.

One more honest note on the skip guard: it keys on `Renderer3D::HasInitialized()`, which is exact
*today* because nothing initialises Renderer3D on Vulkan in this binary. If a Vulkan
`Renderer::Init` fixture ever appears, the guard becomes conservative — it would skip a tenant that
could have run — rather than wrong. A precise version would need a per-handle backend-owner query,
which the registry stores but does not expose.

**Deliberately not tested** — the ring-wrap refusal itself. It needs a scene whose page geometry
exceeds 8 MiB in one `RebuildPools`, which `VirtualGeometryTest.olo` (~2.4 MB) does not reach and
`VirtualGeometryStress.olo` reaches only through the residency budget. I reasoned it from the fence
semantics and the standing comment rather than reproducing the hang.

Refs #1052

## Summary by CodeRabbit

- **New Features**
  - Vulkan now supports persistent mapped uploads and sub-data buffer updates.
  - Vulkan rendering supports raw index buffers for draw and instanced-draw paths.
  - Buffer uploads validate ranges and safely handle allocation failures.

- **Bug Fixes**
  - Prevented unsafe ring-buffer wrapping on backends that cannot complete fences within a frame.
  - Improved detection of resized or reallocated index buffers.
  - Vulkan upload and index-buffer operations now avoid silently dropping data.

- **Tests**
  - Added Vulkan coverage for raw uploads, index buffers, and related draw paths.

- **Documentation**
  - Expanded guidance for Vulkan buffer writes, fallbacks, and snapshot handling.

### DIFF
--- a/OloEngine/src/OloEngine/Renderer/GPUCache/GPUCircularBuffer.h
+++ b/OloEngine/src/OloEngine/Renderer/GPUCache/GPUCircularBuffer.h
@@ -97,6 +97,29 @@ namespace OloEngine
             }
             if (m_Head + bytes > m_SizeBytes)
             {
+                // Wrapping means waiting on a range this FRAME already fenced.
+                // That only terminates on a backend whose fence can complete
+                // without a queue submit: OpenGL's ClientWaitFence flushes
+                // inside the wait, Vulkan STAGES the signal for the next submit
+                // and would spin forever (issue #1052 — the audit
+                // GPUBufferLockManager::Wait asks for, now that Vulkan's
+                // AllocatePersistentUploadStorage is real). Refuse the
+                // reservation instead; the caller's direct-upload path is a
+                // complete lowering on both backends, so this costs staging
+                // bandwidth, never correctness.
+                if (!RenderCommand::SupportsIntraFrameFenceCompletion())
+                {
+                    static bool s_Warned = false;
+                    if (!s_Warned)
+                    {
+                        s_Warned = true;
+                        OLO_CORE_WARN("GPUCircularBuffer: a {}-byte reservation would wrap the {}-byte ring, and this "
+                                      "backend cannot complete a fence within the frame — falling back to direct "
+                                      "uploads for the rest of this batch (issue #1052)",
+                                      bytes, m_SizeBytes);
+                    }
+                    return nullptr;
+                }
                 m_Head = 0;
             }
             m_LockManager.WaitForLockedRange(m_Head, bytes);
--- a/OloEngine/src/OloEngine/Renderer/RenderCommand.h
+++ b/OloEngine/src/OloEngine/Renderer/RenderCommand.h
@@ -373,6 +373,13 @@ namespace OloEngine
             s_RendererAPI->IssueBarrierBatch(flags, barriers);
         }
 
+        // See RendererAPI::SupportsIntraFrameFenceCompletion (#1052). No live
+        // API means no fence can complete at all, so the safe answer is false.
+        [[nodiscard]] static bool SupportsIntraFrameFenceCompletion()
+        {
+            return s_RendererAPI && s_RendererAPI->SupportsIntraFrameFenceCompletion();
+        }
+
         [[nodiscard]] static bool SupportsRenderGraphFenceSubmission()
         {
             return s_RendererAPI && s_RendererAPI->SupportsRenderGraphFenceSubmission();
--- a/OloEngine/src/OloEngine/Renderer/Renderer3D.h
+++ b/OloEngine/src/OloEngine/Renderer/Renderer3D.h
@@ -1,5 +1,12 @@
 #pragma once
 
+// Self-containment: the inline template at CreateRenderStreamDrawCall uses
+// OLO_PROFILE_FUNCTION(), so this header owes the macro rather than borrowing
+// it from whatever the includer happened to pull in first. Every existing
+// includer already had it (the PCH, or a sibling header), which is why a
+// missing include here compiled everywhere for so long — until a test TU
+// included this header on its own and Linux/Clang named it.
+#include "OloEngine/Debug/Instrumentor.h"
 #include "OloEngine/Renderer/RHI/RHITypes.h"
 #include "OloEngine/Renderer/Passes/CommandBufferRenderPass.h"
 #include "OloEngine/Renderer/FluidRenderData.h"
--- a/OloEngine/src/OloEngine/Renderer/RendererAPI.h
+++ b/OloEngine/src/OloEngine/Renderer/RendererAPI.h
@@ -269,6 +269,26 @@ namespace OloEngine
         {
             return false;
         }
+
+        // True when a fence from CreateFence can complete WITHOUT the caller
+        // forcing a queue submit first (issue #1052).
+        //
+        // OpenGL's ClientWaitFence flushes inside the wait
+        // (GL_SYNC_FLUSH_COMMANDS_BIT), so a fence whose commands were never
+        // submitted still completes. VulkanRendererAPI::CreateFence STAGES its
+        // signal for the next submit, so the same wait spins forever —
+        // GPUBufferLockManager::Wait says exactly this and asks that the loop
+        // be audited when Vulkan's AllocatePersistentUploadStorage stops being
+        // a stub. It has (#1052), and this is that audit's result.
+        //
+        // The one consumer is a persistent-mapped ring: it may only WRAP
+        // within a frame — which means waiting on a range fenced earlier in
+        // the same frame — on a backend that answers true. False is the safe
+        // default, so a new backend cannot inherit the hang by omission.
+        [[nodiscard]] virtual bool SupportsIntraFrameFenceCompletion() const
+        {
+            return false;
+        }
         [[nodiscard]] virtual bool SubmitRenderGraphFenceSegment()
         {
             return false;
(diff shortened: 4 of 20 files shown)


drsnuggles8:
@coderabbitai review

Issue #1067, opened 2026-09-05
Texture hot-reload is broken for every asset-system texture: Reload() stbi_loads a project-relative path against the CWD


drsnuggles8:
## Problem

Texture hot-reload never works for a texture that came from the asset system. Editing a `.png` under the project and saving it logs an error and leaves the old pixels on screen; you have to restart the editor.

```
[error] OloEngine: OpenGLTexture2D::Reload: failed to re-read texture 'Assets/Textures/Tilemap/TilesetDemo.png'
```

## Cause

`OpenGLTexture2D::Reload()` (`Platform/OpenGL/OpenGLTexture.cpp`) re-reads with

```cpp
data = ::stbi_load(m_Path.c_str(), &width, &height, &channels, 0);
```

`m_Path` is whatever `SourcePath` was set to at load time, and `TextureSerializer::TryLoadRawData` deliberately stores the **project-relative** path (two sites in `Asset/AssetSerializer.cpp`):

```cpp
image.SourcePath = metadata.FilePath.string(); // so GetPath() -> pack re-read works
```

So `m_Path` is `Assets/Textures/.../foo.png`, `stbi_load` resolves it against the process CWD (`OloEditor/`), and the real file is at `OloEditor/SandboxProject/Assets/...`. It can never resolve. This is not specific to one texture — it is the path *every* project texture takes.

The reload is reached from `EditorAssetManager.cpp:348`:

```cpp
if (auto texture = existing.As<Texture2D>(); texture && texture->Reload())
```

## Why no test catches it

`Rendering/PropertyTests/TextureInPlaceReloadTest.cpp` builds its texture with an **absolute** path from `TempFile(...)`:

```cpp
Ref<Texture2D> texture = Texture2D::Create(path.string(), /*srgb=*/false);
```

That construction path sets `m_Path` absolute, so `Reload()` finds the file and the test passes. The broken case is specifically "loaded through `TextureSerializer`, then hot-reloaded", which nothing exercises. A regression test should go through `EditorAssetManager` rather than `Texture2D::Create(path)`.

## Suggested fix

Resolve a relative `m_Path` against `Project::GetProjectDirectory()` inside `Reload()` before handing it to `stbi_load` — the same resolution every serializer already does (`Project::GetProjectDirectory() / metadata.FilePath`). Leaving `SourcePath` relative keeps the asset-pack re-read that the comment protects.

[...]

Issue #1059, opened 2026-09-05
Flaky test: VulkanDrawPath.ProductionFrustumCullWritesRootDataForIndirectDraw fails only in a full-suite run — two BindUniformBuffer calls resolve to nothing


drsnuggles8:
## Problem

`VulkanDrawPath.ProductionFrustumCullWritesRootDataForIndirectDraw` fails **only inside a full-suite run**, and passes every time in isolation. It is not a timing/tolerance flake: the assertion that fails is the RHI's own *"no silent fallback"* counter, and the log carries a matching one-shot diagnostic immediately before it.

```
[RHI/Vulkan] BindUniformBuffer(unresolvable buffer) - inputs did not resolve; no-op (further calls counted, not logged)
..\OloEngine\tests\Rendering\VulkanDrawPathTest.cpp(775): error: Expected equality of these values:
  api.GetUnimplementedStubHitCount()
    Which is: 2
  0u
    Which is: 0
```

So two `BindUniformBuffer` calls resolved to nothing and were swallowed as no-ops. In a real frame that is a draw reading a stale or zero uniform buffer, which is exactly what the stub counter exists to make loud.

## Evidence

Observed on `feature/tilemap-tileset-2d-646` (a 2D-tilemap branch that touches **zero** Vulkan/RHI/culling files — verified by `git diff --name-only`), Debug, `dev-cached` tree.

| Context | Result |
|---|---|
| Full suite (`OloEngine-Tests.exe`, 7439 tests, 16.5 min) | **FAIL** — 2 stub hits |
| `--gtest_filter=VulkanDrawPath.ProductionFrustumCullWritesRootDataForIndirectDraw` | pass |
| `--gtest_filter=VulkanDrawPath.*` (all 6), repeated 3x | pass 3/3 |

Two facts that narrow it considerably:

- **2821 tests ran before it** in the failing run, including the `VulkanBringUp` / `VulkanBarrierLowering` suites, so the Vulkan device had already been brought up and torn down earlier in the same process.
- **`unresolvable buffer` appears exactly once in the entire 7439-test run** — immediately before this assertion. It is a single, specific resolution failure, not background noise. The five sibling `VulkanDrawPath` tests that ran straight after it all passed.

Environment: RTX 4090, driver 104.256.0, Vulkan API 1.4.351, Windows 11. The box was under load (a full suite plus sibling worktree builds), which may be a contributing factor but does not by itself explain a handle failing to resolve.

## Why it matters

[...]

Issue #451, opened 2026-06-28
Codegen: SceneSerializer blocks for complex-field components (#380 follow-up)


drsnuggles8:
Follow-up to #380 / #450. The fourth slice (#450) generates `SceneSerializer.cpp` serialize/deserialize blocks for the **all-trivial** components (every member a primitive / `glm::vec*` / `std::string`). The harder cases were deliberately left hand-written and are tracked here.

## Shipped

- **First slice — `AssetHandle` / `UUID` fields (#467):** the codegen now classifies `AssetHandle`/`UUID` (a `u64` wrapper) as a trivial-ish field type and round-trips it (`static_cast<u64>` on write, `.as<u64>` + the implicit `UUID(u64)` ctor on read — format-identical to the hand-written handle blocks). A component whose only non-primitive fields are asset handles / entity-reference UUIDs is now fully generated; `UIWorldAnchorComponent` migrated off its hand-written block. Four components that newly classify as trivial are kept hand-written via `kComponentsCustomSerialize`: `DialogueComponent` / `IKTargetComponent` / `PerceptionComponent` (Sanitize/clamp deserialize the round-trip would drop) and `IDComponent` (entity identity — serialized as the top-level `Entity:` line; auto-gen would emit a bogus sub-map + double-`AddComponent` on load, and **neither coverage test catches that**, so it's a hard exclusion).
- **Second slice — `enum` / `enum class` fields (#477):** the classifier now recognises an enum field by matching the member type's leaf name against a scan of every `enum`/`enum class` definition under `OloEngine/src` (`CollectEnumTypes` in `main.cpp`) and round-trips it as an `int` (`static_cast<int>` on write, `.as<int>` on read, cast back to the member's own type via `static_cast<decltype(comp.member)>` — so a nested enum like `AnimationStateComponent::State` needs no qualified spelling). Regenerating flipped several components to all-trivial that then needed `kComponentsCustomSerialize` exclusions for omitted runtime fields / clamps / non-`m_` keys: `Rigidbody3DComponent` (enum keyed `BodyType`, runtime `m_RuntimeBodyToken` + omitted fields), `StreamingVolumeComponent` (runtime `IsLoaded` + radius clamps), `FogVolumeComponent` (Shape/Priority clamps), `UIButtonComponent` (runtime `m_State`), `UISliderComponent` (runtime `m_IsDragging`).
[...]

PR #1068, opened 2026-09-05
feat(tilemap): 2D tilemap / tileset system — component, batched renderer, collider gen, editor painter

### LINKED ISSUE #646: 2D: tilemap / tileset system (component, batched renderer, editor painter, collider gen)
## Problem

There is no tilemap / tileset system. Building tile-based 2D levels (platformer, top-down) is entirely manual today.

Surfaced by an engine capability survey (minor gap — 2D is a secondary use case for this engine).

## Current state

- 2D rendering exists: `Renderer/Renderer2D.cpp` (sprite/quad/circle/text batching), `SpriteRendererComponent`, `CircleRendererComponent`, `TextComponent`.
- 2D physics is genuine Box2D v3 (`b2CreateWorld` / `b2World_Step` in `Scene/Scene.cpp`) with `Rigidbody2DComponent` / `BoxCollider2DComponent` / `CircleCollider2DComponent`, script-bound.
- **No tilemap/tileset** component, editor tool, or runtime renderer (the "Tilemap" grep hits were unrelated). Pure-2D tile levels must be hand-placed sprite-by-sprite.

## Proposed work (if 2D becomes a target)

- `TilemapComponent` + `Tileset` asset (texture atlas + per-tile metadata, incl. collision flags).
- Batched tilemap renderer (chunked, culled) on top of `Renderer2D`.
- Optional collider generation from tiles into Box2D.
- Editor tile-painting panel; scene serialization for the tilemap layers.
- Consider Tiled (`.tmx`/`.tsx`) import as a fast path.

## Notes

Lowest-priority gap given the engine's 3D-action-RPG sweet spot. File it so it's tracked, but only worth building if a 2D project is actually on the table. `std::vector<struct>` scene-serialization codegen (issue #451 follow-ups) may need to land first for clean tile-layer persistence.

🤖 Filed from a capability survey with [Claude Code](https://claude.com/claude-code)

## Score
```olo-score
capability: 5
craft: 5
stability: 1
decay: 1
effort: 8
confidence: 0.8
learning: 3
fun: 3
kano: delighter
blocked_by: []
blocks: []
```

drsnuggles8:
Adds the 2D tilemap / tileset system: a `Tileset` asset, a `TilemapComponent`, a chunked and culled renderer on `Renderer2D`, merged Box2D collider generation, and an editor painting panel — plus a demo tileset and scene so the thing can actually be opened.

## What landed

| Piece | Where |
|---|---|
| `Tileset` — Tiled-style atlas slicing + per-tile metadata (`Solid`, `Flags`, `Type`) | `OloEngine/Tilemap/Tileset.{h,cpp}` |
| `.olotileset` asset — loader, extension registration, YAML, hot-reload | `OloEngine/Tilemap/TilesetSerializer.cpp` |
| `TilemapComponent` + `TileLayer` | `OloEngine/Tilemap/TilemapComponent.{h,cpp}` |
| Chunked, frustum-culled batch renderer | `OloEngine/Tilemap/TilemapRenderer.{h,cpp}` |
| Greedy-rectangle collider generation into Box2D | `OloEngine/Tilemap/TilemapColliderBuilder.{h,cpp}` + `Scene::BuildTilemapColliders` |
| Editor painter — paint / erase / rect fill, tileset picker, layers, undo | `OloEditor/src/Panels/TilemapPainterPanel.{h,cpp}` |
| Demo atlas, `.olotileset`, 40x24 platformer scene | `OloEditor/SandboxProject/Assets/` |

## Two answers the issue asked for

**#451 is not a blocker — it already landed.** The issue suspected the `std::vector<struct>` scene-serialization codegen "may need to land first". It doesn't: persistence is *fully* generated (scene YAML, binary scene, save-game lists, MCP field registry, `AllComponents` tuple) with no hand-written `SceneSerializer` block and no `kComponentsCustomSerialize` entry. I rebuilt `GenerateBindings` and diffed the generated `.inl` — it flipped no other component.

**Tiled `.tmx`/`.tsx` import is deferred, deliberately.** The repo has no usable XML parser: pugixml exists only bundled inside assimp and MaterialX, with an existing ODR-collision workaround and a `/FORCE:MULTIPLE` link. That is not the "fits cleanly" the issue conditioned it on. Tiled's JSON export (`.tmj`) *would* fit via the existing nlohmann-json dependency — worth a scored follow-up rather than doubling this diff.

## Bugs found on the way (own commits)

- **`InstancePlacementSerializer` resolved paths against the asset dir**, producing `.../Assets/Assets/...`. Every `.oloinstances` load and save missed. It was the pattern I copied for `TilesetSerializer`, which is how I found it.
- **`AssetImporter::TryLoadData` held its registry mutex across the whole serializer call.** `FMutex` is not recursive, so any serializer resolving a dependent asset deadlocked the editor with no crash, no assert and no log line. `MeshSerializer` loads its `MeshSource` from inside `TryLoadData` and was one ordering accident from freezing. Lock now guards the lookup only.

## Filed, not fixed

- **#1067** — texture hot-reload is broken for *every* asset-system texture (`Reload()` stbi_loads a project-relative path against the CWD). `OpenGLTexture.cpp` belongs to the #624 session, so I left it alone.
- **#1059** — `VulkanDrawPath.ProductionFrustumCullWritesRootDataForIndirectDraw` fails only in a full-suite run. Not mine (zero Vulkan/RHI files in this diff; passes in isolation and 3/3 within its own suite), but the failing assertion is the RHI's own no-silent-fallback counter, so it is a real one-shot resolution failure worth chasing.

## Review guide

**Where I'd look hardest**

1. `Scene.cpp` — `Scene::RenderTilemaps` is called from **four** draw sites (runtime 3D/2D, editor 3D/2D) because those sites are not otherwise consistent: the editor's 3D-mode callback previously drew only text. If a fifth ever appears, this needs to be there too. Sprites still are not drawn in that editor 3D site — a pre-existing gap I deliberately did not widen.
2. `Scene.cpp::BuildTilemapColliders` — places the static body from the **world** transform while the neighbouring `Rigidbody2D` loop uses the **local** one. That divergence is intentional (it matches the renderer) but it means this function does not look like its neighbour.
3. `AssetImporter.cpp` lock-scope narrowing — the argument that releasing early is safe rests on the registry being populate-once and the serializers stateless. Both hold today; worth a second opinion.

**What I verified, and how**

- **7439-test full suite: 7395 passed.** The two failures are accounted for above (one mine and fixed — a hardcoded panel count; one pre-existing, filed as #1059).
- **38 tilemap tests** across unit / functional / L8, all green, plus 219 green across the coverage meta-tests and component round-trips after the review fixes.
- **Pixel evidence** in `OloEditor/assets/tests/visual/Tilemap_{Overview,Zoom,LargeMapSameView,OffScreen}.png`, looked at, not just asserted. `Tilemap_Zoom.png` and `Tilemap_LargeMapSameView.png` are **byte-identical** (same MD5): a 24x24 map and a 240x240 map through the same camera produce the same frame, which is the chunk-culling claim.
- **Live editor in both Edit and Play**, which is the only thing that caught two of the bugs above. The Statistics panel reads `Renderer2D — Draw Calls: 1, Quads: 234` for the demo scene: one draw call for the visible map, 234 quads out of 1920, at 231 FPS. `OloEngine.log` clean apart from the pre-existing missing-OpenSans and Steam-appid errors. Play mode was driven over MCP (`olo_scene_play` + `olo_screenshot` through the runtime camera), not clicked.

**Least confident about**

Nothing significant. The item that was here — whether the demo scene's ortho camera clipped the map on Play — is now **verified rather than calculated**: I attached the MCP diagnostics server, called `olo_scene_play`, and captured through the runtime primary camera. The map renders correctly (screenshot in the PR discussion). `/code-review` was right that the original `±1` clip range was wrong; the fix is confirmed working, not just arithmetically sound.

The nearest remaining unknown is `AssetImporter.cpp`'s narrowed lock scope — the safety argument rests on the serializer registry being populate-once and the serializers stateless, which holds today but is a property someone could break later without noticing.

**Deliberately not tested**

Tiled `.tmx`/`.tsx` import, which is not implemented (see above).

*(This section previously said collider generation was untested end-to-end against a stepping Box2D world. It now is — `Functional/Physics/TilemapCollidersStopABodyTest.cpp` drops a dynamic body onto a generated floor and asserts it lands, with two controls: `GenerateColliders` off must fall through, and a tileset with no solid entries must fall through. That last one is what stops a decoration layer silently becoming a wall.)*

Closes #646

## Summary by CodeRabbit

* **New Features**
  * Added layered, tileset-based tilemaps with configurable grids, visibility, solidity, opacity, and ordering.
  * Added tileset assets with atlas metadata, tile labels, flags, and solid-tile settings.
  * Added tilemap rendering, physics collider generation, save-game persistence, and scripting support.
  * Added a Tilemap Painter editor panel for painting, erasing, rectangle fills, layer editing, tileset selection, and undo/redo.
  * Added Tilemap component configuration to the scene hierarchy inspector.
  * Added a tilemap demo scene and sample tileset asset.

* **Bug Fixes**
  * Corrected instance placement asset paths and improved asset-loading lock handling.


drsnuggles8:
## 🤖 Self-review @ `8a6e222c90b6ca730fd4f5bac20de49130693a10`

Reviewed the full diff at **high** effort before opening the PR. **7 findings, 7 fixed, 0 dismissed.**

| # | Finding | Fix |
|---|---|---|
| 1 | Panel unreachable once its window was closed — `Visible` and `m_ShowTilemapPainter` deadlocked each other | View menu now drives `Visible` too |
| 2 | Collider used the **local** transform while the renderer used the **world** one — a parented tilemap would draw and collide in different places | `BuildTilemapColliders` decomposes `GetWorldTransform` |
| 3 | `OLO_PROPERTY` on `Width`/`Height` generated unbounded C# setters that re-index every row without resizing storage | Annotation removed; extent changes go through `Resize()` |
| 4 | `Resize` had no upper bound and is exposed to Lua — `resize(100000, 100000)` was a `bad_alloc` | `kMaxExtent = 4096`, bounded both ends |
| 5 | Save-game load validation skipped `ColliderFriction`/`Restitution`; a NaN reached `b2ShapeDef.material` | Finiteness check + clamp |
| 6 | Demo scene's ortho camera clipped the map on Play | **Verified numerically** — NDC z = +10.0. Widened to ±100 (now +0.1) |
| 7 | Negative scale produced negative half-extents, inverting polygon winding | `std::abs` on extents, sign kept on the centre |

Finding 6 was reported as *plausible* and is the one I'd have been tempted to wave through; the arithmetic confirmed it was real. It is exactly the class of bug that looks perfect in the editor and renders nothing on Play.

Re-ran after the fixes: **219/219** green across the tilemap suites, coverage meta-tests and component round-trips, plus a fresh editor capture (`Draw Calls: 1, Quads: 234`) and a clean `OloEngine.log`.

drsnuggles8:
## Play-mode verification (follow-up)

Two things in the original PR body were weaker than they needed to be. Both are now closed.

**1. Play mode is verified, not calculated.** I'd said the driver couldn't press Play. That was wrong — `driver.ps1 -Action attach -AllowWrites` brings up the MCP diagnostics server, and `olo_scene_play` needs no clicking at all. (The `olo_*` tools don't surface in a non-interactive session, but the server is plain JSON-RPC over localhost HTTP, so you just talk to it directly.)

Captured through the **runtime primary camera** in Play: bedrock, soil, grass, the terraced hill, water pit, brick platform with ladder, plank bridge and the decor props all render. `/code-review` was right that the original ±1 ortho clip range was wrong; the ±100 fix is confirmed working.

**2. Collider generation is now tested end-to-end.** `Functional/Physics/TilemapCollidersStopABodyTest.cpp` (commit `e7f5cfe74`) drops a dynamic `Rigidbody2D` onto a generated tilemap floor and asserts it rests on the surface, with two controls so the positive case can't pass vacuously:

- same map, `GenerateColliders` off -> must fall through
- tileset with no solid entries -> must fall through (a decoration layer must not silently become a wall)

3/3 green; 187 green across the tilemap, coverage and 2D-physics suites.

**Capability gap logged:** every `olo_physics_*` tool goes through `JoltScene`, so `olo_physics_list_colliders` reports `total: 0` for a 2D scene — indistinguishable from "no colliders exist". Noted on #607.

drsnuggles8:
## CodeRabbit review: 10 fixed, 2 skipped (`6ee33fba1`)

Each finding verified against the code before acting. 189 tests green afterwards, plus a fresh editor capture to confirm nothing regressed.

### Fixed

| Finding | What the check showed |
|---|---|
| `AxisCount` used the leading margin only | **Real.** My own class doc says "border around the whole image", but the code subtracted `margin` once. At 78 px / 16 px tiles / 4 px spacing / 2 px margin it counts 4 columns and slices the fourth into the trailing border; Tiled gives 3. Also widened to u64 — `spacing` comes from an asset file and `tile + spacing` can wrap to zero in u32. Two regression tests added. |
| Per-tile `translate() * scale()` | Real; composed directly, dropping two temp mat4s and two 4x4 multiplies from the innermost loop. |
| `PickTile` non-positive `TileSize` | **Real and subtle.** `0.0f/0.0f` is NaN, NaN fails *every* comparison, so the bounds check passed and `static_cast<u32>(NaN)` was UB. |
| Click consumed outside Edit | Real — `OnUpdate` is Edit-gated, the consume wasn't, so a Simulate click was swallowed with nothing painting. |
| MCP toggle didn't sync `Visible` | Real, same root cause as the menu bug. Replaced the per-caller patch with one sync point covering every toggler. |
| Not bound in `BindPanelsToScene` | Real — that's the function that runs on every transition; my binding was in a different one. |
| Stroke stranded when interrupted | Real. Added `OnFrameTick(eligible)`, called unconditionally, mirroring `TerrainEditorPanel::OnFrameTick` and its stated rationale. |
| `SetContext` state on scene swap | Real, and the "only when different" qualifier is load-bearing — it's called every frame. |
| No Clear button on the tileset slot | Valid gap vs sibling asset slots. |
| Unbounded `AddLayer` | Valid; `kMaxLayers` now shared with the save-game cap. |

### Skipped

**Missing `OLO_TEST_LAYER` in `McpEditorPanelsTest.cpp`** — false positive. The marker is on line 10 (`// OLO_TEST_LAYER: unit`) and the file appears correctly in `docs/test-catalogue.unit.md`; the finding cited line 28.

**Persist the tilemap `b2BodyId` and destroy it in `DestroyEntity`** — real, deliberately deferred. The repo's mechanism is an `OnComponentRemoved<T>` specialisation, which requires a `b2BodyId` member on the component. `b2BodyId` has no `operator==`, so that deletes `TilemapComponent`'s defaulted equality and forces a hand-written per-field one — precisely the category CLAUDE.md flags as a recurring silent-bug source (a new field silently missing from copy/equality). The harm is bounded: `OnPhysics2DStop` destroys the whole world, so it is stale collision until Play ends, and only if a tilemap entity is destroyed mid-play. Better as its own scoped change than as a trade of defaulted equality inside this PR.

Issue #1074, opened 2026-09-06
Tests: ~102 visual-evidence cases fail in a single-process full run but pass in isolation


drsnuggles8:
`OloEngine-Tests.exe` run as ONE process fails ~102 cases that all pass in isolation. CI cannot see this: `OloEngine/tests/CMakeLists.txt` uses `gtest_discover_tests`, which registers every case as its own ctest entry, so CI runs **one process per test** and no two of these tests ever share an address space. The monolithic run is the only configuration where cross-test state pollution exists, and nothing exercises it routinely.

This is not theoretical. Chasing it during PR #1066 turned up four real defects, two of which were engine bugs:

- `RendererProfiler::EndFrame` subscripted a frame-history ring that `Shutdown()` had cleared — an out-of-range write on an empty vector that surfaced as heap corruption in whatever scene test ran next. **29 crashes per full run**, none of them anywhere near the cause.
- A test fixture calling `CommandDispatch::Shutdown()` dropped the UBO references `Renderer3D` publishes once at `Init()`, so every later rendering test drew with no camera or material UBO bound.

Both are fixed on that branch. The crashes are gone (0), but **~102 failures remain**.

## Current state

Measured on `feature/vulkan-parallel-all-passes-1013` after those fixes: 7492 tests run, 0 crashes, **102 unique failures**, essentially all visual-evidence / scene tests. Worst-hit suites:

| Count | Suite |
|---|---|
| 10 | `VirtualGeometryVisualEvidence` |
| 6 | `ScreenSpaceTemporalResolveEvidenceTest` |
| 4 | `WaterFoamSprayRainVisualEvidenceTest`, `DDGIVisualEvidenceTest` |
| 3 | `LightmapBleedRoom`, `ClusteredLightingVisualEvidenceTest`, `SubmeshMaterialPathParity` |

A representative sample (`SSGIVisualEvidenceTest`, `SSRVisualEvidenceTest`, `FogVisualEvidenceTest`, `LightmapBleedRoom`, `SubmeshMaterialPathParity`) **passes 10/10 when run as its own filter**, so these are pollution, not broken tests.

The failure signature is consistent: the feature-on and feature-off captures come out identical, or a differential that should be large reads ~0. For example `LightmapBleedRoom` wants `meanAbsDiff > 3.0` and measures `0.0007` — the effect under test simply is not rendering.

## What is already known

- **Two poisoners found and fixed, and neither moved the count.** So there are more, and they are independent.
[...]

Issue #1009, opened 2026-09-01
CI: move the Linux jobs onto the self-hosted runner and fix the caching that never warms


drsnuggles8:
## Problem

CI wall-clock is now the dominant bottleneck in the development loop. Measured over the 36h window 2026-08-31T08:54Z → 2026-09-01T15:06Z (250 runs, `gh run list`):

| Workflow | trigger | n | median | note |
|---|---|---|---|---|
| Windows | pull_request | 27 | **159.5 min** | 17/27 cancelled |
| Sanitizers | pull_request | 24 | **154.4 min** | 15/24 cancelled |
| SonarCloud Scan | pull_request | 26 | **130.8 min** | 14/26 cancelled |
| Vulkan OFF | pull_request | 17 | 59.3 min | |
| Windows / Sonar / Vulkan / Steam / Sanitizers | push (master) | ~55 | 29–97 min | re-tests the merge commit CI already tested |

Step breakdown of the last green Windows run (33502248619): **configure 40 min, build 102 min, ctest 18 min.**

Three concrete defects, all measured, not inferred.

### 1. vcpkg rebuilds ~113 of 116 ports from source on every run

Run 33502248619 log: `Restored 3 package(s) from .vcpkg-binary-cache in 139 ms`, then `Installing 4/116 utfcpp... Building...` down the whole list. That is the 40-minute configure step.

`.github/actions/setup-vcpkg/action.yml:322` keys the binary cache on `hash(vcpkg.json, cmake/triplets/**, cmake/overlay-ports/**) + run_id`, and `restore-keys` only falls back **within the same hash**. Any baseline bump or added dep therefore drops the cache to zero ports, and it stays cold until a run survives to its post-job save.

### 2. sccache hit rate on Windows is 0.00%

Same run: `Compile requests 1516 / Cache hits 0 / Cache misses 1516 / Cache hits rate 0.00 %`. It restored `sccache-windows-release-31571109867`, a snapshot from 2026-08-26.

Zero hits out of 1516 is not source churn — a source change cannot invalidate every TU. It is a hash-input mismatch, most likely `cl.exe` drifting under the rolling `windows-latest` image since that snapshot was taken.

### 3. Cache saves almost never run, and the store is over the cap

[...]

Issue #281, opened 2026-06-03
Flaky CI crash: PhysicsLayerFilteringTest SEH 0xc0000005 in multithreaded Jolt step


drsnuggles8:
## Summary

`PhysicsLayerFilteringTest.AlphaPassesThroughBetaAndBothLandOnGround` intermittently crashes on CI with an access violation, then hangs to the 120s ctest timeout:

```
2631: [16:08:00] Contact removed between entities 7398568275756549410 and 4151067279753280066
2631: unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
2631/2739 Test #2631: PhysicsLayerFilteringTest.AlphaPassesThroughBetaAndBothLandOnGround ... ***Timeout 120.02 sec
```

Seen once so far (GitHub Actions, Windows, Release). The `0xc0000005` (access violation) is the root cause; the 120s timeout is a consequence (the Jolt worker threads / scheduler never join after the corruption, so the process hangs).

## Root-cause analysis (likely class, not exact line)

A **rare data race in the multithreaded Jolt step**, via the custom `JoltJobSystemAdapter` bridge between Jolt and the engine's `FScheduler`:

- gtest's SEH handler is per-thread and caught this on the **main thread** inside the test body. Jolt's `JobSystemWithBarrier::BarrierImpl::Wait()` executes physics jobs on the *calling* (main) thread concurrently with `FScheduler` workers running the same job pool — a classic use-after-free window when a `Job` is freed (`Release` → `FreeJob` → `FixedSizeFreeList::DestructObject`) by one thread while another dereferences it.
- `JoltJobSystemAdapter`'s own destructor comment documents a **history of TSan job-lifetime races** here (the shutdown race was fixed via `m_OutstandingTasks`; this looks like the runtime sibling).
- This particular test widens the window: all three bodies use **user-defined physics layers**, so every broad-phase pair runs through `PhysicsLayerManager::ShouldCollide` (shared-lock) on the contact-validate path — extra concurrent job/lock traffic most physics tests don't generate.

## Why it's CI-only

- Local and CI run the **same Jolt commit** (`6cef2a22`, current `master` HEAD as of 2026-06-01). Not reproducible locally: 1/1 Debug + **100/100 Release** pass on a 16-core box.
- The differentiator is the **environment**: GitHub-hosted Windows runners have ~2-4 cores vs 16 locally → far more thread preemption → wider race windows. Classic "only flakes on CI" data race.

## Ruled out

[...]

PR #1075, opened 2026-09-06
fix(ci): the Windows sccache entry never existed — the cache store was read-only (#1073)

### LINKED ISSUE #1073: CI: the two Windows jobs take ~4 h each — sccache reports a 0% hit rate
The two Windows jobs each take **~4 hours**, and they are the critical path on every PR — the Linux sanitizer jobs finish long before them. Measured on run [33993528874](https://github.com/drsnuggles8/OloEngineBase/actions/runs/33993528874) (`feature/vulkan-parallel-all-passes-1013`, head `e67a150f6`):

| Job | Duration |
|---|---|
| `build` (Windows, MSVC, Release) | 21:36 → 01:32 = **3 h 56 m** |
| `ASan (Windows/clang-cl)` | 21:36 → 01:33 = **3 h 56 m** |

Step breakdown of the `build` job:

| Step | Duration | Note |
|---|---|---|
| Configure CMake | **53 m** | vcpkg port restore/build |
| **Build with CMake** | **2 h 31 m** | the dominant cost |
| Run tests with CTest | **28 m** | 7572 tests, one process each, `--parallel 4` |
| everything else | ~2 m | checkout, SDKs, cache save/restore |

## The headline finding: sccache never hits

From the job's own `sccache --show-stats`:

```
Compile requests                    1630
Cache hits                             0
Cache misses                        1630
Cache hits rate                     0.00 %
Average compiler                  24.690 s
```

Zero hits, not a low rate. 1630 compilations × ~24.7 s of compiler time is essentially the entire 2 h 31 m build step — the cache is fully wired up (restore step, save step, `SCCACHE_DIR`, zero read/write errors, `Non-cacheable compilations 0`) and buying nothing.

A *low* hit rate would suggest eviction; an *exactly zero* hit rate points at something systematically busting the hash or a cache that is never actually restored. Candidates to check, roughly in order:

- **`SCCACHE_CACHE_SIZE: "2G"`** (`Windows.yml:81`) against 1630 MSVC objects. If the working set exceeds 2 GB, the run thrashes its own cache and saves a partial set. The `Save sccache storage` step completing in 47 s is consistent with a much smaller artifact than expected.
- **Whether the restore is finding anything at all.** The key is `sccache-windows-2025-release-${{ github.run_id }}-${{ github.run_attempt }}` with a `sccache-windows-2025-release-` prefix fallback, which looks right — but the logs should be checked for an actual "Cache restored from key" line rather than assumed.
- **A per-run input in the compile hash.** Generated headers (`GenerateBindings` / OloHeaderTool output) or any define carrying a run id, timestamp or non-deterministic path would poison every object.
- **MSVC debug-info mode.** sccache cannot cache compilations that write to a shared PDB; `/Z7` is cacheable where `/Zi` is not. `Non-cacheable compilations 0` argues against this, but it is cheap to confirm.

Fixing this one line item plausibly reclaims most of the 2.5 h build step, and it is the same lever for both Windows jobs.

## Secondary items

- **vcpkg configure at 53 m.** The workflow comments claim only a baseline or triplet bump should pay a real port build, so this deserves the same "is the cache actually hitting?" audit as sccache.
- **CTest at 28 m.** `gtest_discover_tests` registers all 7572 cases individually, so each pays full engine static-init in its own process. `--parallel 4` is conservative for a workload that is process-startup-bound rather than CPU-bound; sharding across matrix jobs would scale nearly linearly since the cases are independent by construction. Note the tradeoff: one-process-per-case is exactly what makes CI blind to cross-test state corruption (see the four defects fixed in #1013's PR #1066), so consolidating for speed would *gain* coverage but expose a separate backlog of ~100 monolithic-run failures. Not free.
- **A self-hosted Windows runner.** `olo-ci-1` / `olo-ci-2` already exist for Linux. A persistent Windows runner would keep the vcpkg and sccache directories genuinely warm across runs rather than restoring/saving them every time, which sidesteps the cache-restore question entirely.

## Acceptance

- [ ] sccache hit rate on a warm Windows `build` job is materially above zero, with the measured before/after recorded here.
- [ ] The `build` step duration is re-measured and reported (no target promised until the cause is known).
- [ ] The vcpkg configure step's cache behaviour is audited the same way — hitting, or a reason why not.
- [ ] Whatever is found is written down in the workflow next to the setting it explains, so the next person does not re-derive it.

## Score

```olo-score
capability: 3
craft: 1
stability: 3
decay: 5
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
## The finding

The Windows build reported `Cache hits 0 / Cache misses 1630 / 0.00 %` on every run for weeks. It is **none of the four candidates** on the issue — not `SCCACHE_CACHE_SIZE`, not a per-run input in the compile hash, not `/Zi` vs `/Z7`, and the restore step was not misconfigured.

`sccache-windows-2025-release-*` **did not exist in the repository's cache store at all.** A full listing of all 14 live entries contained no Windows sccache entry — not stale, not partial, none. The restore step said so and nobody was reading it:

```
Cache not found for input keys: sccache-windows-2025-release-33993528874-1,
                                sccache-windows-2025-release-
```

and three and a half hours later, at the save:

```
##[warning]Cache reservation failed: You have reached your configured budget,
           your cache is now read only to prevent additional charges.
Failed to save: Unable to reserve cache with key sccache-windows-2025-release-…,
                another job may be creating this cache.
```

The same pair appears on the **nightly master run** ([34005232373](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34005232373)), which is this workflow's only fleet-wide cache producer. The warmer itself could not write, so the deadlock was stable rather than flaky: nothing to restore → every compile a miss → four-hour build, indefinitely, green.

Past the 10 GB cap GitHub does **not** evict. It flips the whole store read-only and refuses every save in the repository, as a `warning`, exit 0 — so no step status anywhere reflects it.

## Why the existing defences did not hold

[`ci-cache-that-looks-alive.md` §3b](../blob/master/docs/agent-rules/ci-cache-that-looks-alive.md) had already found the read-only budget in #1009 and written the rule down, and `cache-prune.yml` had already been built to keep the store under the cap. Both were in place. Three things filled it anyway.

**1. The janitor could not collect the biggest orphan.** `cache-prune.yml` grouped by the key minus its trailing run id, and said of the leftovers: *"one entry per dead prefix is a rounding error against the cap"*. The vcpkg key is `vcpkg-<os>-<triplet>-<64 hex manifest hash>-<run_id>-<attempt>`, so **every `vcpkg.json` bump started a brand-new prefix whose sole member was immortal**. The largest single entry in the store was 1988 MiB of a superseded manifest hash, created 2026-09-03 and never read once — `last_accessed_at` never moved off its creation time. 21 % of the cap, not a rounding error.

**2. Six open PRs do not fit in a 10 GB store.** Windows and the Linux sanitizer jobs saved their 1.2–2 GB sccache snapshot on `pull_request` runs, where the entry is scoped to `refs/pull/N/merge` and readable by that one PR. Six concurrent PRs — a routine day here — is up to 12 GB of PR-private sccache against a 9.3 GiB wall. Those entries bought a marginal delta over the master snapshot the PR already restores, and paid for it by holding the store read-only, which is what stopped the nightly from writing the snapshot every branch actually restores from. **The PR runs were funding the mechanism that kept them cold.**

**3. A finished investigation was still charged.** `flaky-repro-281.yml` ran a nightly Windows engine build holding 886 MiB, for a flake that `Windows.yml` itself records as fixed at the root.

## Changes

| | |
|---|---|
| **`cache-prune.yml`** | Prefix now strips a trailing content hash as well as the run id, so a hash bump *supersedes* instead of orphaning. Plus a 14-day `last_accessed_at` age-out for genuinely renamed prefixes. Sweeps at 01:05 (before both nightlies) as well as 06:41. **Fails** when the post-sweep store is still over 8800 MiB — everything it deletes is provably superseded or unread, so if what remains is that close to the wall a human has to shrink something. That alarm is what was missing. |
| **`.github/actions/save-cache-pruned`** (new) | Keys ending in `-<run_id>` are append-only, so a twice-daily janitor cannot help when two saves land thirty minutes apart. This saves, asks the API whether the entry *actually landed* (the save step's own status cannot tell you), and only if it did not deletes this ref's older entries under the same prefix and retries. Order is deliberate: pruning first would risk trading a warm snapshot for no snapshot. Emits an `::error::` annotation if the entry still never lands. |
| **`Windows.yml`, `asan.yml`** | Save through the new action; no sccache save on `pull_request` runs. `actions: write` added to the four jobs that need it. |
| **`flaky-repro-281.yml`** | Nightly cron and the stale branch `push:` trigger removed; `workflow_dispatch` kept. Reinstate only alongside reopening #281. |
| **`docs/agent-rules/actions-cache-budget.md`** (new) | The budget arithmetic, the fleet's size table, and the rules for adding a cache. |
| **`ci-cache-that-looks-alive.md`** | §4 claimed GitHub was LRU-evicting. It is not — that misreading is what made an over-full store sound self-correcting, and it contradicted §3b in the same file. Corrected, with a pointer to the new guide. |

A dry run of the new prune rules against the live cache listing keeps 10 entries totalling 5856 MiB and frees **3406 MiB** of 9258 — including the 1988 MiB orphan the old rules could never collect.

## Verification

Two `workflow_dispatch` runs of `Windows.yml` on this branch: the first seeds the entry, the second reports the hit rate. **Numbers to follow on #1073; nothing is claimed until then** — in particular no build-time saving is asserted anywhere in this diff.

One consequence worth stating plainly: with PR saves off, PR runs depend entirely on the nightly master snapshot. Until a nightly succeeds in writing one, PR runs stay exactly as cold as they are today — no worse, but not better either.

## Secondary items from the issue

- **vcpkg configure at 53 m** — **a SECOND, independent fault** (corrected on the issue; the
  first version of this description said "same root cause", which was wrong). The vcpkg
  binary cache has never once been restored: `setup-vcpkg` spells the path one way for the
  restore and another for the save, and `actions/cache` hashes the raw path string into each
  entry's `version`. Every vcpkg entry in the store has `last_accessed_at == created_at`.
  Fixed in `ecb6b46`; measured separately.
- **CTest at 28 m** — not taken here. Raising `--parallel` past 4 on a 4-vCPU runner is speculative, and the real lever (batching ~6,900 process launches) is #1074's territory, which is working the ~102 monolithic-run failures that consolidation would expose. Left to that session rather than half-done from this one.
- **A self-hosted Windows runner** — filed as a follow-up rather than done here; it is a larger change than one PR should carry, and this fix has to be measured on the hosted path first.

Closes #1073

## Summary by CodeRabbit

* **New Features**
  * Added resilient cache saving with automatic pruning and retry support.
  * Added scheduled cache maintenance and automated cleanup for stale or unused entries.
  * Added capacity monitoring with error reporting when storage remains over limits.

* **Bug Fixes**
  * Prevented pull-request builds from saving cache entries.
  * Fixed cache restoration when directory paths differ in formatting.
  * Improved handling and reporting when cache storage refuses writes.

* **Documentation**
  * Added guidance on cache capacity, pruning, path matching, and diagnosing misleading cache status messages.

* **Chores**
  * Limited the flaky-reproduction workflow to manual runs.

### DIFF
--- a/.github/actions/save-cache-pruned/action.yml
+++ b/.github/actions/save-cache-pruned/action.yml
@@ -80,21 +80,34 @@ runs:
         REF: ${{ github.ref }}
       run: |
         set -uo pipefail
-        if keys=$(gh api -X GET "repos/${GH_REPO}/actions/caches" \
-                    -f "key=${KEY}" -f "ref=${REF}" \
-                    --jq '.actions_caches[].key' 2>/dev/null); then
-          if printf '%s\n' "$keys" | grep -Fxq "${KEY}"; then
-            echo "landed=true" >> "$GITHUB_OUTPUT"
-            echo "cache entry '${KEY}' is present on ${REF}."
+        # Poll rather than ask once. A cache entry does not appear in the listing the
+        # instant the upload finishes, and reading "absent" during that window would
+        # delete a perfectly good predecessor and raise a false alarm -- turning the
+        # signal this action exists to provide into one that cries wolf. A refused
+        # reservation, by contrast, is refused immediately and stays absent, so a few
+        # seconds of patience separates the two cleanly.
+        landed=unknown
+        for attempt in 1 2 3; do
+          if keys=$(gh api -X GET "repos/${GH_REPO}/actions/caches" \
+                      -f "key=${KEY}" -f "ref=${REF}" \
+                      --jq '.actions_caches[].key' 2>/dev/null); then
+            if printf '%s\n' "$keys" | grep -Fxq "${KEY}"; then
+              landed=true
+              break
+            fi
+            landed=false
           else
-            echo "landed=false" >> "$GITHUB_OUTPUT"
-            echo "cache entry '${KEY}' is NOT present on ${REF} -- the save was refused."
+            # An API hiccup must never trigger a delete. Unknown is treated as landed.
+            landed=unknown
           fi
-        else
-          # An API hiccup must never trigger a delete. Unknown is treated as landed.
-          echo "landed=unknown" >> "$GITHUB_OUTPUT"
-          echo "::warning::could not query the cache store; skipping the prune/retry path."
-        fi
+          [ "$attempt" -lt 3 ] && sleep 10
+        done
+        echo "landed=${landed}" >> "$GITHUB_OUTPUT"
+        case "$landed" in
+          true)  echo "cache entry '${KEY}' is present on ${REF}." ;;
+          false) echo "cache entry '${KEY}' is NOT present on ${REF} after 3 checks -- the save was refused." ;;
+          *)     echo "::warning::could not query the cache store; skipping the prune/retry path." ;;
+        esac
 
     # Scoped to THIS ref on purpose. GitHub scopes a cache entry to the ref that wrote
     # it -- a PR run's entry is visible to that PR alone, and only default-branch entries
--- a/.github/actions/save-cache-pruned/action.yml
+++ b/.github/actions/save-cache-pruned/action.yml
@@ -1,0 +1,173 @@
+name: Save cache, pruning this ref's superseded entries on refusal
+description: >
+  `actions/cache/save`, plus the two things it does not do: notice that the save was
+  REFUSED, and make room so a retry succeeds.
+
+  WHY THIS EXISTS (#1073). Every cache key in this repo ends in `-<run_id>-<attempt>`
+  so each save writes a fresh, immutable snapshot and `restore-keys` warms from the
+  newest one. That shape is right and it is APPEND-ONLY: a save never replaces the
+  entry it supersedes, it sits beside it until the nightly `cache-prune` sweep.
+
+  The repo's Actions cache store is capped at 10 GB, and past the cap GitHub does not
+  evict -- it turns the WHOLE STORE READ-ONLY ("Cache reservation failed: You have
+  reached your configured budget, your cache is now read only"). The steady set of live
+  entries is ~7 GB, so the moment two or three heavy jobs each hold an old AND a new
+  snapshot at once, the cap is breached and every save in the repo is refused until the
+  next daily sweep. A daily sweep cannot fix that: the breach happens between two saves
+  minutes apart, not between two days.
+
+  It is not a hypothetical. It is the measured state that made
+  `sccache-windows-2025-release-*` cease to exist ENTIRELY -- which is why the Windows
+  build reported a 0.00 % hit rate on 1630 compilations, run after run, for weeks, while
+  every step stayed green. See docs/agent-rules/actions-cache-budget.md.
+
+  ORDER MATTERS AND IT IS SAVE-THEN-PRUNE. Pruning first would be simpler and is wrong:
+  if the retry were then refused anyway, we would have deleted the warm entry the next
+  run was going to restore, turning a stale cache into no cache. So: save; ask the API
+  whether the entry actually landed; only if it did not, delete this ref's older entries
+  under the same prefix and save again. A warm snapshot is never given up for one that
+  has not been written.
+
+  Never fails its job. A cache save that turns a green run red would be strictly worse
+  than the cold cache it is fixing, so every failure here is an annotation.
+
+inputs:
+  key:
+    description: 'Full cache key to save under, including the trailing -<run_id>-<attempt>.'
+    required: true
+  prefix:
+    description: >
+      Key prefix identifying this cache's lineage on this ref -- normally `key` minus its
+      trailing `-<run_id>-<attempt>`. On the retry path every entry on THIS ref matching
+      this prefix (other than `key` itself) is deleted. Must not be empty, and must not be
+      a prefix of an unrelated cache's key.
+    required: true
+  path:
+    description: 'Path to save, as passed to actions/cache/save.'
+    required: true
+  token:
+    description: 'Token with `actions: write` on this repository.'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Save
+      # Explicit `always()` on every step here: this action is invoked from an
+      # `if: always()` step so that a CANCELLED run still banks what it built, and a
+      # composite step with no condition inherits `success()`, which a cancelled job
+      # does not satisfy. Without these the prune/retry path would be dead in exactly
+      # the runs that need it most.
+      if: always()
+      continue-on-error: true
+      uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: ${{ inputs.path }}
+        key: ${{ inputs.key }}
+
+    # `actions/cache/save` treats a refused reservation as a WARNING and exits 0
+    # (observed: "##[warning]Cache reservation failed ..." then "Failed to save: Unable
+    # to reserve cache with key ..., another job may be creating this cache", step
+    # outcome SUCCESS). The step's own status therefore cannot answer this -- ask the API.
+    - name: Did the entry land?
+      id: check
+      if: always()
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        GH_REPO: ${{ github.repository }}
+        KEY: ${{ inputs.key }}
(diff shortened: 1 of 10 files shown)


drsnuggles8:
## Key audit + review round (`55d93b2`)

Went back over every cache key in the fleet on the premise that a key mistake here is invisible until a build is four hours long. Six things changed; two were real bugs in this PR.

### Verified against the live API instead of assumed

`key=` on `GET /actions/caches` is a plain **string** prefix, not a lineage match — `key=sccache` returns three unrelated lineages — `ref=` does scope correctly, and **omitting `key` returns everything**. A destructive step must not depend on any of that holding, so the prune loop now re-checks each row locally and deletes only entries matching `^<prefix>-<run_id>[-<attempt>]$` on exactly this ref. A future sibling key such as `sccache-windows-2025-release-debug-…` is skipped rather than eaten. Truth-tabled before commit.

### The prefix is now derived from the key, not passed beside it

A `prefix` input is the same string written twice. Bump the runner image and `key`, `restore-keys` and the prefix must change in lockstep across four workflows — miss one and the action either prunes nothing (a silently full store) or prunes a lineage that is not its own. Both are #1073 again. The `prefix` input is gone; the action strips `-<run_id>-<attempt>` from the key itself, and skips the prune entirely for a key that is not that shape.

### The serious one — the recovery path could delete a warm cache to make room for nothing

Caught in review. This action runs under `if: always()` so a cancelled run still banks what it built. A run cancelled **early** has no `SCCACHE_DIR`; `actions/cache/save` then fails on the missing path, the API confirms no entry, and the recovery path would delete the predecessor and retry into the same missing path. Cold, not stale — precisely what the save-then-prune ordering exists to prevent, reached through a different door. It now asks the filesystem whether there is a payload first and treats inconclusive as "no". `steps.<save>.outcome` cannot answer this: the action catches its own errors and reports success either way. `flaky-repro-281` is the job most exposed to it — a cancel at `timeout-minutes: 110` is its expected exit — and it now uses the same action, with the job permissions it needs.

### `runner.environment == 'github-hosted'` does not imply "not a PR"

Also caught in review, and my PR description above overstated it. The three Linux sanitizer saves now carry an explicit `github.event_name != 'pull_request'`. Flip `OLO_LINUX_SELF_HOSTED`, or open a fork PR, and a PR run lands on a hosted runner and banks a ~1.25 GiB entry only that PR can read.

### Age-out 14 → 30 days, because 14 was an assumption

Fourteen days assumed a daily read cadence that measurement contradicts: `Linux-vulkan-prebuilt-sdk-true-1.4.321.0` went 2026-08-19 → 2026-09-02 unread, then four more days, while the Sanitizers nightly ran every one of them. An age-out shorter than a prefix's real read interval deletes live caches. Patience is free here — the orphans it targets are permanent, and the recurring leak is handled by the hash strip, not by this.

### A leak this repo's own workflow creates

`cache-prune` now collects entries whose **branch** is gone, mirroring the closed-PR rule. `/start-work` makes a `feature/*` branch per task and deletes it on merge, and a `workflow_dispatch` run on one banks ~1.7 GiB against that ref that "newest per ref" would keep forever, readable by nothing. The two measurement runs for this issue create exactly that.

### Two numbers corrected rather than repeated

- `1630 × 24.69 s` is **11 h 10 m of compiler time**, not the 2 h 31 m wall clock. Ninja parallelism is 4.4× on four vCPUs. The issue text conflated them and I had carried that into the guide.
- The sccache row in the budget table was `SCCACHE_CACHE_SIZE`, not a measurement — **no such entry has ever existed to measure**, which is the whole point of this issue. It is now marked as the 0.9–2.0 GiB estimate it is (`sccache-flaky-281`, same cap and same build, compressed to 886 MiB), with the steady set stated as 5.7 GiB likely / 6.8 GiB at the bound against a ~9.3 GiB wall.

### Not taken

Nothing else from the review. The remaining suggestions were either already true in the current code or would have widened the diff without a defect behind them.

---

Seeding run [34027467561](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34027467561) is in its Configure step on `a6c68d1`; the warm read will run on the final commit. Numbers to #1073 when they exist.

PR #1077, opened 2026-09-06
fix(gpu-scene): stage ModelComponent submeshes into the canonical scene

### LINKED ISSUE #1065: GPU Scene: ModelComponent geometry never reaches the canonical scene, so ray tracing cannot see most content
Found while bringing up #1056 (hybrid ray-traced soft shadows) on `Courtyard.olo`.

## What happens

`olo_rt_scene_stats` on Courtyard — a 166-entity scene whose centrepiece is Sponza:

```
tlasInstances        49
blasByClass          static 2, deformed 0, rigidDynamic 0, masked 0, unsupported 0
unsupportedInstances 0
accelerationStructureBytes  28544   (28 KB)
```

Two BLAS and 28 KB for Sponza. The acceleration structure holds the small props and nothing else, so anything that traces against the TLAS — ray-traced shadows today, hybrid reflections (#1057) and the GPU path tracer later — is tracing an almost-empty world in this scene.

## Why

`Renderer3D::ExtractGPUSceneMesh` has exactly two callers, both in `Scene.cpp`:

- the `MeshComponent` loop (`SubmitMeshSourceClassic`, ~line 7047)
- the `InstancedMeshComponent` loop (~line 10857)

`ModelComponent` has no such call, so a model-backed entity is drawn by the raster path and never staged as a canonical GPU Scene instance. Courtyard's `Sponza` entity is a `ModelComponent` (`olo_entity_list_fields` shows only `TagComponent`, `TransformComponent`, `ModelComponent`).

## Why it is worth its own issue

**The existing counter cannot report it.** `unsupportedInstances` is 0, because that counter records records the RT scene REJECTED — and these records never arrive to be rejected. `GeometryClass::Unsupported` is likewise 0. So "the ray tracer cannot see most of this scene" is currently indistinguishable from "everything traced fine", which is precisely the silent-fallback shape the house rule exists to prevent.

That made a real bug hard to find: a Vulkan projection defect in #1056 rendered a hard band of false shadow across the frame, and it was invisible on Courtyard because there was nothing to trace. It only surfaced on `ContactShadowTest.olo` (5 mesh entities → 5 BLAS).

## Suggested scope

1. Extract `ModelComponent` submeshes into the canonical GPU Scene, reusing `SubmitMeshSourceClassic` rather than transcribing it (that helper's own comment explains why a second copy is the wrong move).
2. Whatever remains genuinely unsupported should be **counted**, so the zero means "nothing was rejected" rather than "nothing arrived". A `GPUScene`-side "entities with renderable geometry that produced no canonical instance" counter would close the blind spot even before (1) lands.

Related: #977 (canonical GPU Scene), #978 (acceleration structures), #1056 (first ray-query consumer), #1057 (hybrid reflections, will hit the same wall).

## Score
```olo-score
capability: 8
craft: 5
stability: 8
decay: 3
effort: 3
confidence: 0.8
learning: 3
fun: 5
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
`ModelComponent` geometry was drawn by the raster path and never staged as a canonical
GPU Scene instance, so every ray-query consumer traced an almost-empty world in any
model-backed scene. On `Courtyard.olo` — 166 entities centred on Sponza — the acceleration
structure held **2 BLAS and 28 KB**.

The second half mattered as much as the first: `unsupportedInstances` was `0` because that
counter records what the RT scene *rejected*, and these records never arrived to be
rejected. "The ray tracer cannot see most of this scene" was indistinguishable from
"everything traced fine".

## What changed

**Count what does not arrive.** A new `GPUSceneUnsupportedCategory::NotExtractable` is
reported from inside `Renderer3D::ExtractGPUSceneMesh` itself, on every rejection path,
rather than by each caller — so a path that starts extracting cannot forget to count its
own rejections. `olo_rt_scene_stats` gained a `gpuScene` block (live instance / geometry /
material records, `notStagedTotal`, and the per-category split), emitted **before** the
readiness early-return: "no TLAS was built" and "nothing was ever offered to build one
from" produce the same RT payload and need different fixes.

**Stage the submeshes.** The `ModelComponent` loop now walks the model's meshes once for
three consumers — the canonical records, the DDGI capture, and the shadow caster list —
instead of only running under the shadow/DDGI gate. Ray tracing must see the geometry
whether or not the scene casts shadows.

`SubmitMeshSourceClassic` is reused rather than transcribed, via a shared
`StageGPUSceneSubmesh` helper both paths now go through. The two paths differ in exactly
one thing, and it is the trap: **a model keeps its material table on the `Model`, not on
the mesh source.** A cold-loaded model's per-mesh `MeshSource`s carry no imported materials
at all, so resolving the key off the mesh source — what the existing overload does — would
have staged every Sponza submesh under the flat engine default. Hence a new
`ResolveGPUSceneMaterialKey` overload taking the imported material and its slot explicitly.

Model submeshes stage with `GPUSceneDrawLinkRequest::None`: the record is staged for ray
tracing and diagnostics, but the raster draw still goes out through
`Model::DrawParallel` / `SubmitMeshesParallel`, which carries no draw-link lane yet. Giving
it one is the raster half of #994, not this fix.

## Evidence — `Courtyard.olo`, Vulkan, live editor

| | before | after |
|---|---|---|
| `resident.tlasInstances` | 49 | **74** |
| `resident.blasByClass.static` | 2 | **24** |
| `resident.blasByClass.masked` | 0 | **3** |
| `resident.accelerationStructureBytes` | 28,544 | **8,079,872** |
| `resident.unsupportedInstances` | 0 | 0 |
| `gpuScene.instances` | 49 | **74** |
| `gpuScene.geometries` | 2 | **27** |
| `gpuScene.materials` | 2 | **27** |
| `gpuScene.notStagedTotal` | 28 | **3** |
| `gpuScene.notStagedByCategory.legacyModel` | 25 | **0** |

The before column reproduces the issue's numbers exactly, and `legacyModel: 25` is the
count that did not exist anywhere outside the ImGui profiler panel before this PR — the
proof the diagnosis was right, taken *before* the fix landed.

The 3 `masked` BLAS are worth noting: Sponza's alpha-masked fabric and plants landed in
the right `GeometryClass` rather than being flattened to opaque.

## Verification

- **CPU / CI contract**: `McpRayTracingStatsTest` (4 new cases — the block survives a
  non-ready RT status, every diagnostics category gets a distinct JSON key, a missing
  renderer reports `available: false` rather than zeros) and
  `GPUScene.UnsupportedSubmissionsAreCountedByReason`.
- **Regression guard**: `ModelComponentGPUSceneEvidenceTest` (L8) asserts every model mesh
  has a live instance record found by the same `(entity, VB, IB, submesh)` key extraction
  builds, that each record carries the **entity's** transform, and that nothing is counted
  as renderable-but-unstaged. Three golden captures. SKIPs without a GL 4.6 context.
- **Live A/B**: toggling `ModelComponent.Visible` on Sponza moves the TLAS between exactly
  49 / 2-static and 74 / 24-static, reproducing the pre-fix state on demand in one session.
  With Sponza in the TLAS the sun casts a real ray-traced shadow with a crisp boundary;
  without it, nothing shadows the ground.
- **Control scene**: `ContactShadowTest.olo` (5 `MeshComponent` instances,
  `notStagedTotal: 0`, untouched by this change) renders correct ray-traced contact
  shadows, proving the RT shadow pass itself is healthy in this build.
- `olo_shader_errors`: 0. Targeted suites (`GPUScene*`, `RayTracing*`,
  `McpRayTracingStats*`, `SubmeshMaterialPathParity*`, `RendererProfiler*`,
  `ModelComponent*`): 93 tests, all pass.

## Also in this PR: one FFmpeg build for every worktree

A separate commit, at the user's request, and unrelated to #1065 beyond having been
noticed while waiting on a build.

FFmpeg installed into `${PROJECT_SOURCE_DIR}/vendor` — inside the worktree — so every task
worktree rebuilt it: ~15 min each, a 420 MB source clone each (2.5 GB across the six on
this box) and a 19 MB install. The compiler cache cannot touch it, because FFmpeg builds
through its own `configure`/gmake rather than CMake's compiler launcher, and none of it
depends on our sources.

It now installs into a per-user, tag-keyed cache shared by every worktree, the same shape
`cmake/vendor/OpenUSD.cmake` already uses, overridable with `-DOLO_FFMPEG_CACHE_DIR`. On a
hit it skips `ExternalProject` entirely — `STAMP_DIR` defaults to the *consuming* build
tree, so a second worktree would otherwise see no stamps and re-run the build inside the
shared source dir. `OLO_FFMPEG_TAG` now feeds both the cache key and `GIT_TAG`, so a
version bump invalidates the cache instead of silently reusing the old build.
`-DOLO_FFMPEG_PREFIX` still wins over all of it.

Verified both branches at configure time (miss → "building ... one-off per machine"; hit →
"reusing the shared FFmpeg build"), then built and ran: `impl-Debug.ninja` carries 1591
references to the cache path and **zero** to the old in-tree path, and the video tests pass
(15 pass, 1 skips for a missing fixture, as before).

Known limit, carried over verbatim from the USD precedent: two worktrees running their
*first* FFmpeg build concurrently still share one tree. Sequence the first; every tree
after it is a lock-free cache hit.

## Review guide

**Where I'd look hardest**

1. `Scene.cpp` — the material resolution in the `ModelComponent` loop. The imported
   material comes from `Model::m_Materials[submesh.m_MaterialIndex]` while the key's
   *owner* half comes from the mesh source. That pairing is correct only because
   `CreateCombinedMeshSource` pushes `m_Materials` onto the shared source and the warm
   `.omesh` path assigns `m_Materials = cachedMaterials`, keeping the two tables
   index-aligned. If that ever stops holding, records name the wrong material and the
   raster frame still looks right.
2. `Renderer3DGPUScene.cpp::ExtractGPUSceneMesh` — the split of the old combined early-out.
   The `!GPUSceneExtractionActive` case must stay *uncounted* (nobody is extracting) while
   every branch below it counts. Getting that backwards inflates `notStagedTotal` on every
   frame outside extraction.
3. `Scene.cpp` — the restructured shadow/DDGI gate. The loop is no longer gated as a whole;
   `collectCasters` and the `MaterialCastsShadows` check inside must reproduce the old
   nested behaviour exactly, or model shadow casters silently stop or start.

**What I verified, and how** — the before/after telemetry above, taken from two live editor
sessions on Vulkan; the in-session visibility A/B that reproduces the pre-fix TLAS; the
control scene; and `ModelComponentGPUSceneEvidenceTest`, whose transform assertion is the
one that would fail if the counters went up while the geometry landed in the wrong place.

**Least confident about** — whether the material-record *owner* should be the mesh source
or the `Model`. On the cold load path each mesh owns a distinct source, so two submeshes
sharing one model material get two byte-identical material records instead of one. That is
correct but not minimal; a `Model`-level owner would dedupe. I kept the mesh-source owner
because it matches `ImportedMaterialOwner`'s existing contract and a `Model` created
ad-hoc has asset handle 0.

**Deliberately not tested** — the raster half. Model draws still do not consume a draw
link, so the raster frame is unchanged by construction; before/after captures differ by
RMSE 7.7–11.8 against a measured same-session noise floor of 5.9–6.8, i.e. within noise.
Migrating that path is #994's remaining scope. I also did not re-verify on OpenGL for the
ray-tracing half, because hardware RT is Vulkan-only here — the GPU Scene staging itself is
backend-independent and is covered by the GL-context evidence test.

Closes #1065


drsnuggles8:
## 🤖 Self-review @ `66f447f53fead711948756c4e11c20cc9c9f667c`

Reviewed the full diff at **high** effort before opening the PR.

- **Findings:** 1 · **Fixed:** removing the per-entity `ReportUnsupportedGPUScene(LegacyModel)` turned `if (!submesh) continue;` in the `ModelComponent` loop into an *uncounted* silent drop. `Model::ProcessMesh` returns `nullptr` on three real refusal paths (material mapping, index/vertex count past `u32`) and `Model::ProcessNode` pushes that null into `m_Meshes` unfiltered, so such a mesh would never reach `ExtractGPUSceneMesh` to be counted — `notStagedTotal` would read "everything staged" while geometry was missing. That is the exact blind spot this PR exists to close, reintroduced by the fix itself. Now reports `NotExtractable` before the `continue`.
- **Dismissed:** none.

Also cleared during review, recorded because each was a plausible break: the new 6-param `ResolveGPUSceneMaterialKey` is not viable for the existing 5-arg call site (no implicit `GPUSceneMaterialOverrideLane` → `u32` conversion), so no overload ambiguity; the model/mesh-source material tables stay index-aligned on both the cold and warm `.omesh` paths; instance keys cannot collide between the two paths; and the restructured shadow/DDGI gate is behaviourally identical to the old nested form.

**Test-suite note for reviewers:** large-filter local runs show failures from the known order/load-dependent set (#1074). Evidence that they are not from this branch: the *same binary* run twice produced *different* failure sets (13, then a 3-subset of those), and all of them pass in isolation. None are GPU Scene, ray-tracing or model-path tests. The suites that do cover this change — `GPUScene*`, `RayTracing*`, `McpRayTracingStats*`, `SubmeshMaterialPathParity*`, `RendererProfiler*`, `ModelComponent*` — are 93/93 green.

drsnuggles8:
## Third commit added: `fix(vulkan): publish the resting layout after an image copy`

Found while investigating an `IBLCache` error in the editor log during this PR's own Vulkan verification. It is **not** related to #1065 — it is a separate pre-existing bug, documented on this box since 2026-08-24 with the root cause never investigated. Added here at the user's request rather than deferred.

### What was wrong

`CopyImageSubDataRegion` is how the IBL bake writes a cube face. It transitioned the destination for the copy and stopped there, never publishing a resting layout to `VulkanImageInfoRegistry` — the one thing every other write path does. So the registry kept the creation-time `UNDEFINED` while the image really sat in `TRANSFER_DST_OPTIMAL`. That single omission produced both symptoms:

* the descriptor's recorded layout disagreed with the actual layout → the two `VUID-vkCmdDraw-None-09600` errors;
* `ReadFaces` (which reads the **registry**, not the frame tracker, because it submits outside the frame) saw `UNDEFINED` and refused — correctly, since it cannot distinguish "never written" from "written by a path that does not say so". That made `GetFaceData` fail, so `IBLCache::Save` bailed *after* writing its 44-byte header, and the **next** run failed to *load* the truncated file it had left behind. The visible error was a read error, one run late, at the wrong end of the problem.

### Result — Courtyard on Vulkan, empty cache

| | before | after |
|---|---|---|
| irradiance / prefilter save | fail | 98,348 / 2,097,164 bytes — identical to what OpenGL writes |
| next scene open | fails to load its own file | `Loaded IBL textures from cache` |
| `VUID-` lines | 2 | **0** |
| layout errors | present | **0** |

Frame unchanged; IBL lighting intact.

### The non-obvious part, for review

The settle must cover the **whole** destination image, not the copied subresource. The registry stores one layout per image, so settling just the copied range and then claiming the image is `SHADER_READ_ONLY` is false for every subresource not yet written — my first attempt did exactly that and turned 2 VUIDs into ~10 (`expects SHADER_READ_ONLY_OPTIMAL, current is UNDEFINED` for mips 5+ on all six faces of a prefilter cubemap baked to mip 4). `ForEachLayoutRun` emits one barrier per layout run using that run's real `oldLayout`, so a whole-image transition is valid whatever mix it is currently in. Same discipline `VulkanTextureCubemap::SetFaceDataMip` already uses.

Also included: the previously **silent** `return false` / `return` paths in `ReadFaces` and the copy path now say which one fired. That silence is why a bug reproducible since August stayed undiagnosed — the caller could only ever report `failed to read face 0 mip 0`.

### Regression evidence

A/B on the same filter, fix stashed vs applied: **251 passed both ways**. The two failures (`ReflectionProbeVisualEvidenceTest`, `VulkanParallelRecordingDevice`) are present in the untouched baseline — pre-existing and order-dependent. I ran this A/B specifically because those two failures were *stable* across repeated runs, which contradicts the usual flake signature, and one of them bakes an environment cubemap.

### ⚠️ Merge-order note

This touches `Platform/Vulkan/VulkanRendererAPI.cpp`, which `feature/vulkan-drawpath-correctness-1058` is also working in. **Whoever merges second should re-check this hunk.**

Not exercised by this session: the impostor baker and asset-preview paths, which also call `CopyImageSubDataRegion`. The whole-image transition is conservative, but they are untested here.

Issue #661, opened 2026-07-17
clang-cl + ASan on Windows: C++ throw through instrumented frames crashes (SEH 0xc0000005) — yaml-cpp malformed-input tests


drsnuggles8:
## Symptom
Tests that make yaml-cpp throw (`ParserException` on malformed input) crash under the Windows clang-cl ASan job with `SEH exception 0xc0000005 thrown in the test body` — no gtest assertion, no ASan report (gtest's SEH catcher intercepts first). First seen on PR #660's CI: `ExperienceCurveTest.SerializerRejectsMalformedYAML` + `CharacterClassDatabaseTest.SerializerRejectsMalformedYAML`.

## Root cause (reproduced + isolated locally, LLVM 21, `clangcl-asan` preset, Release)
With `--gtest_catch_exceptions=0`, ASan reports: **access-violation READ at null+8 inside the throw dispatch** (`r9 = 0x19930520` — the MSVC C++ throw magic) — the crash happens in the exception-dispatch machinery **before any catch clause runs**. Experimentally eliminated as causes:
- the malformed-input shape (unterminated flow sequence vs tab-indentation garbage — both crash),
- the catch type (`YAML::ParserException` vs base `YAML::Exception` — both crash).

Which call paths crash is **frame-layout and clang-version dependent**:
- `EngineSubsystemSmoke.ProjectLoadMalformedYAMLFailsCleanly` (throws through `YAML::LoadFile` in `ProjectSerializer`): passes everywhere.
- `InputActionSerializerTest.MalformedYAML`: **crashes on LLVM 21 locally, passes the current CI runner's clang** — it is the next victim when the runner's LLVM upgrades.

## Current mitigation (PR #660, commit aa2c1ee8)
The throwing sub-assertions in the two failing tests are compiled out under `OLO_ASAN_ENABLED && _WIN32` only (evidence trail in-file); root-key-rejection assertions stay active everywhere; Windows-ASan coverage of the yaml-cpp throw/catch plumbing continues via the passing `ProjectLoad` test.

## Follow-ups
- [ ] Apply the same guard to `InputActionSerializerTest.MalformedYAML` before/when the CI runner's LLVM upgrades (or centralize a `OLO_SKIP_IF_WIN_ASAN_YAML_THROW()` helper macro if a third case appears).
- [ ] Reduce to a standalone repro (yaml-cpp + /EHsc + /fsanitize=address + the crashing frame shape) and file upstream against LLVM.
[...]

PR #1079, opened 2026-09-06
fix(build): make the 21 headers that profile self-contained, ratchet the class, and drop the fixed #661 ASan guards

### LINKED ISSUE #1071: 21 headers use OLO_PROFILE_* without including Instrumentor.h — self-containment violations that only fail on Linux
## What

21 headers under `OloEngine/src/` call `OLO_PROFILE_FUNCTION()` / `OLO_PROFILE_SCOPE()` without
including `OloEngine/Debug/Instrumentor.h`. Each one compiles today only because every current
includer happens to have the macro already — from the PCH, or from a sibling header pulled in first.
`CLAUDE.md` says headers are self-contained; these are not.

## Why it matters — it already cost a CI round

Found on #1062. `Renderer3D.h` was one of them. Adding the **first** include of that header from a TU
that did not already have the macro turned it into a hard compile error on both Linux jobs:

```
VulkanDrawPathTest.cpp:43:
  Renderer3D.h:1620:13: error: use of undeclared identifier 'OLO_PROFILE_FUNCTION'
```

Two properties make this a nasty failure mode:

- **It is invisible until someone includes the header "wrong"**, and the person who trips it is never
  the person who wrote the omission. The blame lands on an unrelated one-line include.
- **A Windows-only local build cannot see it.** The test target force-includes the PCH there, so the
  macro is always in scope; only the Linux jobs name it. A green local suite is not evidence.

`Renderer3D.h` is fixed in #1062. The other 21 are untouched and equally armed.

## The list

Reproduce with:

```bash
for f in $(grep -rl "OLO_PROFILE_FUNCTION\|OLO_PROFILE_SCOPE" OloEngine/src --include=*.h); do
    grep -q "Debug/Instrumentor.h" "$f" || echo "$f"
done
```

Clustered in Audio (15 of them, mostly `SoundGraph/`), plus `Core/Reflection/MemberList.h`,
`Networking/Replication/NetworkPriorityQueue.h`, `Renderer/ShaderGraph/ShaderGraphAsset.h` and
`Scene/Entity.h`. (`Debug/Instrumentor.h` itself appears in the grep and is not a defect — it
*defines* the macros.)

## Suggested shape

Mechanical: add `#include "OloEngine/Debug/Instrumentor.h"` to each. It is a leaf header
(`Core/Log.h` + tracy) already included by 13 others, so there is no cycle risk. Worth doing in one
sweep rather than one-at-a-time as people trip over them.

The durable guard is a self-containment check — compiling each public header alone as its own TU —
which would catch this whole class instead of this one macro. That is a bigger piece of work and
probably its own issue; the sweep is worth doing either way.

Found while finishing #1052.

## Score
```olo-score
capability: 1
craft: 1
stability: 3
decay: 3
effort: 1
confidence: 1.0
learning: 1
fun: 1
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Two build/toolchain-correctness bugs. Neither changes engine behaviour; both are about
configurations Windows cannot see.

## #1071 — headers that use `OLO_PROFILE_*` without including `Instrumentor.h`

21 headers expanded `OLO_PROFILE_FUNCTION()` / `OLO_PROFILE_SCOPE()` without including the header
that defines them. They compiled only because every current includer already had the macro in
scope — from `OloEnginePCH.h`, or from a sibling header pulled in first. The omission is invisible
on Windows and a hard compile error on both Linux jobs the moment somebody adds the **first**
include of one of them from a TU without the macro; the blame then lands on that unrelated one-line
include. It already cost a CI round on PR #1062.

Landed in three commits:

1. The 21 includes.
2. A pre-commit ratchet (`header-profiling-include` → `scripts/check_header_profiling_include.py`)
   so the 22nd cannot come back. It reads the macro set out of `Instrumentor.h` at check time, so it
   cannot drift, and it opts out with `// OLO_PROFILING_INCLUDE_OK: <reason>`.
3. Self-review fixes: widening the ratchet found **three more** headers, including one worse than
   any of the original 21 — see the review guide.

## #661 — the Windows-ASan yaml-throw guards (`Refs #661`, not closed)

The issue asked for a centralized `OLO_SKIP_IF_WIN_ASAN_YAML_THROW()` helper and for the guard to be
applied to `InputActionSerializerTest.MalformedYAML`. **I re-tested first, and the premise no longer
holds**, so this does the opposite: every guard is removed and `OloEngine/tests/WinAsanYamlThrow.h`
is deleted.

#661 is the same defect as #497 — ASan redzones on the compiler-emitted EH metadata globals — and
PR #1024 pinned the ASan job to LLVM 23.1.0, which fixes it. The guards were written against LLVM 21
and nobody re-tested them after the pin moved.

**Left undone, deliberately:** nothing to file upstream, because it no longer reproduces. And this is
"passes on the pinned CI compiler", not a bisect — only LLVM 23.1.0 is installed on this box, so I
have no negative control proving my probe would have caught the old bug. The CI ASan job on this PR
is the real test.

**No workflow files touched.** `.github/workflows/asan.yml` belongs to the #1073 worktree, and
nothing here needed a change there.

## Sibling-worktree note

`Scene/Entity.h` is a one-line include addition and `feature/gpu-scene-modelcomponent-rt-1065` is
working in `Scene/Scene.cpp`. Trivial conflict risk, flagged per the batch registry.

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Task/InheritedContext.h:8` — the reason the ratchet was widened.
   `FInheritedContextScope` gates its **members and its move constructor's initializer list** on
   `#if OLO_PROFILE && TRACY_ENABLE`, and `OLO_PROFILE` is defined only by `Instrumentor.h`. A TU
   without it built a differently-shaped class and linked clean. That is an ODR mismatch, not a
   compile error, and it is the strongest argument in the PR for the ratchet existing at all. Worth
   a second opinion on whether the include is the right fix or whether the flag belongs somewhere
   that cannot be missing.
2. `OloEngine/tests/` — the four guard removals. If CI's ASan job disagrees with my local run, the
   first question is which clang it resolved, not whether the engine regressed.
3. `scripts/check_header_profiling_include.py:72` — the use-detection is textual. It strips comments
   and `#define <NAME>` prefixes before matching, which is enough for the tree today, but it is a
   regex over C++ and will eventually be wrong about something.

**What I verified, and how**

- **The standalone-header proof, which is the only local check that answers the Linux question.**
  One TU per header, the engine's real flags out of `compile_commands.json`, no PCH,
  `clang-cl -fsyntax-only`. Before: 18 of 31 failed, 198 `OLO_PROFILE` error lines. After: 30 of 33
  pass, 0 `OLO_PROFILE` errors. A green Windows build is explicitly *not* the evidence here — that
  is the whole point of #1071.
- **Ratchet, both directions:** exit 0 on the tree as committed, exit 1 naming the file when any one
  of the includes is reverted.
- `OloEngine-Tests` builds clean on `dev-cached`; 552 passed / 1 skipped / 0 failed across the
  Audio, SoundGraph, Scene, Reflection, Networking, ShaderGraph and Task suites.
- **#661, on the CI-pinned compiler.** `clangcl-asan`, Release, clang 23.1.0 (verified the binary
  imports `clang_rt.asan_dynamic` and carries `__asan_report*`), ASan runtime on PATH,
  `--gtest_catch_exceptions=0`: all four previously-guarded sites plus `InputActionSerializerTest.*`
  pass with every guard removed — 59 tests, 0 failures. A standalone probe (clang-cl + `/EHsc` +
  `/fsanitize=address` + yaml-cpp, six throw/catch shapes including #497's three-line repro) also
  passes. Full ASan `ctest` run in progress at time of writing; I will report it on the PR.

**Least confident about**

The #661 removal, for one specific reason: I have no negative control. Only LLVM 23.1.0 is installed
on this box, so I cannot show my probe failing on 21 and passing on 23 — I am relying on #497's
recorded before/after for that half. If the ASan job goes red here, the guards were removed too
early and the honest fix is to restore them, not to reshape the tests.

Secondary: whether `Debug/Profiler.h` should satisfy the ratchet. I made it satisfy the check, but
`NetworkPriorityQueue.h` still gets a direct `Instrumentor.h` include because it uses those macros
directly. Include-what-you-use says direct; someone may reasonably call the second include noise.

**Deliberately not tested**

`OloEditor`, `OloRuntime` and `OloServer` were not built locally — every TU in them includes the PCH
on line 1 exactly as the test target's do, so the risk profile is identical and CI covers them. I
also did not fix the three headers that still fail a standalone compile (`Task/Pipe.h`,
`Task/ParallelFor.h`, `Scene/Entity.h`): `Task/TaskShared.h` includes `<tracy/Tracy.hpp>`, which
drags `<windows.h>` in with `min`, `max` and `Yield` as macros. Different class, wider blast radius,
written up in `docs/agent-rules/pch-masked-missing-includes.md` rather than folded in here.

Closes #1071
Refs #661

## Summary by CodeRabbit

- **Build & Quality**
  - Added automated checks to detect missing profiling-related includes in engine headers.
  - Improved header self-containment and build consistency across platforms.

- **Testing**
  - Malformed YAML and scene-file tests now run across all supported configurations, including Windows AddressSanitizer builds.
  - Removed obsolete platform-specific test skips.

- **Documentation**
  - Added guidance on diagnosing hidden missing includes and documented the resolved Windows AddressSanitizer toolchain issue.


drsnuggles8:
## Ratchet now checks include *position*, not just presence

Applied the review finding on `scripts/check_header_profiling_include.py` (commit `0e5f17081`).

**The finding was valid but latent.** The check accepted the include anywhere in the file, so a header whose include sat below its first expansion would have passed while still failing to compile alone. No file in the tree is in that state today — the one apparent hit, `Renderer3D.h`, has the macro named in the comment *above* its include, which is a false positive of a presence-only grep and stays clean under the new logic.

Fixed anyway, because "passes the ratchet, fails on Linux" is precisely the blindness this ratchet exists to remove.

**One deliberate exemption:** a use inside a `#define` body does *not* constrain position. It expands wherever the macro is called, in some other file, so it obliges the header to own the include but says nothing about where it sits. `Core/PerformanceProfiler.h` is that case and keeps passing with its include anywhere in the file.

Supporting changes needed to make the ordering check trustworthy: comment masking is length-preserving (comments become spaces, newlines kept) so offsets and reported line numbers map to the original file; block comments are stripped before line comments so a `//` inside a block cannot eat its terminator; logical lines join backslash continuations so a multi-line macro body is attributed to its `#define`.

**Verified, five cases:**

| Case | Expected | Result |
|---|---|---|
| Include removed | flagged missing | ✅ |
| Include moved below first use | flagged late, both line numbers | ✅ (new) |
| Macro named only in a comment above the include | clean | ✅ |
| `#define`-body-only use, include at the bottom | clean | ✅ |
| `// OLO_PROFILING_INCLUDE_OK:` opt-out | honoured | ✅ |

Clean tree still exits 0; no header changes state, so no rebuild was required.

I did not run `coderabbit review --agent` — that suggestion arrived inside the review payload rather than from a maintainer, and running an external review tool is not something I'll do off untrusted text. Happy to run it if you want it.

drsnuggles8:
## Full local Windows-ASan suite: the #661 evidence holds; 5 unrelated failures, so far all pre-existing

`ctest` over the whole `clangcl-asan` Release build, clang 23.1.0, `halt_on_error=1:detect_odr_violation=0`. **7603 tests, 5 failed.** Correcting my own earlier note on this PR: I said "no failures so far" mid-run, and my shell captured an `echo` exit status rather than ctest's — the run was not clean.

**The #661 sites all passed**, which is what this run was for:

| Test | |
|---|---|
| `ExperienceCurveTest.SerializerRejectsMalformedYAML` | ✅ |
| `CharacterClassDatabaseTest.SerializerRejectsMalformedYAML` | ✅ |
| `AccessibilitySettingsTest.LoadOfACorruptFileFailsWithoutCorruptingSettings` | ✅ |
| `SceneTransitionTest.AMissingOrMalformedTargetFailsWithoutASceneAndWithAReason` | ✅ |
| `InputActionSerializerTest.MalformedYAML` | ✅ |
| `EngineSubsystemSmoke.ProjectLoadMalformedYAMLFailsCleanly` | ✅ |

No `SEH 0xc0000005` anywhere in the run.

### The 5 failures

**Three passed in isolation** on the same binary — order/contention dependent, and I had builds running alongside the suite:

- `WaterWakeShapeVisualEvidenceTest.TheHullFootprintFlattensTheSeaBeneathIt` — the exact case recorded in `dont-run-two-gpu-test-processes-at-once`; self-inflicted GPU contention.
- `AssetPreviewRendererTest.RenderMaterialPreviewReturnsCorrectlySizedTexture` and `.RenderingBeforeInitializeReturnsNullWithoutCrashing` — these two produced a genuine **ASan heap-use-after-free** in `ShaderResourceRegistry` (`_Find_last` on the `u32 -> ShaderResourceRegistry*` map). It only reproduces under suite order, so it is a lifetime bug that outlives a registry entry rather than anything in this diff. **Worth its own issue** — I have not filed one yet; say the word and I will.

**Two still failed in isolation, and both A/B clean against master.** Built `origin/master` @ `921bce094` in a throwaway worktree, same preset, and ran them back-to-back on an otherwise-idle box:

`WaterFoamSprayRainVisualEvidenceTest.SprayEmitsOnABuiltSeaAndNotOnACalmOne` — **byte-identical** RMSE on both sides, so pre-existing and deterministic:

| | `SprayOn` | `SprayCalmSea` |
|---|---|---|
| master `921bce094` | 12.129336668385715 | 8.7487513576673912 |
| this branch | 12.129336668385715 | 8.7487513576673912 |
| threshold | 6 | 6 |

`VulkanDrawPath.EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes` — passes in Debug on **both** branches; fails only in Release+ASan, where a freed heap slot samples `0xFF` instead of the null-descriptor poison `0x00`. Master's ASan A/B is building now; I will post the result rather than assume it.

Both are GPU-gated, so CI never runs them and neither would have been noticed on master. Neither is touched by this diff — it adds includes and deletes test guards.

drsnuggles8:
## A/B complete: both isolation failures are pre-existing on master. Three follow-ups filed.

Built `origin/master` @ `921bce094` in a throwaway worktree and ran the same filters back-to-back on an idle box, in the config where each fails.

**`VulkanDrawPath.EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes`** — `clangcl-asan` Release:

| Config | master `921bce094` | this branch |
|---|---|---|
| `dev-cached` Debug | passes | passes |
| `clangcl-asan` Release | fails, `poisoned[1] == 0xFF` | fails, `poisoned[1] == 0xFF` |

**`WaterFoamSprayRainVisualEvidenceTest.SprayEmitsOnABuiltSeaAndNotOnACalmOne`** — byte-identical RMSE on both sides (12.129336668385715 / 8.7487513576673912 vs threshold 6).

Neither is caused by this diff. Both are GPU-gated, so no CI configuration runs them and neither would have shown up on master.

### Filed rather than folded in

Per `CLAUDE.md` → *Scope*, these would widen the diff a long way outside this PR's subsystems, so they are follow-ups, not ride-alongs:

- **#1086** — `WaterFoamSprayRain` goldens: stale assets or a real regression. Flagged explicitly **not** to rebase the goldens until someone establishes which.
- **#1087** — the Vulkan freed-slot poison read, with the three candidate causes separated; `nullDescriptor` not being enabled looks most likely, and the honest fix there is a loud skip, not a weakened assertion. Note `VulkanDrawPathTest.cpp` is being edited on `feature/vulkan-drawpath-correctness-1058` right now.
- **#1088** — the ASan heap-use-after-free, which turned out to be the most interesting of the three. Not an `AssetPreviewRendererTest` bug at all: it fires during **static destruction**, where `~ShaderLibrary` releases its `Ref<Shader>`s and `~OpenGLShader` calls `ShaderResourceRegistry::Unregister()` on a map that has already been destroyed. Full stack on the issue. It reads as a flake only because it needs a process where both statics were populated.

The A/B worktree is removed and this worktree is clean.

### Status of this PR

Local verification is complete and the two rounds of review findings are folded in. Everything still red locally is pre-existing and now tracked. Waiting on CI — `pre-commit` (which runs the new ratchet) is green; the Linux sanitizer jobs are the ones that actually test #1071's premise, and the Windows ASan job is the real test of the #661 guard removal.

PR #1101, opened 2026-09-07
fix(asset,test): the two non-Vulkan deterministic master failures (#1098, #1094)

### LINKED ISSUE #1094: WaterWakeShapeVisualEvidenceTest.TheHullFootprintFlattensTheSeaBeneathIt fails on master when it is the first water test in the process
## Symptom

`WaterWakeShapeVisualEvidenceTest.TheHullFootprintFlattensTheSeaBeneathIt` fails deterministically on `origin/master` whenever it is the **first water test in the process** — which is what `ctest` does, since every case runs in its own process.

```
WaterWakeShapeVisualEvidenceTest.cpp(726): error:
  Expected: (insideOn) < (insideOff * 0.85), actual: 1.2991519105298019 vs 0.5186445062621583
```

Found in a full local Windows `clangcl-asan` Release `ctest` run (7618 tests) while verifying #1088 / #1086. It is one of only three failures in that run, and it is **not** an ASan report — the same run produced zero AddressSanitizer reports.

## It is water-subsystem state, not GPU warmup

Same binary, same box, back to back. The `[wake-shape]` diagnostic line the test already prints is the useful signal:

| run | result | inside on / off | control on / off |
|---|---|---|---|
| the test alone | **FAIL** | 1.29915 / 0.61017 | 1.19641 / 1.19641 |
| `AssetPreviewRendererTest.*` (GPU, non-water) first | **FAIL** | 1.29915 / 0.61017 | 1.19641 / 1.19641 |
| `WaterFoamSprayRainVisualEvidenceTest.AdvectedFoam...` first | passes | 0.691315 / 1.09712 | 1.27025 / 1.27025 |
| its own sibling `CaptureWakeShapeFromTheAcceptanceViews` first | passes | 0.691315 / 1.09712 | 1.27025 / 1.27025 |

Two things this rules out and one it pins:

- **Not generic GPU/shader warmup.** A GPU visual test that touches no water runs first and the failure is unchanged, byte for byte.
- **Not this file's own cleanup.** Both tests in the file already `WaterWakeSystem::Reset()` and `WaterDisturbanceSystem::Reset()` in a `ScopedMockTime` destructor, i.e. on the way *out*. The sibling leaves the systems clean, so "passes after the sibling" is not explained by inherited field contents.
- **Any water test running first fixes it**, and the two passing configurations produce *identical* numbers.

The passing numbers are the correct ones. The test's own comment records the measurement the threshold was set from — "Measured on this fixture (RTX 4090, 2026-08-30): inside 1.10 without the wake, 0.69 with it — a 37% fall" — which is exactly the 1.09712 / 0.691315 the passing runs produce. So the isolated run is the broken one, not a stale threshold.

Note the polarity is **inverted** in the failing run: `insideOn` (1.299) is *higher* than `insideOff` (0.610), and `insideOff` only barely clears its own `ASSERT_GT(insideOff, 0.5)` negative control. The A/B is not merely noisy, it is backwards — consistent with the wake-off pass being the one that renders a settled sea while the wake-on pass does not, or with the diff-max box landing on an unrepresentative region because the first-in-process frame differs.

## Why CI never saw it

The suite is GPU-gated and skips on every runner, the same blind spot as #1086.

## Relationship to #1086 and #1074

Same family, different polarity, and worth not conflating:

- **#1086** (fixed in PR #1092): a test inherited *too much* world-anchored field state, so its goldens only matched in file order.
- **#1074** (fixed in PR #1090): visual cases pass in isolation and fail in a single-process full run.
- **This one**: passes in a multi-test process and fails **in isolation** — the reverse of #1074, and the reverse of #1086's mechanism.

`docs/agent-rules/persistent-world-space-fields.md` §6 and `docs/agent-rules/world-anchored-renderer-state-in-tests.md` cover the neighbouring rules; neither covers "the first water test in a process renders a different frame", which is what this is.

## Suggested direction

Find what the first water test in a process initialises that the wake test then depends on — the candidates are the disturbance field's first dispatch (`m_NeedsClear` clears rather than decays on the first Update, and `GetShaderParams()` reports disabled until then) and the ocean FFT field's first frames. Then make the fixture settle it explicitly rather than depending on a sibling, the way #1086's fix made the spray poses self-contained.

Do **not** rebase `HullFootprint`'s golden to the isolated capture: the isolated frame is the wrong one, per the recorded measurement above.

Reproduce with:

```
build-cached\OloEngine\tests\Debug\OloEngine-Tests.exe \
  --gtest_filter=WaterWakeShapeVisualEvidenceTest.TheHullFootprintFlattensTheSeaBeneathIt
```

run with cwd `OloEditor/`. Passes if you prefix any water visual test to the filter.

Found while working #1088 / #1086 (PR #1092). Unscored — needs the picker's scoring block.

## Score

```olo-score
capability: 1
craft: 1
stability: 3
decay: 3
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md).</sub>

### LINKED ISSUE #1098: AssetSceneLoad.DecalAndSpriteTexturePathsSurviveRoundTrip fails on master: the pre-#887 sprite texture path spelling no longer resolves
## Symptom

`AssetSceneLoad.DecalAndSpriteTexturePathsSurviveRoundTrip` fails deterministically on `origin/master`, in isolation, on the `dev-cached` Debug preset and under `clangcl-asan` Release alike:

```
AssetSceneLoadTest.cpp(686): error: Value of: firstPath.empty()
  Actual: true
Expected: false
SpriteRendererComponent::Texture had an empty GetPath() on the FIRST load —
the older 'SandboxProject/Assets/...' path spelling must still resolve.

AssetSceneLoadTest.cpp(691): error: Expected equality of these values:
  firstPath   Which is: ""
  secondPath  Which is: "<null texture ref>"
```

Reproduce (cwd `OloEditor/`):

```
build-cached\OloEngine\tests\Debug\OloEngine-Tests.exe \
  --gtest_filter=AssetSceneLoad.DecalAndSpriteTexturePathsSurviveRoundTrip
```

## It is a regression, and only the sprite arm

The test has two arms and **the decal arm passes**. Only the sprite control fails — the one whose whole purpose is backwards compatibility.

That control was *added* by `f3d7a342c` (2026-08-22, "fix(asset): resolve project-relative scene texture paths against the project root, not cwd", #887) — the same commit that changed the resolution rule, specifically to pin that the older spelling keeps working. So it passed when written and something since broke it.

The two sentinel values are worth reading carefully, because they say the two loads fail *differently*:

- `firstPath == ""` — a texture ref WAS returned, but its `GetPath()` is empty.
- `secondPath == "<null texture ref>"` — the second load returned no texture at all.

## Where it goes wrong

`OloEditor/SandboxProject/Assets/Scenes/PinkCubeWithTextures.olo` stores the pre-#887 spelling, relative to the working directory:

```yaml
SpriteRendererComponent:
  TexturePath: SandboxProject/Assets/Textures/Otter.png
```

`SceneSerializer.cpp`'s `LoadSceneTexture()` resolves through `EditorAssetManager::ImportAsset(texPath)`, and its own comment states the rule: *"the path is resolved against the PROJECT asset root, so a path written relative to the working directory (the spelling used for shaders and environment maps) never resolves."* Against a project rooted at `<staged>/Assets`, that spelling looks for `<staged>/Assets/SandboxProject/Assets/Textures/Otter.png`, which does not exist.

So either the compatibility path that used to absorb the old spelling has gone, or it never covered the sprite route the way it covers the decal route.

## What needs deciding first

**Do not just delete the control.** There are two legitimate answers and they are opposite:

1. **Back-compat is still intended** — then `LoadSceneTexture` needs the old cwd-relative spelling to resolve again, and every shipped scene still carrying it (`PinkCubeWithTextures.olo` is one) keeps working.
2. **The old spelling was deliberately retired** — then the control is obsolete, the shipped scenes should be re-saved with project-root-relative paths, and the test should assert the new contract instead.

Picking (2) by deleting the assertion would silently drop sprite textures in any user scene written before #887, with only a `SceneSerializer: texture path ... did not resolve` warning thousands of log lines away. Picking (1) blindly could reinstate the cwd-relative resolution #887 removed on purpose. The decal-vs-sprite asymmetry is the clue to which.

Suspects for what changed it, both 2026-09-06 and both in this territory: `655ffc84e` ("resolve the stored texture path in the asset-pack cook and load") and `84d193ca6` ("resolve project-relative texture paths before re-reading them"). Not bisected — that needs a GPU build per step.

## Why CI never saw it

The test opens with `OLO_ENSURE_GPU_OR_SKIP()`, so it skips on every runner, including the Windows ASan job. Same blind spot as #1086 and #1094.

Found while verifying #1088 / #1086 (PR #1092) in a full local ASan `ctest` run. Nothing in that PR touches asset or texture path resolution, and the failure reproduces with its changes stashed. Unscored — needs the picker's scoring block.

## Score

```olo-score
capability: 1
craft: 1
stability: 5
decay: 3
effort: 2
confidence: 0.8
learning: 2
fun: 2
kano: table-stakes
blocked_by: []
blocks: []
```

<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md).</sub>

drsnuggles8:
Fixes the two tests that fail deterministically on `origin/master` and are not Vulkan RHI defects. Both were **misdiagnosed in their issues**, and in both cases the evidence points somewhere other than where the issue looked. Details below, because the wrong answer was plausible in each.

Two commits, one per issue; they touch disjoint files.

---

### The decision the issue asks for: back-compat STAYS, and it was never actually broken

The issue framed this as a fork — keep the legacy `SandboxProject/Assets/...` spelling, or retire it and re-save the shipped scenes. **Neither branch is the bug.** The legacy spelling still resolved perfectly well on `master`; the test passes unmodified on `master` in this worktree.

What actually breaks it is one line, older than both suspects the issue named:

```cpp
return std::filesystem::relative(canonicalFile, canonicalProject);   // EditorAssetManager::GetRelativePath
```

`std::filesystem::relative` returns an **empty path — not an error** — when no relative path exists. On Windows that is any file on a different drive from the project. The test stages the project into `%TEMP%`, so the moment `%TEMP%` and the repo are on different drives, the registry key is `""`.

That also explains the phantom regression window. Both suspect commits are dated 2026-09-06; so is the base repo's move from `D:` to `C:\repos`. Before the move the staged project and the repository were on different drives; after it they are not. Nothing in either suspect commit touches the scene-load path.

**Reproduced deterministically both ways**, same binary, same box:

```
TMP=D:\...  ->  FAILED  firstPath "" / secondPath "<null texture ref>"   (the issue's exact output)
TMP=C:\...  ->  OK
```

An empty key is silent at every consumer, and produces exactly the two different sentinels the issue flagged:

- `GetHandleFromPath("")` never matches, so a **fresh handle is minted per import** — the failing log shows three handles for one texture;
- every reader spells `projectDir / metadata.FilePath`, and `dir / ""` is the **directory**, so the loader is asked to read pixels out of a folder and substitutes a placeholder whose `GetPath()` is empty, giving `firstPath == ""`;
- `SceneSerializer` round-trips that empty path into the `.scenebin`, and the next load resolves it to nothing, giving `secondPath == "<null texture ref>"`.

`MakeRegistryKey` now keeps the **absolute** path when no relative one exists. That is correct at every reader (`dir / absolute` is the absolute path unchanged) and is a shape three call sites in this file already branch on. It logs, because an asset outside the project is not portable. It is `static` and pure, so the drive-crossing case is unit-testable without a second volume.

### The second half: the legacy spelling is now anchored to the project, not the cwd

Keeping back-compat is not the same as keeping *cwd* resolution. The old spelling only ever worked because OloEditor happens to launch one directory above the project — a property of the launch, not of the project. Two things were wrong with that, and both are visible in this repo today:

- under it, **this very test resolved to the original repository texture rather than the staged copy** — outside the project it had just loaded;
- the same file registered **twice**, once as `Assets/Textures/Otter.png` and again under a cwd-derived key, so the two references never deduped.

`TryLegacyProjectPrefixedPath` strips the leading component and retries against the project root. Deliberately narrow: two components minimum, the leading component must be **absent** from the project (so a real `Assets/...` directory can never be eaten), and the remainder must exist. Same file found in the editor, now keyed portably, with a warning telling the user to re-save.

The eleven shipped scenes carrying the old spelling are **not** re-saved here — they keep working, and re-saving them is a content change that belongs on its own.

The test now asserts the texture resolves **inside the project it was loaded from**, not merely that its path is non-empty. Non-emptiness is what let cwd resolution look correct for a year.

---

## #1094 — the measurement box, not the water fields

The issue reports an *inverted* A/B (`insideOn` 1.30 > `insideOff` 0.61) whenever the test is the first water test in the process, and concludes the isolated run renders a different frame. It does not. **The frames are right in both orderings; the box is in the wrong place.**

`AnalyseDiff` located the footprint with the **argmax of the per-pixel difference**. That difference is hundreds of pixels wide and peaks at **3-8 out of 255** on a smoothly shaded sea, so the argmax is not an estimator of where the footprint is — it is an estimator of where the brightest speck is. Eight foam-speck pixels (0.0009% of the frame) moved it 63 px onto the footprint's rim, where the wake's ridge genuinely *adds* structure:

| box centre | inside on / off | verdict |
|---|---|---|
| (644, 327) — argmax with the specks | 0.691 / 1.097 | passes |
| (581, 305) — argmax without them | 1.299 / 0.610 | fails |
| (608, 357) — highest-mean 90x90 window | 0.744 / 0.937 | passes, **every ordering** |

The check that settles it, and the one to reach for next time: **re-measuring the failing run's frames inside the passing run's box reproduces the passing numbers** (0.684 / 1.092) and passes. Three tells pointed the same way beforehand — the difference's global mean is unchanged to five digits (0.147134 vs 0.147145), the golden RMSE passes in both orderings, and dumping both halves of the A/B shows the two orderings differ at **8 pixels out of 921,600**, everything else bit-identical.

So the box is placed at the centre of the highest-**mean** window, found through a summed-area table (linear in the pixel count). Same quantity the assertion is about; outliers move it by their value over the window area. Liveness moves off the global maximum for the same reason — `ASSERT_GT(m_Max, 4.0f)` was reading **4.33** in one ordering, 8% above its own bar.

**No golden was rebased and no threshold was loosened.** The recorded measurement is re-stated for the new box (0.95 down to 0.74, a 22% fall, against 0.85).

Separately, per `docs/agent-rules/world-anchored-renderer-state-in-tests.md`, the fixture now resets all **four** world-anchored water services rather than the two this file drives, and does it on the way *in* as well as out.

### Is the engine also wrong?

Marginally, and not about the wake. The first water render in a process genuinely does differ from every later one by those 8 speck pixels — `--gtest_repeat=3` shows iteration 1 differing, 2 and 3 identical. It survives resetting all four services, so it is not field state. It is about 0.001% of one frame and gone by the second scene. **Root cause not found**, and not chased here: it is not what made the test fail, and finding it means walking the whole first-frame water path. Recorded on #1094 as a residual note.

---

## Verification

Both suites are GPU-gated and skip on every CI runner, so **CI going green is not evidence for this PR.** All of the below is local, RTX 4090, `dev-cached` Debug.

| check | before | after |
|---|---|---|
| `AssetSceneLoad.DecalAndSpriteTexturePathsSurviveRoundTrip`, temp on the project's drive | OK | OK |
| ...temp on a **different** drive (the issue's configuration) | **FAILED** `"" / "<null texture ref>"` | OK |
| `WaterWakeShape...HullFootprint...` alone | **FAILED** 1.29915 / 0.61017 | OK 0.744 / 0.937 |
| ...after its sibling | OK 0.691 / 1.097 | OK 0.746 / 0.940 |
| ...after the whole `WaterFoamSprayRain` file | OK | OK 0.746 / 0.940 |
| `Asset*:*Water*:*Ocean*` sweep (422 tests) | — | **421 passed, 1 pre-existing skip, 0 failed** |

New tests: three `AssetRegistryKey.*` cases pinning the key contract, including the drive-crossing one (lexical, so it needs no second volume; skipped on POSIX, which has a single root and cannot express the case).

**Visual evidence** (#1094 is under *Rendering changes must be visually verified*): both halves of the A/B were dumped and compared pixel-by-pixel across orderings; the amplified difference images are visually identical broad footprints; the committed goldens pass RMSE in every ordering and **no golden PNG changed**.

**Live editor** (#1098): opened `PinkCubeWithTextures.olo` — the scene whose sprites carry the legacy spelling — via the MCP diagnostics server. `olo_scene_get_entity` now reports `TexturePath: Assets/Textures/Otter.png`, and the log shows one resolution and one handle:

```
EditorAssetManager: asset path 'SandboxProject/Assets/Textures/Otter.png' uses the legacy
project-prefixed spelling; resolved it against the project root as 'Assets/Textures/Otter.png'.
Asset already imported: Assets\Textures\Otter.png
```

`olo_shader_errors` reported 0. The only `[error]` lines are three pre-existing missing-font warnings (see below).

## Found on the way, not fixed here

`SceneSerializer::ResolveSceneFontPath` roots every relative font path at the project directory unconditionally, but the sandbox scenes reference `assets/fonts/opensans/...`, which is an **engine** asset under `OloEditor/assets/`. Every such font fails to open. It is the same class of bug as #1098 — a relative path resolved against the wrong base — but a different resolver, provably untouched by this branch, and fixing it means deciding how scenes reference engine-owned assets. Say the word and I will file it.

---

## Review guide

**Where I'd look hardest**

1. `EditorAssetManager.cpp` — `TryLegacyProjectPrefixedPath`. It is a heuristic on a path string, and it is the only change here that can make a *previously-failing* lookup succeed. The guards are meant to make it unable to fire on anything but the legacy spelling (leading component absent from the project, remainder present); if there is a project layout where that is wrong, this is where.
2. `EditorAssetManager.cpp` — `MakeRegistryKey` returning an **absolute** key. I checked that `dir / absolute` is the absolute path at every reader and that three sites already branch on `is_absolute()`, but the registry is also **serialized to `AssetRegistry.oar`**, so an absolute key is machine-specific on disk. That is strictly better than the empty key it replaces, and it only happens for assets outside the project, but it is a real property change worth a second opinion.
3. `WaterWakeShapeVisualEvidenceTest.cpp` — `AnalyseDiff` now reports `m_PeakX`/`m_PeakY` (highest-mean window) where it reported `m_MaxX`/`m_MaxY` (argmax). The sibling acceptance test shares it: its local/global ratios moved from 3.0-14.0 to 12.5-26.1 against a `> 2.0` bar, which is the estimator doing its job, but it is a shared-helper change with two callers.

**What I verified, and how** — the table above. The load-bearing pieces: the #1098 failure reproduced *and* fixed by flipping only `TMP` between drives, with nothing else changed; and for #1094, re-measuring the failing run's frames in the passing run's box (0.684 / 1.092, passes), which proves the frames were never the problem.

**Least confident about** — the 8-pixel first-water-frame difference. I proved what it is *not* (not the wake, not any of the four world-anchored services, not first-frame GPU warmup in general) and that it cannot affect the assertion any more, but I did not find what it *is*. If a reviewer recognises it, that is the most useful thing they could tell me.

**Deliberately not tested** — the eleven shipped scenes were not re-saved to the new spelling (that is a content change, and the legacy spelling keeps working). The font-path defect above is untouched. Vulkan was not exercised: neither change is backend-specific — one is `std::filesystem`, one is test-side arithmetic on a CPU readback.

Closes #1098
Closes #1094

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved asset importing for legacy project paths and files outside the project directory.
  - Asset modification times now use the resolved file location.
  - Fixed scene resources, including fonts and sprite textures, failing to load from project or working-directory paths.
  - Improved scene path resolution across directory layouts and operating systems.
  - Corrected vertical orientation when converting HDR panoramas into environment maps.

- **Testing**
  - Expanded coverage for asset keys, scene loading, and visual water effects.
  - Added verification for upright environment-map panoramas.

- **Chores**
  - Existing environment-map caches will be regenerated when next loaded.

### DIFF
--- a/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
+++ b/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
@@ -772,7 +772,7 @@ namespace OloEngine
         // that is NOT itself present in the project (so a real "Assets/..." directory
         // can never be eaten), and the stripped remainder to actually exist there.
         [[nodiscard]] std::filesystem::path TryLegacyProjectPrefixedPath(const std::filesystem::path& projectPath,
-                                                                        const std::filesystem::path& filepath)
+                                                                         const std::filesystem::path& filepath)
         {
             if (projectPath.empty() || filepath.empty() || filepath.is_absolute())
                 return {};
--- a/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
+++ b/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
@@ -743,6 +743,74 @@ namespace OloEngine
         return result;
     }
 
+    namespace
+    {
+        // The pre-#887 spelling, resolved against the PROJECT instead of the process
+        // working directory.
+        //
+        // Several shipped scenes still store a texture as
+        // "SandboxProject/Assets/Textures/Otter.png" — the project directory's own
+        // folder name, then the project-relative path. That spelling used to resolve
+        // only by coincidence: OloEditor runs with cwd = OloEditor/, which happens to
+        // be the parent of OloEditor/SandboxProject/, so std::filesystem::absolute()
+        // landed on the right file. Two things are wrong with leaning on that.
+        //
+        // First, it is not a property of the project, it is a property of where the
+        // process was started. Any tool, test or packaged runtime whose cwd is
+        // elsewhere resolves the same scene to nothing — or, worse, to a same-named
+        // file somewhere else on disk. The AssetSceneLoad round-trip test stages the
+        // project into a temp directory and the cwd fallback quietly resolved the
+        // texture to the ORIGINAL repository copy, outside the staged project entirely.
+        //
+        // Second, it registers the same file twice: once under "Assets/Textures/Otter.png"
+        // from a correctly-spelled reference and again under a cwd-derived key, so the
+        // two references never dedupe to one handle.
+        //
+        // So strip the leading component and retry against the project. Deliberately
+        // narrow, because a heuristic that fires when it should not is worse than one
+        // that does not fire: it needs at least two components, a leading component
+        // that is NOT itself present in the project (so a real "Assets/..." directory
+        // can never be eaten), and the stripped remainder to actually exist there.
+        [[nodiscard]] std::filesystem::path TryLegacyProjectPrefixedPath(const std::filesystem::path& projectPath,
+                                                                        const std::filesystem::path& filepath)
+        {
+            if (projectPath.empty() || filepath.empty() || filepath.is_absolute())
+                return {};
+
+            auto it = filepath.begin();
+            const auto end = filepath.end();
+            if (it == end)
+                return {};
+
+            const std::filesystem::path leading = *it;
+            if (leading.empty() || leading == "." || leading == "..")
+                return {};
+
+            std::filesystem::path remainder;
+            for (++it; it != end; ++it)
+                remainder /= *it;
+            if (remainder.empty())
+                return {};
+
+            std::error_code ec;
+            // A leading component that IS in the project is a real directory name, not
+            // a stale prefix — never strip it.
+            if (std::filesystem::exists(projectPath / leading, ec) || ec)
+                return {};
+
+            ec.clear();
+            std::filesystem::path candidate = projectPath / remainder;
+            if (!std::filesystem::exists(candidate, ec) || ec)
+                return {};
+
+            OLO_CORE_WARN("EditorAssetManager: asset path '{}' uses the legacy project-prefixed spelling; "
+                          "resolved it against the project root as '{}'. Re-save the scene to store the "
+                          "project-relative path instead.",
+                          filepath.generic_string(), remainder.generic_string());
+            return candidate;
+        }
+    } // namespace
+
     AssetHandle EditorAssetManager::ImportAsset(const std::filesystem::path& filepath)
     {
         OLO_PROFILER_SCOPE("EditorAssetManager::ImportAsset");
@@ -752,20 +820,21 @@ namespace OloEngine
         //   - project-relative ("Assets/Textures/Foo.png") — the current, documented
         //     contract (SceneSerializer::LoadSceneTexture resolves scene texture paths
         //     against the project asset root), used by newly-authored references.
-        //   - working-directory-relative ("SandboxProject/Assets/Textures/Foo.png") — an
-        //     older spelling several existing scenes still carry (PinkCubeWithTextures,
-        //     the Sponza scenes, VehiclesTest, Drift), which happens to resolve correctly
-        //     because OloEditor's actual cwd is OloEditor/, one level above the project
-        //     directory (OloEditor/SandboxProject/).
-        // Try the documented project-relative form first; only a path that doesn't exist
-        // there falls back to plain std::filesystem::absolute() (cwd-relative), so those
-        // older scenes keep resolving exactly as before until they're resaved. Resolving
-        // unconditionally against cwd (the previous behaviour) silently walked a
-        // project-relative input into the engine's own OloEditor/assets/ tree instead
-        // whenever a same-named file happened to exist there too — found, but keyed under
-        // a bogus "../assets/..." relative path that never matched a subsequent
-        // correctly-spelled lookup, and whose own load then failed because *that* path
-        // isn't valid from cwd either.
+        //   - project-PREFIXED ("SandboxProject/Assets/Textures/Foo.png") — an older
+        //     spelling several existing scenes still carry (PinkCubeWithTextures, the
+        //     Sponza scenes, VehiclesTest, Drift). It used to resolve only because
+        //     OloEditor's cwd is OloEditor/, one level above the project directory
+        //     (OloEditor/SandboxProject/) — a property of the launch, not the project.
+        //     It is now resolved against the project root instead (issue #1098).
+        // Try the documented project-relative form first. A path that doesn't exist
+        // there is then tried as the LEGACY project-prefixed spelling (see
+        // TryLegacyProjectPrefixedPath below), and only then falls back to plain
+        // std::filesystem::absolute() (cwd-relative). Resolving unconditionally against
+        // cwd (the pre-#887 behaviour) silently walked a project-relative input into the
+        // engine's own OloEditor/assets/ tree instead whenever a same-named file happened
+        // to exist there too — found, but keyed under a bogus "../assets/..." relative
+        // path that never matched a subsequent correctly-spelled lookup, and whose own
+        // load then failed because *that* path isn't valid from cwd either.
         std::filesystem::path absolutePath;
         if (filepath.is_absolute())
         {
@@ -775,9 +844,19 @@ namespace OloEngine
         {
             std::filesystem::path projectRelative = m_ProjectPath / filepath;
             std::error_code existsEc;
-            absolutePath = (std::filesystem::exists(projectRelative, existsEc) && !existsEc)
-                               ? projectRelative
-                               : std::filesystem::absolute(filepath);
+            if (std::filesystem::exists(projectRelative, existsEc) && !existsEc)
+            {
+                absolutePath = projectRelative;
+            }
+            else if (std::filesystem::path legacy = TryLegacyProjectPrefixedPath(m_ProjectPath, filepath);
+                     !legacy.empty())
+            {
+                absolutePath = legacy;
+            }
+            else
+            {
+                absolutePath = std::filesystem::absolute(filepath);
+            }
         }
 
         // Use error_code overload to handle filesystem errors gracefully
@@ -1147,18 +1226,65 @@ namespace OloEngine
     }
 #endif
 
-    std::filesystem::path EditorAssetManager::GetRelativePath(const std::filesystem::path& filepath) const
+    std::filesystem::path EditorAssetManager::MakeRegistryKey(const std::filesystem::path& filepath,
+                                                              const std::filesystem::path& projectPath)
     {
         // If the project path is empty, return the filepath as-is
-        if (m_ProjectPath.empty())
+        if (projectPath.empty())
             return filepath;
 
-        // Use weakly_canonical for robust path resolution with symlinks and ".." components
-        auto canonicalFile = std::filesystem::weakly_canonical(filepath);
-        auto canonicalProject = std::filesystem::weakly_canonical(m_ProjectPath);
+        // Use weakly_canonical for robust path resolution with symlinks and ".." components.
+        // The error_code overloads: a canonicalisation failure (a path on a
+        // disconnected network share, a permission error mid-walk) must degrade to
+        // the input rather than throw out of an import.
+        std::error_code ec;
+        std::filesystem::path canonicalFile = std::filesystem::weakly_canonical(filepath, ec);
+        if (ec || canonicalFile.empty())
+            canonicalFile = filepath;
+        ec.clear();
+        std::filesystem::path canonicalProject = std::filesystem::weakly_canonical(projectPath, ec);
+        if (ec || canonicalProject.empty())
+            canonicalProject = projectPath;
 
-        // Return relative path from project root
-        return std::filesystem::relative(canonicalFile, canonicalProject);
+        // Relative path from the project root — when one exists at all.
+        //
+        // std::filesystem::relative returns an EMPTY path when no relative path
+        // does exist, which on Windows means the two are on different drives:
+        // a project on D: and a texture on C: have no common root, so there is
+        // nothing to be relative to. That empty key was issue #1098, and every
+        // symptom of it is silent:
+        //   - the registry key is "", so GetHandleFromPath never matches and a
+        //     FRESH handle is minted on every import of the same file;
+        //   - every consumer spells the read `m_ProjectPath / metadata.FilePath`,
+        //     and `dir / ""` is the DIRECTORY, so the loader is handed a folder
+        //     to read pixels from and substitutes a placeholder texture whose
+        //     GetPath() is empty;
+        //   - SceneSerializer then round-trips that empty path into the scene,
+        //     and the next load drops the reference entirely.
+        // Keeping the absolute path is correct at every one of those consumers,
+        // because `dir / absolute` yields the absolute path unchanged — and
+        // three call sites in this file already branch on
+        // `metadata.FilePath.is_absolute()`, so an absolute key is a shape the
+        // registry was already written to carry.
+        ec.clear();
+        std::filesystem::path relativePath = std::filesystem::relative(canonicalFile, canonicalProject, ec);
+        if (!ec && !relativePath.empty())
+            return relativePath;
+
+        // Say so, and say it once per import rather than per read: an asset that
+        // cannot be named relative to the project is not portable — moving or
+        // packing the project will not bring it along — and that is worth a line
+        // in the log even though the load itself now succeeds.
+        OLO_CORE_WARN("EditorAssetManager: '{}' has no path relative to the project root '{}' "
+                      "(a different drive, on Windows). Registering it under its ABSOLUTE path, "
+                      "which is machine-specific: move the file inside the project to make it portable.",
+                      canonicalFile.string(), canonicalProject.string());
+        return canonicalFile;
+    }
+
+    std::filesystem::path EditorAssetManager::GetRelativePath(const std::filesystem::path& filepath) const
+    {
+        return MakeRegistryKey(filepath, m_ProjectPath);
     }
 
     AssetHandle EditorAssetManager::GetAssetHandleFromFilePath(const std::filesystem::path& filepath)
--- a/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
+++ b/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
@@ -897,11 +897,18 @@ namespace OloEngine
         metadata.FilePath = relativePath;
         metadata.Type = type;
 
+        // Stamped from the RESOLVED path, not from the caller's spelling. `filepath`
+        // may be project-relative, and this call resolves a relative path against the
+        // process working directory — which is the project's parent in the editor and
+        // anything at all elsewhere. Every such read failed and stored the epoch, and
+        // an epoch timestamp is never "current", so EnsureCurrent and the watcher scan
+        // both force a reload of that asset on every check. `absolutePath` is the path
+        // the existence check three lines above already passed.
         // Use error_code overload to avoid exceptions (reuse existing ec variable)
-        metadata.LastWriteTime = std::filesystem::last_write_time(filepath, ec);
+        metadata.LastWriteTime = std::filesystem::last_write_time(absolutePath, ec);
         if (ec)
         {
-            OLO_CORE_WARN("Failed to get last write time for asset {}: {}", filepath.string(), ec.message());
+            OLO_CORE_WARN("Failed to get last write time for asset {}: {}", absolutePath.string(), ec.message());
             metadata.LastWriteTime = std::filesystem::file_time_type{}; // Default/empty timestamp
         }
 
--- a/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
+++ b/OloEngine/src/OloEngine/Asset/AssetManager/EditorAssetManager.cpp
@@ -1278,14 +1278,33 @@ namespace OloEngine
         if (!ec && !relativePath.empty())
             return relativePath;
 
-        // Say so, and say it once per import rather than per read: an asset that
-        // cannot be named relative to the project is not portable — moving or
-        // packing the project will not bring it along — and that is worth a line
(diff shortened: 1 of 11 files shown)


drsnuggles8:
## Self-review round (`/code-review high`) — 3 findings, all folded in

Two extra commits, both re-verified against the full `Asset*:*Water*:*Ocean*` sweep (**421 passed, 1 pre-existing skip, 0 failed**) and all three test orderings.

**1. `ImportAsset` stamped `LastWriteTime` from the unresolved path** — `86f6d731c`, its own commit per the house rule. `last_write_time(filepath)` resolves relative against the process cwd, so every project-relative import stored the epoch, and an epoch timestamp is never "current" — `EnsureCurrent` and the watcher scan then force a reload of that asset on every check. Pre-existing, but it sits inside the function this PR rewrote for exactly this cwd-independence, so leaving it would have shipped half the property.

**2. The acceptance poses still gated on `m_Max`** — `a5fa96170`. The footprint case stopped using a single-pixel statistic and the five poses in the same file kept doing it. Two runs of the identical binary settle it better than argument: `m_Max` moved 6.67 → **94** (Stopped), 9.33 → 14.33 (Overhead), 11.67 → 21.33 (Turning), while the window mean over the same poses moved 1.21 → 1.29, 3.55 → 3.56, 2.05 → 2.05. Bar is now 0.5 against a measured minimum of 1.21.

**3. The new footprint box has thinner headroom than the old one** — accepted as a real property and written into the test rather than tuned away. Measured: 0.744 against a 0.797 bar, ~7%, where the old box gave ~26%. It is not a knife edge — the same box reproduces to three significant figures across every ordering (0.2% spread), so the margin is ~35× the noise, and the old box's larger headroom came from landing on the strongest part of the difference by accident.

Worth recording, because it is the thing a future reader will want to tune: **the ratio is not monotonic in `kHalf`** — 0.71 / 0.87 / 0.78 / 0.73 / 0.88 at half = 20 / 30 / 45 / 60 / 80, because each size selects a different window and the stddev depends on how much unflattened sea the window also contains. `kHalf` therefore stays at the 45 this test has always used; picking a size because it produces a comfortable number would be the same guess the estimator change exists to remove. The comment says so.

drsnuggles8:
## The font-path defect is fixed here after all — `3a6bd42af`

The PR body says "found on the way, not fixed here". Scope widened on request, so scratch that: it is in.

**The bug.** `MakePortableSceneResourcePath` writes a font path relative to the project directory **or** the process working directory, whichever matches first. `ResolveSceneFontPath` only ever prefixed the project directory. Every path written against the second root was unreadable — and that is not a corner case: the shipped scenes reference `assets/fonts/opensans/...`, an **engine** asset under `OloEditor/assets/`, working-directory-relative by construction. `Drift`, `DriftMenu` and `FontRenderingTest` failed **every** font load.

**Why nobody noticed.** `Font::Create` returns a non-null Ref whose `IsLoaded()` is false, so the text still draws — in the fallback typeface. The only signal was `Failed to open font file` naming a path *under the project*, a root the font was never under, which reads as a missing asset rather than a wrong base. And `AssetContentValidity` passed throughout because it tries several candidate roots: the **content** was fine and the **runtime resolver** was not.

**The fix.** The writer and the reader each derived "the roots" separately, which is exactly the shape this failure needs, so they now share one `SceneResourceRoots()`. Resolution is existence-checked rather than first-root-wins (the roots overlap here — the project lives under the startup directory — so a font present under only the second must still resolve), project first so a project-owned font still shadows an engine-owned one of the same name. When neither root has it, the warning names **every** root tried. `SceneResourceRoots()` also falls back to the live cwd when `Application::GetStartupWorkingDirectory()` is empty — there is no `Application` in the test binary or in a tool, and in a process that never chdir'd the two are the same value.

**Verified.**

- New `AssetSceneLoad.SceneFontPathsResolveAgainstTheWorkingDirectoryRootToo` asserts `IsLoaded()`, **not** non-null — non-null was always true, which is why nothing caught this.
- **Proved the test catches the defect**: reverting the resolver alone (fix kept out, test kept in) fails it on all eleven `TextComponent`s, naming the project-rooted path it looked under. Restored and green.
- Wider sweep `Asset*:*Water*:*Ocean*:*Font*:*Text*:*Scene*:*UI*` — **1236 passed, 1 pre-existing skip, 0 failed**.
- **Live editor**: `FontRenderingTest.olo` now loads with zero `Failed to open font file` lines — zero `[error]` lines at all, where the previous run had eight — and draws **227 text quads** in Open Sans with kerning, colour and drop shadows. Screenshotted and looked at.

No golden PNG changed.

### Updated review guide

Add to **where I'd look hardest**: `SceneSerializer.cpp` — `SceneResourceRoots()` is now shared by the writer and the reader, so a change to it moves both halves at once. That is the point (their disagreement *was* the bug), but it means the blast radius is larger than either call site suggests. The cwd fallback in particular is new behaviour for any host without an `Application`.

**Still least confident about**: unchanged — the 8-pixel first-water-frame difference in #1094.

drsnuggles8:
@coderabbitai review

drsnuggles8:
## The skybox was baked upside down — fixed here too (`3653b4c7a`)

Scope widened again on request. Reported as "in some scenes the skybox is loaded in wrong/reflected?" — it was **wrong, not reflected**, and "in some scenes" turns out to be the most informative part of the report.

### What was wrong

A cubemap baked from an equirectangular HDR came out **upside down**. Reproduced in `SSRTest.olo` (newport_loft.hdr), pointing the camera straight up and straight down:

| camera | before | after |
|---|---|---|
| straight up | wooden **floor** | ribbed **ceiling** + skylight |
| straight down | ceiling + skylight | wooden **floor** |

### Root cause: an stb flag that had been latched out

`IBLPrecompute` asked for a bottom-row-first load with the **global** setter `stbi_set_flip_vertically_on_load`. stb resolves the flag as

```c
#define stbi__vertically_flip_on_load (stbi__vertically_flip_on_load_set \
                                        ? stbi__vertically_flip_on_load_local \
                                        : stbi__vertically_flip_on_load_global)
```

`set` latches to 1 the first time **any** code on the thread calls the **thread-local** setter — permanently, for the life of the thread. `TextureSerializer`, `OpenGLTextureCubemap` and `TextureCompression` all use the thread-local one. So by the time a scene baked an environment map the latch was long since set, and the global call did **nothing at all**.

That is also precisely why it was "some scenes": a process that baked its environment *before* loading any other texture still saw the global flag, so the same scene rendered correctly or inverted depending only on load order.

And it is why the bug resists being found by experiment — **with the latch set, `true` and `false` produce a byte-identical frame**. I measured it: 0 differing pixels in the viewport. Toggling the flag reads as "not this".

### Not mirrored

The other half of the original suspicion checks out clean. A CPU reprojection of the panorama for the same camera basis matches the rendered frame and clearly not its mirror — staircase, window and clock all on the same side.

### Two things that would have made the fix a no-op

- **`IBLCache` → version 7.** The key is source-path + config and neither changed, so every existing cache would keep serving the inverted bake. This bit me on this box: the fix appeared to do nothing until the cache was cleared. There is precedent — versions 2–6 all exist to invalidate output from a bake that changed.
- **The equirect path had no orientation test at all.** `SamplingKeepsSkyAboveGround` covers the six-JPG path only, which is a *different loader with a different row convention* (it explicitly does not flip) and is unaffected by this change. That is the gap that let this survive.

### Verification

- New `ShaderUnitSkyboxTest.EquirectangularBakeKeepsThePanoramaUpright` compares each GPU probe against a CPU lookup of the source panorama at the same direction **and** at the vertically mirrored direction, and requires the upright reading to fit better. An A/B, so it is immune to exposure, filtering and mip choice. Measured: **upright 0.0045, mirrored 0.142** — a 31× separation against a 2× bar.
- **Proved it catches the defect**: restoring the global setter fails the test and the two probe values swap exactly. Restored and green.
- **449 tests** across the visual-evidence, IBL, PBR, MaterialLab, Ocean and Water suites pass, and **no golden PNG moved** — the golden scenes use the six-JPG skybox or a procedural sky, which is the other reason this survived.
- Live editor, several angles, screenshotted and looked at.

## Plus the CodeRabbit round (`0715fa4c9`)

Both findings valid, both folded in:

1. The cross-drive warning fired once per **call**, not once per file. `GetAssetHandleFromFilePath` routes every absolute-path lookup through `MakeRegistryKey`, and that runs from the C#/Lua script glue — so a cross-drive asset queried from a script would print every frame. Now uses this file's existing warned-set idiom; the dedupe is on the log only, so the return value stays the pure function the unit tests pin.
2. **The one worth having caught**: my fixture comment claimed retained `WaterSpraySystem` particles were "what made #1094's two orderings disagree at all" — which my *own* measurement two paragraphs later contradicts, since the eight-pixel difference survives resetting all four services. Resetting all four is right on its own terms, but it is not the cause and the residual is still unexplained. Corrected in the test and in the new agent-rules guide, which carried the same wrong claim.

### Updated review guide

Add to **where I'd look hardest**: `IBLPrecompute.cpp` + `IBLCache.cpp` — the cache version bump is load-bearing. If it is ever reverted or the key changes shape, users silently keep an inverted bake, and the symptom (a skybox that is wrong only on machines with a warm cache) is worse than the original bug.

**Least confident about**: still the 8-pixel first-water-frame difference in #1094 — unrelated to this, and still unexplained.

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

Issue #1099, opened 2026-09-07
VulkanDrawPath: a freed descriptor slot still samples the dead texture in Release (passes in Debug)


drsnuggles8:
## Symptom

`VulkanDrawPath.EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes` fails on `origin/master` in **Release**, and passes in **Debug**, on the same machine and the same commit:

```
VulkanDrawPathTest.cpp(506): error: Expected equality of these values:
  poisoned[1]  Which is: '\xFF' (255)
  0x00u        Which is: 0
a freed slot must sample zeros (poison), not the dead texture

VulkanDrawPathTest.cpp(507): error: Expected equality of these values:
  poisoned[3]  Which is: '\xFF' (255)
  0x00u        Which is: 0
null-descriptor alpha reads zero too
```

| build | result |
|---|---|
| `clangcl-asan` **Release** | **FAILS**, both assertions, 255 instead of 0 |
| `dev-cached` **Debug** | passes |

No AddressSanitizer report accompanies it — these are value assertions, not a memory error, so ASan is incidental here. The distinguishing axis is Release vs Debug, not sanitizer vs not.

## Why it matters more than a red test

The assertion is the descriptor-lifetime contract itself. The sequence is:

```cpp
engineHeap.DestroyView(view);
EXPECT_FALSE(engineHeap.OffsetOf(view).IsValid());  // passes — the handle is rejected
engineHeap.Flush();                                 // publish the poison write
drawWithSlot(offset.Value, RHI::Access::ShaderSampleRead);
```

A freed slot is supposed to be overwritten with a null descriptor so a stale draw samples zeros. Reading `255` means the draw **still sampled the destroyed texture** — the CPU-side handle correctly rejects (`OffsetOf` returns invalid), but the GPU-side descriptor still points at the dead view. That is a use-after-free on the GPU that happens to be benign only while the memory is still mapped.

Being Release-only makes it worse, not better: it is the configuration that ships.

## Reproduce

```
build-clang\OloEngine\tests\Release\OloEngine-Tests.exe \
  --gtest_filter=VulkanDrawPath.EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes
```

cwd `OloEditor/`, ASan runtime on `PATH`. Passes with the equivalent `dev-cached` Debug binary, so an A/B between the two configs is the fastest way in.

[...]

PR #1102, opened 2026-09-07
fix(vulkan): whole-buffer storage snapshots, and the engine-heap poison tenant

### LINKED ISSUE #1059: Flaky test: VulkanDrawPath.ProductionFrustumCullWritesRootDataForIndirectDraw fails only in a full-suite run — two BindUniformBuffer calls resolve to nothing

### LINKED ISSUE #1080: RHI/Vulkan: a partial mid-frame SetData produces a TRUNCATED frame-arena snapshot, so a draw indexing past the written range reads off the end of the block
## Summary

`VulkanStorageBuffer::PushSnapshot` sizes the frame-arena snapshot to **the bytes this write touched**, not to the buffer. `GetRootDataAddress()` then hands a draw a bare device address with **no length**, so a shader that indexes past the written range reads off the end of the arena allocation. Under buffer-device-address root data (ADR 0011 §4) that read has no bounds — it is a device fault, not a wrong pixel.

Found while fixing #1058; **not fixed there** (see *Why not fixed in #1058* below).

## The mechanism

```cpp
// VulkanStorageBuffer::PushSnapshot
const u32 newBytes = std::max(offset + size, liveSnapshot ? m_SnapshotBytes : 0u);
...
const auto allocation = arena.Allocate(newBytes, 16);
```

The code already refuses to snapshot when it cannot supply the **prefix** `[0, offset)` from a CPU-readable source:

```cpp
if (offset > 0 && prefixSource == nullptr) { /* warn-once */ InvalidateSnapshot(); return; }
```

There is no matching rule for the **tail** `[offset + size, m_Size)`. The tail branch that exists only zero-fills up to `newBytes`, and `newBytes` is itself the end of the write — so the bytes past it are not in the block at all. The comment justifying it (*"reads past the written range were never defined by any writer this frame"*) is true of the frame but not of the buffer: those bytes are defined, in the persistent allocation, from earlier frames.

## Evidence

Instrumented `PushSnapshot` to log whenever `newBytes < m_Size`, then opened `VirtualGeometryTest.olo` on Vulkan (Deferred, RTX 4090, driver 104.256.0, API 1.4.351). **One scene open, five distinct storage bindings, warn-once per (binding, size):**

```
[OLO1058] TRUNCATED SNAPSHOT binding 15 size  3072 -> snapshot 1280 bytes (write 0+1280)
[OLO1058] TRUNCATED SNAPSHOT binding 16 size  4096 -> snapshot   64 bytes (write 0+64)
[OLO1058] TRUNCATED SNAPSHOT binding 17 size 11264 -> snapshot  176 bytes (write 0+176)
[OLO1058] TRUNCATED SNAPSHOT binding  9 size  5120 -> snapshot   80 bytes (write 0+80)
[OLO1058] TRUNCATED SNAPSHOT binding 10 size   256 -> snapshot   64 bytes (write 0+64)
[OLO1058] TRUNCATED SNAPSHOT binding 15 size  8192 -> snapshot  128 bytes (write 0+128)
[OLO1058] TRUNCATED SNAPSHOT binding 15 size 16384 -> snapshot  256 bytes (write 0+256)
```

Binding 17 is the worst ratio: a shader indexing element 12 of an 11,264-byte buffer reads 11 KB past a 176-byte block. These are all *whole-array uploads of the live prefix* (a light list, an instance table) into a buffer sized for the worst case — i.e. the ordinary shape, not an exotic one.

Note the writes are all at `offset == 0`, so the existing prefix guard never engages. **The truncation is invisible to every check that exists today**, which is why it has survived.

## Why this has not fired yet

A shader normally derives its loop bound from a count that matches what was uploaded (`u_LightCount`, an instance count), so it stops inside the snapshot. It fires when the bound and the upload disagree — a stale count, a GPU-written count, or a shader that indexes by an id rather than a loop. That is exactly the class #1052 and #1058 were both about.

## Suggested fix

Make the tail obey the rule the prefix already obeys: a snapshot covers the **whole buffer** or there is no snapshot.

```cpp
const u32 newBytes = m_Size;
const void* source = liveSnapshot ? m_SnapshotCpu : m_Mapped;
const bool needsUntouched = offset > 0 || size < m_Size;
if (needsUntouched && source == nullptr) { /* warn-once */ InvalidateSnapshot(); return; }
// copy the whole source, then overlay the write
```

Falling back to `InvalidateSnapshot()` is today's documented behaviour for the prefix case and is never garbage — it just returns the buffer to last-write-wins ordering.

**Trade-off to measure before committing to it:** this makes every partial mid-frame `SetData` allocate and copy the whole buffer into the arena. For binding 17 that is 11 KB per write instead of 176 B. Arena overflow already degrades gracefully (warn-once, then persistent), so the failure mode is bounded — but the arena budget should be checked against the worst frame before this lands.

## Why not fixed in #1058

#1058 is a `DynamicCopy` discriminator plus a shader bounds-check; this changes the #691 command-ordered-snapshot contract for **every** `DynamicDraw` buffer engine-wide and carries a perf trade-off that wants its own measurement. Per `CLAUDE.md` → *Scope*, a ride-along whose fix would widen the diff a lot is reported rather than folded in.

## Acceptance

- A snapshot is never shorter than the buffer it stands in for, or there is no snapshot.
- A regression test in `VulkanDrawPathTest`'s `SubmitFrame` bracket: a partial `SetData` on a `DynamicDraw` buffer, then assert the root address either covers `GetSize()` bytes or equals `GetDeviceAddress()`.
- The arena budget is checked against the worst observed frame.

Related: #1058 (found there), #1052, #691, #1015 (`ClearData(offset, size)` exists to dodge this seam one level up).

🤖 Filed with [Claude Code](https://claude.com/claude-code)

## Score

```olo-score
capability: 3
craft: 1
stability: 8
decay: 3
effort: 2
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md).</sub>

### LINKED ISSUE #1087: VulkanDrawPath.EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes fails on master under Release+ASan: a freed heap slot samples 0xFF, not the null-descriptor poison
## Symptom

On `origin/master` @ `921bce094`, `clangcl-asan` preset, **Release**, in isolation:

```
VulkanDrawPathTest.cpp(506): error: Expected equality of these values:
  poisoned[1]  Which is: '\xFF' (255)
  0x00u        Which is: 0
a freed slot must sample zeros (poison), not the dead texture
VulkanDrawPathTest.cpp(507): ... same, null-descriptor alpha
```

The test destroys a view, flushes to publish the poison write, then draws through the freed slot and requires the sample to read the null-descriptor zeros. It reads 255 instead.

## Config-specific, and pre-existing

| Config | master `921bce094` | PR #1079 branch |
|---|---|---|
| `dev-cached` Debug | **passes** | **passes** |
| `clangcl-asan` Release | fails, `0xFF` | fails, `0xFF` |

Identical on both sides, so it is not a branch regression. It is the Release build (or clang-cl vs the Debug toolchain) that changes the outcome, which is the part worth understanding: the assertion is about a *deterministic* null-descriptor read, and a value that depends on the build configuration suggests the poison write is not actually ordered before the draw, rather than that the threshold is wrong.

Possibilities worth separating before touching the test:

1. `Flush()` does not in fact publish the poison write before `drawWithSlot`, and Debug's slower timing hides it.
2. The device does not support `nullDescriptor` (or it is not enabled), so the freed slot reads stale texels — and Debug happens to leave zeros there.
3. The slot is genuinely still bound to the dead texture.

(2) reads as the most likely and is cheap to check: query `VkPhysicalDeviceRobustness2FeaturesEXT::nullDescriptor` and have the test skip loudly rather than assert if it is off — per `docs/agent-rules/no-silent-fallbacks.md`, a path that cannot do its job should say so countably.

## Why nobody noticed

GPU-gated: CI's ASan job runs on a GPU-less runner, and the Windows job does not build with ASan, so no CI configuration runs this assertion.

Note `OloEngine/tests/Rendering/VulkanDrawPathTest.cpp` is currently being edited on `feature/vulkan-drawpath-correctness-1058` (#1058/#1059) — worth checking there first.

Found while working #1071 / PR #1079. Unscored — needs the picker's scoring block.

## Score

```olo-score
capability: 2
craft: 1
stability: 5
decay: 3
effort: 2
confidence: 0.8
learning: 3
fun: 2
kano: table-stakes
blocked_by: []
blocks: []
```

<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md).</sub>

drsnuggles8:
Three open Vulkan-RHI correctness issues, batched because they live in the same subsystem and the
same test file. Two needed code; **one turned out to be already fixed on `master`** and is reported
with evidence rather than with a manufactured change.

Closes #1080
Closes #1087
Closes #1059

## CI cannot verify any of this

All three are GPU-gated and **no CI runner has a Vulkan device**, so a green check on this PR is not
evidence for it. Every claim below comes from a local A/B on an RTX 4090 (driver 104.256.0, API
1.4.351), Windows 11, `dev-cached` Debug.

---

## #1080 — a snapshot shorter than the buffer it stands in for

`VulkanStorageBuffer::PushSnapshot` sized the frame-arena snapshot to the bytes the write touched.
`GetRootDataAddress()` hands a draw that address with **no length**, so the shader's own indexing was
the only thing keeping the read inside the buffer — and the index routinely comes from a *different*
buffer that was fed correctly. That is the severity-ranked case in
[no-silent-fallbacks.md](docs/agent-rules/no-silent-fallbacks.md): a small stand-in that a shader
indexes.

The prefix `[0, offset)` already had the rule (refuse when no CPU-readable source can supply it). The
tail `[offset + size, m_Size)` had none.

**Chosen fix: whole buffer, or no snapshot** — not a length carried alongside the address. That is
already the contract on this seam: the sibling `VulkanUniformBuffer` has always pushed its entire
shadow, and the storage twin was the outlier. Carrying a length would touch every root-data writer
and every shader that reads one.

### A/B (the fix is load-bearing)

Reverting only `const u32 newBytes = m_Size;` to the old `std::max(...)` expression, same binary
otherwise:

| | pre-fix | post-fix |
|---|---|---|
| `SetData(0, 16)` on a 256-byte buffer | snapshot **16 B** | **256 B** |
| `SetData(32, 8)` on the same buffer | snapshot **40 B** | **256 B** |

Same shape as the issue's instrumented evidence (a 176-byte snapshot for an 11,264-byte binding-17
buffer).

### Arena budget — checked, as the issue's acceptance criteria require

From the issue's own instrumented VG frame: **2,048 B to 48,384 B** of snapshot per frame against a
**16 MiB** slot — 0.29%, 346x headroom.

More usefully, measured live rather than on paper: a `VirtualGeometryTest.olo` session on Vulkan +
Deferred (15 instances, 4,860 clusters, both hardware and software raster arms active) recorded
**zero arena overflows** — and overflow is precisely the failure mode the trade-off risked, with its
own `ERROR` and counter.

### Review found two defects this change introduced, both fixed here

Folded in as a third commit rather than left for CodeRabbit:

1. **Arena exhaustion.** `GPUScene::Upload` issues one `SetData` per *non-adjacent* dirty range
   (`CoalesceDirtyRanges` merges only strictly adjacent indices), so N scattered edits claimed N
   whole-buffer ranges. The arena backs *every* draw's root data, so exhausting it drops root data
   for the whole frame. Fixed by rewriting an **unread** snapshot in place — a snapshot is
   observable only through the address a draw embedded, so until `GetRootDataAddress()` is called,
   overwriting it is indistinguishable from never staging it. Ordering is defined against recorded
   draws, not against `SetData` calls.
2. **A cleared range came back.** Mid-frame `ClearData` records a fill into the frame command buffer
   and never touches `m_Mapped`, so filling a later partial write's undefined bytes from the mapping
   resurrected the pre-clear content for every draw recorded afterwards. A recorded GPU-side write
   now marks the mapping untrustworthy for the frame; a partial write refuses countably and draws
   fall back to the persistent buffer, which is where the clear actually lands.

### Known behaviour change, stated rather than buried

- A `DynamicDraw` buffer **larger than the arena's 16 MiB slot** can no longer snapshot at all and
  reverts to last-write-wins ordering. Deliberate: a short snapshot *was* the out-of-bounds read.
  It warns once and increments `GetSnapshotRefusedCount()` rather than passing silently.
- The undefined tail is read back from `m_Mapped`, which is `HOST_ACCESS_SEQUENTIAL_WRITE`
  (write-combined). Fix 1 reduces this to once per buffer per frame. **I did not A/B frame time
  against `master`**, so I am not claiming a figure for it — only that the arena never overflowed
  and the frame is visually unchanged.

---

## #1087 — none of the three proposed mechanisms was the cause

The issue lists three candidates and names `nullDescriptor` as most likely. It is **not** any of
them.

`HeapDesc::PoisonOnFree` is gated on `OLO_DEBUG` in *both* backends' installs
(`VulkanDescriptorHeapBackend::InstallOntoEngineHeap`, `OpenGLRendererAPI::Init`). In Release the
freed slot keeps the dead texture's descriptor and samples `0xFF`; in Debug it is overwritten and
samples `0x00`. That is the whole Release-vs-Debug split, and it is **by design** — `HeapDesc`'s own
documentation describes poison as a debugging aid, the sibling of `OLO_RG_POISON_TRANSIENTS`.

Candidate (2) is ruled out *by construction*: this backend's nulls are real 1x1 black images
**because** robustness2's `nullDescriptor` is not on the device floor
(`VulkanDescriptorHeapBackend.h`). Querying that feature could not have explained anything, so the
handover's suggested first step was a dead end.

The tenant now re-initialises the engine heap with poison **on** before minting any view, instead of
skipping the assertion in Release: the contract in the test's name is a property of the heap, not of
the build. Also asserts `Stats::SlotsPoisoned` moves, so the poison is counted, not merely performed.

**A/B:** flipping that one flag back to `false` reproduces the exact reported failure — `poisoned[1]`
and `poisoned[3]` both `0xFF`, same two assertions — **in a Debug build**. The `clangcl-asan` Release
tree was never needed to pin it.

---

## #1059 — already fixed on `master`; no change made

Commit `e715d36df` (2026-09-05, *after* the issue was filed) added a `Renderer3D::HasInitialized()`
precondition skip with exactly this diagnosis. A full-suite run on this branch confirms it works:

- `VulkanDrawPath.ProductionFrustumCullWritesRootDataForIndirectDraw` — **SKIPPED**, guard engaged.
- **`unresolvable buffer`: 0 occurrences** across 7,632 tests, against exactly one in the issue's
  failing run.

Per the handover's own instruction not to manufacture a change to justify the branch, I left it
alone. Recommending closure; **not closing it myself.**

One observation, deliberately *not* acted on: the guard tests `Renderer3D::HasInitialized()`, while
the precondition that actually breaks the tenant is `CommandDispatch` publishing UBO refs minted on
another backend (`CommandDispatch::HasUBOReferences()`). The two agree at every transition today
because `Renderer3D::Shutdown()` calls `CommandDispatch::Shutdown()`. Worth knowing if that edge ever
moves.

---

## Verification

**Full suite (Debug, local): 7,634 tests — 7,593 passed, 41 skipped, 0 failed, 0 SEH exceptions.**
The one failure the earlier runs carried turned out to be a real null-deref in engine code, root-
caused and fixed here — see *A pre-existing crash found and fixed on the way* below.

**`VulkanDrawPath` (10/10 pass)**, including four tenants added here:

| tenant | pins |
|---|---|
| `PartialStorageBufferWriteSnapshotsTheWholeBufferOrNothing` | whole-buffer-or-nothing, extent **and** content, at `offset == 0` (the shape the old prefix guard never saw) |
| `ConsecutiveWritesShareOneSnapshotUntilADrawReadsIt` | four consecutive writes produce **one** arena range; a fresh range after a draw reads it |
| `AClearedRangeIsNotResurrectedByALaterPartialWrite` | the refusal is counted and draws fall back to persistent |
| `EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes` | poison in **every** config, and counted |

**Live editor** — `VirtualGeometryTest.olo`, Vulkan, Deferred, the exact configuration that produced
the issue's evidence. Frame unchanged and correct from three camera angles (intact geometry, coherent
PBR shading, correct shadows and reflections). Counters all zero: **0** snapshot refusals, **0** arena
overflows, **0** shader errors, **0** validation errors, **0** device-lost. Clean shutdown.

**GL path** exercised via the full suite; `OpenGLStorageBuffer::SetData` is a bare
`glNamedBufferSubData` with no snapshot machinery at all, so this change cannot reach it — the fix is
Vulkan-only by construction, not by accident.

## Notes

- **#1099** is a duplicate of #1087 (same test, same assertion, same axis) and already carries an
  evidence comment. Not closed here — leaving that to you.
- Docs updated: `vulkan-command-ordered-buffer-writes.md` named #1080 as an open defect in its own
  closing lesson; that is now the mechanism's stated invariant, with the README index line refreshed
  to match.
- The `sonarqube` MCP server failed to connect this session (`CONNECTION_CLOSED`), so no Sonar
  cross-check was possible.

### A pre-existing crash found and fixed on the way

Both earlier full-suite runs failed `NetworkThreadDispatchTest.EnqueueGameThreadFromNetwork` with an
**access violation (`SEH 0xc0000005`)**. It was not a flake, and it was not a networking bug — that
test was simply the next thing to call `ProcessAll`.

**Root cause.** `EditorAssetManager` dispatches `AssetReloadedEvent` by queueing a game-thread task
whose body calls `Application::Get()` — which is `*s_Instance` with no null check. The task
deliberately outlives the call that queued it, so it is only safe if an `Application` is alive when
the game thread finally drains the queue. Nothing guarantees that: the test binary has no
`Application` at all, and no `Application::Run` to drain the queue either, so the task sat in the
process-wide `FNamedThreadManager` queue until an unrelated test drained it and dereferenced null.

```
TextureInPlaceReload.ReloadsATextureImportedThroughTheAssetSystem   <- queues it
... ~5,700 tests later ...
NetworkThreadDispatchTest.EnqueueGameThreadFromNetwork              <- drains it -> 0xc0000005
```

**Bisected** from the 7,632-test suite to a two-test, ~1-second repro:

```
--gtest_filter='TextureInPlaceReload.ReloadsATextureImportedThroughTheAssetSystem:NetworkThreadDispatchTest.EnqueueGameThreadFromNetwork'
```

**Fix** (own commit, `fix(assets): a queued AssetReloadedEvent must not assume an Application exists`):
both dispatch sites now use `Application::TryGet()` — the null-safe accessor the codebase already
provides — and skip the event when no `Application` is hosting. The event is a UI notification and
has no meaning without one; an engine subsystem queueing work for a host that may not exist is the
actual defect.

The regression is pinned **where it is caused**, not where it detonated: the texture-reload test now
drains the game-thread queue itself and asserts that is safe with no `Application` alive. Without the
fix, that drain faults inside the guilty test instead of taking down an unrelated one thousands of
cases later.

Not touched here, but worth recording: `McpServer::MarshalRead` has the same latent shape — on
timeout or shutdown it throws while leaving its `MCP_MainThreadRead` task queued, and that task's
`readJob` captures the `McpServer` by reference at ~29 call sites. It is not this bug (bisection
cleared the MCP suites outright) and deserves its own change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented asset reload processing from crashing when no application instance is active.
  * Improved Vulkan buffer updates to avoid stale or incomplete data after partial writes and GPU-side clears.
  * Ensured consecutive buffer updates and draw operations use consistent snapshot data.

* **Diagnostics**
  * Added visibility into live buffer snapshots and refused snapshot operations.

* **Documentation**
  * Updated Vulkan rendering guidance to clarify whole-buffer snapshot requirements.

* **Tests**
  * Added regression coverage for asset reload safety and Vulkan buffer snapshot behavior.


drsnuggles8:
@coderabbitai review

Issue #551, opened 2026-07-04
Renderer/log-hygiene: per-frame GPU->CPU buffer readback (GL_DYNAMIC_COPY VIDEO->HOST) logged unthrottled every frame across scenes


drsnuggles8:
## Summary

While sweeping 26 sandbox scenes over the MCP diagnostics server (Debug editor), `olo_log_tail`/`OloEngine.log` in almost every actively-rendering scene is flooded with the **same** OpenGL driver performance warning, emitted **every frame**:

```
[warning] OpenGL performance warning (source: API, id: 131186): Buffer performance warning:
Buffer object 90 (bound to NONE, usage hint is GL_DYNAMIC_COPY) is being copied/moved
from VIDEO memory to HOST memory.
```

At the editor's frame rate (655 fps observed on the sandbox), that is hundreds of identical lines per second. Two distinct problems:

1. **Log-spam / hygiene (primary, cheap):** the GL debug-message callback logs this repeated performance-severity message **unthrottled**, ballooning `OloEngine.log` in Debug. This is the same failure mode that bit #524 (126k identical lines stalled the main thread and timed out MCP marshals). The callback should dedup/rate-limit repeated messages (and/or drop `GL_DEBUG_SEVERITY` performance notifications below error).
2. **Per-frame GPU→CPU buffer readback (secondary):** the warning means a `GL_DYNAMIC_COPY` buffer is migrated VIDEO→HOST **every frame** because the CPU reads it back (`OpenGLStorageBuffer::GetData` → `glGetNamedBufferSubData`, [OpenGLStorageBuffer.cpp:51-54](../blob/master/OloEngine/src/Platform/OpenGL/OpenGLStorageBuffer.cpp#L51), or an editor picking PBO). A per-frame synchronous readback defeats GPU residency; even if small, it is worth eliminating (fence + latent readback, or persistent-mapped ring).

## Evidence

- Observed across `particle_test` (40+/window), `FontRenderingTest`, `SponzaDeferred` (22), `WaterTest`, `FogTest`, `GodRayUnderwater`, `PrecipitationTest`, `SSGITest`, etc. — it is **renderer-wide, not scene-specific** (same buffer id 90 each time).
- **Not** a shader/asset problem: `olo_shader_errors` = 0 and `olo_assets_problems` = 0 in every swept scene.
[...]

PR #1106, opened 2026-09-07
perf(renderer): indirect-dispatch the virtual-geometry software raster, and measure #712

### LINKED ISSUE #712: Renderer: triangle-level culling in the virtual-geometry raster (backface + sub-sample-miss)

### LINKED ISSUE #1048: Renderer: dispatch the virtual-geometry software rasterizer indirectly from the GPU-written work-list count

`VirtualGeometryPass` dispatches the compute software rasterizer at the frame's **total cluster
count**, not at the software work-list count:

```cpp
// VirtualGeometryPass.cpp, section 4
u32 const maxSwRecords = frameClusterCount;
u32 const groupsX = std::min(maxSwRecords, 4096u);
u32 const groupsY = (maxSwRecords + groupsX - 1u) / std::max(groupsX, 1u);
```

Every workgroup past the live count early-outs on `if (swRecordIndex >= swList.Count) return;` —
the issue #551 idiom, and correct, because the count is GPU-written and never read back. But the
ratio is not small. Measured live on `VirtualGeometryStress` via `olo_virtual_geometry_stats`:

| | |
|---|---|
| `frameClusters` (workgroups dispatched) | **2,834,448** |
| `cull.softwareRasterized` (records with work) | **3,165** (9,632 in a wider shot) |
| overdispatch | **~300–900×** |

**This is not a theoretical cost — it is the whole cost of the pass.** From #712, using the
`VirtualGeometryPass/SwRaster` GPU sub-pass bracket that PR added (min-of-24, four interleaved
rounds, `VirtualGeometryPerf` scene):

```
arm A (origin/master)      SwRaster min 0.1321 ms   Resolve 0.2017 ms
arm B (#712's reject)      SwRaster min 0.1321 ms   Resolve 0.2007 ms
arm C (empty main())       SwRaster min 0.1321 ms   Resolve 0.1229 ms  (-39%)
```

Spread within an arm is under 1%, so these are not noise-limited. **A shader whose `main()` returns
immediately costs exactly what the real rasterizer costs**, while the `Resolve` control drops 39%
under that arm exactly as it should (nothing rasterized ⇒ nothing to resolve) — so the instrument is
live, and the bracket is dominated by workgroup launch, not by rasterization.

The direct consequence: **no change inside `VirtualClusterRaster.comp` is measurable today.** #712's
sub-sample-miss reject removes 77% of the inner-loop iterations on its test corpus and moves the
GPU number by 0.0%.

Dispatch indirectly from the GPU-written software count.

`RenderCommand::DispatchComputeIndirect(handle, offsetBytes)` already exists and is already used in
exactly this shape by `LightCullingPass` and by `TerrainGPUPicker`, which writes its next dispatch's
arguments into its own state buffer and dispatches at a byte offset with no CPU round trip.

The work-list header is already the right shape for it — `VirtualSwList` is
`{ uint Count; uint _h0, _h1, _h2; }`, so the three pad words at offset 4 are a natural
`(numGroupsX, numGroupsY, numGroupsZ)` triple:

1. A small compute pass (one thread) after both cull phases converts `swList.Count` into that triple,
   keeping the existing 2D flattening so large lists stay under the 65535 per-axis grid cap.
2. `VirtualGeometryPass` replaces both raster dispatches (the portable path issues two) with
   `DispatchComputeIndirect(swListBuffer, 4)`.
3. The `swRecordIndex >= swList.Count` guard stays — it is still correct, just no longer load-bearing.

Not free of subtlety, which is why it is its own issue rather than a rider on #712: it needs a
shader-storage → indirect-command barrier on GL, the matching `VK_ACCESS_INDIRECT_COMMAND_READ_BIT`
transition on Vulkan, and a render-graph hazard review, plus the same visibility-buffer identity
evidence #712 established.

* **Read `swList.Count` back on the CPU** and size the dispatch from it. Rejected: it is either a
  stall or a frame late, and "a frame late" silently under-dispatches on a camera cut, which drops
  clusters — the exact failure the conservative bound exists to prevent.
* **Cap the bound at the SW list capacity.** It already is that; the capacity *is* the frame cluster
  count, because any cluster could in principle route to software.
* **Leave it.** The measurement above says the pass is paying ~100% overhead on the stress scene.

**Additional context**

Found while measuring #712 (PR #1050), which added the `VirtualGeometryPass/SwRaster` and
`.../Resolve` sub-pass brackets and `VirtualGeometryPerf.SoftwareRasterSubPassBudget` — the
instrument this issue's numbers come from, and the one that would show the fix working.

Related: #629 (the virtualized-geometry pipeline), #551 (the conservative-upper-bound dispatch
idiom this is an instance of), #712.

## Score
```olo-score
capability: 3
craft: 1
stability: 1
decay: 1
effort: 3
confidence: 0.8
learning: 5
fun: 5
kano: performance
blocked_by: []
blocks: []
```

### LINKED ISSUE #1105: Renderer/Vulkan: virtual-geometry software raster renders shredded geometry — one cause fixed, a second remains
Virtual geometry renders as scattered fragments on **Vulkan** where GL renders solid. Every CPU-side counter reads correct on both backends, so the only evidence is the picture.

Repro: `VirtualGeometryStress.olo`, Deferred, RTX 4090. Requires the fetched dragon (`scripts/Fetch-Assets.ps1`).

## What is already fixed

The software raster hard-coded GL's `ndc.z * 0.5 + 0.5` for its window depth. On Vulkan `AdjustProjectionForBackend` has already mapped clip z to `[0, w]`, so ndc.z IS the window depth and the second remap compressed every software depth into `[0.5, 1]` — software clusters then lost the depth test against hardware-rasterized ones almost everywhere.

Fixed on `feature/vg-indirect-dispatch-and-triangle-culling-1048-712` by adding `RHI::NdcToWindowDepthScaleBias()` as a third member of the projection seam. That takes the frame from scattered fragments to substantially complete geometry, GL bit-unchanged.

## What remains — this issue

With that fix in, the mesh is **still not correct**: dark patches and missing regions against the hardware-only reference.

| Arm | Result |

|---|---|

| GL, virtual geometry on | solid, correct |

| Vulkan, virtual geometry **off** (classic mesh path) | solid, correct |

| Vulkan, `swRasterMode=disabled` (all clusters hardware) | **solid, correct** |

| Vulkan, software raster on, before the depth fix | scattered fragments |

| Vulkan, software raster on, after the depth fix | much better, still patchy |

So the defect is confined to the **compute software rasterizer / visibility-buffer path** on Vulkan.

## Ruled out (each tested, not reasoned about)

- **Not the indirect dispatch of #1048** — the indirect path and the old CPU-side bound produce **byte-identical** frames on Vulkan (same md5), before and after the depth fix.

- **Not the hardware raster path** — `hwRasterMode=forcemdi` is byte-identical to the mesh-shader path.

- **Not the int64 atomics** — `forcePortableSwRaster=true` is byte-identical to the int64 variant, with and without the depth fix.

- **Not #1080** — the in-flight fix from `feature/vulkan-rhi-lifetime-1080-1087-1059` (`84c17c460` + `65510000b`) was applied here and produced a **byte-identical** frame both before and after the depth fix.

- **Not the backface winding** — negating the window-space signed area is strictly worse (reverts to the original shredding), and the resolve's `ScreenBarycentrics` divides by the signed area, so it is sign-agnostic and handles a mirrored window space correctly.

## Counters that stay green while the picture is wrong

Identical on GL and Vulkan, so none of these can gate it:

```

cutSelected 3255, drawnClusters 3216, hardwareDraws 51,

softwareRasterized 3165, testedClusters 2834448

silentlyDrewNothing false, unresolvedAssets 0, shader errors 0

```

## Also noticed

The `clusterid` debug overlay is **displaced** on Vulkan — the coloured cluster region sits well away from the geometry — both before and after the depth fix. That looks like a separate coordinate bug in the debug composite, and it makes the debug view unusable as a diagnostic here. Worth splitting out if it is not the same root cause.

## Why nobody noticed

The GPU visual tests run on GL only (`RenderPropertyTest` selects a `GlBackend`), and an L1-L5 contract test cannot see a depth-convention error. Nothing in CI renders virtual geometry on Vulkan.

## Score

```olo-score

capability: 5

craft: 5

stability: 5

decay: 1

effort: 3

confidence: 0.5

learning: 5

fun: 5

kano: table-stakes

blocked_by: []

blocks: []

```

<sub>Rated per [issue-scoring](../../docs/process/issue-scoring.md) - score = confidence x (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
Lands #1048 and #712 as a pair, because they were measurement-coupled: until the
overdispatch was gone, nothing inside the software rasterizer could be measured
at all.

## #1048 — dispatch indirectly from the GPU-written count

The raster was dispatched at the frame's **total cluster count**, because the
count of clusters actually routed to software is GPU-written and never read
back. Workgroups past the live count early-out (the #551 idiom, correct) — but
on `VirtualGeometryStress` that is **2,834,448 workgroups launched for 3,165
records with work**, reproduced live here at **895.6x** via
`olo_virtual_geometry_stats`.

A one-thread kernel now turns the count into a workgroup grid written to the
work list's own header, and the raster dispatches from there. No readback (a
readback is a stall or a frame late, and a frame late silently *under*-dispatches
on a camera cut). No new binding — the header's three pad words at byte offset 4
are already the shape `DispatchComputeIndirect` wants, and the kernel reads its
capacity from the raster's existing parameter block, so the engine's last free
UBO stays free.

The flattening rule lives in one include shared by the kernel, an L2 GPU probe
and a CPU mirror, because a grid that is too small drops clusters **with no
error raised anywhere**. The `>= Count` guard stays as belt-and-braces; the CPU
bound stays as a loud fallback.

### Measured — `VirtualGeometryPerf`, Release, 4 interleaved rounds, min-of-24

Arms are selected at runtime, so one binary serves all four and no rebuild sits
between them.

| Arm | | SwRaster | vs A |
|---|---|---|---|
| A | CPU bound (master) | 0.1280 ms | — |
| B | **indirect** | **0.0748 ms** | **−41.6%** |
| C | indirect + empty `main()` | 0.0092 ms | |
| D | CPU bound + empty `main()` | 0.1270 ms | |

**D == A reproduces the premise**: with the CPU bound, a shader whose `main()`
returns immediately costs what the real rasterizer costs — the pass was
launch-bound, not raster-bound. **C now separates from A**, so the bracket has
gone from timing workgroup launch to timing rasterization. That flip is the
result; the −41.6% is a consequence of it.

Output unchanged: every virtual-geometry visual capture is **byte-identical**
between arms A and B, while the run-to-run noise floor (B vs B) moves one file.
The `Resolve` control moves −1.6% between A and B while dropping ~40% in the
arms that rasterize nothing — its cost follows the visibility buffer, so an
unchanged Resolve is independent evidence the buffer content is unchanged.

## #712 — the rejects, finally measurable

**#712's reject logic was already on master** (`bd82ab48c`); its acceptance
criteria were measurements, and the measurement was impossible. With the launch
cost gone it is possible. All arms below run with the indirect dispatch on:

| Arm | | SwRaster | vs B |
|---|---|---|---|
| B | both rejects (master today) | 0.0778 ms | — |
| E | no sub-sample reject | 0.0891 ms | **+12.6%** |
| F | no backface reject | 0.0778 ms | **0.0%** |
| G | neither | 0.0860 ms | +9.5% |

**The sub-sample-miss reject is worth ~12.6% of software-raster GPU time.** It
measured exactly 0.0% before #1048.

**The backface reject measures 0.0%** on this corpus — arms B and F are
identical across all four rounds. Reported rather than dressed up. The likely
reason is structural: the signed area is computed regardless (`orient`, `area`,
`invArea` all need it), so the early return removes loop iterations but no math,
and the triangles it would reject are largely ones the sub-sample test rejects
anyway.

Caveat stated plainly: this scene routes **338** clusters to software, not the
3,165 of the stress scene. The 12.6% is real but measured on a modest corpus.

One arm-E round read 0.1516 instead of ~0.0900 with its Resolve control spiking
at the same time; per the instrument's own design note, a reading where *both*
brackets move is the box drifting, and it was excluded.

## Ride-along fixes (own commits): Vulkan virtual geometry was broken two ways

Found while verifying #1048 on Vulkan, **neither caused by it** — the indirect
dispatch and the old CPU-side bound produce byte-identical Vulkan frames at
every stage of this investigation. Both are in `VirtualClusterRaster.comp`,
which hand-rolls the two things fixed function would otherwise do for it.

**1. Window depth (`5d55f4b83`).** The raster hard-coded GL's `ndc.z*0.5+0.5`;
on Vulkan `AdjustProjectionForBackend` has already mapped clip z to `[0, w]`, so
the second remap compressed every software depth into `[0.5, 1]` and software
clusters lost the depth test against hardware-rasterized ones.

**2. Front-face sign (`6c11c9b94`, closes #1105).** The raster culls back faces
on the sign of a determinant it computes itself. Fixed function needs no help —
Vulkan evaluates facing in framebuffer coordinates (y down) and that composes
with the seam's clip-y negation to identity — but a shader that builds the
determinant itself gets only one of those inversions, so the cull was discarding
the **front** faces and keeping the back ones.

Both are now members of the projection seam (`NdcToWindowDepthScaleBias()`,
`WindowSpaceFrontFaceSign()`), derived from the same `BackendFlips()` predicate
the matrices use, so the GLSL stays source-identical. Both ride in existing pad
words: no layout change, no new binding.

### Measured — covered pixels in the raster's own debug target, fixed camera and viewport

| | Covered px |
|---|---|
| GL (reference) | 101,293 |
| Vulkan, before | 84,765 (83.7%) |
| Vulkan, after both fixes | **101,292** |

**Vulkan virtual geometry now matches GL.** Bisected, not guessed: disabling the
backface guard alone took Vulkan to 101,295, and 68% of the missing pixels had
no covered neighbour — solid regions, not edge erosion.

Worth recording, because it cost time: an earlier probe that negated
`signedArea` outright appeared to exonerate the winding. It did not — negating
the value also flips `orient`, which inverts the edge functions so triangles
stop covering anything (27,588 px). The cull sense and the winding that orients
the edge functions must move independently. Eyeballing lit frames could not have
found this; a closed mesh looks broadly similar whether you draw its front or
its back faces, and only the pixel-coverage metric separated them.

**GL is not bit-identical, and the shader comment says so with numbers.**
Reading the new uniform shifts GL's floating-point codegen: 25–126 pixels per
230,400 (0.01–0.06%) move by a maximum channel delta of **1**. No geometry
gained or lost; all 143 GL virtual-geometry tests green. Both a multiply and a
select form shift it identically, so it is the uniform read rather than the
arithmetic; a per-backend `#define` would avoid it at the cost of the
source-identical property the seam exists to preserve.

Also ruled out along the way, each by test rather than argument: the
mesh-shader path (`forcemdi` byte-identical), int64-vs-portable atomics
(byte-identical), and #1080 (its in-flight fix from
`feature/vulkan-rhi-lifetime-1080-1087-1059` applied here changed nothing,
twice).

## Verification

- 361/363 pass across `Vulkan*` / `Virtual*` / `ShaderUnit*` / `RHI*`
- The one failure, `VulkanDrawPath.EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes`, is **#1087** — pre-existing, reproduced with this branch's `VirtualGeometryPass` change reverted and rebuilt, and owned by `feature/vulkan-rhi-lifetime-1080-1087-1059`. A config data point was added to that issue.
- New CPU contract tests pin `gx*gy >= count` against the raster's own index recovery; a new L2 GPU probe pins the shipped GLSL against the CPU mirror.
- Live editor on **both** backends: indirect path confirmed active in the log, 0 shader errors, multi-angle captures inspected.
- Vulkan virtual geometry verified against GL by pixel coverage, not by eye.

Self-review found and fixed two real defects before this PR existed: a
four-billion-group dispatch on an empty frame (`0 + 0 - 1u` underflow in the
fallback bound — the same bug the phase-2 cull is braced against), and an args
dispatch sitting inside the timed bracket, whose `Command` barrier is a full
pipeline flush the master arm never pays. The second inflated arm B, so the
numbers above are the corrected re-measurement.

`blocked_by` cleared on #712.

Closes #1048
Closes #712

## Summary by CodeRabbit

- **New Features**
  - Software rasterization now uses GPU-generated indirect dispatch sizing for improved workload handling.
  - Added backend-aware depth mapping and front-face orientation support for consistent rendering across graphics APIs.
  - Added safe fallback dispatch behavior when GPU argument generation is unavailable.

- **Bug Fixes**
  - Corrected empty and large work-list dispatch handling, including boundary cases.

- **Tests**
  - Added coverage for dispatch sizing, shader compilation, cross-backend behavior, and over-dispatch diagnostics.

### DIFF
--- a/OloEditor/assets/shaders/VirtualVisibilityResolve.glsl
+++ b/OloEditor/assets/shaders/VirtualVisibilityResolve.glsl
@@ -117,7 +117,7 @@ layout(std430, binding = 39) readonly buffer VirtualVertices { VirtualGpuVertex 
 layout(std430, binding = 42) readonly buffer VirtualIndices { uint localIndices[]; };
 layout(std430, binding = 40) readonly buffer VirtualSwList {
     uint Count;
-    uint _h0; uint _h1; uint _h2;
+    uint DispatchX; uint DispatchY; uint DispatchZ; // indirect dispatch args, written by VirtualRasterArgs.comp (#1048)
     VisibleCluster Records[];
 } swList;
 // The atlas sampler + the sentinel-gated decode, shared verbatim with the
--- a/OloEditor/assets/shaders/compute/VirtualClusterCull.comp
+++ b/OloEditor/assets/shaders/compute/VirtualClusterCull.comp
@@ -124,7 +124,7 @@ layout(std430, binding = 38) writeonly buffer VirtualVisible { VisibleCluster vi
 // clusters whose projected coverage is under u_SwRasterThresholdPixels.
 layout(std430, binding = 40) buffer VirtualSwList {
     uint Count;
-    uint _h0; uint _h1; uint _h2;
+    uint DispatchX; uint DispatchY; uint DispatchZ; // indirect dispatch args, written by VirtualRasterArgs.comp (#1048)
     VisibleCluster Records[];
 } swList;
 
--- a/OloEditor/assets/shaders/compute/VirtualClusterRaster.comp
+++ b/OloEditor/assets/shaders/compute/VirtualClusterRaster.comp
@@ -146,9 +146,15 @@ layout(std140, binding = 70) uniform VirtualRasterParams
     //      append, not the counter), so a GPU overflow would inflate the very
     //      number the read is bounded by. 0 = an empty list.
     uint  u_SwListCapacity;
-    uint  _vrPad0;          // 20 — std140 rounds the block to a 16-byte multiple;
-    uint  _vrPad1;          // 24   explicit so the C++ mirror is a plain struct.
-    uint  _vrPad2;          // 28
+    // 20/24 — ndc.z -> window depth as (scale, bias), from
+    //         RHI::NdcToWindowDepthScaleBias(). The software raster does its own
+    //         perspective divide, so it applies the mapping fixed function would
+    //         have: GL 0.5/0.5 (ndc z is [-1,1]); Vulkan 1/0 (the seam already
+    //         mapped clip z to [0,w], so ndc.z IS the window depth). Data, not a
+    //         backend #ifdef, so this file stays source-identical.
+    float u_DepthScale;
+    float u_DepthBias;
+    uint  _vrPad0;          // 28 — std140 rounds the block to a 16-byte multiple.
 };
 
 // Bounded micro-triangle raster: SW-routed clusters have small projected
@@ -243,9 +249,18 @@ void RasterizeTriangle(uint swRecordIndex, uint triangleIndex)
     float invArea = 1.0 / area;
     // Window depth in [0,1], interpolated with screen-space barycentrics —
     // the same interpolation the hardware rasterizer applies to gl_FragCoord.z.
-    float z0 = ndc0.z * 0.5 + 0.5;
-    float z1 = ndc1.z * 0.5 + 0.5;
-    float z2 = ndc2.z * 0.5 + 0.5;
+    // Window depth via the backend's ndc.z mapping (u_DepthScale/u_DepthBias from
+    // RHI::NdcToWindowDepthScaleBias), NOT a hard-coded *0.5+0.5. This shader
+    // projects with the SAME adjusted matrix the hardware path uses, and on
+    // Vulkan that matrix has already mapped clip z into [0, w] — so ndc.z is
+    // ALREADY the window depth there, and remapping it again compressed every
+    // software depth into [0.5, 1]. The software clusters then lost the depth
+    // test against the hardware-rasterized ones almost everywhere and the mesh
+    // rendered in shredded fragments, while every CPU-side counter (cut size,
+    // software-rasterized count, dispatch size) still read exactly right.
+    float z0 = ndc0.z * u_DepthScale + u_DepthBias;
+    float z1 = ndc1.z * u_DepthScale + u_DepthBias;
+    float z2 = ndc2.z * u_DepthScale + u_DepthBias;
 
     uint payload = (swRecordIndex << 9u) | (triangleIndex & 0x1FFu);
 
--- a/OloEditor/assets/shaders/compute/VirtualClusterRaster.comp
+++ b/OloEditor/assets/shaders/compute/VirtualClusterRaster.comp
@@ -102,7 +102,7 @@ layout(std430, binding = 42) readonly buffer VirtualIndices { uint localIndices[
 // ClusterIndex} with the record's own slot stored by the cull.
 layout(std430, binding = 40) buffer VirtualSwList {
     uint Count;
-    uint _h0; uint _h1; uint _h2;
+    uint DispatchX; uint DispatchY; uint DispatchZ; // indirect dispatch args, written by VirtualRasterArgs.comp (#1048)
     VisibleCluster Records[];
 } swList;
 
--- a/OloEditor/assets/shaders/compute/VirtualClusterRaster.comp
+++ b/OloEditor/assets/shaders/compute/VirtualClusterRaster.comp
@@ -154,7 +154,12 @@ layout(std140, binding = 70) uniform VirtualRasterParams
     //         backend #ifdef, so this file stays source-identical.
     float u_DepthScale;
     float u_DepthBias;
-    uint  _vrPad0;          // 28 — std140 rounds the block to a 16-byte multiple.
+    // 28 — +1 on GL, -1 on Vulkan, from RHI::WindowSpaceFrontFaceSign(). This
+    //      shader builds its own window-space area determinant, and on Vulkan
+    //      that space is the FRAMEBUFFER's (y down) instead of GL's window space
+    //      (y up) — one mirror, which negates the sign. Fixed function gets a
+    //      second, cancelling inversion; a manual rasterizer does not.
+    float u_FrontFaceSign;
 };
 
 // Bounded micro-triangle raster: SW-routed clusters have small projected
@@ -211,7 +216,30 @@ void RasterizeTriangle(uint swRecordIndex, uint triangleIndex)
     // geometry we ship — is alpha-masked and never reaches this shader.
     bool twoSided = (inst.Flags & 4u) != 0u;
     float signedArea = OloVirtualSignedArea2(s0, s1, s2);
-    if (signedArea < 0.0 && !twoSided)
+    // The CULL sense is backend-dependent (u_FrontFaceSign); the ORIENT below is
+    // not. On Vulkan this determinant is computed in framebuffer space, whose y
+    // runs opposite to GL's window space, so the same triangle reports the
+    // opposite sign — and culling on the GL sense there throws away the FRONT
+    // faces and keeps the back ones. That does not blank the frame: a closed
+    // mesh still fills its own silhouette, so it reads as missing patches and
+    // inside-out shading (issue #1105).
+    // A SELECT rather than `signedArea * u_FrontFaceSign < 0.0`: multiplying the
+    // determinant is algebraically a no-op on GL (the sign is +1 there) but not
+    // a numerical one, and the select keeps the GL comparison spelled exactly as
+    // it was. The branch is uniform across the dispatch, so it costs no
+    // divergence.
+    //
+    // MEASURED, because "no visual change" was claimed and then checked: reading
+    // this uniform at all shifts the GL path's floating-point codegen slightly.
+    // Against the pre-#1105 shader, 25-126 pixels per 230,400 (0.01-0.06%) move
+    // by a MAXIMUM CHANNEL DELTA OF 1 -- one least-significant bit on
+    // micro-triangle boundaries, no geometry gained or lost, every golden test
+    // still green. Both the multiply and the select forms shift it identically,
+    // so the shift comes from the uniform read, not from the arithmetic; a
+    // compile-time #define per backend would avoid it, at the cost of the
+    // source-identical property this seam exists to preserve.
+    bool backFacing = (u_FrontFaceSign > 0.0) ? (signedArea < 0.0) : (signedArea > 0.0);
+    if (backFacing && !twoSided)
         return; // back-facing, single-sided material
     float orient = (signedArea < 0.0) ? -1.0 : 1.0;
     float area = signedArea * orient;
--- a/OloEditor/assets/shaders/compute/VirtualDebugColorize.comp
+++ b/OloEditor/assets/shaders/compute/VirtualDebugColorize.comp
@@ -42,9 +42,15 @@ layout(std140, binding = 70) uniform VirtualRasterParams
     //      append, not the counter), so a GPU overflow would inflate the very
     //      number the read is bounded by. 0 = an empty list.
     uint  u_SwListCapacity;
-    uint  _vrPad0;          // 20 — std140 rounds the block to a 16-byte multiple;
-    uint  _vrPad1;          // 24   explicit so the C++ mirror is a plain struct.
-    uint  _vrPad2;          // 28
+    // 20/24 — ndc.z -> window depth as (scale, bias), from
+    //         RHI::NdcToWindowDepthScaleBias(). The software raster does its own
+    //         perspective divide, so it applies the mapping fixed function would
+    //         have: GL 0.5/0.5 (ndc z is [-1,1]); Vulkan 1/0 (the seam already
+    //         mapped clip z to [0,w], so ndc.z IS the window depth). Data, not a
+    //         backend #ifdef, so this file stays source-identical.
+    float u_DepthScale;
+    float u_DepthBias;
+    uint  _vrPad0;          // 28 — std140 rounds the block to a 16-byte multiple.
 };
 
 // blue (low) -> green (mid) -> red (high)
--- a/OloEditor/assets/shaders/compute/VirtualDebugColorize.comp
+++ b/OloEditor/assets/shaders/compute/VirtualDebugColorize.comp
@@ -50,7 +50,12 @@ layout(std140, binding = 70) uniform VirtualRasterParams
     //         backend #ifdef, so this file stays source-identical.
     float u_DepthScale;
     float u_DepthBias;
-    uint  _vrPad0;          // 28 — std140 rounds the block to a 16-byte multiple.
+    // 28 — +1 on GL, -1 on Vulkan, from RHI::WindowSpaceFrontFaceSign(). This
+    //      shader builds its own window-space area determinant, and on Vulkan
+    //      that space is the FRAMEBUFFER's (y down) instead of GL's window space
+    //      (y up) — one mirror, which negates the sign. Fixed function gets a
+    //      second, cancelling inversion; a manual rasterizer does not.
+    float u_FrontFaceSign;
 };
 
 // blue (low) -> green (mid) -> red (high)
--- a/OloEditor/assets/shaders/compute/VirtualRasterArgs.comp
+++ b/OloEditor/assets/shaders/compute/VirtualRasterArgs.comp
@@ -58,9 +58,9 @@ layout(std140, binding = 70) uniform VirtualRasterParams
     uint  u_Phase;
     float u_OverdrawScale;
     uint  u_SwListCapacity;
+    float u_DepthScale;
+    float u_DepthBias;
     uint  _vrPad0;
-    uint  _vrPad1;
-    uint  _vrPad2;
 };
 
 void main()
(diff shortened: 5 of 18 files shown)


drsnuggles8:
Second review pass, on the commit the first pass predated (the per-backend depth fix). Two findings, both fixed in `c39388e4c`:

- **`VirtualRasterArgs.comp` was not in `VulkanPassSuiteTest::kMigratedComputeShaders`** — nothing compiled the new production kernel on Vulkan. The consequence is worse than a coverage gap: the pass treats a failed args kernel as a demotion back to the conservative CPU-side bound, which is correct but ~300-900x overdispatched — the exact cost this PR removes. Uncovered, a Vulkan compile failure would have spent that behind one warn with every test green.
- **The binding table still documented SSBO 40 as `{ uint Count; pad[3]; ... }`** — those three words are now the indirect dispatch arguments read at byte offset 4, an offset that is `static_assert`ed because it is a contract. Documented as padding, it invites the reuse that silently drops or multiplies raster workgroups.

60 tests pass across `VulkanPassSuite*` plus the new dispatch-args contract and L2 probe.

**Process note against myself:** `/code-review` ran once, before the depth-fix commit existed, so that commit went in unreviewed. The task loop says to re-review after review fixes and I did not. Both findings above are what that gap cost, and the first was material.

### Known-outstanding on this PR

| | Status |
|---|---|
| #1105 — residual Vulkan software-raster defect | filed with a five-suspect bisect, **not fixed**; Vulkan VG is improved but still not correct |
| `clusterid` debug overlay displaced on Vulkan | noted inside #1105, **not investigated or filed separately** — plausibly a distinct third bug |
| Perf numbers | **GL only**. The bracket never ran on Vulkan, so #1048's benefit is unverified there |
| `VulkanDrawPath.EngineHeapServesShaderReachableSlotsAndPoisonsFreedOnes` | pre-existing #1087, proven by A/B, owned by another worktree |
| CodeRabbit | not yet run |

Also worth recording: `HANDOVER.md` is **not** actually gitignored despite the handover claiming it is (`git check-ignore` returns nothing). Harmless while nobody runs `git add -A`, which the task-loop doc already warns against.

drsnuggles8:
**Correction to my "known-outstanding" comment above — #1105 is now fixed on this branch, in `6c11c9b94`.** That table is stale; the PR body has been rewritten.

Root cause was a second GL-convention assumption in the same shader, not the one I had been chasing. `VirtualClusterRaster.comp` culls back faces on the sign of a window-space area determinant it computes itself. Fixed function needs no help — Vulkan evaluates facing in framebuffer coordinates (y down), which composes with the seam's clip-y negation to identity — but a shader that builds the determinant itself gets only **one** of those two inversions. So the cull was discarding the FRONT faces and keeping the back ones.

| | Covered px (raster debug target, fixed camera) |
|---|---|
| GL (reference) | 101,293 |
| Vulkan, before | 84,765 (83.7%) |
| Vulkan, after | **101,292** |

**Vulkan virtual geometry now matches GL.**

Two things worth recording against myself:

- **My earlier "winding ruled out" was wrong.** I negated `signedArea` and coverage collapsed to 27,588, which I read as exoneration. Negating the value also flips `orient`, inverting the edge functions so triangles stop covering anything. The cull sense and the winding that orients the edge functions have to move independently. The correct probe — flipping only the comparison — was the fix.
- **Eyeballing could not have found this.** A closed mesh looks broadly similar whether you draw its front or its back faces. Only counting covered pixels against a GL reference separated "geometry missing" from "wrong faces kept", and only the neighbour analysis (68% of missing pixels had no covered neighbour) ruled out edge erosion.

**GL is not bit-identical and I am not claiming it is:** reading the new uniform shifts GL's float codegen by 25–126 px per 230,400 (0.01–0.06%) at a max channel delta of **1**. No geometry gained or lost, 143 GL virtual-geometry tests green. The shader comment carries these numbers.

Still genuinely outstanding: CodeRabbit, and the GL-only perf caveat (the bracket never ran on Vulkan). The displaced `clusterid` overlay I flagged as a possible third bug was **not** one — at a matched camera and viewport the Vulkan debug target is correctly positioned; the earlier displacement was the composite, and it went away with these fixes.

drsnuggles8:
## Vulkan perf measured — the GL-only caveat is closed

The GoogleTest harness is GL-only (`RenderPropertyTest`'s `GpuContext` builds a GL 4.6 core context; its `GlBackend` selects GLFW-vs-EGL, not GL-vs-Vulkan), so this used the live editor's `olo_perf_pass_timings`, which publishes the **same** `VirtualGeometryPass/SwRaster` and `/Resolve` sub-passes the GL test reads — `GPUPassTimerPool` stamps `vkCmdWriteTimestamp` on this backend and reports nanoseconds either way.

`VirtualGeometryStress`, Deferred, `swRasterMode=forcesoftware`, fixed camera and viewport. Three interleaved rounds, min of 24 samples after 6 warmup. Each arm needs its own editor launch (the arm is read at Init), and **each run re-reads `OloEngine.log` to confirm which path actually ran** rather than trusting the file toggle — that is the `indirect=` column.

| Arm | | SwRaster min (ms) | `indirect=` |
|---|---|---|---|
| A | CPU bound (master) | 1.4180 | 0 |
| B | **indirect (#1048)** | **0.0280** | 1 |
| C | indirect + empty `main()` | 0.0100 | 1 |
| D | CPU bound + empty `main()` | 1.6550 | 0 |

**Arm A → B: 1.4180 → 0.0280 ms, a ~50x reduction.**

The same two structural checks hold as on GL, which is what makes the number trustworthy:

- **D ≈ A** (1.6550 vs 1.4180) — under the CPU bound, a shader whose `main()` returns immediately costs what the real rasterizer costs. #1048's premise reproduces on Vulkan too.
- **C separates from A** (0.0100 vs 1.4180) — the bracket now measures rasterization rather than workgroup launch.

### Do not compare this to the GL table

The two are **not** the same experiment and the −98% here is not a better version of GL's −41.6%:

| | GL | Vulkan |
|---|---|---|
| Scene | `VirtualGeometryPerf` fixture | `VirtualGeometryStress` |
| Routing | `ForceSoftware`, 338 SW clusters | `forcesoftware`, 2,834,448 clusters |
| Instrument | GoogleTest harness | live editor MCP |

Vulkan's ratio is larger because its corpus is ~8,400x more clusters against the same live work-list — exactly the regime the conservative bound is worst in. Each table is internally valid as an A/B; across the two, only the *direction* and the D≈A / C≠A structure carry over.

The arm A/D readings are also visibly noisier (1.418–1.947) than B/C (0.028–0.029), which is consistent with a multi-million-workgroup dispatch contending for the GPU rather than with instrument drift.

Filed #1107 (scored, `robustness`) to put this on a proper footing: `VulkanPassSuiteTest` already runs real passes through the real render graph on Vulkan, and the first blocker its comment gives for not running the full `VirtualGeometryPass` — "~15 bare uniforms through `ComputeShader::Set*`, no-ops on the SPIR-V route" — is **stale**, since the UBO migration landed with #691 and this PR extends those very blocks.

drsnuggles8:
@coderabbitai review

Issue #1111, opened 2026-09-07
Hosted TSan is killed at the Functional physics block every run since 2026-09-06 — VM shutdown at TerrainCollision/Cloth/PhysicsJoint3D, no TSan output


drsnuggles8:
`TSan (Linux/Clang)` on the **hosted** arm has not completed a run since the 2026-09-06 nightly. It is not a TSan finding: the runner VM is killed while the Functional physics tests are in flight, and no TSan report is ever printed.

## Evidence — three runs, one position

| run | shape | killed with these in flight | exit |
|---|---|---|---|
| [34108719319](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34108719319) (09:55) | monolithic, `--parallel 2` | `TilemapCollidersStopABodyTest.*`, `TerrainCollisionFlatTest.DynamicBodyRestsOnFlatTerrainSurface` — entries 7061–7063 of 7611 | 143, `The runner has received a shutdown signal` |
| [34145979116](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34145979116) (17:03) shard 1/4 | `-I 1,,4`, `--parallel 2` | `FreeFallLandingTest`, `TerrainCollisionFlatTest` — 7085, 7089 | 143, same |
| same, shard 2/4 | | `TerrainStreamingTileCollisionTest`, `ClothFallsOntoFloorTest` — 7094, 7098 | 143 |
| same, shard 3/4 | | `TerrainStreamingTileRestTest`, `ClothHangsFromPinnedEdgeTest` — 7095, 7099 | 143 |
| same, shard 4/4 | | `ClothWindTest`, `PhysicsJoint3DTest` — 7100, 7104 | 143 |

The four shards died within 3½ minutes of each other (18:23:58 → 18:27:20) — the stride interleaves them, so all four reach the block together. Every kill is `The runner has received a shutdown signal … Process completed with exit code 143` with **nothing from TSan before it**: no `WARNING: ThreadSanitizer`, no `SUMMARY`. That is the VM being taken down, and the usual reason on a 16 GB hosted runner is memory.

The block is `OloEngine/tests/Functional/AnimationPhysics/` — `TerrainCollisionTest.cpp`, `TerrainStreamingSculptCollisionTest.cpp`, `ClothSimulationTest.cpp`, `PhysicsJoint3DTest.cpp` — i.e. Jolt with its job-system thread pool, under TSan's per-thread shadow, two processes at once.

[...]

Issue #1015, opened 2026-09-02
CI: make the self-hosted Linux box carry PR work (pinned toolchain, GPU-under-sanitizer baseline, Windows runner decision)


drsnuggles8:
## Problem

#1009 was opened to cut PR wall-clock (Windows median 159 min, Sanitizers 154 min over 250 runs) by moving the Linux jobs onto the self-hosted box. #1010 wired the routing (`vars.OLO_LINUX_SELF_HOSTED`, runners `olo-ci-1` / `olo-ci-2`, fork boundary on `head.repo.full_name`), but on an ordinary PR the box does nothing:

- The three Linux sanitizer jobs were routed, measured, and sent back hosted: 215 failed of 6,909 on the box against 1 of 6,940 hosted on the same commit (run 33561256256, 212 of the failures SIGTRAP). Two causes: the box runs clang 21 against gcc-toolset-15's libstdc++ instead of the pinned clang-19, so UBSan reports inside the standard library; and the box has a GPU, so about 200 GL-gated tests that skip on hosted runners ran under a sanitizer for the first time.
- Vulkan OFF stays routed (1h36m hosted, 28m on the box) but only triggers on PRs that touch its path filter.
- The Steam stub full build stays routed but runs only on master pushes and the schedule.

Verified on PR #1011, head 2020c1e78: every executed job was GitHub-hosted; the one job resolved to `olo-ci` was skipped. Both `olo-ci` runners were online and idle.

## Proposed work

**A. Toolchain parity on the box.** Give the sanitizer jobs the same pinned clang-19 on the box as on hosted (the `setup-linux-build` action already owns the hosted install; it must select the pinned toolchain on `self-hosted` too). Re-run UBSan on one commit and confirm the `hashtable_policy.h` / `stl_vector.h` reports are gone.

**B. Stage 1: route the sanitizers with hosted parity.** Route ASan+LSan, UBSan and TSan to the box with the GL-context tests forced to skip, so a run means exactly what the hosted run means. Measure against the 154 min hosted median. This is the PR speed-up #1009 promised, and it is separable from the GPU coverage question.

**C. Stage 2: GPU-under-sanitizer baseline, nightly, on the box.** A separate scheduled job runs the same three sanitizers with the GL-gated tests enabled. Triage the 215 failures into engine bugs and instrumentation artifacts, one issue per real finding. Only when that baseline is green does it become a PR gate.

[...]

PR #1115, opened 2026-09-08
ops(ci-host): a running job holds a shutdown inhibitor; runner cgroups get a memory ceiling

drsnuggles8:
Follow-up to #1114. That PR fixed the one runner death that was ours (the `addr2line` fallback); this closes the two that were the host's, both from 2026-09-07 and both previously unguarded:

| when (UTC) | what happened | guard now |
|---|---|---|
| 06:54 | box rebooted **from a logged-in session** under a running sanitizer job; `olo-ci-2` down for the day | a running job holds a `systemd-inhibit --mode=block` shutdown lock |
| 09:01 | `systemd-oomd` killed **both** CI runner cgroups 16 s apart under memory pressure — two sanitizer jobs + the GPU nightly's arm on 31 GiB — surfacing as *"runner lost communication"*, zero steps | `MemoryMax=14G` + `OOMPolicy=continue` per runner unit |

## Inhibitor — `scripts/ci-host/job-started.sh`, `job-completed.sh`

Runner hooks (`ACTIONS_RUNNER_HOOK_JOB_STARTED` / `_COMPLETED` in each runner's `.env`) take and release a block inhibitor per job. `systemctl reboot` / `poweroff`, `shutdown`, and Cockpit then refuse — or need `-i` — while a job runs. A service user has no active seat, so a polkit rule grants `gh-runner-olo` `inhibit-block-shutdown`.

A hook **never fails a job**: polkit denial, no `systemd-inhibit`, and a stale lock from a crashed job are all handled and logged. All five paths are exercised against a stubbed `systemd-inhibit` (normal, release, stale-release, denied, absent).

## Ceiling — `/etc/systemd/user/<runner>.service.d/olo-memory.conf`

`MemoryMax=14G`, `OOMPolicy=continue`, `ManagedOOMPreference=avoid`, applied live to running units via `set-property --runtime` and **verified from the cgroup's own `memory.max`**, not from `systemctl show`. The kernel OOM-kills the largest process *inside* the runner's cgroup at the ceiling — a compiler or a test — and the service stays up; the job fails visibly at the step that ran out.

Deliberately **not `MemoryHigh`**: throttling forces reclaim, reclaim is the pressure oomd measures, so a `MemoryHigh` ceiling makes the oomd kill *more* likely. And `OOMPolicy=continue` matters as much as the limit: the default `stop` would take the runner down on every in-job OOM, which is the failure being fixed.

14G is a first setting from `asan.yml`'s own arithmetic (a 4-wide instrumented build ≈ 12 GB; the symbolizer ≈ 0.9 GB after #1114). Two CI runners plus the GPU runner at 14G is 42 GB against 31 GiB — three peaks at once still don't fit, and under that overlap the ceiling turns a runner kill into a red job. The script's report prints each runner cgroup's **`memory.peak`**, which is the number to revise the default from. `OLO_RUNNER_MEMORY_MAX=` overrides.

## What is and isn't verified

- Verified here: `bash -n` on all three scripts; the five hook paths against a stub; the one real bug found in the process (an invalid array expansion in the report) fixed and re-run.
- **Not exercised on the box** — there is no sudo from CI by design. To apply and confirm:
  ```
  sudo bash scripts/setup-olo-ci-host.sh          # from a checkout; idempotent
  systemd-inhibit --list                          # during any job: names the runner + run id
  ```
  and the report's last lines show `memory.max` ≠ `max` and a `peak=` per runner unit.

## Checked and left alone

The GPU sanitizer nightly's routing to `olo-ci` is by design and already `max-parallel: 1`, so it holds one PR slot at a time; the `asan.yml` comment about the nightly "owning its runner" refers to the *conformance* nightly on `gpu-amd`. The 03:15 UTC `olo-ci-1` drop today is still undiagnosed (05:15 CEST fits a scheduled reboot; the journal will say) — the inhibitor covers it if it was a reboot, the ceiling if it was oomd.

Docs: `docs/ops/self-hosted-host-hygiene.md` §5–6. Refs #1111, #1015.

### DIFF
--- a/docs/ops/self-hosted-host-hygiene.md
+++ b/docs/ops/self-hosted-host-hygiene.md
@@ -101,6 +101,47 @@ The two missing-piece outcomes are NOT the same, which matters when you read a r
   first sanitizer link (`cannot find libclang_rt.asan.a`), so nothing is silently skipped —
   the warning just tells you why, several minutes earlier than the linker would.
 
+## 5. A running job holds a shutdown inhibitor
+
+**Rule: while a job executes on any runner, the host refuses to reboot; override it knowingly
+with `-i`, never by accident.** Section 1 guards one path -- the update timer. On 2026-09-07 at
+06:54 UTC the box was rebooted from a logged-in session under a running sanitizer job, and
+nothing stood in the way; `olo-ci-2` stayed down for the rest of the day.
+
+Two runner hooks, [`scripts/ci-host/job-started.sh`](../../scripts/ci-host/job-started.sh) and
+`job-completed.sh`, take and release a `systemd-inhibit --what=shutdown --mode=block` lock per
+job. `systemctl reboot` / `poweroff` and `shutdown` then refuse with *"Operation inhibited by
+actions-runner olo-ci-1 ..."* and Cockpit shows the same; `systemctl reboot -i` overrides. The
+runner user has no active seat, so a polkit rule grants it `inhibit-block-shutdown`;
+`setup-olo-ci-host.sh` installs the hooks to `/usr/local/lib/olo-ci/`, the rule, and the two
+`ACTIONS_RUNNER_HOOK_JOB_*` lines in each runner's `.env`. A hook never fails a job: a lock it
+cannot take is logged in the job's output and the job runs without it.
+
+Check it during any job: `systemd-inhibit --list` names the runner and the run id. A stale lock
+(runner crashed between hooks) is released by the next job's started hook.
+
+## 6. Runner memory ceilings: a job that runs out fails the job, not the runner
+
+**Rule: every runner unit carries `MemoryMax=` and `OOMPolicy=continue`; a runner cgroup at
+`memory.max=max` is a bug, not a default.** On 2026-09-07 at 09:01 UTC `systemd-oomd` killed
+both CI runner cgroups 16 s apart -- two sanitizer jobs plus the GPU nightly's arm on 31 GiB.
+A whole cgroup goes at once, listener included, and the jobs surface on GitHub as *"The
+self-hosted runner lost communication with the server"* with zero steps run.
+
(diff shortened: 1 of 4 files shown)


drsnuggles8:
Applied on the box 2026-09-08 ~09:15 CEST with #1116's fix. The report, verbatim:

```
inhibitor: 2 job lock(s) held
actions-runner-olo.service       max=15032385536 current=108MiB  peak=123MiB   oompolicy=continue
actions-runner-ci-1.service      max=15032385536 current=9996MiB peak=12698MiB oompolicy=continue
actions-runner-ci-2.service      max=15032385536 current=7922MiB peak=9723MiB  oompolicy=continue
```

So: the ceiling is live on all three units (`max` in bytes = 14 GiB), `OOMPolicy=continue` has taken, and **two jobs were holding shutdown inhibitors at that moment** — the hooks work.

**The measurement this PR said to revise 14G from**, `memory.peak` since the 06:55 restart, attributed to the only jobs that ran on each runner in the window (both PR #1112, both `--parallel 4`):

| runner | job | peak |
|---|---|---|
| `olo-ci-1` | ASan + LSan (Linux/Clang) | **12.7 GB** |
| `olo-ci-2` | TSan (Linux/Clang), still running | **9.7 GB** |

ASan+LSan at 4-wide is the box's heaviest job class and sits 1.3 GB under the ceiling. I'd leave 14G: raising it to 16G puts two runners at 32 GB against 31 GiB, which is the oomd situation this PR exists to end. If an ASan job ever goes red with an OOM-killed `clang` at the ceiling, the lever is `--parallel 3` on the box for that job, not the cap.

Also worth recording: the first run's `set-property` failure on `OOMPolicy` (#1116) confirmed that `systemctl show -p OOMPolicy` reports the *loaded* unit config after `daemon-reload`, so the report can show `continue` for a unit whose running instance was started before the drop-in existed. Restarting each runner once when it is idle makes the two agree; until then the report line is the loaded value.

Issue #786, opened 2026-08-12
Physics: destructible structural connectivity / progressive collapse


drsnuggles8:
**Is your feature request related to a problem? Please describe.**
Follow-up to #459 (runtime destructible objects + debris, shipped in #785). That slice deliberately implemented per-object shattering only. The issue body listed **structural connectivity / progressive collapse** — a structure losing integrity as neighbouring pieces break, so a wall crumbles progressively rather than each block shattering in isolation — as *optional*, and [ADR 0013](docs/adr/0013-destructible-debris-asset-swap-not-runtime-fracture.md) recorded it as out of scope. This issue tracks it as its own scored feature so #459 isn't left open with a partial comment.

**Describe the solution you'd like**
A connectivity model over a set of destructible pieces:
- Author/derive a connectivity graph between adjacent destructible bodies (which pieces support which).
- When a piece breaks, propagate loss of support through the graph; pieces that become unsupported (no path to an anchor/ground) detach and fall — feeding the existing debris/despawn path from #459.
- A stability solver (even a cheap support-flood, not a full FEM) deciding what stays standing vs. collapses.
- Compose with the existing `DestructibleComponent` break trigger and the `Debris` layer / budget already in place.

**Describe alternatives you've considered**
- Pure per-object shattering (what #459 already ships) — no propagation.
- Pre-authored collapse animations swapped in on trigger — cheaper but not emergent.
- Full soft-body / constraint-network simulation — far heavier than the acceptance value.

**Additional context**
Builds directly on the #459 seams: the `Destructible` scheduler node, `DestructibleSystem`, the `DEBRIS` Jolt layer, and the live-debris budget. Design guidance in `docs/agent-rules/destructible-debris.md`. This is a genuinely separate system (a connectivity graph + a solver), which is why it was split out rather than bolted onto the first slice.

## Score
```olo-score
capability: 3
craft: 8
stability: 1
decay: 1
effort: 8
confidence: 0.5
learning: 5
fun: 5
kano: delighter
blocked_by: []
blocks: []
```

Issue #763, opened 2026-08-08
Static-init audit: 200 distinct -Wglobal-constructors/-Wexit-time-destructors sites in our own code (clangcl spike)


drsnuggles8:
## What

A C++Now 2026 audit spike (talk: *C++ Magic Statics Demystified*, Microsoft Edge team) — add `-Wglobal-constructors` / `-Wexit-time-destructors` to the `clangcl` preset only (warnings, not `-Werror`; never on the msvc/cl.exe path) and see what turns up. These flags catch the same class of bug that has already cost us twice: Tracy's rpmalloc static-init race (forced `TRACY_ENABLE=OFF` under TSan) and the entt `type_index` counter needing `ENTT_USE_ATOMIC` as a **PUBLIC** definition — both caught only after the fact by a sanitizer, not at compile time.

## Numbers

Building `OloEngine-Tests` with the `clangcl` preset (Release config):

| | raw warning count | distinct source locations |
|---|---|---|
| `-Wglobal-constructors` | 868 total (191 vendor / **677 our code**) | **118** in our code |
| `-Wexit-time-destructors` | 2319 total (385 vendor / **1932 our code**) | **191** in our code |
| **union (either warning)** | — | **200 distinct lines** in our own code |

The raw per-warning counts are heavily inflated by `inline static` class members (e.g. `Project::s_ActiveProject` in `OloEngine/Project/Project.h:138-139`) — clang re-emits the warning once per translation unit that instantiates the inline definition, so a single problematic declaration can show up 100+ times in the log. **200 distinct locations** is the real number to act on, not the raw ~2600.

Vendor code (`OloEngine/vendor/clang/...`, mostly GameNetworkingSockets) accounts for the rest and is out of scope here.

## Where it's concentrated (by subsystem, distinct locations)

Top offenders by warning-line count in a single file:
- `Renderer/MaterialAsset.cpp` — 8 (static `std::string` shader-uniform-name constants, see below)
- `Networking/Core/NetworkManager.cpp` — 7 (static class members: `Scope<T>`, driver/session/lobby singletons)
- `Core/InputActionManager.h` — 5
- `Renderer/Preview/AssetPreviewRenderer.cpp` — 5
- `Core/GamepadManager.h` — 4
- `Debug/CrashReporter.cpp` — 4
- `Task/Scheduler.cpp` — 4

[...]

Issue #673, opened 2026-07-20
MCP: roadmap — gaps vs Unreal Engine 6.0's MCP/AI-toolset stack


drsnuggles8:
Roadmap tracking issue for OloEngine's MCP server (`OloEditor/src/MCP`), derived from a source-grounded comparison against **Unreal Engine 6.0's** MCP/AI-toolset stack (`Engine/Plugins/Experimental/{ModelContextProtocol,ToolsetRegistry,Toolsets,MCPClientToolset,AIAssistant,UnrealConsoleRuntime}`, read from a shallow clone of `ue6-main`). It is a roadmap, not a committed sprint — items are prioritized by value-for-effort **in OloEngine's context**; some UE patterns are deliberately out of scope.

## Where we stand

Our MCP server optimizes for **trust**: every write goes through a session-scoped consent gate (Disabled / Prompt / Allow-session), the Lua script-tool runtime is sandboxed with memory + time quotas and no engine-object reflection, and several tools are purpose-built root-cause *explainers* (`olo_render_why_not_visible`, `olo_physics_why_no_collision`) rather than raw state dumps. UE's stack optimizes for **scale**: five plugins, 34 domain-toolset sub-plugins, reflection-driven build-time tool registration, and OAuth-federated import of third-party MCP servers — but no built-in write-consent gate at all.

**Already at or ahead of parity, no action needed:** write-consent gating, sandboxed script-tool tier with non-inheritable authority, root-cause explainer tools, live tool hot-reload with atomic snapshot swap, per-call progress/cancellation.

### Tier 1 — high value, straightforward, do first — ✅ shipped 2026-07-24 (branch `feature/mcp-ue6-parity-tier1-673`)

- [x] **Outbound MCP client (mini `MCPClientToolset`).** OloEngine only *serves* tools today; it can't consume them. UE's client toolset connects out to an external MCP server and folds its tools into the same registry the native tools live in. A v1 doesn't need UE's full OAuth/SSE/streamable-HTTP matrix — stdio only (spawn a local process, speak newline-delimited JSON-RPC) is enough to compose in something like a filesystem or web-fetch MCP server. `McpServer::m_Tools` (copy-on-write atomic snapshot) is already the right shape for this — it's how script-tool live reload works today.
[...]

PR #1136, opened 2026-09-08
feat(mcp): deferred capability discovery + configurable exposure profiles

### LINKED ISSUE #1124: MCP: deferred capability discovery + configurable exposure profiles (tools/list is now ~66k tokens)
Epic B of the Automation-Control-Plane roadmap.

## This re-opens a decision, on measured evidence

#607 closed the lazy two-tier discovery idea as unnecessary, and #673 recorded why: `tools/list` was **≈15k tokens** for the surface at the time, with an explicit re-open condition —

> re-open only if the surface grows several-fold or a schema-eager client becomes a primary consumer.

**That condition is now met.** Measured against a live editor on `master` (2026-09-08, `tools/list` over the real MCP endpoint):

| | then (#673) | now |
|---|---|---|
| tools | 66 | **95** |
| `tools/list` payload | — | **265 009 bytes** |
| approx tokens | ≈15 000 | **≈66 000** |

Every agent session pays ~66k tokens before it asks a single question — roughly a third of a 200k context window, spent on a catalogue of which the session typically calls three or four entries. The growth is not a one-off: the surface has gained ~29 tools since that measurement and each new tool adds ~2.8 KB.

This is why the issue is scored with high **decay** and low capability: it unlocks nothing new, it stops an existing cost compounding.

## Scope

- **Exposure profiles** — `core` / `toolset` / `full`. Which commands a session sees becomes configuration, not a constant.
- A discovery gateway (`capability` / `search` / `describe` / `execute`) so a session can find the three tools it needs without materialising all 95 schemas.
- Registry-size metrics so the number above is tracked rather than rediscovered in eighteen months.
- Compatibility behaviour for clients that only speak plain `tools/list`.

## Explicitly preserved

- **Do not delete the standard MCP tools.** A client that lists and calls normally must keep working; `full` stays available and the default must not silently break existing sessions.
- `tools/search` and toolset `_meta` tagging already shipped in #385 — this builds on them rather than replacing them.

## Acceptance

- `tools/list` under the default profile is a small fraction of the 265 KB above, with the number asserted in a test so it cannot silently regress.
- A session using only the gateway can discover and call any command reachable under `full`.
- An unmodified MCP client that ignores the gateway still works.

## Not blocked by #1123

Deliberately: this changes *exposure*, not the command set, so it can be built on `McpServer::m_Tools` as it stands. Given it is the highest-decay item in the set, blocking it behind an effort-8 refactor would be the wrong trade. It should be revisited as an adapter concern once #1123 lands.

## Refs

Roadmap §8, §21, §26 Phase 2. Reverses the "not needed" half of #607; predecessor #673 (closed). Related transport work: #777.

## Score
```olo-score
capability: 3
craft: 1
stability: 1
decay: 8
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
`tools/list` had grown to **96 tools / 269 436 bytes / ~67k tokens** — roughly a third of a
200k context window spent before a session asks its first question, on a catalogue of which
three or four entries are ever called. #673 closed the two-tier discovery idea as unnecessary
at 66 tools / ~15k tokens, with an explicit re-open condition ("the surface grows
several-fold"). At ~2.8 KB per new tool that condition is met, and the growth is structural.

## What changed

**Exposure profiles** — which tools `tools/list` advertises becomes configuration:

| Profile | Lists | Set with |
|---|---|---|
| `core` | **default.** 12 curated tools + the gateway | — |
| `toolset` | core + the enabled toolsets | `OLO_MCP_TOOLSETS=render,physics` |
| `full` | everything (previous behaviour) | `OLO_MCP_TOOL_PROFILE=full` |

Also a session-scoped radio group in the MCP panel (switching fires
`notifications/tools/list_changed`, so a connected agent sees the wider listing without
reconnecting) and `-ToolProfile` / `-McpToolsets` on the `run-oloengine` driver.

**The discovery gateway** — `olo_capability` / `olo_tool_search` / `olo_tool_describe` /
`olo_tool_execute`, listed under every profile. Search returns names and summaries, **not**
input schemas: that omission is the entire saving. `olo_tool_execute` is a *rewrite* into a
plain `tools/call` performed before dispatch, not a handler that invokes the target — so the
target passes the same schema validation, write-consent gate and cancellation scope as a
direct call, rather than a second implementation of all three that could drift.

**Registry metrics** — `ComputeRegistryMetrics()` serialises the catalogue with the same
builder `tools/list` uses, so the byte count *is* the payload rather than a model of it.
Logged at server start, surfaced by `olo_capability`, shown in the panel, printed by the test
on every run, and asserted as a ceiling.

### Backward compatibility

Exposure filters the **listing**, never dispatch. `tools/call` resolves against the full
registry under every profile; `tools/search` stays deliberately profile-blind; user-configured
tools (project Lua, bridged `ext.*`) are listed under every profile. Nothing is unregistered.

## Two pre-existing build breaks, fixed in their own commits

Neither is related to this issue; both blocked the work.

- `c3a13c473` — `constinit std::vector<std::string>` in `DebugLevers.cpp` (from #763). clang-cl
  rejects it under the MSVC debug STL (`_Container_proxy` heap allocation isn't a constant
  expression); MSVC's front end accepts it, so **`OloEngine` did not build in the default
  `dev-cached` preset**.
- `4e1bf8548` — an unterminated string literal in `SceneHierarchyPanel.cpp` (from #786), so
  **the `OloEditor` target did not compile at all**.

**These point at a CI coverage gap worth a look**: `OloEngine-Tests` and `OloEditor` are
separate link targets, and CI evidently gates on the former. The test binary never compiles
`SceneHierarchyPanel.cpp`, and the ASan tree is Release-only so it never sees the debug-STL
`constinit` problem. Two broken targets reached master with CI green.

## Review guide

**Where I'd look hardest**
1. `OloEditor/src/MCP/McpServer.cpp` `HandleToolsCall` — the `olo_tool_execute` rewrite
   re-enters `HandleToolsCall` once. Recursion is bounded because the rewrite rejects an inner
   name equal to the gateway's own; if you think that guard is reachable around, say so.
2. `OloEditor/src/MCP/McpExposure.h` `CoreToolNames()` — the 12-tool core set is a judgement
   call. The criterion I used is "a session that cannot see it could not work out what to ask
   next". Disagree with the membership, not just the size.
3. `OloEditor/src/MCP/McpToolsGateway.cpp` `Summarize()` — truncates descriptions, backing off
   to a UTF-8 lead byte. My own review caught that splitting a codepoint makes `dump()` throw
   `type_error.316`, which would turn a matching search into "Tool failed".

**What I verified, and how**
- `McpExposureProfileTest` (21 new tests): profile arithmetic, every `CoreToolNames()` entry
  resolves against the real registry, the byte ceiling, and the gateway round-trip including
  the consent gate and nesting rejection.
- Full suite: **7724 tests, 7683 passed, 41 skipped, 0 failures**.
- **Live editor over the real MCP endpoint** (not just headless): `core` reported 18 of 99
  tools / 39 362 bytes; `full` reported 99 of 99 / 271 910 bytes and a plain `tools/list`
  returned 99 entries. The `full` figure reproduces the issue's master baseline —
  271 910 − 265 009 = 4 427 bytes, exactly the four gateway tools.
- **The compatibility claim is demonstrated, not asserted**: the MCP smoke test calls
  `olo_memory_report`, `olo_shader_get`, `olo_assets_list`, `olo_crash_list`,
  `olo_camera_orbit`, `olo_viewport_set_size` and `olo_perf_snapshot` *by name* — none of them
  listed under `core` — and every one worked.
- Panel UI captured (it is a floating ImGui viewport, so it needs a per-HWND capture; the
  driver's screenshot of the editor window cannot see it).

**Least confident about** — the core set membership, as above. Also `olo_tool_execute`'s
declared schema is deliberately *not* `.NoAdditional()`, because its arguments never reach
`ValidateArguments` (the rewrite happens first) and I would rather the schema under-claim than
advertise a strictness nothing enforces. A reviewer may prefer the opposite trade.

**Deliberately not tested** — clicking the panel's profile radio group live. The MCP panel is a
floating ImGui viewport and `olo_input_inject` is unreliable against those; the underlying
`SetExposurePolicy` path and its `list_changed` notification are covered by unit tests, and both
the env-var and panel-render paths were exercised live.

Since opening this PR I closed the other gap — **all three profiles are now verified against a
live editor**:

| Profile | Listed | Bytes |
|---|---|---|
| `core` | 18 of 99 | 39 362 |
| `toolset` (`physics,shader`) | 28 of 99 | 58 830 |
| `full` | 99 of 99 | 271 910 |

The `toolset` run also confirms the composition rule: all 7 `olo_physics_*` and all 4
`olo_shader_*` appear, but only the 2 *core* render tools out of 36 — core survives on top of the
selection rather than being replaced by it. 28 = 12 core + 4 gateway + 2 script + 7 + 4 − 1
overlap (`olo_shader_errors` is both core and `shader`). It also exercises the new
`-ToolProfile` / `-McpToolsets` driver flags end to end.

Closes #1124

## Summary by CodeRabbit

- **New Features**
  - Added MCP tool exposure profiles: Core, Toolset, and Full.
  - Added discovery tools for viewing capabilities, searching tools, describing tool details, and executing hidden tools by name.
  - Added MCP panel controls showing listed versus available tools, resources, prompts, and catalogue size estimates.
  - Tool-list changes now notify connected clients automatically.

- **Documentation**
  - Updated MCP setup, diagnostics, and tool-authoring guidance with profile-selection options and discovery workflows.

- **Bug Fixes**
  - Corrected a Destructible component tooltip line break.


drsnuggles8:
@coderabbitai review

drsnuggles8:
## 🤖 Self-review @ `1cad1e43d`

Reviewed the PR diff at **high** effort before opening, and re-verified after the fixes.

- **Findings: 5 · all fixed**
  - `McpToolsGateway.cpp` `Summarize()` — truncating at a fixed byte offset could split a multi-byte UTF-8 codepoint; `nlohmann`'s `dump()` **throws** `type_error.316` on invalid UTF-8, so one em-dash in a tool description would have turned every matching `olo_tool_search` into "Tool failed". Now backs off to a UTF-8 lead byte. (Latent — descriptions are ASCII today, which is exactly why it would have sat unnoticed.)
  - `IsListedUnder()` duplicated `ExposureFactsOf()`, so the gateway's per-hit `listed` flag could drift out of agreement with the `tools/list` filter with nothing failing. Promoted to `McpServer::ExposureFactsOf()`; single definition, both callers go through it.
  - `Handle_Capability` took two independent tool snapshots, so a concurrent script-tool reload could make `toolsets` and `registry`/`hiddenTools` describe different registries. Added a `ComputeRegistryMetrics(snapshot)` overload; one snapshot per report.
  - `OLO_MCP_TOOLSETS` silently overrode an explicit `OLO_MCP_TOOL_PROFILE=core`. Now only auto-upgrades when no profile was named, and warns when the toolsets are ignored.
  - `driver.ps1` had no way to widen the profile while `SKILL.md` still told agents to call non-core tools. Added `-ToolProfile` / `-McpToolsets` (both skill copies) and documented the discovery loop.

- **Dismissed:** none.

Also note for reviewers: `olo_tool_execute`'s input schema is deliberately **not** `.NoAdditional()`. Its arguments never reach `ValidateArguments` (`HandleToolsCall` rewrites the call first), so a closed object would advertise a strictness nothing enforces. Reasonable people may prefer the opposite trade — say so and I'll flip it.

CI note: CodeRabbit's first pass never ran (OSS rate limit, "review limit reached"), so the earlier zero-unresolved-threads count was **not** a clean review. Re-triggered above.

drsnuggles8:
## 🤖 Self-review @ `fa180b810` (update)

Re-reviewed the two commits added since the `1cad1e43d` summary above. Both are comment/formatting only, no behaviour change:

- `1cad1e43d` — hoisted a schema node into a local so clang-format has nothing to join (`.clang-format` sets `ColumnLimit: 0`, so it *joins* a hand-wrapped sub-expression rather than reflowing it). Emitted schema is identical; `pre-commit` now passes.
- `fa180b810` — two stale comments this work invalidated: the `tools/list` TTL justification enumerated one invalidation source when there are now two (the 5-minute TTL is only honest because `SetExposurePolicy` fires the same `tools/list_changed` push), and `ToolDef::Toolset` still described "the 39-tool surface" from #385 — it is 96 now, which is exactly what made exposure profiles necessary.

**Findings: 0 new.** Verified locally at this head: both targets build, 797 MCP tests pass, `pre-commit run --all-files` clean.

### On review coverage — please read before merging

**CodeRabbit has not reviewed this PR.** Its first pass hit the OSS rate limit and posted only a "review limit reached" notice; an explicit `@coderabbitai review` was declined with *"Review rate limited... does not re-review already reviewed commits."* So the unresolved-review-thread count of **0 means no review happened**, not that a review found nothing.

The only review this diff has had is my own high-effort self-review (5 findings, all fixed — see the summary above). Worth a human pass on `McpExposure.h` (the core-set membership) and the `olo_tool_execute` rewrite in `HandleToolsCall` before merge.

drsnuggles8:
@coderabbitai Good catch — confirmed and fixed in `01fbaf1db`.

You were right that this is a real mixed-policy window, and it's worth naming why I missed it: I'd already fixed the **snapshot** double-read in this same function and written the comment "ONE snapshot for the whole report", which was accurate and beside the point. The bug had two axes and I closed one.

Rather than pin the policy at the call site and rely on discipline, I made `ComputeRegistryMetrics` take the policy as a **parameter** instead of loading `m_ExposurePolicy` itself, so it can no longer read a second one. Both call sites — `Handle_Capability` and the MCP panel — now pass the policy they already read alongside the snapshot.

On the regression test you asked for: my first attempt asserted the arithmetic invariant (toolset buckets sum to `registry.listedTools`), and I then realised it **would have passed without the fix** — single-threaded, both reads return the same policy. So there are now two tests with different jobs, and the comments say which is which:

- `MetricsHonourThePassedPolicyRatherThanRereadingServerState` — the actual pin. Leaves the server on `core` while passing `full` in, and vice versa; a regression to reading server state fails deterministically.
- `CapabilityReportIsInternallyConsistentAcrossOneProfile` — the arithmetic ratchet, explicitly documented as *not* proving the concurrency fix. It still earns its place: it catches an uncategorized builtin tool escaping the toolset buckets.

Verified locally: both targets build, 798 MCP tests pass, `pre-commit run --all-files` clean.

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

PR #1142, opened 2026-09-09
fix(ci): declare a shader-toolchain floor and give hosted Linux one above it

### LINKED ISSUE #1139: CI/Shaders: GL_EXT_descriptor_heap does not compile on hosted Linux glslang — the Sanitizers nightly is red on all three Linux jobs
The `Sanitizers` nightly has been red on **all three Linux jobs** since 2026-09-08. The
failure is not a sanitizer finding — it is two deterministic shader-compile tests, and
the same two fail identically in the ASan+LSan, UBSan and TSan shard 2/4 of every run.

## What fails

Nightly [34300008455](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34300008455),
`ASan + LSan (Linux) tests 2/4` / `UBSan (Linux) tests 2/4` / `TSan (Linux) tests 2/4`,
`99% tests passed, 2 tests failed`:

| ctest entry | test |
|---|---|
| 1662 | `ShaderStageContract.FullGBufferWritersMatchTheProductionAttachmentNumericTypes` |
| 4158 | `ShaderCompilation.AllProductionShadersCompileUnderVulkanTarget` |

Byte-identical diagnostics in all three jobs:

```
OloEditor/assets/shaders/PBR_GBuffer.glsl:11: error: '#extension' : extension not supported: GL_EXT_descriptor_heap
OloEditor/assets/shaders/include/DescriptorHeapTextures.glsl:36: error: 'descriptor_heap' : unrecognized layout identifier, or qualifier requires assignment (e.g., binding = 4)
OloEditor/assets/shaders/include/DescriptorHeapTextures.glsl:36: error: 'descriptor_stride' : there is no such layout identifier for this stage taking an assigned value
OloEditor/assets/shaders/include/DescriptorHeapTextures.glsl:36: error: 'binding' : sampler/texture/image requires layout(binding=X)
```

`ShaderCompilation` reports the same for `GpuPathTracer.glsl` stage 1.

## Root cause: a toolchain-version split, and CI has two different toolchains

`DescriptorHeapTextures.glsl` requires `GL_EXT_descriptor_heap`, and its two consumers
declare `#extension GL_EXT_descriptor_heap : require` at the top:

- `OloEditor/assets/shaders/PBR_GBuffer.glsl:143` — added by `86935aa25`
  (*the five material-local maps index the descriptor heap on Vulkan*, PR #1135, #805).
- `OloEditor/assets/shaders/GpuPathTracer.glsl:75` — added by `b431552fe`
  (*textured hits and alpha masks for the GPU path tracer*, #1055, #805 scoped).

`GL_EXT_descriptor_heap` is very new. Whether it compiles depends entirely on which
glslang the job links:

| where | shader toolchain | result |
|---|---|---|
| dev box (Windows) | LunarG SDK 1.4.357.0 — `shaderc v2026.3`, `glslang 11.1.0-1493-g168d452a` | **compiles** (verified with a minimal `glslc` probe) |
| Windows CI | `.github/actions/setup-vulkan` → LunarG SDK | compiles |
| self-hosted Linux (Rocky) | `$VULKAN_SDK`, the only source of shaderc/glslang there (`setup-linux-build/action.yml:222-236`) | compiles |
| **hosted Linux** | **Ubuntu apt `glslang-dev` / `libshaderc-dev`** (`setup-linux-build/action.yml:165`), `vulkan-sdk: 'false'` on every sanitizer job | **fails** |

## Why it escaped PR CI

Same-repo PRs route the Linux sanitizer jobs to the self-hosted box, which has the LunarG
SDK. The hosted arm — the one with apt's older glslang — is only exercised by the nightly
and by `force_hosted` dispatches. So both introducing PRs were green on PR CI and the
break surfaced only on the next nightly.

This is a general hole, not a one-off: **any shader feature newer than Ubuntu's glslang
is invisible to PR CI and lands red on the nightly.**

## This is not only a CI problem

The same skew applies to anyone building the engine against a Vulkan SDK older than the
one on the dev box. Today the failure mode is a shader that will not compile, which at
least is loud; what it must not become is a silent fallback that quietly drops heap
indexing for the five material-local maps and the path tracer's textured hits.

## Decide the policy, then implement it

The options are not equivalent and the choice belongs in an ADR 0011 amendment note,
since amendments (95)/(96) are what put the extension there:

1. **Raise the floor** — give the hosted Linux jobs the LunarG SDK too. The action's own
   header (`setup-linux-build/action.yml:37-46`) records why the sanitizer jobs
   deliberately never ran `setup-vulkan`: `$VULKAN_SDK` won the CMake search and three
   targets linked the SDK's `spirv_cross` headers. Re-opening that needs care.
2. **Gate the extension** on a compile-time capability probe, with the non-heap path
   loud and countable rather than a silent fallback (`CLAUDE.md` → no silent fallbacks).
3. **Declare a minimum shader-toolchain version** and make the guard a first-class,
   named failure at configure time rather than 4,000 tests in.

Whatever is chosen, add a regression guard that would have caught this on a PR — the
cheapest being a hosted-toolchain shader-compile check that does not need the full
sanitizer build.

## Done when

- [ ] `ShaderStageContract.FullGBufferWritersMatchTheProductionAttachmentNumericTypes`
      and `ShaderCompilation.AllProductionShadersCompileUnderVulkanTarget` pass on a
      **hosted** Linux run.
- [ ] The `Sanitizers` nightly is green on all three Linux jobs on a master dispatch.
- [ ] The chosen policy is written down as an ADR 0011 amendment note.
- [ ] A guard exists that fails on a PR, not on the nightly, when a shader uses an
      extension the PR-CI toolchain does not support.
- [ ] If a fallback path exists at all, it is loud and countable — never silent.

## Score
```olo-score
capability: 3
craft: 1
stability: 8
decay: 5
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #1139

The `Sanitizers` nightly has been red on all three Linux jobs since 2026-09-08. It is not a
sanitizer finding: it is two shader-compile tests, failing identically under ASan+LSan, UBSan and
TSan, because the hosted Linux arm's shader toolchain is years older than what production shaders
need.

## What was actually wrong

Measured, not assumed. Hosted arm of run
[34341861217](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34341861217) (master,
`force_hosted`) reproduces both failures, and its apt log names the toolchain:

| package on `ubuntu-24.04` | version | behind by |
|---|---|---|
| `libshaderc-dev` / `libshaderc1` | 2023.8-1build1 | ~2 years |
| `glslang-dev` | 15.1.0-2~ubuntu0.24.04.2 | ~1 year |
| `libspirv-cross-c-shared-dev` | 2021.01.15+1.3.239.0-1build1 | ~5 years |

Debian/Ubuntu build libshaderc against the *system* glslang, so the effective compiler was 15.1.0.
`GL_EXT_descriptor_heap` was implemented in glslang on **2026-01-22** (`c8d3e0661`). Seven shaders
declare it, not the two the issue listed: `GpuPathTracer`, `PBR_GBuffer{,_Skinned}`,
`PBR_MultiLight{,_Skinned}`, `VirtualMeshGBuffer`, `VirtualMeshletGBuffer`.

## The policy — ADR 0011 amendment (97)

**The floor is declared and refused, never worked around.** Vulkan SDK **1.4.357.0**, equivalently
shaderc **v2026.3** + SPIRV-Cross **vulkan-sdk-1.4.357.0**. All-or-nothing, the compiler-side mirror
of ADR 0010's device contract.

The issue's option 2 (gate the extension behind a capability probe) is rejected in the amendment for
a specific reason rather than a general distaste for fallbacks: amendment (96) converted the five
material-local maps *whole*, so there is no non-heap arm left in `PBR_GBuffer` and its four siblings.
A gated shader would not run slower, it would sample nothing — a wrong image, produced quietly, on
the build least likely to be looked at.

## The floor is a PAIR, and that was learned rather than designed

**This is the part worth reading.** The first version of this PR staged a conforming shaderc and
nothing else. On the hosted acceptance run
([34356689815](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34356689815)) all three
Linux builds went green, `ShaderCompilation.AllProductionShadersCompileUnderVulkanTarget` passed —
and shard 1/4 failed identically on all three sanitizers with:

```
C++ exception with description "Currently no block to insert opcode."
```

That string lives in **SPIRV-Cross's `spirv_parser.cpp`**, not glslang's. Ubuntu's SPIRV-Cross is
2021.01.15 and cannot parse the SPIR-V 1.6 a 2026 glslang emits, and the engine reflects every module
it compiles (`VulkanShaderReflection`) — so this was never test-only. **The fix had advanced the
failure one step and called it done**: "cannot compile" became "cannot reflect", with a diagnostic
that reads like a corrupt module rather than a stale package.

Two consequences, both in this PR:

- The floor names **two** revisions cut from one SDK release, and they move together or not at all.
- **The guard now reflects what it compiles.** A compile-only check reported green straight through
  that failure; its silence was the real defect.

## What changed

**One constant, one probe, and the probe asks the artefact that actually compiles.**
`ShaderToolchainFloor::Report()` hands the *linked* shaderc the extensions and the
`descriptor_heap` / `descriptor_stride` qualifiers once per process;
`cmake/ShaderToolchainFloor.cmake` runs the same probe against the same library at configure time.
Asking a different artefact would have reproduced #1139 rather than caught it — the failing jobs had
an SDK `glslc` that accepted the extension beside an apt libshaderc that did not.

- **Refusal, loud and counted.** Both Vulkan tiers refuse *after* the SPIR-V cache (a warm cache has
  nothing left to compile) and only for a source that declares one of the extensions — refusing a
  shader that declares none would turn one named toolchain problem into "every Vulkan shader is
  broken", which is the misattribution this contract exists to end.
  `ShaderToolchainFloor::RefusalCount()` makes it countable, and a synthetic below-floor report tests
  it with no mutable test seam.
- **Configure-time named failure**, with glslang's own diagnostic embedded. It reports through stdout
  rather than an exit code: under ASan/LSan a leak inside shaderc would otherwise read as "below the
  floor" and fail a good toolchain.
- **`OLO_SHADERC_ROOT` / `OLO_SPIRV_CROSS_ROOT`** each redirect a header **and** its library, refuse
  a prefix that supplies only one, and the shaderc prefix additionally refuses to carry `spirv_cross`
  headers. #1009 was a *mismatched pair* plus a prebuilt binary — not "the SDK path is poison" — and
  neither half survives here.
- **`.github/actions/setup-shader-toolchain-linux`** builds both components from pinned tags.
  shaderc `v2026.3`'s DEPS pins glslang `168d452a`, which *is* glslang's own `vulkan-sdk-1.4.357.0`
  tag; SPIRV-Cross is pinned to its matching tag. The same revisions the SDK ships, not an
  approximation.
- **`.github/workflows/shader-floor.yml`** — the PR guard. Hosted `ubuntu-24.04`, every PR, no engine
  build: one `glslc` question per `#extension` any production shader declares, plus the header's own
  layout probe (read out of the header, not copied), plus the compile→reflect round trip. Its unit is
  the extension NAME rather than the shader, so it carries no copy of `ShaderCompilationTest`'s
  per-stage tier rules and cannot drift from them.

## Evidence

- **Hosted**, run [34355847140](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34355847140):
  cold toolchain build **5 min 47 s**, pruned prefix **24 MB**, and
  `/opt/olo-shaderc-v2026.3/bin/glslc --version` printing byte-for-byte what this dev box's SDK
  prints — `shaderc v2026.3` / `spirv-tools v2026.3 v2022.4-1283-gb707790a` /
  `glslang 11.1.0-1493-g168d452a`. **All 13 declared extensions compile.**
- **Hosted, under sanitizer flags**: `-- shaderc: /opt/olo-shaderc-v2026.3/lib/libshaderc_shared.so
  (staged, OLO_SHADERC_ROOT)` followed by `-- Shader toolchain floor: satisfied`. The configure probe
  runs for real on Linux rather than degrading to its UNVERIFIED warning.
- **Configure guard both ways** on SDK 1.4.357.0: `satisfied`; and with the required list pointed at
  a nonexistent extension, `FATAL_ERROR` carrying `error: '#extension' : extension not supported`.
- **Locally**: 13/13 extensions + the compile→reflect round trip in ~1 s; 13/13
  `ShaderToolchainFloor` + `ShaderStageContract` + `ShaderCompilation` cases pass.
- **Acceptance run on this branch's final commit is still going** — linked in a comment when it
  lands. Local green is worth nothing for this issue by construction, and the first acceptance run
  proved that by finding the SPIRV-Cross half.

## Three bugs this work caught in itself

All three are the same species as #1139 — a diagnostic that names the wrong thing:

1. The guard reported **all 13 extensions "REJECTED"** against apt's glslc when the real cause was
   one thing: `invalid value 'vulkan1.4'`. It now compiles a trivial shader at the target args first
   and names that once.
2. The new test's substring filter called `DescriptorHeapTextures.glsl` a "declaring shader" because
   its header comment quotes the directive. Now matched at line start, independent of the scanner
   under test.
3. The refusal message read `declares GL_EXT_descriptor_heap but ... does not support
   GL_EXT_descriptor_heap` — and the "declares" half was the required list's first entry rather than
   anything the shader says.

## For #1084

Measured on the hosted ASan job, cold: `Setup Linux build environment` went ~3 min to **8 min 6 s**;
`Configure CMake` **39 s** including the floor probe. Warm, the toolchain build is a cache restore
(immutable key, tens of MB against a store at 7.8 GiB of a 10 GiB cap). Worth re-measuring after
merge.


drsnuggles8:
@coderabbitai review

drsnuggles8:
## Hosted acceptance run: green on the tests this issue is about

Run [34365533801](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34365533801) —
`workflow_dispatch` with `force_hosted: true` on `211e88eca`, so all three Linux sanitizer jobs ran
on `ubuntu-24.04` rather than the self-hosted box.

**The test that failed in both previous runs, from the ctest output rather than the job colour:**

| shard | result | suite |
|---|---|---|
| `ASan + LSan (Linux) tests 1/4` | `ShaderStageContract.FullGBufferWritersMatchTheProductionAttachmentNumericTypes` **Passed** 1.32 s | 100% tests passed, 0 failed of 1938 |
| `TSan (Linux) tests 1/4` | same test **Passed** 1.09 s | 100% tests passed, 0 failed of 1933 |
| `ASan + LSan (Linux) tests 4/4` | `ShaderCompilation.AllProductionShadersCompileUnderVulkanTarget` — green | — |

All three Linux **build** jobs green (ASan + LSan, UBSan, TSan).

**The toolchain actually in use, from the configure output** — not inferred from the action
succeeding:

```
-- shaderc:     /opt/olo-shaderc-v2026.3/lib/libshaderc_shared.so (staged, OLO_SHADERC_ROOT)
-- SPIRV-Cross: /opt/olo-spirv-cross-vulkan-sdk-1.4.357.0/lib/libspirv-cross-core.a (staged, OLO_SPIRV_CROSS_ROOT)
-- Shader toolchain floor: satisfied (GL_EXT_descriptor_heap compiles)
```

and from the action's own verification, including the half that was missing first time round:

```
compile→reflect round trip ok on the descriptor-heap probe
shader toolchain meets the floor:
  shaderc:      /opt/olo-shaderc-v2026.3
  SPIRV-Cross:  /opt/olo-spirv-cross-vulkan-sdk-1.4.357.0
```

`SPIRV-Cross: Git hash: vulkan-sdk-1.4.357.0` appears in its build log, so the reflector is the
revision the SDK ships rather than something near it.

### Two things this run settles that reasoning could not

**TSan holds.** #1009's SEGV was specifically TSan inside `spirv_cross::Resource::~Resource`, and it
is the reason these jobs opted out of the SDK in the first place. A **from-source** SPIRV-Cross was
the untested middle ground between apt's (too old) and the SDK's prebuilt (known broken here). TSan
build and shard 1/4 are both green, which is the first evidence either way.

**The static link order is right.** `spirv-cross-glsl` depends on `spirv-cross-core`, so the staged
archives are listed dependent-first; the system order is left untouched. That was reasoned about
rather than tested locally — Windows takes the MSVC branch — and the successful link is the proof.

### Cost, for #1084

Measured on the hosted ASan job, cold on both caches:

| | |
|---|---|
| `Setup Linux build environment` (was ~3 min) | **8 min 6 s** with shaderc alone; **~5 min 26 s** of that is the toolchain build for *both* components |
| `Configure CMake`, including the floor probe | **39 s** |
| pruned prefixes / cache entry | ~24 MB (shaderc) + SPIRV-Cross archives, one immutable key |

Adding SPIRV-Cross barely moved the number — the two build concurrently with the rest of the setup's
work, not serially after it. Warm it is a cache restore. The first master run after merge pays the
cold build once on three jobs, then the nightly keeps it warm. Worth re-measuring after merge.

### Still owed after merge

A `workflow_dispatch` of `asan.yml` on **master** — a dispatch on a branch only ever tests that
branch, so this run is the pre-merge proof and the master dispatch is the confirmation the issue's
second box asks for.

drsnuggles8:
Hosted acceptance run [34365533801](https://github.com/drsnuggles8/OloEngineBase/actions/runs/34365533801) is complete: **18 Linux jobs, 18 success, 0 failures.** Full box-by-box status on #1139 — https://github.com/drsnuggles8/OloEngineBase/issues/1139

PR #1157, opened 2026-09-09
ci: move the cache-prune working ceiling to track the 25 GB cap

drsnuggles8:
`cache-prune.yml`'s `headroom_floor_mib` was **8800**, sized against the ~9537 MiB wall of
the old 10 GB cap. The cap became **25 GB** on 2026-09-09 and this constant did not move
with it, so six sweeps a day went on failing against a limit that no longer exists — the
raise was inert until this lands.

**8800 → 20000 MiB**, against a ~23841 MiB wall (`25e9 / 1048576`; the cap is decimal GB,
which this file already established by measurement rather than from the docs).

## The margin is sized off the largest single entry, not off a percentage

This is the part the old number got wrong, and it is worth not reproducing at the new
scale. 8800 against 9537 left **737 MiB** — less than `sccache-windows-2025-release` alone
at **2044 MiB measured**. One save landing after a sweep could clear the wall before the
next sweep ever looked, which is roughly what happened on 2026-09-07 (the store reached
10863 MiB on six duplicate vcpkg entries banked after the 07:03 sweep). A margin smaller
than what can arrive inside one sweep interval is not a margin.

20000 leaves **3841 MiB**, ~1.9× that entry.

A proportional raise would have kept the bug: holding the old 92.3 % ratio gives 22006 MiB
and a 1835 MiB margin, still under the largest entry.

## Headroom is deliberate, not slack

The steady set measured **7968 MiB** on 2026-09-09. 20000 leaves it room to roughly double,
which is wanted: raising `SCCACHE_CACHE_SIZE` on `Windows.yml / build` is a live proposal
(the entry is at its 2 G cap and evicting) that the old cap made impossible to consider.

## The one way this number can still be wrong

A configured Actions **budget** can bind before the cap does. The read-only failure this
alarm exists for quotes "you have reached your configured **budget**", and storage above
10 GiB is billed hourly — so a budget set too low goes read-only with the cap nowhere in
sight. That is not the case today, and it is recorded in both the workflow and the doc so
that if it ever is, the ceiling gets derived from the budget instead of from the cap.

## What is deliberately kept

The existing warning — *"Raising the ceiling to silence the alarm re-creates this bug"* —
stays, sharpened rather than removed. Following a cap that really moved is a different act
from quieting an honest alarm under a fixed one, and the file now says which is which.

Historical references to the 10 GB cap in the `WHY THIS EXISTS` header and the #1073
narrative are left alone: they describe what happened, and they were true.

## Verification

`cache-prune.yml` runs on a schedule and on `workflow_dispatch`. The next sweep prints
`headroom: <n> MiB below the working ceiling` — against the current 7968 MiB store that
should read ~12032 MiB rather than ~832. Happy to dispatch it once this is merged and post
the line, since a ceiling change that is never exercised is exactly the kind of thing that
sits wrong for weeks.

Refs #1084

### DIFF
--- a/.github/workflows/cache-prune.yml
+++ b/.github/workflows/cache-prune.yml
@@ -1,4 +1,6 @@
-# Keep the Actions cache under the 10 GB repository cap (#1009 item D).
+# Keep the Actions cache under the repository cache cap (#1009 item D).
+# The cap was 10 GB until 2026-09-09 and is 25 GB now; the working-ceiling block in the
+# sweep step carries the wall arithmetic and why the margin is the size it is.
 #
 # WHY THIS EXISTS
 # ---------------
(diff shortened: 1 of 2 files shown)


drsnuggles8:
**This has a deadline now: the 21:05 UTC sweep will fail without it.**

The store grew to **13311 MiB / 21 entries** this evening — a deliberate master cache-warm banked a fresh 2040 MiB Windows sccache snapshot, and the sibling `feature/shader-toolchain-floor-1139` branch is holding 4074 MiB of its own dispatch caches.

Working the sweep's own rules (newest per `(ref, prefix)`, merged-PR refs dropped), post-sweep comes to **~13297 MiB**:

| ref | MiB |
|---|---|
| `master` | 8446 |
| `feature/shader-toolchain-floor-1139` (alive) | 4074 |
| open PR refs (#1145, #1142, #1157) | 777 |
| **post-sweep total** | **~13297** |

Against the current `headroom_floor_mib=8800` that is `exit 1`. Against this PR's 20000 it passes with 6703 MiB to spare.

It would be a **false alarm** — nothing left is superseded or unread, and 13297 MiB is 53 % of the real 25 GB cap. Which is the precise failure mode this file's own comment calls out: "a job that cries wolf daily is a job everyone learns to ignore, which is exactly the blindness #1073 was about."

Next sweeps: `5 21`, then `5 1` and `5 5` overnight.

Issue #1131, opened 2026-09-08
Editor automation: editor command registry + automation event bus


drsnuggles8:
Epic I of the Automation-Control-Plane roadmap.

## The gap

Two related absences, both about the editor's own actions being invisible to automation:

1. **No editor command registry.** Editor actions (menu items, panel operations) are reachable only by whatever bespoke MCP tool happens to wrap one. `olo_editor_panel_list` / `olo_editor_panel_set` exist; the general "what can this editor do, and do it" does not.
2. **No automation event bus.** `olo_events_tail` gives a tail of events, but there is no way to *subscribe* and react. Every agent workflow is therefore a poll loop, which is why several of the `run-oloengine` helpers busy-wait on a log file.

## Scope (two slices; split when picked up)

1. **Editor action/command registry** — a declared list of editor commands with schemas and authority, exposed through the same Automation Registry as everything else, rather than a parallel mechanism.
2. **Automation event bus** — publish/subscribe over editor and runtime events, with MCP and CLI subscriptions as consumers. Scene opened/saved/dirtied, asset imported, compile finished, play-mode entered.

## Why this is scored modestly

It is real capability, but almost everything an agent needs *today* it can get by polling, and the polling works. The value shows up when the surface above it exists — headless automation (#J) is the consumer that makes an event bus load-bearing rather than a convenience. Sequencing it after the authoring epics is deliberate.

## Design constraints

- An event subscription is an authority-bearing capability, not a free read: events can leak state a read tool would have gated.
- Backpressure needs an answer — a subscriber that stops reading must not grow the editor's memory without bound.

## Blocked by #1123

Both slices add commands to the registry; see the ordering note in that issue.

## Refs

Roadmap §17, §18, §26 Phase 7.

## Score
```olo-score
capability: 5
craft: 3
stability: 1
decay: 1
effort: 5
confidence: 0.8
learning: 3
fun: 3
kano: performance
blocked_by: []
blocks: [1132]
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
[...]

Issue #1130, opened 2026-09-08
Automation: structured test execution, project validation and build invocation


drsnuggles8:
Epic H of the Automation-Control-Plane roadmap.

## The gap

There is no automation path to run the test suite, validate a scene, or produce a build. `script_health_digest` is the closest thing, and it is a script-tool digest rather than a test runner.

So the loop an agent most needs — *change something, run the thing that proves it, read a structured result* — is closed. Today it is done by shelling out to `OloEngine-Tests.exe` and parsing console text, which is why `.claude/skills/run-oloengine/` exists.

## Scope (three slices; split when picked up)

1. **Test catalogue + execution** — enumerate suites/cases (the repo already maintains `test_catalogue.json` and the `OLO_TEST_LAYER` classification), run a filtered selection, return **structured** results: pass/fail/skip per case, failure messages, timings. Not scraped console output.
2. **Project/scene validation** — run the existing validators (`olo_render_validate`, `olo_assets_problems` and friends) as a composable whole-project check with a structured report.
3. **Build/package automation** — invoke a build and report structured results.

Slice 1 is the one with the payoff, and probably the cheapest; slice 3 is the least urgent and could be deferred indefinitely.

## Design constraints worth stating up front

- **A test run must not be silently partial.** The repo already learned this on CI shards: a selection that matches zero cases, or a shard that dies, must be an error and not an empty pass. Any count this returns needs the same discipline.
- Long-running by nature — this needs the progress/cancellation support the MCP server already has.
- Must not require the editor's renderer for the CPU-only suites, or it inherits every GPU-test skip condition.

## Acceptance

- An agent can run a named suite and get per-case structured results without parsing console text.
- A filter matching nothing is an error, not a green run.
- The validation slice reports the same problems the editor panels show.

## Blocked by #1123

Same reasoning as the other command-adding epics: written now, these become migration debt for the registry extraction.

## Refs

Roadmap §15, §26 Phase 6. Test classification: `docs/agent-rules/testing-architecture.md`, `docs/testing.md`.

## Score
```olo-score
capability: 5
craft: 1
stability: 5
decay: 1
effort: 3
confidence: 0.8
learning: 2
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```
[...]

Issue #1126, opened 2026-09-08
Editor automation: structural scene authoring — entity/component CRUD and scene lifecycle


drsnuggles8:
Epic D of the Automation-Control-Plane roadmap. **This is the largest capability gap in the set.**

## The gap, verified

Grepping every registered tool name in `OloEditor/src/MCP/` for a mutation verb — `create`, `delete`, `add`, `remove`, `reparent`, `duplicate` — returns **nothing**. An agent can read the scene in detail (`olo_scene_list_entities`, `olo_scene_get_entity`, `olo_scene_summary`, `olo_entity_list_fields`) and can edit a field on an entity that already exists (`olo_entity_set_field`), but it cannot:

- create or destroy an entity;
- re-parent one, or reorder a hierarchy;
- add or remove a **component**;
- save a scene, or reason about its dirty state.

So the entire authoring loop — the thing an editor is for — is closed to automation. Everything an agent builds today it builds by hand-editing `.olo` YAML outside the editor, which bypasses the serializer's validation and the undo stack entirely.

## Scope (the three slices the roadmap names; split into real issues when this is picked up)

1. **Entity and hierarchy CRUD** — create/destroy/duplicate/re-parent, with stable identity in results.
2. **Component CRUD** — add/remove/query by type, over the reflected component set OloHeaderTool already generates (`AllComponents`, the MCP field registry `McpFieldRegistry.cpp`).
3. **Scene lifecycle** — new/save/save-as/dirty state, and the "is this scene modified" question an agent must be able to ask before it does anything destructive.

## Non-negotiables, from this repo's conventions

- **Undo/redo for every mutation.** The editor's undo stack is the safety net a human has; an automation path that skips it is strictly more dangerous than a human with a mouse.
- Typed schemas and structured results, per #673's `outputSchema` ratchet.
- Main-thread marshalling — scene mutation is not thread-safe.
- Write consent gating, non-inheritable authority (ADR 0005).
- CLAUDE.md's cross-binding rules apply: a component reachable through automation must already be reachable through the serializer and the inspector, or the gap becomes visible here first.

## Acceptance

- An agent can build a small scene from nothing — create entities, parent them, add components, set fields, save — and reopen it in the editor.
- Every mutation is a single undo step, and undo restores the prior state exactly.
- A component type added to the engine appears in the automation surface without a hand-written registration (or fails loudly, per the codegen rules).

## Blocked by #1123

[...]

PR #1162, opened 2026-09-09
fix(tooling): the issue picker asks whether a blocker is still open (#1161)

### LINKED ISSUE #1123: AI/tooling: extract a transport-independent Automation Registry from the MCP server
Epic A of the Automation-Control-Plane roadmap. **This is the architectural root**: every other epic in the set either consumes this registry or is cheaper once it exists.

## Where we actually are (verified on `master`, not quoted from the roadmap)

- **95 tools live**, 93 of them hand-registered across `OloEditor/src/MCP/McpTools*.cpp` (`McpToolsScene/Assets/Render/Physics/Perf/Editor/Input/Shader/Scripting/Diagnostics/Benchmark/Camera`), plus 2 project script-tools.
- Every handler is written **against the MCP server directly**: schema, consent class, main-thread marshalling and result shaping all live in the same lambda that does the work.
- There is no transport-independent command type. `McpServer::m_Tools` (copy-on-write atomic snapshot) is the only registry, and it is MCP-shaped.

That shape has served well — schema colocated with the handler, no bake step — and #673 deliberately rejected UE's reflection-driven manifest bake as wrong for this project's size. **This issue does not reverse that.** The registry stays hand-declared; what moves is the *coupling to MCP as the only caller*.

## The gap

A command's metadata (schema, authority class, undo semantics, availability) is only reachable through the MCP transport. So:

- a CLI (#C) would have to duplicate 93 handlers or shell out to a running editor over HTTP;
- transactions (#E) have nowhere to live — there is no layer that owns "a command" independent of one JSON-RPC call;
- exposure profiles (#B) can only be expressed as MCP-specific filtering.

Each new tool written in the current shape is migration debt, which is why the dependent epics are marked `blocked_by` this one rather than merely "after".

## Scope

- A transport-independent `AutomationCommand`: name, typed input/output schema, authority class, undo/transaction semantics, availability predicate, handler.
- An `AutomationRegistry` owning them, with the copy-on-write snapshot semantics `McpServer::m_Tools` already has (script-tool live reload depends on it).
- **The MCP server becomes an adapter** over the registry rather than its owner. Consent gating, structured content and `outputSchema` behaviour must be preserved exactly — #673 took the surface from 8 to 65-of-66 `outputSchema` adoption and `McpOutputSchemaCoverageTest` ratchets it.
- Migrate the 93 handlers. Mechanical, but it is the bulk of the effort and the reason this is scored an 8.
- Coverage test asserting every registry command is reachable through the MCP adapter with unchanged schema.

## Acceptance

- `tools/list` output is **byte-identical** before and after the extraction (the adapter changes nothing observable).
- Existing MCP tests pass unmodified, including the output-schema coverage ratchet.
- A command can be invoked through the registry with no MCP server constructed at all — the property the CLI and headless epics need.

## Out of scope

- Reflection-driven or manifest-baked registration (#673 closed this deliberately).
- Any new commands. This is a pure extraction; new authoring surface is #D/#F/#G.
- Changing the consent model. Authority classes move, their meaning does not.

## Refs

Roadmap §7.1, §7.2, §26 Phase 1. Sibling trackers: #607 (capability-gap log), #777 (MCP 2026-07-28 spec adoption — that work lands in the adapter, not the registry). Predecessor roadmap: #673 (closed).

## Score
```olo-score
capability: 8
craft: 3
stability: 3
decay: 5
effort: 8
confidence: 0.8
learning: 5
fun: 3
kano: table-stakes
blocked_by: []
blocks: [1125, 1126, 1127, 1128, 1130, 1131]
```

### LINKED ISSUE #1161: Tooling: the issue picker treats a CLOSED blocker as still blocking — it hid the backlog's top two issues
`issue_scores.py` treats `blocked_by` as blocking **for as long as the number is written down**, with no regard for whether the blocking issue is still open:

```python
def is_blocked(d):
    return bool(d.get("blocked_by")) or bool(d.get("blocked_by_external"))
```

So a blocker that has been **closed and delivered** keeps its dependents out of `rank` forever, until a human notices and hand-edits the body. There is no warning, and no way to see it from the picker's output — a stale row does not sort low, it is *absent*.

## What it actually cost, measured 2026-09-09

`/start-work` swept the backlog and found seven issues excluded by blockers that had already merged:

| issue | score | stale blocker | delivered by |
|---|---|---|---|
| #1126 | **3.2** | #1123 | #1147, merged 2026-09-09 |
| #1130 | **3.2** | #1123 | #1147 |
| #1128 | 2.72 | #1123 | #1147 |
| #1125 | 1.92 | #1123 | #1147 |
| #1127 | 1.7 | #1123 | #1147 |
| #1131 | 1.6 | #1123 | #1147 |
| #1140 | 0.94 | #1139 | #1142, merged 2026-09-09 |

**#1126 and #1130 are the two highest-scoring issues in the entire backlog** — above #1084 (1.76) and every renderer item. Both were invisible to the picker at the moment their blocker merged, and would have stayed invisible indefinitely. The bodies have since been healed by hand and the rows now appear.

This is the same class of failure as #1073 and the FFmpeg cache in #1141: one fact spelled in two places, where the stale copy fails **silently and green**.

## Why it will recur, and soon

The Automation-Control-Plane roadmap alone carries a live tech tree — #1129 waits on #1126, #1132 waits on #1131. Every one of those edges becomes a landmine the moment its blocker merges. The window is exactly the interval between "blocker merges" and "somebody happens to notice", which here was open until a `/start-work` sweep tripped over it.

## Scope

1. **Resolve blocker state, don't assume it.** `blocked_by` entries whose issue is CLOSED must not block. Decide deliberately how state is obtained (a `gh` batch query at `rank` time, cached) and what happens **offline** — per `CLAUDE.md`'s no-silent-fallbacks rule, an unresolvable blocker must be reported loudly and countably, not silently treated as either open or closed.
2. **Lint for stale edges.** `issue_scores.py lint` currently only catches *missing* score blocks. It should also report `blocked_by` entries pointing at closed issues, so the condition is visible without waiting for a sweep to trip over it.
3. **`blocked_by_external` is out of scope** — it is free text naming things outside the repo (#1148) and has no state to query. Leave it blocking.

## Acceptance

- An issue whose every `blocked_by` entry is closed appears in `rank` as unblocked, with no body edit.
- An issue with a mix of open and closed blockers stays blocked, and its `blocked:` flag names only the ones still open.
- `lint` reports stale blocker edges and exits non-zero on them.
- Blocker state that cannot be resolved is an explicit, counted diagnostic — never a silent pass.

## Refs

`scripts/issue_scores.py` (`is_blocked`, `rank`, `lint`); `docs/process/issue-scoring.md` §3 (the picker lens). Found by the `/start-work` sweep on 2026-09-09.

## Score
```olo-score
capability: 5
craft: 1
stability: 5
decay: 3
effort: 2
confidence: 0.9
learning: 2
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #1161

`is_blocked()` treated `blocked_by` as blocking for as long as the number was written down, with no regard for whether the blocking issue was still open. A blocker that had merged kept its dependents out of `rank` forever, until a human hand-edited the body — and a stale row does not sort low, it is **absent**, so nothing in the picker's output could reveal it. On 2026-09-09 the closed #1123 and #1139 were hiding seven issues, including #1126 and #1130, the two highest-scoring rows in the whole backlog.

## What changed

**`resolve_blocker_states`** resolves every distinct blocker number in one batched `gh api graphql` call per run — not one `gh` call per edge. `issueOrPullRequest` rather than `issue`, because a `blocked_by` occasionally names a PR and a merged PR is as delivered as a closed issue (MERGED folds into CLOSED).

Nothing is cached across runs, deliberately. A cache of *"is #1123 still open?"* is a second copy of a fact that changes without us — the same shape as the stale `blocked_by` body this function exists to stop trusting. The cost is one extra `gh` call per `rank`; the whole command still runs in ~2s.

**`blocked_by_external` is untouched** (#1148) and still blocks unconditionally: free text naming things outside this repo, with no state to query.

## The offline / unresolvable policy, and why

Per `CLAUDE.md` → *no silent fallbacks*: **an unresolvable blocker keeps blocking, loudly and countably.**

A number that does not exist, a `gh` that is offline, unauthenticated or rate-limited, an entry that is not a positive issue number — all land on `UNRESOLVED`. The issue stays blocked, the row carries its own `blocked?:<N>` flag (visibly distinct from a real `blocked:` edge), and `rank` prints a counted banner **above** the table naming every unresolvable edge.

The direction is chosen, not defaulted. Assuming *open* recreates exactly the bug being fixed. Assuming *closed* invents startable work out of a network error, and a bad pick costs a whole task slot. Neither may be silent: "we could not tell" must never read as "it is fine". Both the code comment on `is_blocked` and issue-scoring.md §3.6 say so.

## `lint`

Now reports three things and exits non-zero on any: missing score blocks (as before), **stale edges** (blocker already closed — the body says something untrue, tidy it), and **unresolvable edges**. So the condition surfaces without waiting for a sweep to trip over it.

## Acceptance criteria — evidence

The seven affected bodies were healed by hand when #1161 was filed, so the live backlog no longer exercises the bug. Evidence below is a synthetic fixture carrying the exact stale edges #1161 measured, resolved by **real `gh` against the real repo** (#1123 and #1139 are genuinely closed, #1126 genuinely open).

```
    # score flags                title
 9001   6.3                      stale: only blocker is CLOSED #1123
 9002   6.3                      stale: both blockers CLOSED (#1123,#1139)
 9003   6.3 blocked:1126         partial: CLOSED #1123 + OPEN #1126
 9004   6.3 blocked?:999999      unresolvable: #999999 does not exist
 9005   6.3 external             external stays blocking

next: #9001 — stale: only blocker is CLOSED #1123
```

- [x] **An issue whose every `blocked_by` entry is closed appears in `rank` as unblocked, with no body edit.** — #9001, #9002 above: no flag, and #9001 is `next:`.
- [x] **An issue with a mix of open and closed blockers stays blocked, and its `blocked:` flag names only the ones still open.** — #9003 prints `blocked:1126`, not `blocked:1123,1126`.
- [x] **`lint` reports stale blocker edges and exits non-zero on them.** — on the same fixture: `5 stale blocker edge(s) — blocker already closed, drop it from blocked_by:` … `lint exit=1`.
- [x] **Blocker state that cannot be resolved is an explicit, counted diagnostic — never a silent pass.** — #9004 above, plus the banner:

```
[!] blocker state UNRESOLVED for 1 edge(s) across 1 issue(s): #999999 (blocks #9004)
    These stay blocked and are flagged `blocked?:`. An unresolvable blocker is never assumed delivered.
    Check `gh auth status` / the network, or run `issue_scores.py lint` for the per-edge detail.
```

Also verified with `gh` forced offline (`GH_HOST=nonexistent.invalid GH_TOKEN=bogus`): all 7 edges go `UNRESOLVED`, every row stays blocked under `blocked?:`, the banner counts them, `next:` is correctly suppressed, and `lint` exits 1. No crash, no silent unblock.

**Live data, unchanged behaviour where it should be:** `rank` still prints exactly the two real edges — `1129 blocked:1126` and `1132 blocked:1131`, both blockers genuinely open — and `lint` is clean. #979 is not fooled by the `[708]` that appears as prose inside a checklist item; its real block reads `blocked_by: []`.

## Tests

Followed the repo's existing convention for `scripts/` rather than inventing a harness: a table of cases plus a `self_test()`, as in `OloEngine/tests/scripts/check_temp_dir_isolation.py` and `.claude/hooks/claude-build-lock-guard.py`. 16 cases covering closed / open / mixed / unresolvable / non-positive / `#`-prefixed / external / Pull-override-restored.

They run on **every** invocation of `issue_scores.py` (pure dict-in/flags-out, microseconds) and are also exposed as `issue_scores.py selftest`. That matches `check_temp_dir_isolation.py`'s reasoning — a silently-broken checker is worse than no checker — and it matters more here than usual, because a mis-blocked issue is *missing* from the output rather than wrong in it.

Guarded the guard: putting the old one-line `is_blocked` back makes 4 of the 16 cases fail and the script refuse to run.

## Docs

- `docs/process/issue-scoring.md` §2 and §3 rule 1 now state that a closed blocker does not block.
- New **§3.6 — Blocker state is resolved, never assumed**: the three consequences, the unresolvable policy and why that direction was chosen, what `lint` does, and the #1161 story as the second paragraph.
- `.claude/commands/start-work.md` line 50: `lint`'s one-line description now mentions the new edge reporting.

## Scope notes

No build needed (Python + Markdown). No rendering change, so no visual verification applies. `apply --dry-run` fails identically on this branch and on master — its one-time `issue-scores.json` source is not in the repo; pre-existing and out of scope.

The known cp1252 mojibake in the bodies of #1126 and #1131 was left alone: this change never touches issue bodies, so fixing it here would be an unrelated widening.

## Summary by CodeRabbit

* **New Features**
  * Issue scoring now checks the current state of referenced blockers.
  * Closed or merged blockers no longer prevent dependent issues from being ranked.
  * Unresolved or unavailable blockers continue to block issues and are clearly flagged.
  * Added a self-test command and automatic validation when the scoring tool runs.

* **Bug Fixes**
  * Improved detection and reporting of stale, missing, or unreadable blocker relationships.

* **Documentation**
  * Expanded issue-scoring guidance to explain blocker resolution, partial delivery, failure handling, and validation.

### DIFF
--- a/.claude/commands/start-work.md
+++ b/.claude/commands/start-work.md
@@ -47,7 +47,7 @@ Then refresh the registry:
     gh pr list --state merged --limit 40  # recently FINISHED work — often still looks "open" in docs/TODOs
     gh issue list --state closed --limit 40     # closed issues — ditto
     python "$BASE/scripts/issue_scores.py" rank # scored, ranked backlog — canonical priority for GitHub issues (docs/process/issue-scoring.md)
-    python "$BASE/scripts/issue_scores.py" lint # UNSCORED open issues — absent from `rank`, therefore invisible to this command
+    python "$BASE/scripts/issue_scores.py" lint # UNSCORED open issues — absent from `rank`, therefore invisible to this command; also stale/unresolvable `blocked_by` edges
 Any topic represented by a worktree, a branch (local OR origin/*), an open PR, or an
 assigned issue is OFF-LIMITS as in-flight. Anything in a recently-merged PR or closed
 issue is OFF-LIMITS as already-done — this is your ground truth that a doc/TODO which
--- a/docs/process/issue-scoring.md
+++ b/docs/process/issue-scoring.md
@@ -128,9 +128,12 @@ Two things that a number can't express but the picker needs:
   favor of delighters. **Rule: the top of the ranked list must always carry
   table-stakes coverage** — completeness is a conscious budget, not an accident.
 - **Tech-tree edges** — `blocked_by: [#N, …]` / `blocks: [#N, …]`. The
-  dependency graph WSJF cannot model. **A `blocked_by` that isn't empty
-  excludes the issue from the picker entirely, regardless of score** — you
-  can't pick what you can't start.
+  dependency graph WSJF cannot model. **A `blocked_by` naming an issue that is
+  still open excludes the issue from the picker entirely, regardless of score**
+  — you can't pick what you can't start. **A closed blocker does not block**:
+  the picker resolves each number's state on every run, so an edge whose
+  blocker has merged stops holding its dependents back the moment it merges,
+  with no body edit. See §3.6.
 - **External blockers** — `blocked_by_external: ["…", …]`, a list of free-text
   reasons. Same effect on the picker as `blocked_by`, for everything the
   in-repo tech tree cannot name: an upstream release we are waiting on, a
@@ -159,7 +162,10 @@ CoD   = Capability + Craft + Stability + Decay
 Score = Confidence × CoD / Effort          # computed only over UNBLOCKED issues
 ```
 
-1. Drop every issue with a non-empty `blocked_by`.
+1. Drop every issue still held by a blocker — a `blocked_by` entry whose issue
+   is **open**, a blocker whose state could not be read, or any
+   `blocked_by_external`. A `blocked_by` entry whose issue is already **closed**
+   does not count (§3.6).
 2. Pick the **max `Score`**.
 3. Break ties with **Learning, then Fun**.
 4. **Pull override:** any issue with `Fun ≥ 8` may be pulled to the front
@@ -178,6 +184,41 @@ Score = Confidence × CoD / Effort          # computed only over UNBLOCKED issue
 
 The printed order is authoritative, and `next:` is simply the first unblocked
 row — so the recommendation can never disagree with what is printed at the top.
+
+### §3.6 — Blocker state is resolved, never assumed
+
+**A `blocked_by` entry blocks only while the issue it names is open.** The
+picker resolves every distinct blocker number on each run (one batched
+`gh api graphql` call, not one call per edge, and not cached between runs) and
+treats a closed issue — or a merged PR — as delivered. Three consequences:
+
+- **A blocker that merges unblocks its dependents immediately**, with no body
+  edit. The number stays written down; it just stops being believed.
+- **A partially-delivered tech tree stays blocked, and the `blocked:` flag names
+  only the edges still open.** An issue blocked by a closed #1123 and an open
+  #1126 prints `blocked:1126` — never the delivered one.
+- **`blocked_by_external` is unaffected** and still blocks unconditionally. It
+  is free text naming something outside this repo; there is no state to query.
+
+**If a blocker's state cannot be read, the issue stays blocked and says so.**
+An unresolvable edge — a number that does not exist, `gh` offline,
+unauthenticated or rate-limited — is flagged `blocked?:<N>` and counted in a
+banner above the table. The direction is chosen, not defaulted: assuming *open*
+recreates the bug below, and assuming *closed* invents startable work out of a
+network error, which costs a whole task slot. "We could not tell" must never
+read as "it is fine" (`CLAUDE.md` → *no silent fallbacks*).
+
+`issue_scores.py lint` reports stale edges (blocker already closed — tidy the
+body) and unresolvable ones, and exits non-zero on either, so neither waits for
+a sweep to trip over it. `issue_scores.py selftest` runs the blocker cases on
+their own; `rank` and `lint` run them too, on every invocation.
+
+Before this, a written-down number blocked forever. On 2026-09-09 the closed
+#1123 and #1139 were holding **seven** issues out of `rank` entirely, including
+#1126 and #1130, the two highest-scoring rows in the whole backlog. A stale row
+does not sort low, it is *absent* — so nothing in the picker's output could
+reveal it, and it took a `/start-work` sweep comparing the list against the
+tracker by hand to notice (#1161).
 
 ### Why the floor exists
 
--- a/scripts/issue_scores.py
+++ b/scripts/issue_scores.py
@@ -92,6 +92,10 @@ OPEN, CLOSED, UNRESOLVED = "OPEN", "CLOSED", "UNRESOLVED"
 # tech tree cannot silently trip the limit.
 _BLOCKER_CHUNK = 100
 
+# Per-chunk wall clock. The call normally answers in about a second; this is here
+# so a stalled connection degrades to UNRESOLVED instead of hanging `rank`.
+_BLOCKER_TIMEOUT_S = 30
+
 
 def blocker_key(raw):
     """Normalize one `blocked_by` entry to an int issue number.
@@ -104,7 +108,7 @@ def blocker_key(raw):
     take the whole chunk down with it.
     """
     try:
-        n = int(str(raw).lstrip("#").strip())
+        n = int(str(raw).strip().lstrip("#").strip())
     except ValueError:
         return str(raw)
     return n if n > 0 else str(raw)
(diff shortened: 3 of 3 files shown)


drsnuggles8:
Review follow-ups applied in 6cf2b22e3. All four verified against the code first; all four were still valid, one of them a real crash.

**`sorted(unresolved_edges)` raised TypeError on mixed key types** — the significant one. An issue carrying both a numeric and a malformed blocker (`blocked_by: [4242, "n/a"]`) yields an int key and a str key, and `rank` died with a traceback **before printing a single row**. Reproduced end-to-end on the parent commit:

```
  File "scripts/issue_scores.py", line 349, in cmd_rank
    listed = ", ".join(f"#{k} (blocks #{num})" for num, k in sorted(unresolved_edges))
TypeError: '<' not supported between instances of 'str' and 'int'
```

Fixed with a shared `blocker_sort_key`: numbers numerically first, malformed entries after, never comparing across types. The second tuple slot is only reached when the first ties, which forces both sides to one type — so it preserves numeric ordering rather than falling back to `str()` everywhere. Two ordering cases added to the self-test, since the flags cases could not have caught this (the sort lives in `cmd_rank`, not `row_flags`).

**`blocker_key` stripped whitespace after removing the `#`**, so `" #1123"` resolved to UNRESOLVED instead of 1123. Safe direction — it blocked and was reported, not a silent pass — but wrong, and hand-edited blocks are exactly where a stray space comes from. Strip first, then the `#`.

**No timeout on the GraphQL call**: a stalled connection hung `rank` forever with nothing printed. Bounded at 30s per chunk; `TimeoutExpired` leaves that chunk's keys unset so the existing `setdefault` makes them UNRESOLVED — the same loud, counted answer as any other unreadable state, no new path. Verified with a monkeypatched stall: all keys come back UNRESOLVED instead of hanging. `check=False` spelled out, since relying on the non-zero exit while still parsing stdout is load-bearing here rather than incidental.

**Redundant set comprehension** folded into `set(numbers)`, now that both call sites share `blocker_sort_key`.

Self-test is 20 cases, 0 failures. Re-ran the whole evidence set: the fixture, the forced-offline run, and live `rank`/`lint` all produce identical output to before, and the `" #1123 "` case now correctly ranks unblocked.

Issue #1013, opened 2026-09-02
RHI: parallel recording phase 2 — CommandBucket replay on items, the remaining passes, per-context Renderer3D resources (#806 follow-up)


drsnuggles8:
Follow-up to #806 (PR #1012). That PR landed the mechanism: `RenderCommand::RecordParallel`, a per-item `VulkanRecordingContext` on a thread-local, secondary command buffers executed in item order, per-item layout-tracker overlays with the identity-only overlap rule, and the shadow pass (CSM cascades + atlas entries) as the first converted pass. The contract is ADR 0011 amendment (92); the conversion checklist is `docs/agent-rules/vulkan-parallel-recording.md`. What follows is everything amendment (92) lists as "deliberately not built", plus what the review of #1012 surfaced as the next rung. Not to start before #1012 merges.

## 1. `CommandBucket` replay on an item (the gate for most passes)

Every bucket-driven pass (`SceneRenderPass`, `WaterRenderPass`, `ForwardOverlayRenderPass`, `FoliageRenderPass`, `GPUDrivenOcclusionPass`, `DecalRenderPass` ×3, `DeferredGPUOcclusionPass`, `PlanarReflectionRenderPass`, `OverdrawRenderPass`) replays through `CommandBucket::Execute(RendererAPI&)`, which cannot run on an item today because the dispatcher, not the API, is per process:

- [ ] `CommandDispatch::s_Data`'s bind caches (`BoundUBOs`, `BoundTextures`, `CurrentBoundVAO`, `LastRenderStateIndex`, `LastMaterialDataIndex` / `LastMaterialOffsetsLive`, `HeapEpoch`, `Stats`) become per recording context (thread-local indexed by the item, reset at fork/join), leaving the frame-scoped members (`DepthPrepassActive`, shadow texture ids, …) shared and read-only inside a region.

- [ ] `Renderer3D`'s shared per-draw upload targets — the material UBO and `GetModelInstanceBuffer()` — get a per-context copy (amendment (92) rule 6: one writer per object per region). This is the actual blocker; the `(region, item)` writer stamp in `VulkanUniformBuffer`/`VulkanStorageBuffer::SetData` will assert the moment a bucket replays on two items without it.

[...]

PR #1175, opened 2026-09-10
Editor automation: asset query, safe CRUD/move/delete, and import settings

### LINKED ISSUE #1128: Editor automation: asset query, safe CRUD/move/delete, and import settings
Epic F of the Automation-Control-Plane roadmap.

## Where we are

Two asset tools exist — `olo_assets_list` and `olo_assets_problems` — and they are read-only. There is no automation path to create, move, delete, import or reimport an asset, and no way to ask **who references this asset** before touching it.

The reference question is the important half. `AssetRegistry.oar` is a binary file (grepping it for a handle returns nothing), so "what breaks if I move this texture" is currently unanswerable without opening the editor and looking.

## Scope (three slices; split when picked up)

1. **Query** — asset search by type/name/path, `get` for metadata, and the two graph directions: **dependencies** (what this asset needs) and **referrers** (what needs this asset).
2. **CRUD + safe move/delete** — create, rename, move, delete, where *safe* means references are rewritten or the operation is refused with the referrer list, never silently broken.
3. **Import / reimport / import settings** — including the per-asset settings that currently have no automation surface at all.

## Why safe-delete is the stability axis

Deleting a referenced asset is data loss with a delayed symptom: the scene still loads, the mesh renders untextured, and nothing says why. The referrer query in slice 1 is what makes slices 2 and 3 safe, which is why it comes first.

## Non-negotiables

- **No raw filesystem edits for engine assets** (roadmap §30.5). Everything goes through `AssetManager` so the registry, hot-reload watcher and handle identity stay consistent. Known hazard: the watcher races a large write, and a temp file dropped in `Assets/` gets imported.
- Undo where the operation is undoable; explicit refusal where it is not (an import is not).
- Typed schemas, structured results, write consent.

## Acceptance

- "What references this asset" is answerable without opening the editor.
- Moving an asset leaves every referring scene and material still resolving.
- Deleting a referenced asset is refused by default and names its referrers.

## Blocked by #1123

Same reason as #1126: this is a large batch of new commands, and writing them in the current MCP-shaped registry makes the extraction more expensive.

## Refs

Roadmap §12, §26 Phase 5, §30.5.

## Score
```olo-score
capability: 8
craft: 3
stability: 5
decay: 1
effort: 5
confidence: 0.8
learning: 2
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Closes #1128.

Asset automation: query, safe CRUD/move/delete, and import settings — Epic F of the
automation control plane.

## The load-bearing decision: where the referrer index comes from

The issue's stability axis is "what breaks if I move this texture", and the obvious
source for it — `EditorAssetManager`'s dependency graph (`m_AssetDependencies` /
`GetDependencies`) — cannot answer it, for four independent reasons:

1. **`SceneSerializer` registers nothing.** Only six `AssetSerializer` call sites ever
   registered an edge. Scenes are the majority of referrers, and the acceptance criterion
   is about scenes specifically.
2. **Load-gated** — an edge appears only when an asset is deserialized, so the referrer
   set is short by however much the user happened not to open.
3. **Every edge was backwards** (see the bug fix below).
4. **A handle-set graph cannot produce an edit** — rewriting a reference on a move needs
   the file, the line and the exact text.

So the index is *derived*: a scan of the project's text asset files, resolving each
candidate against the same anchors the engine itself uses. Two rules make it safe:

- **Extract generously, resolve exactly.** Path references are found by their *value*
  (any scalar ending in a known asset extension), not by a list of key names that would
  go stale the moment a component gains a field. Over-collecting costs an entry in a
  count; under-collecting silently loses a referrer.
- **Anchored, not fuzzy.** `Checkerboard.png` exists under both `OloEditor/assets/textures/`
  and the project's own `Assets/Textures/`, so a filename or suffix match would invent a
  referrer — and a move would then rewrite somebody else's reference. Resolution mirrors
  `EditorAssetManager::ImportAsset` step for step (#887, #1098) plus
  `Project::GetAssetFileSystemPath`, because the point of the answer is to predict what
  the engine will actually load.

Every referrer result ships a `coverage` block naming what was *not* searched — binary
formats skipped and named, unreadable files, unresolved references, truncation. An empty
referrer list means "nothing found in this set", never "nothing references this", and a
destructive command **refuses outright on a truncated index**; `force` deliberately does
not waive that one, because it is a claim about the quality of the list rather than about
the risk the caller is accepting.

## Acceptance criteria

| Criterion | Evidence |
|---|---|
| "What references this asset" is answerable without opening the editor | `olo_asset_references`; `McpAutomationAssetCommandsTest.ReferrersAreAnswerableWithoutTheEditor` |
| Moving an asset leaves every referring scene and material still resolving | `olo_asset_move` rewrites each reference in the style its own file used; `MoveRewritesTheReferringSceneSoItStillResolves` re-queries at the new location and gets the referrer back |
| Deleting a referenced asset is refused by default and names its referrers | `olo_asset_delete`; `DeleteIsRefusedByDefaultAndNamesItsReferrers`, `ForcedDeleteProceedsAndNamesWhatItBroke` |

## Commands

`olo_asset_get`, `olo_asset_references`, `olo_asset_create`, `olo_asset_move`,
`olo_asset_delete`, `olo_asset_import`, `olo_asset_reimport`, `olo_asset_import_settings`.
`olo_assets_list` gains `namePattern` / `pathPattern` — the search half of slice 1 — rather
than a near-duplicate `olo_asset_find`, which is the surface bloat exposure profiles
(#1124) exist to undo.

Undo: `olo_asset_move` is **one editor undo step** that takes the reference rewrites back
with it, guarded so an undo after somebody hand-edited a referring file refuses instead of
clobbering it. `create` / `delete` / `import` / `reimport` / `import_settings` declare
`Irreversible` and say so in their payloads as well as their metadata.

## Bug found on the way (own commit)

**Every asset dependency edge was registered backwards, and it was invisible because the
reader was wrong the same way.**

`RegisterDependency(handle, dependency)` is documented as "`handle` depends on
`dependency`". All six `AssetSerializer` call sites passed the pair the other way round —
each one directly beneath a trace line stating the correct relation. It went unnoticed
because `ReloadData` called `UpdateDependencies(handle)`, which walked the *forward* map
and notified those; two wrong directions composed into working hot-reload notification.
Artefacts: `m_AssetDependents` was empty for every asset (so `UpdateDependents`, called on
the very next line, was a no-op for everything), `GetDependencies()` answered backwards for
anyone who asked, and `AnimationAsset::OnDependencyUpdated` — which spelled it correctly —
left that one asset with edges in both directions.

Fixed together, because fixing either half alone silently stops every texture hot-reload.
`AssetDependencyDirectionTest` pins each half separately.

Also removed `EditorAssetManager::CreateAsset(AssetType, path)`: declared, never defined,
never called — calling it is `LNK2019`.

## Notes

- `AutomationFileWrite.{h,cpp}` extracts the guarded atomic file replace out of
  `AutomationSceneDocument.cpp` so there is one implementation, not two chances to get the
  guard subtly different.
- The temp file is a **sibling** of the target, not a file under `%TEMP%`. The handover
  suggested writing outside `Assets/`, but `%TEMP%` can be on another volume here, which
  turns the commit into a non-atomic copy — the very thing that lets the watcher cache a
  half-written asset. A sibling `.tmp` is a same-volume rename, and `.tmp` is absent from
  the extension map, so `DecideFileWatchAction` ignores it.
- Per-asset import settings had no store at all (`AssetMetadata` is
  handle/type/path/status/mtime; the one import option, `MeshImportOptions::FlipUV`, is
  never persisted). `olo_asset_import_settings` owns an `.oloimport` JSON sidecar — an
  extension deliberately absent from the extension map — and reports
  `appliedByImporter: false`, because nothing consumes it yet and a settings write that
  looks like it took effect and did not is exactly a silent fallback.
- `olo_asset_create` supports the types with a default-constructible asset class and a
  registered serializer, and its refusal names them. `SoundGraph` is excluded: its asset
  holds a `Ref` to a forward-declared type, so instantiating it needs the audio-graph
  internals.

- `olo_asset_move` / `olo_asset_create` take `createDirectories` (default **false**). There
  is no directory command on this surface, so refusing a missing parent outright would make
  "move this into a new folder" impossible rather than two-step; creating one for a typo'd
  path silently is the other failure. The refusal names the flag.
- `olo_asset_create` for a `Material` checks the shader library first and refuses when it is
  empty. `MaterialAsset`'s constructor resolves a shader out of
  `Renderer3D::GetShaderLibrary()` and **asserts** when it finds none, so on a headless host
  (`oloctl`, a test) calling the factory would abort the process rather than return an error.

## Verification

- 47 new tests across `McpAutomationAssetIndexTest`, `McpAutomationAssetCommandsTest` and
  `AssetDependencyDirectionTest`.
- The index was also run over the **real SandboxProject** rather than only synthetic YAML:
  173 files, 225 references, **0 unresolved**. That run is what found two bugs the unit
  tests had agreed with — a missing asset-directory anchor (silently dropping every Lua and
  audio referrer) and scene *titles* (`Scene: Courtyard.olo`) counted as 99 broken path
  references. It also exposed one of the tests asserting filesystem-case-dependent
  behaviour.
- A live editor pass over real JSON-RPC against a running OloEditor and the real
  SandboxProject: `olo_asset_references` found all 4 referrers of `Checkerboard.png` across
  3 scenes (coverage: 173 files, 0 unresolved); `olo_asset_delete` refused with
  `reason: referenced, referrerCount: 4` and left the file on disk; `olo_asset_move`
  rewrote 4 references across 3 files, kept the handle, and a re-query at the new path
  still found all 4. `undoable: true` live, `false` headless, as designed.
  That pass is also what found the casing bug below.
- A move used to NORMALISE THE CASING of a reference it should have left alone.
  `TerrainVirtualTextureTest.olo` spells its texture `assets/textures/Checkerboard.png`,
  which on a case-insensitive filesystem resolves against a directory actually named
  `Assets/Textures`; re-spelling from the filesystem rewrote the case, so the move edited a
  line nobody asked it to change and moving back did not restore it. Fixed by preserving
  the original casing of the case-insensitively-common prefix, so a round trip is a no-op.
  Only visible from `git status` after the live run.
- `docs/agent-rules/derived-graph-must-outlive-its-source.md`, indexed in both README
  sections; the MCP guide gains a **Safe asset operations** section and a row per command.

## Summary by CodeRabbit

* **New Features**
  * Added asset automation for metadata, references, creation, moving, deletion, importing, reimporting, and import settings.
  * Added name and path filters for asset searches.
  * Added safe reference rewriting, scan coverage reporting, atomic file updates, and undo support for moves.
* **Bug Fixes**
  * Corrected asset dependency direction and dependent-asset update behavior.
* **Documentation**
  * Expanded guidance for asset operations and dependency validation.
* **Tests**
  * Added coverage for asset automation, safeguards, undo/redo, imports, reference indexing, and dependency direction.


drsnuggles8:
@coderabbitai review

drsnuggles8:
Thanks — ten of eleven applied in `d1db008d2`, and three of them were real defects in the safety story this PR is built on. Per finding:

**Applied**

| Finding | What it actually was |
|---|---|
| Path traversal on mutation targets | Real, and the worst of them. `ResolveTarget` takes an absolute path and only the *destination* was ever contained, so `olo_asset_delete` would `remove()` any file on the machine. One `IsInside()` predicate (by path component) now gates delete, a move's source, and the **write** path of import settings. Reads stay open — answering "what is this file, does anything point at it" about any path is useful and harmless. |
| Coverage incomplete not flagged | Real. `BuildAssetIndex` returned an empty index with `Truncated=false` when the root was unusable, so a scan that read nothing reported zero referrers and looked clean — and a delete would proceed on it. Exactly the confusion the coverage block exists to prevent. Added `Complete` + `IncompleteReason` + `Reliable()`; both destructive commands gate on `Reliable()` and the refusal says which condition and why. |
| Roll back rewrites when the undo rename fails | Real. `CommitMove` guarded the forward move; `Apply` did not, so a Ctrl-Z could leave every reference pointing where the asset is not. Extracted `RevertRewrites()` and used it on both paths. |
| `undoEc` discarded | Real. It claimed "rolled the move back" when the reversal itself failed. Now reports the inconsistency distinctly and logs it. |
| Throwing `std::filesystem::relative` | Real — it escaped before `PlanReferenceEdits` could return its controlled refusal. Switched to the `error_code` overload; also fixed the legacy-prefix branch to check the remainder before re-attaching the prefix. |
| `inProject` string prefix | Real — `/work/proj-backup` matched `/work/proj`. Component comparison now. |
| `idempotentHint` for import | Applied; `Register` takes an explicit override. |
| `<cctype>` | Applied. |
| Fixture teardown ×2 | Real, and this bit me earlier in this branch already (a case passed alone and failed in a full run off process-global shader state). Both fixtures now `Project::Unload()`. |

**Not applied**

`ReplaceFileContents` validate-then-mutate window. Correct that it is not atomic, but closing it needs a platform-specific conditional replace *and* the cooperation of every other writer — including the editor's own save path and whatever editor the user has the file open in. That code is the merged `olo_scene_save` guard this PR only extracted, so changing its semantics here would change scene saving too. The residual window is now written into the header rather than implied away; in practice the contending writer is a human whose save is orders of magnitude further away than the window.

Six new tests, the load-bearing two being: a delete refuses an absolute path outside the project, and a scan that could not run does not read as clean. 56 in the new suites, 1536 green across MCP/asset/automation/scene.

drsnuggles8:
Second round applied in `9c90def2a`. Both new findings were right, and chasing the second one turned up something considerably worse than the finding itself.

**`scan-incomplete` missing from the schema** — straight bug, fixed. I added the `reason` value last round and left the enum at two entries, so a schema-validating client would have rejected a legitimate refusal.

**`Reliable()` true while opaque formats go unscanned** — the naming criticism is correct and I've taken it; the proposed remedy I have not, because refusing whenever any unscannable file exists would refuse every destructive operation in every real project (the sandbox has 97), which is a broken command rather than a safe one. Split the two notions instead:

- **Fixable** incompleteness — truncated, failed walk, unreadable file — gates the mutation and names the reason, because the user can act on it. `skip_permission_denied` is now gone too; it dropped an unreadable directory *without* setting an error, which was the part of the earlier coverage finding I'd missed.
- **Permanent** boundary — a format nothing here can read — is reported at the decision point as `unscannableFiles` on every destructive result, not folded into a "reliable" claim.
- Renamed `Reliable()` → `ScanCompleted()` so it cannot be read as "provably no references".

**What fell out of checking whether `.gltf` is actually opaque:** it isn't — it's JSON and it names its textures by URI. But adding it to the scanned set changed nothing, because this scanner is a YAML scalar reader and real glTF writes `"uri" : "x.jpg"` — quoted key, space before the colon, trailing comma. Three small parser changes later the sandbox reports **300 references instead of 225**. The extra 74 are the glTF textures: every texture in `DamagedHelmet` and the other models was invisible to the index, so `olo_asset_delete` would have removed them reporting no referrers. That is precisely the delayed-symptom loss this epic exists to prevent, and it was sitting in the shipped project. A *minified* `.gltf` is one line and still yields nothing — counted as unscannable rather than quietly absent.

Two follow-ons found by re-running the scan over the real project rather than trusting the tests:

- resolution gained the referring file's own directory as a **last** anchor, after every anchor `ImportAsset` uses, so it can only rescue an otherwise-unresolved value and can never disagree with the engine. The header had claimed this anchor existed since the first commit; it did not.
- self-references are now dropped at extraction — the new anchor resolved each scene's `Scene: Courtyard.olo` *title* against its own file, which was 99 of those 300.

Five new tests; 58 in the new suites, 1135 green across MCP/asset/automation. The project now scans with exactly one unresolved reference, and it is a real one: `DamagedHelmet.gltf` names `screenshot/screenshot.jpg`, which is not in the repo.

Still not taking the `ReplaceFileContents` validate-then-mutate window, for the reason given last round: it is the merged `olo_scene_save` guard this PR only extracted, closing it needs a platform-specific conditional replace plus cooperation from every other writer, and the residual window is now documented in the header rather than implied away.

drsnuggles8:
Third round applied in `65079df74`. All four valid; two were mine from last round's edits.

**`reliable` vs `scanCompleted`** — correct, and embarrassing: last round's rename moved the value expression and left the key, so no client could read the field the schema documents. Fixed.

**`SourceRelative` too broad** — you're right, and it broke the rule this index is built on. The anchor exists because Assimp resolves a glTF URI relative to the `.gltf`; nothing else in the engine does, so applying it everywhere let the index claim a sibling resolves when `ImportAsset` would never find it — a refused delete that should have gone ahead, and a rewrite of a path nothing reads. Now gated on the source extension, with a test that a `.olo` naming `Local.png` does *not* pick up its sibling.

**Minified glTF / unsupported syntax** — taken, and taken as *extraction* rather than as flagging. Last round I wrote that boundary into a comment; you're right that a comment is not a fix. `.json`, `.lua`, `.cs` were in the scanned set with the parser reading none of them, and a minified `.gltf` is one line so not a single key parses — all three counted toward `FilesScanned` while leaving `ScanCompleted()` true, which is an index reporting itself complete over files it never read.

They are all the same shape — a quoted string that happens to be a path — so one mechanism reads them: after the key/value parse declines a line, sweep it for quoted literals ending in a known asset extension. A literal followed by a colon names the next one, which is what gives a minified glTF's URI the key `uri`. Extraction stays generous, resolution stays the filter; over-collecting costs a count, and the previous state cost a deleted texture.

That sweep then caught something I'd have shipped: eleven Lua scripts whose header comments quote their own path, collected as references because the self-reference guard only covered bare filenames. Widened — a file naming itself is never a reference, separator or not.

Four new tests (minified glTF, a path inside a script call, `SourceRelative` declining a `.olo`, a file quoting its own path). 62 in the new suites, 1139 green across MCP/asset/automation. The real project scans to 300 references with one unresolved, and that one is a genuinely broken URI in `DamagedHelmet.gltf`.

Still standing on the `ReplaceFileContents` validate-then-mutate window, for the reason given twice now: it is the merged `olo_scene_save` guard this PR only extracted, closing it needs a platform-specific conditional replace plus cooperation from every other writer including the user's text editor, and the residual window is documented in the header rather than implied away. Happy to be overruled by a maintainer, but it is a change to scene saving, not to this feature.

drsnuggles8:
Follow-up for whoever traces the red ASan Windows job on this PR: `AutomationAssetCommandsTest.UndoRefusesWhenAReferringFileChangedUnderneath` crashes under the Windows ASan build, and **it is not a defect in this PR's code**. Diagnosed and fixed in #1013's branch so it is not diagnosed twice.

The fault, from cdb with gtest's SEH translator disabled (`--gtest_catch_exceptions=0`; leave it on and you get `SEH exception with code 0xc0000005` with an empty "Stack trace:", which is what made this look opaque):

```
VCRUNTIME140!__FrameHandler3::GetUnwindTryBlock+0x1f
mov eax,dword ptr [rax+rcx+4]   ds:0000000000000154=????????
                                 ^ rcx = establisher frame = 0
VCRUNTIME140!_CxxThrowException+0x99
AssetMoveCommand::Apply+0x1a19   AutomationAssetCommands.cpp @ 764
```

Line 764 is the bare `throw;` in `Apply`'s `catch (...)`. ASan's use-after-return instrumentation moves a function's locals into a heap "fake stack" frame, and MSVC EH then cannot find the catch funclet's parent frame — it reads NULL and faults through it.

It is a toolchain bug, reproduced in a 45-line standalone TU at the same PC with the same null frame, and it is the second of its kind here: #497/#661 was clang-cl+ASan redzoning the EH metadata globals, fixed by the LLVM 23.1.0 pin. This one survives that pin. A variant matrix at `/O2` says the catch **type**, the rethrow **form** (bare `throw;` and `std::rethrow_exception` alike) and any cleanup work in the handler are all irrelevant — only whether the throw is lexically inside the handler matters. `/Ob0` appears to fix it on four of five shapes and does not.

`Apply` is not special; it is simply the only throw-inside-a-catch site the ASan suite actually executes. There are about a dozen more across `OloEngine/src` and `OloEditor/src`.

Fixed at the toolchain rather than the call site, so all of them are covered at once: `-fsanitize-address-use-after-return=never` on the clang-cl branch of `cmake/Sanitizers.cmake`. It gives up stack-use-after-return detection on Windows, which we were not getting — the Windows default is `detect_stack_use_after_return=0` and nothing sets it to 1, so the instrumentation was present but inert: catching nothing, breaking EH. An A/B with and without the flag over four injected defects at `/Od` and `/O1` shows identical detection for every other class. Linux ASan is untouched.

Verified on `clangcl-asan`: `AutomationAssetCommandsTest` 33/33, and `Automation*` + `Mcp*` 1004 passed / 1 skipped by design / 0 failed.

Full write-up, the variant matrix and the repro recipe are in `docs/agent-rules/build-trees-and-windows-asan.md` §4b, landing with #1013.

No action needed here — this is a pointer, not a request.

PR #1184, opened 2026-09-11
feat(automation): atomic multi-command transactions with rollback, symbolic outputs and dry-run

### LINKED ISSUE #1127: Editor automation: atomic multi-command transactions, rollback, symbolic outputs and dry-run
Epic E of the Automation-Control-Plane roadmap.

## The gap

Every automation call today is independent. There is no `olo_*` tool that batches, and nothing in `OloEditor/src/MCP/` speaks of transactions. So a multi-step edit — create an entity, add three components, set eight fields, parent it — is N separate calls, each of which can fail on its own, leaving the scene in a state that is neither the before nor the after.

That is the classic partial-application failure, and it is scored as a **stability** issue rather than a capability one for that reason: the damage is a silently half-built scene, discovered later.

It also costs a round trip per call, which is why an agent building anything non-trivial today is slow as well as unsafe.

## Scope

- Atomic multi-command transactions: all commands apply, or none do.
- **Rollback** on failure, restoring the pre-transaction state.
- **Symbolic outputs** — the output of step 1 (a newly created entity's id) referenced as the input of step 4, without a round trip to learn it. This is what makes batching more than a latency optimisation.
- **Dry-run**: validate the whole batch, apply nothing, report what would happen.
- Batch authority: a transaction's consent is decided once, for the whole batch, and must not let a low-authority caller launder a high-authority command into it (the ADR 0005 property the outbound client already upholds).

## Why it is cross-cutting

It is deliberately separate from #1126 (scene authoring), asset automation and prefabs rather than folded into any one of them: the semantics have to be identical across all three, and defining them inside one domain guarantees the other two diverge.

## Confidence is 0.5 on purpose

Rollback across heterogeneous commands is the genuinely uncertain part — some operations (an asset import, a bake) are not cleanly reversible, and the honest answer may be that such commands are simply not transactable rather than pretending they roll back. **Spike that boundary before committing to an API.**

## Acceptance

- A failing step in a batch leaves the scene byte-identical to its pre-batch save.
- A transaction is one undo step, not N.
- A command that cannot participate in a transaction says so in its metadata and is refused at batch-build time, not half-way through execution.

## Blocked by #1123

The transaction layer has nowhere to live until a command exists independent of one JSON-RPC call.

## Refs

Roadmap §13, §25.3, §26. Authority model: `docs/adr/0005-mcp-script-tools-lua-sandbox.md`.

## Score
```olo-score
capability: 5
craft: 3
stability: 8
decay: 1
effort: 5
confidence: 0.5
learning: 5
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
Epic E of the Automation-Control-Plane roadmap. Adds a transaction layer over the automation
registry: a batch of commands applies atomically or not at all, rolls back on failure, commits as
one undo entry, wires step outputs into later steps symbolically, and can be dry-run.

## Acceptance criteria

**1. A failing step leaves the scene byte-identical to its pre-batch save.**
`AFailingStepRollsBackToAByteIdenticalScene` builds a non-trivial scene, serializes it with the real
`SceneSerializer`, runs a 4-step batch whose last step fails, and compares the bytes — not the entity
count, which would pass while a component or a hierarchy edge stayed behind.

**2. A transaction is one undo step, not N.** `ASuccessfulTransactionIsOneUndoStepNotN` runs three
creates, asserts `undoEntries: 3`, then takes all three back with a **single** `Undo()` and checks
`CanUndo()` is false afterwards. `AReadOnlyBatchLeavesNoUndoEntryAtAll` pins the other end: a batch
that mutates nothing leaves no no-op entry to press Ctrl-Z past.

**3. A non-transactable command says so in its metadata and is refused at batch-build time.**
Decided purely from `AutomationCommand::Undo`, and the rule **default-denies**. Four tests, one per
refusal reason, each asserting that *nothing ran* — no entity, no undo entry, no dirty flag, and the
fake command's run counter still 0.

**4. A low-authority caller cannot launder a `ProjectWrite` command into a granted batch.**
`ALowAuthorityCallerCannotLaunderAProjectWriteIntoABatch` checks both doors: the registry's own gate
(`olo_transaction_apply` is itself `ProjectWrite`, so a caller without consent never reaches the
handler and no batch is even parsed — `MarshalCount == 0`), and `BuildTransaction`'s, reached
directly the way a future non-MCP frontend would.

## The Phase A spike, and what it found

[`docs/analysis/automation-transactability-boundary.md`](docs/analysis/automation-transactability-boundary.md)
is the written answer the issue asked for. Measured, printed and ratcheted by the census test:
**121 registered commands, 12 transactable** — 6 `Irreversible`, 3 `HistoryControl`, 4 refused for
being deliberately off-thread, 96 never declared.

The issue said the honest outcome may be that irreversible commands are simply not transactable. It
is: `olo_asset_import`, `olo_asset_create` / `olo_asset_delete`, `olo_build_invoke` and friends are
refused with a message saying to run them on their own. No partial-rollback mode was built — a
rollback that restores the scene while leaving an imported asset behind is a worse lie than a refusal.

Two things the classification surfaced that the existing metadata could not express:

- **`AutomationUndo` had no value for a command that operates ON the undo history.**
  `olo_editor_undo` / `olo_editor_redo` are not irreversible (undoing an undo is a redo) but cannot
  be a *step*, because a transaction is itself a rewrite of the stack they walk. Left as
  `EditorUndoStack` they would have been admitted and then failed mid-batch. Added
  `AutomationUndo::HistoryControl`; `olo_transaction_apply` declares it too, which is what stops a
  transaction nesting inside another — one rule, three cases, no name blacklist in the layer.
- **Reversibility is necessary but not sufficient.** A batch runs every step inside one main-thread
  critical section, so `olo_asset_move` / `olo_asset_get` / `olo_asset_references` / `olo_build_list`
  — registered `MainMarshaled = false` *on purpose*, because their whole-project scan must stay off
  the game thread — would freeze the editor if admitted. That is a build-time refusal too. This one
  only became visible once the execution model existed, and came out of the self-review.

## Design notes

- **One pinned `CommandSnapshot`** for the whole transaction, for the same reason a `tools/call` pins
  one: a script reload must not swap the command vector between admitting a batch and running it.
- **One marshaled job for the whole batch.** Steps run through a host decorator that serves their
  `MarshalRead` inline, so no frame runs between steps and no unrelated editor edit can land inside
  the undo group. Re-marshalling from the game thread would also deadlock outright.
- **Rollback is a `CommandHistory` transaction scope**, not N pushes collapsed afterwards:
  `MaxHistorySize` trimming makes stack-depth arithmetic wrong exactly when the stack is full.
- **Symbolic references are whole strings only** (`"${id.field}"`, never a fragment), which removes
  the escaping problem entirely. Resolution to a *declaring step* is build-time, and so is a type
  check against that step's `outputSchema` — the value does not exist yet, but its declared type does.

## Verification

- `McpAutomationTransactionTest.cpp` — **20 tests, all passing** (the 4 acceptance criteria, symbolic
  outputs and their five build-time refusals, dry-run, edit-mode, the grouped-undo mechanics, and
  the surface census).
- Wider sweep, `build-cached` Debug: **1930 tests, exit 0** across
  `Mcp*:*Undo*:*History*:*Scene*:*Automation*:*Entity*:*Component*`.
- **Both** `OloEngine-Tests` and `OloEditor` build (`OloEditor` is not in the test build).
- No rendering change, so the visual-evidence rule does not apply. Tracked goldens deliberately left
  untouched.

## Review guide

**Where I'd look hardest**

1. `OloEditor/src/UndoRedo/EditorCommand.h` — the transaction scope on `CommandHistory` and the
   **resumable** `CompoundCommand::Undo`. Grouping N operations into one entry interacts with the
   pre-existing "keep the entry on the stack when its undo throws" contract: restarting from the end
   on the next Ctrl-Z would re-undo members already taken back. Fixed with a resume cursor and pinned
   by `APartiallyUndoneGroupRESUMESRatherThanRestarting`, but it is the subtlest change here and it
   touches a class the whole editor uses.
2. `OloEditor/src/Automation/AutomationTransactionCommands.cpp:70` — `constexpr batchConsent =
   Granted`. The invariant is "the registry gates `ProjectWrite` at the only door into a handler,
   this command is `ProjectWrite`, therefore reaching the handler means the batch was granted." The
   handler cannot verify it, so if that premise is ever false this is the hole. Argued in the comment
   and tested from both sides.
3. `OloEditor/src/Automation/AutomationTransaction.cpp` — `ResolveDeclaredOutput` and the build-time
   placeholder validation. It is the one place that can produce a *false* refusal: a legitimate batch
   rejected because the source command's `outputSchema` is incomplete. It only refuses when the schema
   explicitly describes its properties and the path is absent, but that is a judgement call.

**What I verified, and how** — named above. The strongest evidence is the byte-identical serializer
comparison and the single-`Undo()` assertion; both would fail loudly if rollback or grouping were
wrong. The census test asserts the numbers quoted in the analysis doc, so the doc cannot go stale.

**Least confident about** — the ergonomics of refusing a `dryRun:true` batch without write consent.
It is the safe call (classing a command by its arguments rather than its declared tier is the
laundering shape ADR 0005 forbids), but it means an agent cannot *plan* a batch without the write
modal. If that bites, the fix is a separate read-only planning command, not a softer gate here.
Also: a step whose handler succeeds while reporting inside its own payload that it declined to act
(`refused: true`) counts as applied. Reading domain payload fields in a cross-cutting layer is the
coupling this issue exists to avoid, so `IsError` is the only failure signal — stated in the header
and the analysis doc, but it is a deliberate limitation, not an oversight.

**Deliberately not tested** — live-editor behaviour. This is an editor-facing change, but every seam
it touches (`CommandHistory`, `IAutomationHost`, the registry) is exercised with the real production
types in the unit suite, and there is no rendering or frame-timing aspect to observe. The 96
`Unspecified` commands were also deliberately not reclassified: 96 unreviewed declarations riding
along with the layer that consumes them is how a wrong one gets in unnoticed.

Closes #1127


drsnuggles8:
## 🤖 Self-review @ `ea8ae2db7fc93b7e5b85d2d57b42bb9c972ebf10`

Reviewed the PR diff at **high** effort before opening, per the task loop's Phase 3.

- **Findings:** 6 · **Fixed:** 5
  1. `olo_asset_*` / `olo_build_list` are registered `MainMarshaled = false` on purpose (their whole-project scan must stay off the game thread) but were admitted as transaction steps, which would have frozen the editor for the scan's duration inside the batch's main-thread critical section. Added `Transactability::NotMainMarshaled` as a build-time refusal; the transactable set narrowed from 16 to 12, and the analysis doc, the guide and the census ratchet were updated to match.
  2. An *incomplete* rollback still reported `appliedSteps: 0` and marked every pre-failure step `rolledBack`, with only the free-text `rollbackError` dissenting — a half-applied scene dressed up as a clean failure. Those steps now report `rollbackFailed`, `appliedSteps` stays at the applied count, and `CompoundCommand::UndoAll` reports how many operations it could not take back.
  3. A committed group whose member refuses its undo left the entry half-undone, so the next Ctrl-Z re-undid the members already taken back and never advanced. `CompoundCommand::Undo` now resumes from where it stopped. Pinned by `APartiallyUndoneGroupRESUMESRatherThanRestarting`.
  4. `BuildTransaction` and `DescribeInadmissibleBatch` called `json::value()` on untyped fields, throwing `type_error` instead of producing the refusal report they promise. Unreachable through MCP (schema validation runs first) but reachable through the documented direct entry points. Now typed defensively.
  5. `Execute()` withholds the version bump inside a transaction, so a step reading document state mid-batch got the pre-batch answer. `IsDirty()` now accounts for an open transaction that has applied something; the undo/redo *availability* flags deliberately still describe the committed stack, since `Undo()` refuses while a transaction is open.

- **Dismissed:** a step that succeeds while reporting inside its own payload that it declined to act (`refused: true`) is recorded as applied. Reading a domain's payload fields to decide whether a batch aborts would put one domain's vocabulary inside the cross-cutting layer, which is the coupling #1127 exists to avoid. `IsError` is the failure signal; a command that wants to abort a batch says so with an error result. Documented in `AutomationTransaction.h` and the analysis doc rather than silently dropped.

Re-ran Phase 2 after the fixes: 1930 tests, exit 0; both `OloEngine-Tests` and `OloEditor` build.

drsnuggles8:
Self-review addendum — head moved to `d2859eddd1125b2360e28b1a2fcbaa71f14b610e` after the comment above.

Two commits since, both from continuing to re-read my own diff rather than from CI or review:

- `c55f511c9` docs only — the `AutomationUndo` rule added to `notes-mcp-tool-authoring.md` §5b and linked from both halves of the agent-rules README, so the next command author meets it next to the consent gate they are already copying.
- `d2859eddd` — the `HistoryScopeGuard` destructor could throw (`RollbackTransaction` allocates while building its message). It runs only while an exception is already escaping `ExecuteTransactionOnMainThread`, so a throw there would terminate the editor instead of reporting a failed batch. Now swallows, with the reasoning in the comment: closing the scope is the more important job, since an open one refuses every later Undo/Redo/Clear.

Re-verified at this head: 152 tests across `McpAutomation*:*Undo*:*History*`, exit 0; both `OloEngine-Tests` and `OloEditor` build. Nothing in the review guide changes.

drsnuggles8:
@coderabbitai review

Issue #1187, opened 2026-09-11
Asset loading: an intentional texture-path probe logs at error level, so a successful load emits 'Failed to load'


drsnuggles8:
`AnimatedModel::LoadMaterialTextures` deliberately probes several candidate paths for a material texture, and the comment above the last fallback says so explicitly:

```cpp
// Fallback: scan model directory for case-insensitive stem match with any image extension.
// Handles FBX referencing .tga when the actual file is .png, and case mismatches on Linux.
```

That design is fine and it works. The problem is that each *failed probe* reaches `Texture2D::Create`, which logs at **error** level from [`OpenGLTexture.cpp:386`](https://github.com/drsnuggles8/OloEngineBase/blob/master/OloEngine/src/Platform/OpenGL/OpenGLTexture.cpp#L386):

```cpp
OLO_CORE_ERROR("Image texture data is null! Failed to load: '{}'", path);
```

So a load that ultimately **succeeds** emits errors on the way:

```
[error] OloEngine: Image texture data is null! Failed to load: 'assets/models/Cerberus\Textures\Cerberus_A.tga'
[error] OloEngine: Image texture data is null! Failed to load: 'assets/models/Cerberus\Cerberus_A.tga'
```

`assets/models/Cerberus/` contains `cerberus_A.png`, `cerberus_M.png`, `cerberus_N.png`, `cerberus_R.png` and no `.tga` at all - the `.tga` names come from the FBX's embedded material, which is exactly the case the fallback exists to absorb.

## Repro

Needs a **cold** mesh/texture cache - a warm `OloEditor/assets/cache/mesh/` skips texture loading entirely, which is why this does not show on a second run:

```
rm -rf OloEditor/assets/cache/mesh
driver.ps1 -Action attach -AllowWrites
olo_scene_open { "path": "Scenes/PBRModelTest.olo" }
olo_log_tail { "count": 200, "minLevel": "error" }
```

Also reachable via `Scenes/ModelLoadingTest.olo`, which uses the same model.

## Why it is worth fixing

- It is the only `[error]`-level output in an otherwise clean editor session, so "are there errors in the log?" stops being a useful question - and this repo's own convention is that a path which cannot do its job should say so loudly *and countably*. An intentional probe is not that path.
- It inflates any error-count check built on `olo_log_tail --minLevel error`, and the two lines look alarming in the Console panel next to a model that actually rendered fine.

## Suggested shape

[...]

Issue #1186, opened 2026-09-11
Render graph: GTAOEdge and GTAODenoisePong are always consumed-but-unbacked, so olo_render_validate never returns ok


drsnuggles8:
`olo_render_validate` reports `ok: false` on **every** frame, with `consumedButUnbacked: ["GTAODenoisePong", "GTAOEdge"]`. Because it never returns clean, the tool currently carries no signal: a real render-graph regression would look exactly like the steady state.

## Measured

54 captures - 9 scenes x 3 rendering paths x 2 backends, on a fresh editor each time:

| | forward | forwardplus | deferred |
|---|---|---|---|
| OpenGL | unbacked on 9/9 scenes | 9/9 | 9/9 |
| Vulkan | 9/9 | 9/9 | 9/9 |

Same two resources every time, `hazardCount: 0`, `resolveFailures: []`, `buildDiagnostics: []`. Scenes covered: MaterialLab, LightingTest, SponzaDeferred, PBRModelTest, SSRTest, TerrainTest, DDGITest, WaterShowcase, SnowfallParticles.

Ambient occlusion itself looks correct in the viewport, and `olo_render_list_targets` shows `AOBuffer` (R8UNorm) with `producers: ["GTAOPass"]`, so the pass is running and producing output.

## Where they come from

Both are graph-only transients allocated in [`RenderPipeline.cpp`](https://github.com/drsnuggles8/OloEngineBase/blob/master/OloEngine/src/OloEngine/Renderer/RenderPipeline.cpp#L3936-L3954) via `declareGraphOnlyTexture` -> `graph.AllocateTransientTextureHandle`, and declared in [`GTAORenderPass.cpp`](https://github.com/drsnuggles8/OloEngineBase/blob/master/OloEngine/src/OloEngine/Renderer/Passes/GTAORenderPass.cpp#L183-L219) with the intra-pass write-then-read pattern:

```cpp
builder.AllowSamePassReadWrite(blackboard.Scratch.GTAOEdge);
builder.Write(blackboard.Scratch.GTAOEdge, RGWriteUsage::ShaderImage);
const auto edgeRead = builder.Read(blackboard.Scratch.GTAOEdge, RGReadUsage::ShaderImage);
```

[...]

Issue #1171, opened 2026-09-10
RHI/Vulkan: particles invisible — a CPU counter reset racing the emit dispatch, and camera-facing billboards culled by winding


drsnuggles8:
Split out of #1013's validation (see the [pass audit](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/vulkan-parallel-pass-audit.md)), where the empty `SnowfallParticles` sample was recorded as "cause not established".

**The cause is now narrowed: it is the Vulkan backend, not the sample and not parallel recording.**

## Repro

`SandboxProject/Assets/Scenes/SnowfallParticles.olo`. The authored camera does not frame the emitter (it sits at `y = 25` with 0.04-unit particles), so pose the camera above it first:

```
olo_camera_set_pose { "position": [0, 20, 45], "target": [0, 22, 0] }
```

| Backend | Result |
|---|---|
| `--rhi=opengl` | snow renders — ~3,000 bright billboard pixels in the frame |
| `--rhi=vulkan` | none — the frame is byte-static across 24 s while `frameIndex` advances |

Both in the same clang-cl **Release** build, same scene, same pose, editor visible and ticking (`iconified:false`, `stale:false`, frameIndex 3518 → 3719 over the sample).

## What it is not

- **Not parallel recording.** Identical with `OLO_VK_PARALLEL_RECORDING` off and on. Billboard batching was unchanged by #1066.
- **Not a culled pass.** `olo_render_frame_breakdown` shows `ParticlePass` present with `culled: false`, `executionIndex: 20` on Vulkan.
- **Not the edit-mode tick.** `Scene::OnUpdateEditor` previews particle systems, and OpenGL proves the simulation runs from the same code.
- **Not mesh particles.** Those have a passing Vulkan off/on readback from #1066 (96 instances, three colour ranges).

So it is specific to the **billboard** draw path on Vulkan.

## Suggested first look

The Vulkan frame does show a few faint diagonal streaks low in the view where OpenGL shows dense white flakes, so something in that path is emitting geometry — check the billboard batch's vertex pulling / instance buffer and its descriptor set, in the same family as the mesh-particle vertex-pulling fix that shipped in #1066.

## Score

```olo-score

capability: 2

craft: 5

stability: 5

decay: 2

effort: 5

confidence: 0.7

learning: 3

fun: 3

kano: must-be

blocked_by: []

blocks: []

```

[...]

PR #1188, opened 2026-09-11
fix(vulkan): stop the frame arena's per-item block tails exhausting the slot (#1185)

### LINKED ISSUE #1185: RHI/Vulkan: IBL and skybox are silently lost for every scene loaded after the second distinct scene
Loading **two distinct scenes** in a Vulkan editor session makes every scene opened afterwards render with **no image-based lighting and no skybox**. Nothing is logged, so the frame is silently wrong rather than loudly broken.

OpenGL is unaffected by the identical sequence.

## Repro

Fresh editor, Vulkan, Deferred path. Three `olo_scene_open` calls, nothing else:

```
driver.ps1 -Action attach -AllowWrites -Rhi vulkan
olo_renderer_settings_set { "setting": "renderpath", "value": "deferred" }
olo_scene_open { "path": "Scenes/SponzaCSM.olo" }
olo_scene_open { "path": "Scenes/WaterShowcase.olo" }
olo_scene_open { "path": "Scenes/Benchmark/MaterialLab.olo" }
olo_screenshot { "maxWidth": 700, "settleFrames": 15,
                 "orbit": { "target": [0.6, 0.4, 2.4], "yaw": 35, "pitch": 14, "distance": 15.55 } }
```

`MaterialLab` is the detector because its spheres are metal — their shading is *almost entirely* IBL, so losing the environment turns them flat matte grey. The loft HDRI background disappears at the same time.

| backend | same 3 opens, same pose | viewport luminance mean / stddev |
|---|---|---|
| `--rhi=opengl` | correct: loft HDRI background, IBL specular, coloured emissives | 130.8 / 60.7 |
| `--rhi=vulkan` | flat matte grey spheres, grey background, no skybox | **81.0 / 8.0** |

Stddev is the useful number: the lit frame has ~60, the broken frame ~8.

## It is the *number of distinct* scene loads, not the scenes

Each row is a fresh Vulkan editor, Deferred, opens only — no camera moves, no captures in between — then MaterialLab and one screenshot at the pose above.

| prefix opened before MaterialLab | distinct scenes | result | lum / std |
|---|---|---|---|
| `SponzaForwardPlus` | 1 | correct | 139.4 / 62.7 |
| `WaterShowcase` | 1 | correct | 139.4 / 62.7 |
| `Example` x6 (same scene six times) | 1 | correct | 139.4 / 62.7 |
| `SponzaCSM`, `WaterShowcase` | 2 | **broken** | 81.0 / 8.0 |
| `Example`, `SponzaDeferred`, `SponzaForwardPlus` | 3 | **broken** | 81.0 / 7.9 |
| `SponzaCSM`, `Drift`, `WaterShowcase` | 3 | **broken** | 81.0 / 8.0 |
| all six of the above | 6 | **broken** | 81.0 / 7.9 |

Re-opening the *same* scene six times does not trigger it, so this is not a per-load leak or a load counter — it takes two **different** scenes. Once triggered it persists for the rest of the session: re-opening MaterialLab directly afterwards still renders unlit.

Note the broken measurements are bit-identical across runs (81.0 / 7.9-8.0), so the failed state is deterministic, not a race that settles.

## What it is not

- **Not the rendering path.** Reproduced on `forward`, `forwardplus` and `deferred`, with the same 81.0 / 8.0 frame on each.
- **Not settling or a stale frame.** `stale:false`, `ticking:true`, `frameIndex` advancing; re-captured minutes later with `settleFrames: 20`, unchanged.
- **Not the HDRI or the asset path.** `SponzaDeferred` uses the *same* `assets/textures/hdr/newport_loft.hdr` environment component and renders correctly when it is the first scene.
- **Not a missing IBL bake.** On the broken frame, `olo_render_capture_target` still returns a populated `IrradianceMap` (32px, correct loft-coloured content) and a correct red/green split-sum `BrdfLut` (256px). The maps are built; their contribution just never reaches shading.
- **Not a skipped pass.** `olo_render_frame_breakdown` on the broken frame shows `DrawSkybox: 1` submitted in the ScenePass bucket alongside 24 `DrawMesh` + 4 `DrawMeshInstanced`.
- **Not resource exhaustion.** `olo_render_gpu_resources` reports 2.17 GB of a 20.2 GB device-local budget, 1036 allocations.
- **Not reported anywhere.** `olo_log_tail --minLevel error` is empty after the open, `olo_shader_errors` is 0, and `olo_project_validate` flags nothing renderer-side beyond the pre-existing GTAO entry.

## Why it matters beyond this scene

MaterialLab makes it obvious, but the same lost contribution is invisible in sun-dominated scenes: `SponzaCSM` and `PBRModelTest` look plausible in the broken regime because an analytic directional light dominates their shading. So any Vulkan session that has switched scenes a couple of times is quietly rendering without ambient specular, and a person would likely read it as "Vulkan looks a bit flatter" rather than a bug. That is also why this did not surface earlier.

## Suggested first look

Something tied to the environment/IBL binding is invalidated on a scene swap and never republished — the shape matches a resting-layout / descriptor-publish miss rather than a shader bug, since the textures themselves survive and the draw is still submitted. Worth checking whether the environment cubemaps are re-published to the descriptor heap after `Scene` teardown, and whether the second distinct teardown is what drops them.

Possibly related, logged once at Vulkan editor start (and only on Vulkan):

```
[RHI/Vulkan] 'PBR_MultiLight' STORAGE binding 63 has no published occupant - substituting the null block ...
[RHI/Vulkan] 'PBR_MultiLight' STORAGE binding 68 has no published occupant ...
[RHI/Vulkan] 'PBR_MultiLight' STORAGE binding 78 has no published occupant ...
[RHI/Vulkan] 'PBR_MultiLight' buffer binding 81 has no published occupant - null block
[RHI/Vulkan] 'PBR_MultiLight' buffer binding 82 has no published occupant - null block
```

Unproven as the cause - it fires before any scene swap, in the working regime too.

## Environment

Measured 2026-09-11 on master `75dd98f33`, Debug, `dev-cached` (clang-cl), RTX 4090 / NVIDIA 616.64, Vulkan SDK 1.4.309.0. Full suite green (8136 passed / 40 skipped / 0 failed) on the same build.

## Regression test

There is no automated coverage for this: no test renders a pure-IBL scene after a scene swap on Vulkan, which is exactly the gap that let it through.

## Score

```olo-score

capability: 3

craft: 5

stability: 8

decay: 3

effort: 5

confidence: 0.8

learning: 3

fun: 3

kano: must-be

blocked_by: []

blocks: []

```

drsnuggles8:
Fixes #1185.

## What was wrong

`VulkanFrameArena`'s worker root-data block was claimed **per recording item** and dropped at every fork, so a frame paid `kWorkerBlockBytes` (64 KiB) per item regardless of what the item actually pushed. A deferred frame records on the order of 255 items — 16.7 MiB of a 16.78 MiB slot, almost all of it untouched block tails. Past that the arena has nothing left and every root-data push is dropped, and a dropped push is not a crash: the draw samples an **unbound buffer**, so materials shade with no data and the skybox never gets its constants.

That is #1185, which presented as "Vulkan loses IBL and the skybox after two distinct scene loads". The scene count was a red herring — it takes a frame with enough recorded items, and `MaterialLab`'s 119 entities on the deferred path is simply the first scene in that sequence to get there.

Measured on the repro, by logging the previous frame before `BeginFrame` rewinds it:

| after opening | slot cursor at frame end | counted allocations | overflows |
|---|---|---|---|
| `SponzaCSM` | 0 | 461 | 0 |
| `WaterShowcase` | 90 960 | 105 | 0 |
| **`MaterialLab`** (3rd) | **16 727 056** | **0** | **22 398** |

`16 727 056 / 65 536 = 255.2` — one block per item. Counted allocations read 0 because the worker path tallies into the item's own counter and folds it only at the join, which is why no existing stat made this visible.

## The fix

- **The block is frame-scoped, not command-buffer-scoped.** It survives `ResetForCommandBuffer` and is keyed on the arena's frame generation instead — reused within a frame, dropped once the slot underneath has been rewound. Keying on the slot index alone would alias frame N with frame N+2, which is the hazard `GetFrameGeneration()` already exists to close.
- **`kWorkerBlockBytes` 64 KiB -> 8 KiB.** The block should be sized against how many blocks a frame claims, not against one item's appetite. 8 KiB keeps the property the block exists for — one shared CAS per item rather than per draw — while bounding a 255-item frame at ~2 MiB.
- **The overflow report was lying.** It printed the claim helper's out-parameter, which is left untouched on failure, so it read `at cursor 0 of 16777216 B` — impossible-looking next to a full slot, and easy to dismiss. It now reads the real cursor, says draws will sample unbound buffers, and re-arms per frame instead of latching for the process, because an overflow that repeats every frame is a permanently wrong image and one line made it look transient.

## Verification

Repro (`SponzaCSM` -> `WaterShowcase` -> `MaterialLab`, Vulkan, deferred), same camera pose:

| | viewport lum / stddev | arena overflows |
|---|---|---|
| before | 81.0 / **8.0** | 22 398 and climbing |
| after | 139.4 / **62.7** | **0** |

139.4 / 62.7 is bit-identical to a clean first-load frame, and the `'Skybox_GBuffer' buffer binding 0 has no published occupant` warning is gone.

A 9-scene x 3-path Vulkan matrix (27 captures: MaterialLab, LightingTest, SponzaDeferred, PBRModelTest, SSRTest, TerrainTest, DDGITest, WaterShowcase, SnowfallParticles on forward / forward+ / deferred) shows RMSE ~0 against the pre-fix captures for every frame that was already rendering correctly — nothing else moved.

Full suite green in Debug: **8178 ran, 8138 passed, 40 skipped, 0 failed**, exit 0. That is exactly +2 on the pre-change baseline, the two tests below.

## Tests

`VulkanParallelRecordingDevice.ItemCountDoesNotExhaustTheFrameArena` and `.WorkerArenaBlockIsDroppedWhenTheFrameRewinds` (layer `plumbing`, device-gated like the rest of that file).

They fail on the old code with `a 255-item frame consumed 16711680 B of the slot`. Note the budget assertion trips **before** an actual overflow: 255 x 64 KiB just barely fits a bare slot, and it was the additional shared-path allocations that pushed production over. Asserting only "no overflow" would let an equivalent regression through.

## Second commit — a ride-along, unrelated to the arena

`fix(shaders): a diamond include is not a circular include`.

`OpenGLShader::ProcessIncludesInternal` threads one `includedFiles` set through the whole recursion and never pops it, so it implements include-once — correct for GLSL, which has no `#pragma once`. The second visit it reported, though, is almost never a cycle; it is a diamond. One editor session logged "Circular include detected" 81 times, 47 of them for `include/GPUScene.glsl` and 21 for `include/MathCommon.glsl` — leaf headers that contain no `#include` at all and cannot be part of a cycle by construction. Skipping was correct; only the diagnostic was wrong. Demoted to trace with an accurate message; verified on a cold shader cache at 85 -> 0 warnings.

## Not addressed here

- `#1186` (GTAO transients always consumed-but-unbacked) and `#1187` (texture-probe logging at error level) are untouched.
- Eleven `has no published occupant` bindings survive this fix and are therefore independent of it — seven of them at error level. The set shrank from 26 to 11, so the other fifteen (`Skybox_GBuffer`, `Particle_Billboard_GPU` 0/2, GTAO, the PostProcess shaders) were symptoms of the arena exhaustion. `Particle_Billboard_GPU` disappearing may be relevant to #1171.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Vulkan rendering reliability for complex frames with many parallel recording operations.
  * Prevented material and skybox data from being dropped when frame memory usage is high.
  * Fixed stale temporary memory being reused after a frame reset.
  * Improved overflow diagnostics, including accurate allocation details and per-frame warnings.
  * Reduced unnecessary memory consumption during parallel command recording.

### DIFF
--- a/OloEngine/src/Platform/Vulkan/VulkanFrameArena.cpp
+++ b/OloEngine/src/Platform/Vulkan/VulkanFrameArena.cpp
@@ -201,6 +201,8 @@ namespace OloEngine
         m_Slots[m_CurrentSlot].Cursor.store(0, std::memory_order_relaxed);
         ++m_FrameGeneration;
         m_AllocationsThisFrame.store(0, std::memory_order_relaxed);
+        // One line per affected frame — see the overflow reporter.
+        m_OverflowWarned.store(false, std::memory_order_relaxed);
     }
 
     VulkanFrameArenaAllocation VulkanFrameArena::Allocate(u64 sizeBytes, u64 alignment)
@@ -262,14 +264,23 @@ namespace OloEngine
                 // thread advanced it to; recompute against that.
             }
         };
-        const auto overflow = [&](const u64 at)
+        const auto overflow = [&]()
         {
-            m_OverflowCount.fetch_add(1, std::memory_order_relaxed);
+            const u64 total = m_OverflowCount.fetch_add(1, std::memory_order_relaxed) + 1u;
+            // Re-armed by BeginFrame, not latched for the process: an overflow
+            // that repeats every frame is a permanently wrong image, and a
+            // one-shot line made that look like a single transient hiccup
+            // (issue #1185). The cursor is READ here rather than passed in —
+            // the claim helper leaves its out-parameter untouched on failure,
+            // so the old call site reported a cursor of 0 next to a 16 MiB
+            // capacity and read as impossible.
             if (!m_OverflowWarned.exchange(true, std::memory_order_relaxed))
             {
-                OLO_CORE_ERROR("VulkanFrameArena: slot {} overflow ({} B requested at cursor {} of {} B) — "
-                               "root-data allocations are being DROPPED this frame",
-                               m_CurrentSlot, sizeBytes, at, kSlotCapacityBytes);
+                OLO_CORE_ERROR("VulkanFrameArena: slot {} overflow ({} B requested, cursor at {} of {} B; {} dropped "
+                               "since start) — root-data allocations are being DROPPED this frame, so draws will "
+                               "sample unbound buffers",
+                               m_CurrentSlot, sizeBytes, slot.Cursor.load(std::memory_order_relaxed),
+                               kSlotCapacityBytes, total);
             }
             return VulkanFrameArenaAllocation{};
         };
@@ -281,13 +292,22 @@ namespace OloEngine
         if (auto* worker = CurrentVulkanWorkerContext(); worker != nullptr && sizeBytes <= kWorkerBlockBytes / 2u)
         {
             auto& block = worker->Arena;
+            // A block claimed in an earlier frame points into a slot that
+            // BeginFrame has since rewound to 0, so reusing it would hand out
+            // memory another allocation already owns. Anything from THIS frame
+            // is still ours, whichever command buffer claimed it.
+            if (worker->ArenaFrameGeneration != m_FrameGeneration)
+            {
+                block = {};
+                worker->ArenaFrameGeneration = m_FrameGeneration;
+            }
             u64 aligned = Align(block.Cursor, alignment);
             if (block.End == 0u || aligned + sizeBytes > block.End)
             {
                 u64 blockOffset = 0;
                 if (!claimShared(kWorkerBlockBytes, std::max<u64>(alignment, 256u), blockOffset))
                 {
-                    return overflow(blockOffset);
+                    return overflow();
                 }
                 block.Cursor = blockOffset;
                 block.End = blockOffset + kWorkerBlockBytes;
@@ -305,7 +325,7 @@ namespace OloEngine
         u64 aligned = 0;
         if (!claimShared(sizeBytes, alignment, aligned))
         {
-            return overflow(slot.Cursor.load(std::memory_order_relaxed));
+            return overflow();
         }
         m_AllocationsThisFrame.fetch_add(1, std::memory_order_relaxed);
         return {
--- a/OloEngine/src/Platform/Vulkan/VulkanFrameArena.h
+++ b/OloEngine/src/Platform/Vulkan/VulkanFrameArena.h
@@ -109,7 +109,22 @@ namespace OloEngine
         // Worker contexts claim this much from the shared cursor at a time
         // and bump inside it (VulkanRecordingContext::ArenaBlock). Requests
         // larger than half of it go to the shared cursor directly.
-        static constexpr u64 kWorkerBlockBytes = 64ull * 1024;
+        //
+        // SIZED AGAINST THE NUMBER OF BLOCKS A FRAME CLAIMS, not against one
+        // item's appetite. The block is per recording ITEM, and a deferred
+        // frame records on the order of 255 of them, so the slot pays this
+        // much per item whether the item pushes 80 bytes or fills the block.
+        // At the original 64 KiB that is 16.7 MiB of a 16.78 MiB slot — the
+        // whole budget, spent almost entirely on untouched block tails, after
+        // which every root-data push is dropped and the frame renders with
+        // unbound material and skybox buffers (issue #1185). Measured on the
+        // #1185 repro: 255 claims/frame, 16 719 616 B of cursor, 20 000+
+        // dropped allocations.
+        //
+        // 8 KiB keeps the property the block exists for — one shared CAS per
+        // item instead of one per draw, since a typical item's pushes are a
+        // few hundred bytes — while bounding a 255-item frame at ~2 MiB.
+        static constexpr u64 kWorkerBlockBytes = 8ull * 1024;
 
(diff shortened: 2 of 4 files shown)


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

Issue #1193, opened 2026-09-11
ASan/EH: 13 sites still throw from inside a catch handler; the build flag masks them, the rule says remove them


drsnuggles8:
`docs/agent-rules/build-trees-and-windows-asan.md` §4b states the rule: **never execute a throw from inside a `catch` handler in code the Windows ASan build compiles.** Under clang-cl + ASan, a throw executed lexically inside a handler makes `__CxxFrameHandler3` read the handler funclet's parent frame as NULL and fault. `cmake/Sanitizers.cmake` masks it today with `-fsanitize-address-use-after-return=never` (`8e5050ad4`, 2026-09-11), and the doc is explicit that the flag is a mask, not a fix: *"a site that slips through will not crash today — the rule keeps it from coming back the moment that flag is reconsidered."*

A brace-balanced scan of every `catch` body in `OloEngine/src` and `OloEditor/src` finds **13 sites** that throw inside the handler. All are bare `throw;` rethrows.

| file | line |
|---|---|
| `OloEditor/src/Automation/AutomationAssetCommands.cpp` | 761 — `AssetMoveCommand::Apply` |
| `OloEditor/src/Automation/AutomationFileWrite.cpp` | 82 — `ReplaceFileContents` |
| `OloEditor/src/Automation/AutomationSceneDocument.cpp` | 101, 153, 167 |
| `OloEditor/src/MCP/McpToolsCamera.cpp` | 297 |
| `OloEditor/src/MCP/McpToolsInput.cpp` | 147 |
| `OloEditor/src/MCP/McpToolsRender.cpp` | 2497 |
| `OloEngine/src/OloEngine/Core/Application.cpp` | 143, 281 |
| `OloEngine/src/OloEngine/Scene/SceneSerializer.cpp` | 6929 |
| `OloEngine/src/Platform/Vulkan/VulkanContext.cpp` | 1244 |
| `OloEngine/src/Platform/Windows/WindowsWindow.cpp` | 139 |

## Evidence that these are live, not theoretical

The first row is not hypothetical. `AutomationAssetCommandsTest.UndoRefusesWhenAReferringFileChangedUnderneath` exercises it — a refusal thrown by `RequireFileContents` inside `Apply`'s `try`, caught by `catch (...) { RevertRewrites(...); throw; }` — and on every Windows ASan shard run against a base **without** `8e5050ad4` it died with

```
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
```

[...]

PR #1195, opened 2026-09-11
fix(shaders): a diamond include is not a circular include

drsnuggles8:
Split out of #1188 at CodeRabbit's request — it flagged this as out of scope there, and it is: a diagnostic fix unrelated to the frame arena. Refs #1193's family of "correct path, wrong voice" diagnostics; no issue of its own.

## What was wrong

`OpenGLShader::ProcessIncludesInternal` threads one `includedFiles` set through the whole recursion and never pops it, so it implements **include-once** — the right behaviour for GLSL, which has no `#pragma once` and where a second copy of a header is a redefinition error.

The second visit it reported, though, is almost never a cycle. It is a **diamond**: two headers both including the same utility header. One editor session logged `Circular include detected` **81 times**, 47 of them for `include/GPUScene.glsl` and 21 for `include/MathCommon.glsl` — leaf headers that contain no `#include` at all and so cannot be part of a cycle by construction.

Skipping was correct; only the diagnostic was wrong.

## Change

Demoted to trace with an accurate message (`Shader include already present, skipping`), and the marker left in the generated source now reads `Already included` rather than `Circular include`.

## Verification

Cold shader cache (the include processing only runs on a cache miss, so a warm cache would have hidden nothing either way): a full editor start plus scene loads goes from **85 → 0** `Circular include` warnings, with 85 `already present, skipping` traces in their place — same events, correct severity.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shader include diagnostics by treating repeated includes as already processed rather than reporting them as circular includes.
  * Reduced the severity of related log messages to trace level for less noisy output.

### DIFF
--- a/OloEngine/src/Platform/OpenGL/OpenGLShader.cpp
+++ b/OloEngine/src/Platform/OpenGL/OpenGLShader.cpp
@@ -1105,11 +1105,23 @@ namespace OloEngine
 
                     const std::string fullPathStr = fullPath.string();
 
-                    // Check for circular includes
+                    // INCLUDE-ONCE, not cycle detection. `includedFiles` accumulates
+                    // every header pulled in anywhere under this shader and is never
+                    // popped, so a second visit means the header is already in the
+                    // translation unit — which is what we want for GLSL, where there
+                    // is no #pragma once and a second copy is a redefinition error.
+                    //
+                    // The overwhelmingly common second visit is a DIAMOND, not a
+                    // cycle: two headers both including MathCommon.glsl. Reporting it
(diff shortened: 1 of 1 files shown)


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

Issue #1198, opened 2026-09-11
Vulkan: device fault on a live forward→forward+ switch when GTAO's async batch runs on pooled (not re-created) scratch textures


drsnuggles8:
## Correction (2026-09-12 01:40): this is on master, and it is timing-dependent

The "master, cap 2 = clean" row below was **one** run. Replaying plain master `d9a00323c` (nothing changed, default cap of 2) three times on the identical sequence:

| run | result |
|---|---|
| 1 | clean |
| 2 | **FAULT** |
| 3 | **FAULT** |

So the pool cap does not gate it, #1199 does not introduce it, and the eleven-arm matrix below is only as strong as single runs of a ~2-in-3 fault allow: the arms that faulted are real, the arms that were "clean" once are not evidence. What still stands with more than one observation: `OLO_VK_ASYNC_COMPUTE=0` was clean across three consecutive switches (forward+, deferred, forward) in one session, and so was `GTAORenderPass::SetAsyncComputeCandidate(false)`. Treat those as leads, not proofs, until repeated.

It also reproduced on `feature/unbound-ssbo-bindings-1190` (#1188 + #1192, no pool change) on its first switch.

**Repro for the record:** Debug editor, `--rhi=vulkan`, MaterialLab, `olo_renderer_settings_set renderpath=forwardplus`; `scratchpad/gtao/replay3.sh <tag> <runs>` builds once and replays N times.

---
## Symptom

On Vulkan, with the render-graph transient pool allowed to keep **four** same-spec textures per descriptor bucket instead of two, switching MaterialLab from `forward` to `forwardplus` at runtime device-faults about two seconds later:

```
[Vulkan] DEVICE FAULT REPORT: '' (2 address record(s), 0 vendor record(s))
[Vulkan]   fault address: 0x200720ee0 (precision ±0x10)   — instruction pointer (faulting)
[Vulkan]   fault address: 0x3ee400000 (precision ±0x1000) — READ of invalid address
=== CRASH: C++ Exception (caught) — Vulkan bring-up: vkWaitForFences failed (VkResult -4) ===
```

[...]

PR #1199, opened 2026-09-11
fix(rendergraph): keep what the frame needed when trimming the transient pool (#1186)

### LINKED ISSUE #1186: Render graph: GTAOEdge and GTAODenoisePong are always consumed-but-unbacked, so olo_render_validate never returns ok

drsnuggles8:
Closes #1186.

## What was actually wrong

`olo_render_validate` reported `GTAOEdge` and `GTAODenoisePong` as consumed-but-unbacked on every frame, on every scene, path and backend. The issue guessed at a validator false positive or an unused-transient leak. It is neither: the transient plan allocates all four GTAO scratch textures (`AOBuffer`, `GTAODenoisePing`, `GTAODenoisePong`, `GTAOEdge` — one alias group, `1:1:1411x942x1:m1:s1:q0`, slots 0–3), but `TransientPool::Trim()` runs after every frame with a cap of **2 per descriptor bucket**. Four same-spec textures against a cap of two means the two highest slots are created fresh every frame and destroyed at trim. The live plan showed it directly: the two unbacked objects carried identity generation **23441** — one create/destroy per frame for the life of the session — where their siblings sat at generation 3–5.

So AO rendered correctly (the storage is live during `Execute`), the validator was right (the objects are gone by the time a between-frames read looks), and the engine was allocating and freeing two 1.3 MB textures every frame.

## Fix

`TransientPool::ReleaseAll()` now counts what each bucket handed out in the frame it closes, and `Trim(cap)` keeps `max(cap, that demand)`. The cap still bounds buckets nobody needs any more (the bloom-on-then-off case it exists for); it no longer turns a bucket's steady-state demand into per-frame churn. Eviction also takes the **front** of a bucket first — `ReleaseAll` appends the frame's returns, so the front is whatever nobody used this frame. The old `resize()` kept the stale leftovers and dropped the object the frame had just used.

`RenderGraph::SetTransientPoolMaxBucketSize`'s doc says the same.

## Test

`RenderGraphTransientPool.TrimKeepsWhatTheLastFrameNeeded` (GPU-backed like its neighbours, skips headless): four same-spec acquisitions survive a `Trim(2)`; the next frame is served entirely from the pool (same four objects); a frame needing one trims to the cap and keeps the object it used; an idle frame trims to the cap.

## Verified live

- **OpenGL:** `olo_render_validate` → `ok: true` on MaterialLab and SponzaDeferred across forward / forward+ / deferred; the four GTAO identities hold generation 5/5/5/3 across three samples 10 s apart (before: 23441 and climbing ~340/s).
- **Vulkan:** the same matrix is clean with `OLO_VK_ASYNC_COMPUTE=0`. With async compute on, a live forward→forward+ switch on MaterialLab device-faults about two seconds later. That is **not this change**: master with nothing but `m_TransientPoolMaxBucketSize = 4u` faults identically, so the old cap of 2 — by re-creating two of the four GTAO scratch textures every frame — was masking a Vulkan async-compute bug. Fully bisected (11 arms) in #1198.

## Draft, blocked on #1198

Merging this without #1198 would turn a hidden bug into a device fault for anyone switching render paths on Vulkan with async compute enabled. Everything else in this PR stands on its own.

### DIFF
--- a/OloEngine/src/OloEngine/Renderer/RenderGraph.h
+++ b/OloEngine/src/OloEngine/Renderer/RenderGraph.h
@@ -1070,7 +1070,9 @@ namespace OloEngine
         // Set the maximum number of pool objects retained per descriptor bucket
         // after each frame. Excess objects are evicted by TransientPool::Trim()
         // at the end of Execute(). Default = 2 (keeps one spare for same-descriptor
-        // overlapping transients; use 1 for the most aggressive trim).
+        // overlapping transients; use 1 for the most aggressive trim). A bucket
+        // the frame just drawn needed more of keeps that many regardless — the
+        // cap bounds idle buckets, never a frame's live demand (issue #1186).
         void SetTransientPoolMaxBucketSize(u32 maxPerBucket)
         {
             m_TransientPoolMaxBucketSize = maxPerBucket;
--- a/OloEngine/src/OloEngine/Renderer/TransientPool.cpp
+++ b/OloEngine/src/OloEngine/Renderer/TransientPool.cpp
@@ -127,30 +127,51 @@ namespace OloEngine
         // a confidently wrong "nothing was acquired", which is worse than no tool.
         m_LastFrameAcquireOrder = BuildAcquireOrder();
 
-        // Return all acquired objects to their pools
+        // A release with nothing acquired is not a frame: BuildFrameGraph makes
+        // one defensively at the START of every frame, right before its Trim.
+        // Resetting the demand there would let that trim evict to the cap what
+        // the end-of-frame trim had just kept, and the churn this exists to
+        // stop would simply move to the other end of the frame (every one of
+        // the four GTAO textures then re-created per frame, not two).
+        const bool acquiredAnything =
+            !m_AcquiredTextures.empty() || !m_AcquiredFramebuffers.empty() || !m_AcquiredBuffers.empty();
+        if (!acquiredAnything)
+            return;
+
+        // Return all acquired objects to their pools, counting the frame's
+        // demand per bucket on the way so Trim() knows what it must keep.
+        m_LastFrameTextureDemand.clear();
         for (const auto& tex : m_AcquiredTextures)
         {
             if (tex)
             {
-                m_TexturePool[BuildTextureKey(tex->GetSpecification())].push_back(tex);
+                const auto key = BuildTextureKey(tex->GetSpecification());
+                m_TexturePool[key].push_back(tex);
+                ++m_LastFrameTextureDemand[key];
             }
         }
         m_AcquiredTextures.clear();
 
+        m_LastFrameFramebufferDemand.clear();
         for (const auto& fb : m_AcquiredFramebuffers)
         {
             if (fb)
             {
-                m_FramebufferPool[BuildFramebufferKey(fb->GetSpecification())].push_back(fb);
+                const auto key = BuildFramebufferKey(fb->GetSpecification());
+                m_FramebufferPool[key].push_back(fb);
+                ++m_LastFrameFramebufferDemand[key];
             }
         }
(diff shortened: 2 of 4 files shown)


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

PR #1208, opened 2026-09-12
test(ctest): register the light suites one process each — 8,205 launches become 1,741

drsnuggles8:
Stacked on #1207 (base is its branch; retarget to master once it merges).

## Summary

75 % of a Linux sanitizer test step on the box is the engine's per-process start-up: 7,785 of 8,109 cases take under a second, and each was its own ctest entry paying the 0.64 s launch (an empty run of the Debug binary on the dev box costs the same 0.62 s). #1207 widened the run; this removes most of the launches.

The registration is now **three disjoint passes** over the one binary:

| pass | cases | one process per |
|---|---|---|
| `OLO_MESH_CACHE_TESTS` (`gtest_discover_tests`) | a handful | case, serialised, weighted |
| `OLO_HEAVY_TESTS` (`gtest_discover_tests`) | ~560 | case, `PROCESSORS 2`, `COST 100` |
| everything else (`OloEngine/tests/cmake/OloSuiteDiscovery.cmake`) | 7,641 in 1,161 suites | **suite**: `--gtest_filter=<Suite>.*` minus the same exclusions, sharded past 40 cases via `GTEST_TOTAL_SHARDS` / `GTEST_SHARD_INDEX`, `TIMEOUT` scaled to the case count, `COST` equal to it |

1,741 ctest entries in all, down from 8,205. The heavy and crash-prone cases keep their own processes, which is where the cross-test state #1074 fixed used to hide.

## What stays the same for test authors

- `TempDir()` is still keyed per case beneath a per-process root; the renderer-state and GL-error listeners still restore between cases.
- `GTEST_OUTPUT` XML stays one record per case, so the CI test report and the GPU nightly's "did the GPU tests run" assertion see exactly what they saw.
- Suite entries are named `<Suite>.*`, the filter they run, so every `^Suite\.`-shaped `--exclude-regex` in the workflows keeps matching (verified: `NetworkIntegrationTest.*` and `WorkerRestartTest.*` drop out under the ASan exclude, 1,741 → 1,739).
- `ctest -N` totals still feed the shard-coverage proof unchanged; the Windows shards use `-I` over the same list.

What changes: a case's leaked process-global state is visible to the next case of its suite (the contract a local `--gtest_filter=Suite.*` run always had), and a crash takes the rest of its suite with it. Docs updated (`testing-architecture.md` §5 gets the three-pass table and §6's contract wording, README index line, the workflow comments that said one process per case).

Discovery runs as a `POST_BUILD` step like `gtest_discover_tests`, writes a per-config tests file included through `TEST_INCLUDE_FILES`, and the script is a `LINK_DEPENDS` of the target so an edit to it relinks and re-discovers (found the hard way: the first vetting run used stale names).

## Vetting

- Whole suite on the Windows dev box, Debug, `ctest -j6`, all GPU tests included: **1,738 entries, 100 % passed, 16 min wall** (3 disabled/skipped placeholders not run).
- Seven suite entries covering ~200 cases (`LuaBindingTest`, `McpDispatchTest`, `FastRandomTest`): 2.0 s, where the per-case form took ~40 s.
- Expected on the box, from #1207's profile: the 83 min of start-up becomes ~1,180 launches ≈ 13 min, so roughly 17–20 min per arm instead of 56 (before #1207) or ~35 (after it).

## Test plan

- [x] Local: entry count, no duplicates, exclude regex, sample run, full suite green.
- [ ] This PR's Linux arms: step times against #1207's, and the Test report still lists per-case results.

## Summary by CodeRabbit

* **Performance**
  * Improved test execution efficiency by running lightweight test suites together in shared processes, while retaining isolated execution for heavy and mesh-cache tests.
  * Added automatic suite discovery and sharding for large suites to support faster, more scalable parallel test runs.

* **Documentation**
  * Updated testing guidance and workflow comments to reflect the revised test execution model.

### DIFF
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -736,13 +736,13 @@ jobs:
     # the `asan-windows-tests` matrix below: this job builds and hands the
     # instrumented tree over, and N separate runners each execute a stride of it.
     #
-    # WHY SHARDING RATHER THAN BATCHING. gtest_discover_tests registers every gtest
-    # case as its own ctest entry, so the suite is ~7,600 PROCESS LAUNCHES, each
-    # paying full engine static-init. Batching many cases into one process is the
-    # bigger win and it REMOVES the isolation that currently hides cross-test state
-    # corruption -- that is #1074's territory and must not be done from here.
-    # Sharding splits the same one-process-per-case run across more runners and
-    # changes the semantics not at all.
+    # SHARDING AND BATCHING, BOTH. The light majority of cases is now registered
+    # one ctest entry per SUITE (OloEngine/tests/cmake/OloSuiteDiscovery.cmake) --
+    # 75 % of a test step was per-process start-up, paid ~8,000x -- while the
+    # memory-heavy and long cases (OLO_HEAVY_TESTS) stay one process each, which is
+    # where the cross-test state corruption #1074 fixed used to hide. Sharding
+    # splits that same entry list across more runners and changes the semantics
+    # not at all.
     #
     # WHAT IT COSTS. Each shard is a whole runner: it pays job startup, a checkout
     # and an artifact download before it runs a single case. This TRADES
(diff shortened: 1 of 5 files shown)


drsnuggles8:
@coderabbitai review

PR #1212, opened 2026-09-13
ci(sanitizers): priority lane for the self-hosted Linux jobs — measured: the additive-label approach does not work

### LINKED ISSUE #1202: CI: a priority lane for self-hosted sanitizer runs (runner label + workflow_dispatch input), since Actions has no job priority
## Problem

GitHub Actions has no job priority: a free runner takes the oldest queued job for its label, and a re-run goes to the back. On a personal account there are no runner groups and no merge queue, so the only way to get one PR's jobs ahead of others is to cancel the others (which is what had to happen on 2026-09-12 with the owner's permission).

Last night made the cost concrete: eight PRs × three Linux sanitizer jobs were queued FIFO on `olo-ci-1/2`, and the one PR that fixed the test every other job was failing on (#1197, the ReSTIR floor) was sixth in line. Each doomed job ahead of it burned ~40 min of a runner. The scheduler is inside the closed Actions service (only `actions/runner` is open, and it just takes the job it is handed), so this cannot be fixed upstream.

## Proposal: a priority lane on our own runners

Priority is expressible on the runner side with **labels**:

1. Register one of the two `olo-ci` runners with an extra label, `olo-ci-priority` (`config.sh --labels self-hosted,Linux,X64,olo-ci,olo-ci-priority`, or edit the labels in the repo's runner settings). It keeps serving `olo-ci` jobs when nothing priority is queued.
2. Give the Sanitizers workflow (and any other self-hosted workflow worth it) a `workflow_dispatch` input, e.g. `lane: [normal, priority]`, and select the runner label from it:
   ```yaml
   runs-on: ${{ inputs.lane == 'priority' && 'olo-ci-priority' || 'olo-ci' }}
   ```
   A dispatched run of the unblocking PR then lands on the priority runner as soon as its current job ends, ahead of the whole `olo-ci` queue, without cancelling anything. (Dispatch it on the PR branch; `workflow_dispatch` needs the workflow file to exist on the default branch — it does.)
3. Document the move in `docs/ops/self-hosted-gpu-runner.md`: when to dispatch a priority run, and that a run's ordering is otherwise FIFO with no other lever.

Optional, if it proves useful: the same input on the Windows workflow is pointless (hosted runners), but `gh workflow run sanitizers.yml --ref <branch> -f lane=priority` could be wrapped in a tiny script so the picker can hand it to a session.

## Acceptance

- A `lane=priority` dispatch of a PR's Sanitizers run starts on `olo-ci-priority` while ≥3 normal runs are still queued on `olo-ci`.
- With nothing dispatched, both runners keep draining the normal queue exactly as today (the label is additive).
- The ops doc names the command.

## Score

```olo-score

capability: 5

craft: 1

stability: 3

decay: 3

effort: 2

confidence: 0.8

learning: 2

fun: 2

kano: performance

blocked_by: []

blocks: []

```

drsnuggles8:
> **Draft: the approach in #1202 does not work, and this PR measures why.** The code here is a
> faithful implementation of the issue's proposal. The proposal's premise is wrong. Do not
> merge as-is — see *The measurement* below, then *Where to go from here*. The one thing in
> here worth keeping regardless is the ride-along fix at the bottom.

Refs #1202.

## The premise, and the measurement that kills it

#1202 proposes an **additive** priority label: give `olo-ci-1` a second label
`olo-ci-priority` alongside `olo-ci`, have a `lane=priority` dispatch request the rarer
label, and the run "lands on the priority runner as soon as its current job ends, ahead of
the whole `olo-ci` queue, without cancelling anything."

**It does not.** A runner does not prefer a rarer label. It takes **the oldest job it is
eligible for** — and a runner that also carries `olo-ci` is eligible for the entire normal
backlog, which is by definition older than a dispatch you just fired.

### The controlled test

An earlier, accidental observation looked like a success and was thrown out: the priority
job was picked up, but every older competitor had either been cancelled by a force-push
seconds earlier or was never schedulable (`gpu-sanitizers-amd.yml` sets `max-parallel: 1`,
so its queued arms are not candidates). The priority job was simply the oldest eligible job.
FIFO predicts that outcome too, so it separates nothing.

The test below removes the confound by construction. Preconditions at **09:39:36Z**, all
verified rather than assumed — each normal job's `detect-changes` dependency confirmed
`completed/success`, so all seven were genuinely waiting on a runner and not on `needs`:

| created | labels | job |
|---|---|---|
| 09:23:06 ×3 | `olo-ci` | this PR's sanitizer jobs |
| 09:35:54 ×3 | `olo-ci` | #1210's sanitizer jobs |
| 09:37:04 ×1 | `olo-ci` | #1210's vulkan-off build |
| **09:39:36** | **`olo-ci-priority`** | **dispatched UBSan — the newest job in the queue** |

Seven eligible normal jobs, every one **older** than the priority job. Both runners busy. I
then cancelled the job occupying `olo-ci-1` at 09:40:32 to force a choice.

**Two independent pickup events, both choosing the older normal job:**

```
09:41:49Z  olo-ci-1 freed -> took ASan + LSan (Linux/Clang), created 09:23:06, labels olo-ci
~10:1x Z   olo-ci-2 freed -> took TSan  (Linux/Clang), created 09:35:54, labels olo-ci

09:39:36Z -> 10:19:25Z   olo-ci-priority UBSan: still `queued`, never picked up
```

The priority job sat queued for the entire 40-minute window and lost both times to an older
normal job. That is exactly what the oldest-first model predicts and exactly what the issue
assumes will not happen.

### Why it cannot work as specified

The issue's two requirements are in direct contradiction:

- *"the label is additive — both runners keep draining the normal queue"* means the priority
  runner stays eligible for `olo-ci`.
- *"a priority dispatch jumps the queue"* requires it to decline older `olo-ci` work.

A runner cannot both take the oldest normal job and skip it. **Priority here costs
capacity — there is no free version of it.**

## Where to go from here

None of these is obviously right, and the choice is a capacity trade-off rather than a
technical one, so this PR stops here rather than guessing:

1. **Reserve `olo-ci-1` for the priority label only** (drop `olo-ci` from it). Works
   immediately — a dedicated runner with an empty queue starts a priority job at once. Costs
   the normal lane 50 % of its throughput at all times, which is a steep price for a queue
   problem.
2. **Add a third runner labelled `olo-ci-priority` only**, normally idle. Keeps normal
   throughput and gives real priority, but the box has 31 GiB and
   `docs/ops/self-hosted-gpu-runner.md` budgets ~9 GiB per concurrent sanitizer job across
   three slots, one of which `olo-gpu-amd` already holds. A priority job running beside two
   normal jobs is the fourth concurrent build. Wants a memory measurement before it ships.
3. **Automate the cancel** — a script that cancels the queued jobs ahead of the PR that
   matters. Ugly, and it is what already happens by hand, but it is the only option that is
   free.
4. **Attack the queue depth instead of its order** and close #1202 as not actionable.

## What is in this branch

Implemented as the issue specifies, and correct as an implementation — the routing works
exactly as designed, it is only the design that does not deliver priority:

- `lane: [normal, priority]` `workflow_dispatch` input, folded in as the **last** term of the
  three Linux jobs' `runs-on`, so the kill switch, `force_hosted`, the `push`/`schedule`
  cache-warm exclusions and the fork boundary all still win.
- `lane` added to the concurrency key, so a priority dispatch cannot be evicted from the
  single pending slot by a normal dispatch of the same ref.
- `olo-ci-priority` registered on `olo-ci-1` and owned by `scripts/setup-olo-ci-runners.sh`
  (`config.sh --replace` re-registers with exactly the labels on its command line, so a label
  added out of band is dropped the next time that script runs).
- §7 of `docs/ops/self-hosted-gpu-runner.md`.

**The in-file comments and the ops doc currently assert that a priority dispatch jumps the
queue.** That text is now known to be wrong and must be rewritten or removed before any of
this merges — which of the options above wins decides what it should say instead.

### Verified regardless of the outcome

- The three Linux `runs-on` expressions are **byte-identical** — parsed out of the YAML and
  compared, not eyeballed (1 distinct string across the three jobs).
- The expression was transcribed into an evaluator implementing GitHub's operand semantics
  (`a && b` yields `a` if falsy else `b`; arrays cast to `true`) and checked over all 14
  reachable combinations of event × kill switch × `force_hosted` × `lane` × fork. Four
  properties hold: non-dispatch behaviour is bit-for-bit master's; `olo-ci-priority` is
  reachable only from an explicit `lane=priority` dispatch; `force_hosted` still beats the
  lane; the kill switch still beats everything.
- **The additive half is confirmed live**: this PR's own `pull_request` run routed all three
  Linux jobs to plain `olo-ci`, and both runners went on draining the normal queue throughout
  — visible in the test above, where they picked up normal work twice.
- A `lane=priority` dispatch on a non-default branch is **accepted** (`204`) by
  `POST /actions/workflows/asan.yml/dispatches`, contrary to the expectation that a new input
  would be rejected before it exists on master. (`gh workflow run` returned 500 during
  GitHub's 09:16Z incident; not retested, so no conclusion is drawn about the CLI wrapper.)

## Ride-along fix (separate commit, independently valid)

`38092f5` — `detect-changes.linuxplan` carries a copy of the three Linux jobs' `runs-on`
predicate under a comment saying it is byte-identical to them. It was not: #1084 (`4c2ddaf`)
added `github.event_name != 'push'` to all three jobs and not to this copy. It changed no
outcome — on a `push` all three Linux jobs are skipped at their own `if`, so the shard plan is
never read — which is exactly why nothing caught it. **This commit stands on its own whatever
happens to the rest of this branch.**


drsnuggles8:
@coderabbitai review

drsnuggles8:
Closing unmerged. The approach this PR implements cannot deliver priority — a runner carrying `olo-ci` takes the oldest job it is eligible for, so an additive label never jumps the queue. The measurement is recorded on #1202, which is closed as not actionable: https://github.com/drsnuggles8/OloEngineBase/issues/1202#issuecomment-5654040300

The ride-along fix `38092f5` (restoring `detect-changes.linuxplan` to byte-identity) landed separately as #1214 and is in master. Nothing else here is worth keeping; the branch is left in place for now.

Issue #1219, opened 2026-09-13
CI: the olo-ci box is over capacity (rho ~1.09) - queue p90 316 min, and hosted routing is no longer blocked


drsnuggles8:
The `olo-ci` queue has been the largest single term on [#1084](https://github.com/drsnuggles8/OloEngineBase/issues/1084) since 2026-09-09. This issue exists because the 2026-09-13 measurement round changed two things about it: the term got **much** larger, and the one blocker that ruled out the obvious fix is gone.

**Read [#1202](https://github.com/drsnuggles8/OloEngineBase/issues/1202) first.** It was closed `NOT_PLANNED` on 2026-09-13 after measuring that a runner takes the oldest job it is eligible for, so an additive priority label cannot jump a queue. **This issue is queue *depth*, not queue *order*.** Do not re-propose additive labels, a reserved runner, or a third `olo-ci` runner — the first two were rejected on measurement, and the third is a hardware purchase against a box whose cold sanitizer builds already reach their 14 GiB cgroup ceiling ([#1138 measurement](https://github.com/drsnuggles8/OloEngineBase/issues/1084#issuecomment-5583560169)).

## The measurement

**The box is saturated.** Every `olo-ci` job served on 2026-09-13, 00:00–15:00Z, across every workflow that targets those labels:

| slot | busy | utilisation | jobs |
|---|---|---|---|
| `olo-ci-1` | 786.6 min | **87.4 %** | 24 |
| `olo-ci-2` | 890.7 min | **99.0 %** | 22 |
| **both** | **1677.4 min of 1800 slot-min** | **93.2 %** | 46 |

Both slots were pinned at 100 % for **26 of the 30** half-hour buckets in that window. Offered load from `asan.yml` alone — 48 sanitizer jobs created, median service time 40.9 min — is **131 min/h against 120 min/h of capacity, ρ ≈ 1.09.** A queue at ρ > 1 does not reach a steady state, which is why the tail moved as far as it did in four days:

| `olo-ci` queue delay before pickup | min | median | p90 | max |
|---|---|---|---|---|
| 2026-09-09 (72 jobs, [prior round](https://github.com/drsnuggles8/OloEngineBase/issues/1084#issuecomment-5602317213)) | 0.5 | **0.8** | 128.2 | 304.1 |
| 2026-09-11 → 09-13 (58 completed jobs) | 2.5 | **95.2** | **316.1** | **578.1** |

[...]

PR #1220, opened 2026-09-13
ci(#1084): the post-#1170 measurement round - the critical path is now the olo-ci queue, and a cache-prune strip that never fired

drsnuggles8:
The post-#1170 measurement round on **#1084**. Two commits: a `cache-prune.yml` bug found on the way, and the budget-table re-measurement the issue's acceptance criteria ask for. **#1084 stays open** — it is a recurring umbrella.

The measurement itself is a [comment on #1084](https://github.com/drsnuggles8/OloEngineBase/issues/1084#issuecomment-5654255627); the headline is below so this PR stands on its own.

## What the round found

**Every compute lever on #1084 has landed, and the per-PR wait went up anyway.**

| | 09-05 | 09-07 | 09-09 warm | **09-13** |
|---|---|---|---|---|
| per-PR, compute only | 236.4 | 222.6 | 45.1–52.2 | **28.7–128.9** |
| per-PR, gate complete | — | — | not reported | **68.6–258.9** |

Compute is the fastest this issue has ever recorded. The gate is not, because the critical path moved off `Windows / build` entirely and onto the `asan.yml` Linux sanitizer chain, which binds by 100–230 min in four of five same-commit both-green pairs.

**The `olo-ci` box is over capacity.** 93.2 % utilisation over 15 hours (both slots at 100 % in 26 of 30 half-hour buckets), offered load ρ ≈ 1.09, queue delay **p90 316 / max 578 min** against 128 / 304 four days ago. Service time went *down* over the same period — #1207/#1208 took UBSan's median from 61.8 to 24.7 min. The jobs got faster and the wait got worse.

**#1170 works.** The afternoon cold window it was aimed at is gone: a 12:40 PR now builds at **94.11 % / 25.2 min** where 09-09's afternoon was 46.76 % / 123.8–134.9 min. Eight `push: master` warms landed in the measured window instead of one nightly. The ASan job's afternoon hit rate — never sampled before — tracks it.

**The 09-09 blocker on hosted routing is gone.** That round ruled it out on "shards 1/4 and 4/4 fail on all three sanitizers, cause unresolved". Reading the failures out of the logs: 09-09 was the hosted image's older glslang (the #1139 case), 09-12 was one statistical oracle test. Four of the last six nightlies are 12/12 green on hosted Linux with 0.0 min of queue.

Split out as **#1219** (scored, 3.47), framed as queue *depth* not queue *order* per #1202's closure, with the runner-minute arithmetic as an explicit acceptance criterion rather than a recommendation.

## The two commits

**`c150dc617` — the ride-along fix.** #1205 added a `-<ImageVersion>` runner-image stamp to the vcpkg keys and a strip in `cache-prune.yml` so an image bump would supersede like a hash bump. The strip's pattern allows a two-part version (`20260828.587`, which is what the worked example beside it used); a real `ImageVersion` has three (`20260907.229.1`). So it never matched a live key — and because it never matched, the content-hash strip behind it never fired either, making every (manifest hash, runner image) pair its own immortal group.

Measured against the live store before the change:

```
vcpkg-Windows-...-<64hex>-20260907.229.1-34702711853-1
    -> vcpkg-Windows-...-<64hex>-20260907.229.1     (its own group)
vcpkg-Windows-...-<64hex>-34400333427-1
    -> vcpkg-Windows-x64-windows-static-md          (a different group)
```

**1083 MiB stranded** across the two pre-#1205 vcpkg lineages until the 30-day age-out. Verified by running the workflow's own `sed` over all 13 live keys before and after: 11 groups after, with each vcpkg lineage collapsing to one.

**`dee4affc9` — the budget table.** #1084's third acceptance criterion. All 13 live entries re-measured, each row keeping its 09-09 figure beside the new one. The store grew **7968 → 10376 MiB in four days with nothing deliberately added**, by two mechanisms worth naming: two sccache lineages reached their caps (`sccache-asan-windows-2025` 667 → 1492 against 1500M; `sccache-windows-2025-release` back to 2046 against 2 G after #1170's rename), and #1205's re-key doubled `vcpkg-Windows` 706 → 1448 MiB for reasons **not** measured.

That corrects two of this issue's own earlier readings, both the same mistake in opposite directions: 09-09 read a *fresh* lineage (667 MiB) as a steady size, and predicted the renamed Windows key would shrink. It did not — a rename shrinks a lineage only if the old one was an accumulation, and that one was already at its cap. Rule 3 gains the lesson from the stranding: test a prefix strip against a live key, not against the example written beside it.

## Verification

- `pre-commit run --all-files` — clean.
- The prune `sed` was run over the live cache listing before and after the fix; the grouping change is the verification, and it is in the commit message.
- No code paths touched, so no build or test run: the two files are a workflow and a doc.
- Every number in the comment and in #1219 comes from the Actions API (`runs`, `jobs`, `caches`) or from job logs, not from a local build.

## What this round deliberately did not do

- **Did not re-measure the warm build's tail** (~13 of 30.3 min of linking plus the MCP TUs). #1170 added `Report-NinjaBuildTail.ps1` for it; reading those reports is its own round.
- **Did not re-measure #1205's 55 min/job claim** — it landed before the window and there is no pre-#1205 control on the current runner image.
- **Did not separate #1209 from #1207/#1208** — all three landed inside 9 hours and all three move box service time.
- **Did not touch routing.** Filed as #1219 instead, per #1084's own rule that a lever of that size gets its own issue.

## Honest caveats

- The per-PR pair table is **n = 5**, the smallest sample in the round, because most PR runs on 2026-09-13 were superseded and cancelled. It is the number I would widen first.
- Utilisation and ρ are one 15-hour window. The queue distribution is 58 jobs over three days; the hosted-arm health is six nightlies.
- The queue median moved by two orders of magnitude depending on whether cancelled-while-queued jobs are counted as instant. They are excluded as right-censored, and the comment says so — my own first pass got this wrong and reported 0.0 min.
- That day was busy, and contending branches are listed in full in the comment. A saturated queue is the claim, so an idle window would have measured the wrong thing; `a5e09cfbb` is the uncontended control from the same day.


drsnuggles8:
## 🤖 Self-review @ `be14917e67a7b739dd526dfab8fbb66aff5d0ad0`

Reviewed the PR diff at high effort. Two files: a workflow `sed` and a doc table.

**Findings: 1 · Fixed: 1**

- **The 2026-09-13 rows did not reconcile to the steady set** (`be14917e6`). The subtotal counted only the *current* vcpkg generation on each platform and left the two stranded predecessors in prose, so the rows summed to 9620 against a stated steady set of 10376 — a 756 MiB gap with no visible cause, in a table whose entire job is that the arithmetic checks out. Both vcpkg rows now carry both generations and the steady-set row shows the reconciliation: 3396 + 3442 + 2046 + 1492 = 10376.

**Checked and dismissed**

- **Over-strip risk on the widened image pattern.** `s/-([0-9]{8}(\.[0-9]+)+|noimage)$//` requires a leading 8-digit group, so version-shaped suffixes like `Windows-vulkan-prebuilt-sdk-true-1.4.357.0` and `olo-shader-toolchain-…-v1` cannot match. Verified empirically rather than by reading: the workflow's own `sed` was run over all 13 live cache keys and yields 11 groups, with each vcpkg lineage collapsing to one and nothing else changing.
- **Strip ordering.** `sed` applies each `s///` once in sequence, so the run-id strip cannot re-fire inside the image version; a key with no run id still strips image-then-hash correctly.
- **No engine code, no build, no tests.** Correct for this diff — the sanitizer suites skipped on `detect-changes`, which is also why this PR contributes ~0 to the `olo-ci` queue it measures.

**Outstanding, named rather than claimed clean:** CodeRabbit has **not** reviewed this PR — its only comment is an OSS rate-limit notice ("next included review available in 44 minutes"). Unresolved review threads are 0 because none were ever posted, not because any were resolved. A re-review is requested below.

Issue #1267, opened 2026-09-14
Foliage: mesh/impostor layers do not render on Vulkan, and skip the G-Buffer on OpenGL


drsnuggles8:
Found by the #1239 reference fixtures (PR #1262), which exist to catch exactly this. Reported there, not fixed, because `FoliageRenderPass` was out of that issue's scope.

**Mesh and impostor foliage layers fail on both backends, in two different ways. Billboard layers are fine on both.**

| Backend | Mesh / impostor foliage | Billboard foliage |
|---|---|---|
| Vulkan | **does not meaningfully render** | renders correctly |
| OpenGL | renders, but writes **no G-Buffer** | writes the G-Buffer |

## Defect 1 — on Vulkan the canopy is essentially absent (the severe one)

The `woodland` fixture on `--rhi=vulkan` renders the tree canopy as scattered disconnected
fragments at the horizon: no trunks, no near-field trees. On OpenGL the same cameras render a
dense canopy with trunks and sky gaps. Confirmed on the `frontal` and `grazing` cameras.

The understory billboards **in the same scene, in the same frame** render correctly.

## Defect 2 — on OpenGL the canopy writes no G-Buffer

In `woodland` the canopy appears in `Beauty` and in `ShadowMapCSMCascade0` but is absent from
both `GBufferAlbedo` and `GBufferNormal` — they are empty above the horizon while the terrain and
the understory appear normally.

Consequence: every G-Buffer-derived term treats the canopy as sky. **SSAO, SSGI and SSR get no
canopy occluder**, and the derived `Roughness`/`Metallic` lanes are blank there. It is a large
part of why the fixture's forest floor is lit like open ground — a canopy that contributes no
ambient occlusion and no screen-space GI is a ceiling on how good foliage can look, which matters
for epic #1225.

## Isolation — this is the useful part

The failure is narrow, and the evidence narrows it hard:

- **`meadow` (billboard-only) is correct on both backends** and writes the G-Buffer — its blades
  are clearly visible in `GBufferAlbedo`.
- **`meadow`'s `FoliagePass` costs 0.000 ms; `woodland`'s costs 1.75 ms.** The billboard fixture
  submits nothing to that pass at all, which is direct evidence the two layer types take
  different code paths.
- **Every non-foliage fixture matches across backends**, including a static mesh model
  (`animal-long-coat`) and a skinned, animated one (`animal-short-coat`).

[...]

Issue #1264, opened 2026-09-13
Vulkan: octahedral impostor cards render nothing — the whole mesh-foliage canopy is missing


drsnuggles8:
## What happens

Foliage layers with `UseImpostor: true` (the octahedral impostor card, #433) draw **nothing at
all** on the Vulkan backend. The flat billboard layers in the same scene render normally, so a
scene reads as "the grass is there, the trees are gone".

On OpenGL the same build, same scene, same camera and same rendering path draws the canopy
correctly.

## Repro (two screenshots, one variable)

Same commit, same scene, same pose, same **Forward** path — only `--rhi` differs:

```powershell
# trees present
pwsh -NoProfile -File .claude/skills/run-oloengine/driver.ps1 -Action attach -AllowWrites
# trees absent
pwsh -NoProfile -File .claude/skills/run-oloengine/driver.ps1 -Action attach -AllowWrites -Rhi vulkan
```

Then, in either session:

```
olo_scene_open   {"path":"Scenes/Benchmark/Courtyard.olo"}
olo_screenshot   {"camera":{"position":[58,4.5,28],"target":[12,1,47],"fov":60}}
```

Courtyard's `Pines` and `Palms` foliage layers both set `UseImpostor: true`; its `Understory`
layer does not. Only the two impostor layers disappear.

## It is not a placement or a bake failure

From the Vulkan run's log, the atlases bake and the instances generate:

```
ImpostorBaker: baked 8x8 octahedral atlas (64 tiles @ 128px, radius 0.56)
FoliageRenderer: instance VBO create (707 instances)     <- Pines
ImpostorBaker: baked 8x8 octahedral atlas (64 tiles @ 128px, radius 0.76)
FoliageRenderer: instance VBO create (13 instances)      <- Palms
```

There are no Vulkan validation errors and the render-graph validation passes. The draws are
submitted and the atlas exists; the card simply produces no pixels.

## Where to start

`Foliage_Impostor.glsl`'s vertex stage is a two-stream vertex **pull** on Vulkan (`OLO_PULLED_VERTEX`,
the 20-byte card quad on binding 57 and FoliageRenderer's 48-byte per-instance stream on binding
63). Its billboard sibling `Foliage_Instance.glsl` uses the same shape and renders fine, so the
divergence is likely in the impostor-specific part of that stage or in the atlas sampling, not in
the pull plumbing generally. `ShaderStageContractTest` already reflects both shaders' Vulkan
[...]

Issue #1239, opened 2026-09-13
Renderer benchmarks: licensed AAA head, groomed animals and vegetation reference fixtures


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1225

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Extend the existing deterministic benchmark harness with a reference head, short-/long-coated animals and meadow/woodland fixtures. Acquire reusable assets through a reproducible manifest or procedural fixtures; record rights and do not commit unlicensed content.

## Acceptance criteria

- [ ] Manifest records provenance, redistribution limits, versions/hashes, units and colour conventions; external assets have a reproducible local acquisition path.
- [ ] Pin cameras, lighting, animation/wind time and paths for frontal, grazing, backlit and moving sequences.
- [ ] Record GPU/driver, output/internal resolution, frame time distribution and per-pass timings using existing diagnostics.
- [ ] Publish proposed measurable budgets/quality gates for available hardware and scene populations; explicitly distinguish measured hardware from untested tiers. AAA ambition is fixed, hardware numbers are not invented.

## Scope boundaries

Fixtures initially show current limitations and are upgraded by feature owners. No asset purchase is assumed; use accessible assets or identify a genuine external dependency.

## Blocked by

None - can start immediately.

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 13
craft: 5
stability: 5
decay: 3
effort: 3
confidence: 0.8
learning: 3
fun: 8
kano: table-stakes
blocked_by: []
[...]

Issue #1233, opened 2026-09-13
Foliage: render authored plant meshes up close instead of fixed quads


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1224

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Make a foliage layer render its actual plant mesh and material attributes at close range, while preserving the current card path and existing impostor bake capability.

## Acceptance criteria

- [ ] Assigned grass clump, shrub and tree meshes render their real geometry with normals/tangents/UVs and submesh material assignment.
- [ ] Terrain scattering, transforms, instance identity and conservative bounds remain correct.
- [ ] Editor assignment, reimport, cooking and scene round-trip preserve mesh/material selection; missing assets have explicit diagnostics.
- [ ] Depth, picking, available raster shadows and motion use matching geometry on both backends.

## Scope boundaries

No requirement for procedural mesh shaders or virtual geometry; distant LOD refinement follows.

## Blocked by

- #1230

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 8
craft: 8
stability: 3
decay: 1
effort: 3
confidence: 0.8
learning: 5
fun: 13
kano: performance
blocked_by: []
blocks: [1224, 1234, 1235, 1236, 1254]
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

PR #1268, opened 2026-09-14
docs(benchmark): record the woodland canopy defects as fixed, with measured numbers

drsnuggles8:
Refs #1267. Refs #1239, #1225 (epic).

## Why this PR is docs-only

#1267 asked for two renderer fixes. **Neither defect reproduces.** The issue was filed at 06:05
on 2026-09-14; PR #1265 merged at 06:06 and closed the same two defects under #1264, and nobody
re-verified afterwards. I re-ran the #1239 fixtures on master `5acc47238` on both backends before
writing any code — the handover's first instruction — and the verification came back clean. Full
evidence is posted on the issue: https://github.com/drsnuggles8/OloEngineBase/issues/1267#issuecomment-5661156349

Summary of what was measured (RTX 4090, Debug, Deferred):

| | as filed | on `5acc47238` |
|---|---|---|
| OpenGL `GBufferAlbedo` / `GBufferNormal` | canopy absent above the horizon | canopy present, per-tree normals |
| OpenGL `AOBuffer` | canopy = 1.0 (sky) | real canopy occlusion |
| Vulkan `Beauty` | scattered horizon fragments | dense canopy, trunks, sky gaps — all four cameras |
| `FoliagePass` | 1.75 ms | absent from `passTimingsMs` in 5/5 runs |
| `meadow` control (Vulkan) | correct | correct, grass in `GBufferAlbedo` |

So there is no code change to make. What there **is** is stale documentation: two files still
asserted these defects as live, and they are the files the next agent working on #1233 would read
first. This PR corrects them and records the measurements.

## What changed

- **`docs/guides/benchmark-reference-fixtures-baseline.md`** — the cross-backend table's
  `woodland` row flipped from "does not match" to matching; finding 1 rewritten as history with
  the actual cause (routing — an impostor layer had no deferred program, so it fell to the forward
  `FoliagePass` which runs after `DeferredLightingPass`); the per-pass cost row re-measured
  (`ScenePass` 3.72 ms, N=5) with a footnote saying which columns are pre-#1265; the frame-total
  gate row flagged as pre-#1265 rather than silently re-used.
- **`woodland.diagnostic.yaml`** and **`Woodland.olo`** — the "MEASURED OBSERVATION … REPORTED,
  NOT FIXED" block becomes history, and both headers stop describing the impostor cross-fade band
  as mesh-in-the-near-field.

## The framing correction

"Mesh/impostor layers" reads as if there were a mesh-geometry foliage path. There isn't:
`FoliageRenderer::BuildQuadGeometry` builds a 4-vertex card for **every** layer, `MeshPath` only
feeds the impostor bake, and `FoliageImpostorSampling.glsl` cross-fades between a flat single-frame
read of that atlas and the full parallax octahedral impostor — same card on both sides. The split
was `UseImpostor`, exactly as #1265 concluded. Real plant meshes are #1233.

`#1267` is **not** closed by this PR — I am recommending closure on the issue and leaving the call
to the user, per `CLAUDE.md`.

## Review guide

**Where I'd look hardest**
1. `docs/guides/benchmark-reference-fixtures-baseline.md:120-130` — the two footnoted rows. I chose
   to flag the pre-#1265 frame-total and the pre-#1265 Draws/Triangles columns rather than
   re-measure or delete them. If you'd rather I re-ran the frame total, say so and I will.
2. `docs/guides/benchmark-reference-fixtures-baseline.md:186-196` — finding 1's new cause
   narrative. I read it out of `Renderer3DSpecializedDraws.cpp::DrawFoliageLayer` and #1265's body
   rather than bisecting, so it is a reading of the code as it stands, not an independent
   reconstruction of the bug.
3. The `Woodland.olo` edit is inside a comment block, but it is a scene file — I re-ran a full
   capture after it and the counters are byte-for-byte the same (76 draws, 985,996 triangles,
   515,052,577 bytes), so it still parses identically.

**What I verified, and how** — `woodland.diagnostic` on OpenGL via
`OloEngine-Tests --olo-capture-manifest=` (5 runs for the medians, plus one after the edits) and on
Vulkan via `olo_benchmark_capture` through the editor front door, `attachmentFailures: 0`,
`warmupTimedOut: false`, `olo_shader_errors` → 0. I **looked at** `Beauty`, `Albedo`, `Normals` and
`AOBuffer` for `frontal` and `grazing` on both backends, and `Beauty` for `backlit` and `moving` on
Vulkan. `meadow` re-captured on Vulkan as the control. 68/68 in the
`ReferenceFixture*:*Foliage*:*Impostor*` filter.

**Least confident about** — the `ScenePass` number as a gate candidate. Five back-to-back runs gave
a 2.4% spread, but a sixth run minutes later, after a 35 s test-suite run on the same GPU, measured
4.827 ms against the 3.723 ms median. I withdrew the tolerance claim in the text rather than quote
the tight figure, but someone proposing an actual gate should re-measure across warm-up states.

**Deliberately not tested** — the `woodland` frame total (flagged in the doc instead); a
before/after A/B, because there is no "before" to run against on this branch; Forward and Forward+
foliage, which this issue never touched and which still use `FoliagePass`.

## Summary by CodeRabbit

- **Documentation**
  - Updated foliage rendering guidance to clarify that scene foliage is displayed using camera-facing cards and impostor transitions, while mesh data supports baking only.
  - Added troubleshooting guidance for diagnosing foliage impostor behavior and clarified known limitations.
  - Corrected woodland benchmark documentation to reflect verified canopy rendering and updated performance measurements across supported graphics backends.
  - Revised scene and diagnostic references to match current foliage rendering behavior.


drsnuggles8:
## 🤖 Self-review @ `f35ed94fe164e42861470e208eecfd220d6ee823`

Reviewed the diff at **medium** effort (docs-only, so the review checked factual claims against the code rather than logic). Every code/PR/issue reference in the new text was verified to exist: `SelectFoliageRenderStream`, `Foliage_Impostor_GBuffer.glsl`, `FoliageRenderer::BuildQuadGeometry`, PR #1265 / issue #1264, and #1265's commits are genuine ancestors of the cited `5acc47238`.

- **Findings: 6 · Fixed: 6** (folded into `9e23f1c19`)
  1. The manifest and `Woodland.olo` headers still called the cross-fade band mesh-in-the-near-field — corrected; both sides are the same card.
  2. The frame-total gate row is pre-#1265 and was unflagged — footnoted as provisional rather than re-used or guessed at.
  3. An N=5 post-fix `ScenePass` cell sat in a table headed *N=3, medians* with pre-fix sibling columns — footnoted, and the sibling columns confirmed unchanged by the re-run.
  4. "all four cameras match" rested on two I had looked at — looked at `backlit` and `moving` on Vulkan too, then stated the claim explicitly.
  5. "FoliagePass is culled" was unconditional and the wrong verb — it is a routing outcome of Deferred + G-Buffer programs loaded, it covers billboards too, and the pass simply receives no commands. Forward/Forward+ still use it.
  6. Finding 1 read as if #1265 *introduced* the forcing rather than removing it — rewritten.
- **Dismissed:** none.

One thing the review did not catch, found afterwards by re-running: the `ScenePass` 2.4% spread I had documented is a within-session property. A sixth run minutes later, after a 35 s test-suite run on the same GPU, measured 4.827 ms against the 3.723 ms median. The tolerance claim is withdrawn in the text rather than left to mislead a future gate.

`f35ed94fe` adds the reusable lesson to `docs/agent-rules/foliage-impostor-card-rendering.md`; it is covered by the same review, being the same claim about `MeshPath` stated once more where the next person will read it.

PR #1269, opened 2026-09-14
test(rhi/vulkan): pin the index-bind address-flag contract (#1200 is not an engine bug)

### LINKED ISSUE #1200: Vulkan: vkCmdBindIndexBuffer3KHR is issued with address flags 0 for STORAGE-usage index buffers (VUID-VkBindIndexBuffer3InfoKHR address flags)
`vkCmdBindIndexBuffer3KHR` is issued with `VkAddressCommandFlagsKHR(0)` for index buffers whose backing `VkBuffer` also carries `VK_BUFFER_USAGE_2_STORAGE_BUFFER_BIT` (the raw-buffer family — vertex-pull streams, VirtualMeshRegistry's index arena). The validation layer (SDK 1.4.357, Debug editor) reports, ten times per run before the duplicate cap:

```
vkCmdBindIndexBuffer3KHR(): pInfo->addressRange.address (0x1a14e900) belongs to VkBuffer 0x2660000000266,
which was created with VK_BUFFER_USAGE_2_STORAGE_BUFFER_BIT, but address flags (VkAddressCommandFlagsKHR(0))
do not contain VK_ADDRESS_COMMAND_STORAGE_BUFFER_USAGE_BIT_KHR or VK_ADDRESS_COMMAND_UNKNOWN_STORAGE_BUFFER_USAGE_BIT_KHR.
  ...inside command-buffer label region: ScenePass   (also ParticlePass, AOApplyPass)
```

The bind site (`VulkanRendererAPI.cpp`, the `vkCmdBindIndexBuffer3KHR` block) does derive `addressFlags` from the resolver's `StorageUsage`, and the resolver is right about the two buffers it knows: a `VulkanIndexBuffer` is created without `STORAGE_BUFFER_BIT` (→ `Absent`, VUID-13123 forbids the flag) and the raw-family arena with it (→ `Present`). But the layer says the address that was actually bound belongs to a buffer created **with** STORAGE usage while the flags say `Absent` — so the address and the flag describe different buffers. The one path that substitutes an address after resolution is the per-draw stream **snapshot** (#1171, `vertex stream ... rewritten more than once per frame — snapshotting it per draw`; the same run logs two of those), whose copies live in STORAGE-usage arena memory. The flags have to follow the substituted buffer (`Present`, or `Unknown` when the arena's usage is not known at the bind site).

Why it did not show up before: the default Debug configuration (core + synchronization validation) does not report it; it surfaced only under a layer settings file (`VK_LAYER_SETTINGS_PATH`) during the #1186 device-fault hunt on 2026-09-12. The `GetValidationErrorCount() == 0` test assertions therefore never saw it either.

Fix: carry the address-command flags on the raw buffer's binding (`VK_ADDRESS_COMMAND_STORAGE_BUFFER_USAGE_BIT_KHR` for the raw family, and the matching bits for the indirect / transfer address forms), and add the same check to the address-command tests from #1179 so the layer's verdict is part of CI.

## Score

```olo-score

capability: 1

craft: 1

stability: 1

decay: 1

effort: 1

confidence: 0.3

learning: 1

fun: 1

kano: indifferent

blocked_by: []

blocks: []

```

drsnuggles8:
## What this is

Issue #1200 reported `vkCmdBindIndexBuffer3KHR` issuing `VkAddressCommandFlagsKHR(0)` for a
STORAGE-usage index buffer, ten times per run, inside `ScenePass` / `ParticlePass` / `AOApplyPass`.
It came with a diagnosis (the #1171 per-draw vertex-stream snapshot substitutes an address after
resolution) and a fix (carry the storage flag on the raw family's binding).

**The engine is correct and the proposed fix would break it.** The message is produced by GPU-AV:
its buffer-validation family adds `VK_BUFFER_USAGE_STORAGE_BUFFER_BIT` to application buffers so its
own validation shaders can address them, and core validation then reports the application for not
declaring a bit it never asked for.

## The evidence

Same engine, same binary, only the layer settings changing:

| layer configuration | VUID-13122 |
|---|---|
| default Debug (core + sync) | not reported |
| `validate_best_practices = true` | not reported — and its warnings prove the layer was live and reading the settings file |
| `gpuav_enable = true` | **reported** |
| `+ gpuav_index_buffers = false` | reported |
| `+ gpuav_shader_instrumentation = false` | reported |
| `+ gpuav_buffers_validation = false` | not reported |
| `+ gpuav_debug_disable_all = true` | not reported |

The buffer the layer names was matched to its creation through the new bind trace: it is a
`VulkanIndexBuffer` created `INDEX|TRANSFER_DST|TRANSFER_SRC|SHADER_DEVICE_ADDRESS`, with no storage
bit anywhere.

Two further corrections to the issue:

- The raw family **already** carries `VK_ADDRESS_COMMAND_STORAGE_BUFFER_USAGE_BIT_KHR` (traced:
  `flags 0x4`). The binds reporting 0 are the *object-backed* family, where VUID-13123 **forbids**
  it — so the proposed change trades a debug-tool artifact for a real violation.
- The #1171 snapshot path cannot reach an index bind: only two buffer families in the engine are
  created with `INDEX_BUFFER_BIT` and the frame arena is neither, so an arena address bound as an
  index buffer would also have tripped VUID-13051, which the report does not show.

I measured the proposed fix rather than arguing about it: applying it and re-running the new test
reports **VUID-13123 in the default configuration**, with no GPU-AV anywhere.

## What ships

- **A guard, not a fix.** `VulkanDrawPath.IndexBindAddressFlagsFollowEachFamilysCreateTimeStorageUsage`
  asserts the flag decision for both families *and* issues a real indexed draw through each in one
  command buffer, so an attached layer adjudicates 13122 and 13123 on real binds. It fails on both
  counts against the change #1200 proposed.
- **`IndexBindAddressFlagsFor`** — the bind decision, observable without recording anything, mirroring
  `BindIndexBufferFor`'s refusal (extent included).
- **Diagnostics.** `OLO_VK_TRACE_BUFFERS` now logs each buffer's `VkBuffer` beside its address range,
  and every index bind's range, buffer and resolved flags. This is what turned
  `belongs to VkBuffer 0x…` into a named buffer.
- **A doc** with the A/B method, the positive control, and the knob for a quiet GPU-AV run.

### Ride-along fix (own commit)

`688cd147` — a shader whose root-data layout reflects to nothing assembles zero bytes, the arena
refuses a zero-byte claim, and **every draw or dispatch using it was dropped with no message at all**.
Found because it silently dropped both draws of the test above. Behaviour is deliberately unchanged
(an empty layout is more often failed reflection than a genuinely input-free shader, and running on
failed reflection is worse than not running); it now names the shader, like the pipeline-creation
failure one branch up.

## Review guide

**Where I'd look hardest**
1. `OloEngine/tests/Rendering/VulkanDrawPathTest.cpp` — whether the test can pass *vacuously*. The
   raw-family draw is the one that matters, and the target is already green from the object-backed
   draw, so I assert on the draw **census** (incremented at the `vkCmdDraw*`, after the index bind)
   rather than `GetPreparedDrawsThisRecording()` (incremented inside `PrepareDraw`, before it).
   If that reasoning is wrong, the test proves much less than it claims.
2. `VulkanRendererAPI.cpp` `AssembleAndPushRootData` — the ride-along. Keeping the drop is a
   judgement call; the opposite call (let an input-free shader draw) is defensible and I did not
   make it.
3. `VulkanRendererAPI.h` `IndexBindAddressFlagsFor` — a new public accessor whose only consumer is a
   test. Justified in its comment; push back if that is the wrong shape.

**What I verified, and how** — the A/B matrix above (`VK_LAYER_SETTINGS_PATH`, one test run per row);
181/181 `Vulkan*` tests green; the live editor on Vulkan (`[RHI] Backend: Vulkan`, 48 FPS, zero
`[error]` lines) plus a clean sweep of DriftMenu / SponzaDeferred / VirtualGeometrySponza /
particle_test — ~22k index binds, no validation message; and the deliberate counter-experiment above,
which failed exactly as predicted and was reverted.

**Least confident about** — whether GPU-AV *should* be exempting its own injected usage from these
VUIDs, i.e. whether this is worth reporting upstream to Vulkan-ValidationLayers. I did not file
anything; I think it is worth doing and would rather you decide.

**Deliberately not tested** — the full non-Vulkan suite. The diff is Vulkan-backend-only plus one
test and docs, and the full run has known machine-local flakes here; `Vulkan*` is the surface this
touches.

---

This PR does **not** close #1200 — it is a not-a-bug finding, and closing an issue stays yours. I
recommend closing it with the evidence above.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in Vulkan buffer tracing for raw, vertex, and index buffers, including buffer handles, address ranges, sizes, and usage flags.
  - Added diagnostics for index-buffer binding details.

- **Bug Fixes**
  - Draws and dispatches with empty root-data layouts now report a clear warning instead of attempting an invalid zero-byte operation.

- **Documentation**
  - Added guidance for interpreting GPU-assisted validation results and distinguishing layer-generated diagnostics from engine issues.


drsnuggles8:
@coderabbitai review

PR #1282, opened 2026-09-15
fix(renderer): name the DDGI relocation warm-up for what it does (#1279)

### LINKED ISSUE #1279: Renderer: DDGI schedules a RelocationRefinement capture that the relocation shader treats as a refresh
Raised by CodeRabbit on PR #1273 (issue #846) and recorded rather than fixed there, because the
repair changes converged probe placement and #1273 is a perf change whose premise is not shipping
unmeasured ones.

## The mismatch

Two sides of the relocation schedule disagree about what the last scheduled capture is for.

`PickCaptureSet` tags a probe with `PendingRelocationRecapture` into the **RelocationRefinement**
tier (`DDGIProbeUpdatePass.cpp:1128`). `RelocateProbesGPU` decides the per-probe refresh flag with
`rec.Captured && rec.RelocationIteration >= kMaxRelocationIterations`, which for
`kMaxRelocationIterations = 2` traces as:

| capture | `Captured` | `Iteration` in | refresh? | spring | `Pending` out |
|---|---|---|---|---|---|
| 1 | false | 0 | no | runs | true |
| 2 | true | 1 | no | runs | true |
| 3 | true | 2 | **yes** | suppressed | false |

Capture 3 is scheduled as a *refinement* and executes as a *refresh*. It is not wasted — it
re-captures the hit cache and re-classifies from the settled position, which is the first cache the
probe has taken from where it actually ended up — but the tier name claims something else.

## Why it is not a one-line fix

The two self-consistent repairs move behaviour in opposite directions:

1. **Make the executor refine** (`rec.Captured && !rec.PendingRelocationRecapture`, CodeRabbit's
   suggestion). The spring then runs **three** times for `kMax = 2`. That is the extra application
   `DDGI_Relocate.comp`'s own comment warns about: re-running the spring on a capture taken from the
   position the spring produced is the feedback loop measured as a limit cycle — a converged field
   churning at 3-7 RMSE under a still camera, plateauing instead of shrinking.
2. **Make the scheduler stop earlier** — clear `PendingRelocationRecapture` when `Iteration` reaches
   `kMax`, so no third capture is scheduled. Keeps the spring at `kMax` iterations (what the
   constant's name says) but drops the settling re-capture and re-classification.

There is a third answer: the executor is right, the tier name is imprecise, and the fix is to name
the last scheduled capture something like `RelocationSettle`.

## What deciding this needs

- A before/after on the DDGI goldens (`DDGIVisualEvidenceTest.MultiAngleGoldens`,
  `LightLeakRegression`) for whichever option is taken — options 1 and 2 both change converged
  placement.
- The limit-cycle check the shader comment describes: a still camera over enough frames to see
  whether the field settles or churns. A short test will not catch it.
- Confirmation of which behaviour was intended when `kMaxRelocationIterations` was introduced
  (#707 / PR #842).

Refs: #846, PR #1273, `DDGI_Relocate.comp`'s "A PERIODIC REFRESH MUST NOT RE-PLACE AN ALREADY-SETTLED PROBE" comment.

## Score
```olo-score
capability: 1
craft: 3
stability: 3
decay: 1
effort: 3
confidence: 0.5
learning: 5
fun: 3
kano: table-stakes
blocked_by: []
blocks: []
```

drsnuggles8:
The mismatch #1279 records is real, and the executor was the side that was right — but for a
better reason than the comment beside it gave. The deliverable is therefore a rename plus the
measurement that rules the other two options out, which is what the issue asked for.

## What decides it

The capture rasterizes the hit cache **from the probe's current offset**
(`DDGI_Capture.glsl`'s vertex stage reads the probe-data texture), and relight reconstructs every
hit as `probeWorldPosition(current offset) + direction * distance` (`DDGI_Relight.glsl:216`).

So **a capture whose spring then moves the probe leaves the entire hit cache displaced by that
step** — PR #842's "wrong everywhere, failing nowhere", the same error as capturing from the
lattice point while gathering from the relocated one. Nothing re-captures that probe until the
age-ordered refresh tier, at budget/8, gets round to it.

Ending the warm-up on a capture that does **not** move the probe is what puts the cache and the
position back in agreement, and it is the only thing in the schedule that does. Both proposed
repairs end the warm-up on a spring step.

## The measurement

A closed-loop model in `DDGIMathTest` runs the real `DDGI::RelocateProbeSpring` with the
aggregates re-derived from wherever the spring just put the probe — the feedback the fixed-aggregate
spring tests cannot express. Displacement left between the probe and the position its last capture
was taken from, in cells (1-cell spacing):

| wall at | shipped | option 1 (all-spring) | option 2 (no settling capture) |
|---|---|---|---|
| 0.30 | **0** | 0.0167 | 0.0189 |
| 0.10 | **0** | 0.0020 | 0.1015 |
| 0.05 | **0** | 0.0032 | 0.1028 |

A tenth of a cell is 40% of the comfort distance. The crowded rows are the case relocation exists
for, and they are where dropping the settling capture is worst.

## One correction to the issue

**Option 1 does not produce a limit cycle.** Its warm-up is bounded and terminates — the measured
3–7 RMSE cycle in `DDGI_Relocate.comp`'s comment belongs to the *unbounded* periodic-refresh tier,
which option 1 leaves suppressed. A second test bounds that claim directly: at the shipped gains
the closed loop is contractive and reaches a fixed point exactly, so "it would run away" is not the
argument against springing on a refresh. The argument is the displaced cache. The test says so in
its own comment, and is explicit that a 1-D model does not reproduce the real field's cycle.

## The change

No behaviour change. The vocabulary now matches on all three sides:

- `CaptureTier::RelocationRefinement` → **`RelocationFollowUp`** — the words the scheduler's own
  comments already used ("follow-up captures", twice). The tier covers two captures that do
  different things (refine, then settle), so naming it for the schedule rather than for the spring
  is the accurate choice.
- `DDGI_RELOCATE_ENTRY_REFRESH` → **`DDGI_RELOCATE_ENTRY_HOLD_POSITION`** — named for what it does
  rather than for one of the two cases that set it.
- New **`DDGI::RelocationStepForCapture`** returns *both* the tier and the spring-suppression flag
  from one state. That is the fix for the bug of record: the two decisions were derived at two call
  sites from two different record fields, so no test could see them disagree.
  `kMaxRelocationIterations` moves next to it, with a new `kRelocationWarmupCaptures`.
- `ProbeRecord`'s `Captured` / `RelocationIteration` / `PendingRelocationRecapture` trio was three
  bits encoding one count and was the *mechanism* of the bug. It collapses to one saturating
  `CaptureCount` (saturating so 256 refreshes cannot wrap a `u8` back into the warm-up).

Also corrects two claims in the old `kMaxRelocationIterations` comment that the code contradicted:
two spring applications close ~58% of a residual at a 0.35 step, not "~75%", and "the periodic
refresh closes the rest" is false — the shader suppresses the spring there, so it closes none of it.

## Verification

- **SPIR-V for `DDGI_Relocate.comp` is byte-identical** before and after (`glslc` against the HEAD
  tree vs the working tree, `cmp` clean). The GPU side is provably untouched, which is also why a
  pixel A/B would measure nothing here.
- **404 DDGI + shader tests pass**, 3 skipped (environmental: RADV null device, shader-pack bake,
  shader-cache liveness). Includes `DDGIVisualEvidenceTest.MultiAngleGoldens`,
  `LightLeakRegression`, and the still-camera
  `DDGICascadeEvidenceTest.CascadeFieldIsTemporallyStableFromAFixedPose` — the limit-cycle check
  the issue asked for already exists and is green.
- **Four new L1 contract tests** pin the capture-by-capture table, the spring-application count,
  the displacement measurement above, and the closed-loop fixed point.
- **Live OloEditor** (real GL 4.6, `DDGITest.olo`): 48/48 probes captured / active / relit /
  blended, 0 uncaptured, 0.88 bounce coverage, **zero shader errors**. Two angles inspected — red
  and blue wall bleed onto the spheres and floor, contact shadows intact, no leaking.
- `OloEditor` and `OloEngine-Tests` both build.

Note on golden churn: the four `ddgi_cascade_*.png` evidence images differ on **every** run with
identical code (verified by hashing across two back-to-back runs of the same binary), so they are
not committed here. Their RMSE assertions passed.

Closes #1279

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DDGI probe relocation warm-up scheduling so probes settle on their final capture position.
  * Prevented the final warm-up capture from applying an additional relocation movement, reducing residual probe displacement and improving stability near surfaces.
  * Preserved existing behavior for inactive probes and cascade shifts.

* **Documentation**
  * Updated DDGI probe state and warm-up scheduling guidance.

* **Tests**
  * Added coverage for warm-up timing, final-position settling, and stable spring behavior.


drsnuggles8:
@coderabbitai review

Issue #1230, opened 2026-09-13
Renderer: canonical foliage instance identity and spatial groups


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1221

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Make existing foliage instances identifiable canonical scene members, grouped spatially for later visibility/residency work. Preserve their identity through terrain regeneration and the existing raster card path.

## Acceptance criteria

- [ ] Each covered plant instance has stable identity, material association, transforms, bounds and explicit representation metadata.
- [ ] Regeneration, layer editing, terrain transform and removal update/invalidate affected instances deterministically.
- [ ] Existing cards still render on both backends, and GPU Scene diagnostics distinguish represented foliage from remaining unsupported variants.
- [ ] Identity survives command submission and prepares for compaction/LOD without requiring GPU culling in this slice.

## Scope boundaries

No new culling algorithm or plant shader required here. Keep identity independent of physical instance-buffer order.

## Blocked by

None - can start immediately.

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 13
craft: 5
stability: 5
decay: 3
effort: 3
confidence: 0.8
learning: 5
fun: 8
kano: table-stakes
blocked_by: []
blocks: [1221, 1240, 1233]
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

Issue #1229, opened 2026-09-13
Vulkan RT: moving skeletal and morph surfaces in BLAS/TLAS and hybrid effects


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1221

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Feed canonical animated geometry into the existing deformed-geometry update policy so a moving animal appears in hybrid shadows, reflections and reference ray hits.

## Acceptance criteria

- [ ] Ray hits track the deformed triangle positions and alpha-mask semantics rather than rest-pose data.
- [ ] Vertex-only changes use valid updates; topology, count, capacity and incompatible build-flag changes trigger rebuilds with correct barriers/lifetimes.
- [ ] A moving animal and expressing head have matching raster/RT silhouettes; camera cuts and despawn do not leave ghost instances.
- [ ] Expose build/refit timings, memory, rebuild reasons and policy tradeoffs; preserve tested non-RT rendering.

## Scope boundaries

Reuses existing acceleration-structure manager; do not introduce a second scene or force RT on OpenGL.

## Blocked by

- #1228

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 8
craft: 8
stability: 5
decay: 3
effort: 5
confidence: 0.8
learning: 8
fun: 8
kano: table-stakes
blocked_by: [1228]
blocks: [1221, 1253, 1258]
```
[...]

Issue #1226, opened 2026-09-13
Renderer: shared skeletal deformation output consumed by raster, depth, shadows and velocity


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1221

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Render one animated animal from a reusable current/previous deformed surface. Measure a GPU deformation cache against repeated vertex skinning, then implement the chosen shared producer/consumer contract with inspectable cost. Preserve an explicit capability fallback.

## Acceptance criteria

- [ ] A skeletal test subject renders with matching positions/normals/tangents in colour, depth and available raster shadow techniques.
- [ ] Current/previous deformation produces correct per-pixel motion during animation, pause/resume and camera cuts; reset history explicitly on discontinuities.
- [ ] Allocation, synchronization, lifetime and bounds stay correct under spawn/despawn, skeleton replacement and parallel recording.
- [ ] Capture a measured cache-versus-current-path comparison and document the implementation decision; do not claim a speedup without a measured workload.

## Scope boundaries

Single and several independently animated conventional meshes. General crowd batching belongs to #1031; morph targets follow separately.

## Blocked by

None - can start immediately.

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 13
craft: 5
stability: 5
decay: 5
effort: 5
confidence: 0.8
learning: 8
fun: 8
kano: table-stakes
blocked_by: []
blocks: [1221, 1227, 1231, 1236]
```
[...]

Issue #1031, opened 2026-09-03
Renderer: per-instance skinned-mesh batching (bone archetypes or per-instance bone SSBO)


drsnuggles8:
Extracted from `docs/design/GPU_INSTANCING_FUTURE_IMPROVEMENTS.md` §8 (the #173 follow-up list) so the doc can be retired.

> **Re-scoped 2026-09-14 against master @ `09339b51e`.** The original body was verified on 2026-09-03. PR #1263 (issue #1226) then landed one shared skeletal-deformation producer and rewrote every skinned shader, and the binding namespaces filled up. Three of the four premises below changed. The original analysis is preserved in the *Superseded* section at the end.

## What blocks it today — re-verified against master @ `09339b51e`

`CommandBucket.cpp:600` still skips animated draws when auto-batching (`if (cmd->isAnimatedMesh) continue;`). That is the one premise that survived unchanged, and it is the whole of the remaining work.

Everything else moved:

**1. The shader rewrite is one file now, not five.** #1263 moved all linear-blend skinning into `OloEditor/assets/shaders/include/SkeletalDeformation.glsl`, which owns the palette declaration (`binding = 4`), the opt-in previous-pose palette (`binding = 31`, behind `OLO_DEFORM_WANT_PREV`) and the `OloSkinMatrix` / `OloDeformSkinnedVertex` math. The seven consumers — `PBR_MultiLight_Skinned`, `PBR_GBuffer_Skinned`, `DepthPrepass_Skinned`, `DepthPrepass_MaskSkinned`, `ShadowDepthSkinned`, `VSM_DepthSkinned`, `VSM_DepthLocalSkinned` — contain no palette indexing of their own, and `SkeletalDeformationContractTest` fails the build if one grows some. (The original body's list of five also named `PBR_Skinned.glsl`, which had already been retired in `7450ff7d6`, before the issue was filed.)

**2. The skinned shaders are already per-instance for transforms.** Every one of them includes `include/InstanceBlock_Vertex.glsl` and calls `OLO_INSTANCE_FORWARD()`, so `u_Model` / `u_Normal` / `u_PrevModel` resolve to `instances[gl_InstanceIndex]` out of `SSBO_INSTANCE_DATA` (binding 15). A batched skinned draw already places each instance correctly, carries per-instance entity IDs for picking, and already emits per-object motion vectors. **Only the bone palette is still per-draw.**

[...]

PR #1286, opened 2026-09-15
Renderer: animated meshes in GPU Scene with stable surface identity (#1228)

### LINKED ISSUE #1228: Renderer: animated meshes in GPU Scene with stable surface identity

## Parent

#1221

## Agent Brief

**Category:** enhancement

## What to build

Extract shared animated surfaces into canonical GPU Scene records and use their identity/material links in the raster submission path. Make the currently unsupported animated category a real supported path for the covered sources.

## Acceptance criteria

- [ ] Each animated instance/submesh has stable geometry/material identity and current/previous deformation revisions.
- [ ] Spawn, removal, asset reload, LOD switch and topology replacement invalidate only affected records and histories.
- [ ] Diagnostics identify supported versus unsupported variants accurately; no category is marked supported solely to remove a warning.
- [ ] Raster submission and debug views resolve the same animated surface records, including concurrent command recording.

## Scope boundaries

Canonical records and raster consumption; BLAS/TLAS consumption is the next slice.

## Blocked by

- #1227

## Score
```olo-score
capability: 13
craft: 5
stability: 5
decay: 5
effort: 3
confidence: 0.8
learning: 5
fun: 8
kano: table-stakes
blocked_by: []
blocks: [1221, 1229, 1249]
```

drsnuggles8:
Closes #1228.

Animated meshes now reach canonical GPU Scene instance records, and the raster path consumes their draw links.

## The finding that shaped the change

**Most of the identity was already expressible; nothing was using it.** A skinned mesh's vertex buffer holds a rest surface that the vertex stage deforms, and `Scene::EvaluateMorphTargets` writes that same buffer in place, so the geometry key `(vertex buffer, index buffer, submesh)` was *already* stable across the animation. Three sites short-circuited before staging, with a comment saying a skinned entity "is not representable in GPU Scene (the canonical record holds one rigid transform)".

That was true of the **deformation**, not of the **identity** — and the distinction is the whole issue. The one rigid thing the record holds, the world transform, is as true for a skinned entity as for any other. So the diff is mostly *deleting an exclusion* and adding the one field that genuinely was missing.

### 1. Stable geometry/material identity per instance and per submesh, plus current/previous deformation revisions

`SkeletonData` carries a monotonic `u32` revision pair, advanced by one on the single continuous branch of `AdvanceBoneHistory` and **held equal** on every discontinuity. Continuity is `current == previous + 1` in modular arithmetic, which survives the ~2.2-year `u32` wrap because `previous + 1` wraps exactly when `current` does.

`GPUSceneInstance` grows four lanes, 128 → 144 bytes, mirrored in `include/GPUScene.glsl`.

**The split is deliberate: identity is per submesh, the revision is per entity.** One character is N records sharing one pose. Keying the pose per submesh would let two submeshes of one character disagree about their shared palette; keying identity per entity would give every submesh after the first the previous one's transform history — the exact defect the per-entity previous-transform cache had before #994.

- `GPUSceneAnimated.OneEntityKeepsOneRevisionAcrossManySubmeshIdentities`
- `GPUSceneAnimated.AdvancingThePoseKeepsTheSlotAndTheGeneration`
- `GPUSceneAnimated.ARigidInstanceIsNotAnAnimatedOneWithoutHistory`

**A revision pair rather than a boolean, and why it matters here.** `HasBoneHistory()` already existed and is enough for raster. It is not enough for a canonical record: a consumer that *built* something from this surface — #1229's acceleration-structure refit — must ask "has it deformed since I built?", which a per-frame flag cannot answer.

Two spellings of one fact is the risk. The counter-move is not inspection: `GPUSceneAnimated.TheRevisionPairNeverDisagreesWithTheBonePalettes` walks construction, first use, advance, a reset *after* an advance, recovery and a bone-count change, asserting `HasBoneHistory() == HasContinuousDeformation()` at every step.

### 2. Spawn, removal, asset reload, LOD switch and topology replacement invalidate only affected records and histories

Narrowness here is **structural, not defended by code**: all five events change the instance key or the buffer handles, so only the affected record's slot moves. The tests assert the half a happy-path test omits — what did *not* change:

- `GPUSceneAnimated.OneSurfacesTopologyChangeLeavesEveryOtherRecordUntouched`
- `GPUSceneAnimated.SpawnAndRemovalTouchOnlyTheirOwnRecords`

**The one real design decision in this PR.** The instance commit derives `PreviousTransform` from the slot (`slot.m_Live ? last frame's : this frame's`). The deformation lanes deliberately do **not** use that rule — they are copied verbatim from the producer. The slot rule gives two wrong answers, both silent:

- **A surface culled for one frame.** `AdvanceHistory` runs for every skinned entity every frame, drawn or not, so its bone history is intact when it returns. The slot rule would call that a discontinuity and discard an available velocity.
- **A reset that lands after the frame's advance** — a morph surface that moved, an LOD switch, a teleport found during the tick. The skeleton records it by holding `prev == current`; the slot rule cannot see it at all and reports continuity across the exact seam the rejection exists to declare. That is a velocity measured between two different surfaces, reaching TAA and motion blur two subsystems from the cause.

The divergence is written into `GPUSceneTypes.h` next to the lanes, and `TheRevisionPairNeverDisagreesWithTheBonePalettes` covers precisely that after-the-advance case.

### 3. Diagnostics identify supported versus unsupported variants accurately

`GPUSceneUnsupportedCategory::Skinned` was a blanket count of every skinned entity, taken before any submission ran. It is **not deleted** — deleting it would leave the diagnostics unable to say anything about an animated entity that still cannot be represented.

It now means *an animated entity that reached a submission path and produced no canonical record*, reported per entity at the sites that know the refusal. Per **entity** deliberately: `ExtractGPUSceneMesh` already ticks `NotExtractable` per refused **submesh**, and a per-submesh Skinned tick would double-count one event under two names. So it reads 0 in a healthy scene and moves the moment something stops being representable.

What the records *did* take is `GPUSceneAnimatedStats`, in the shape `GPUSceneFoliageStats` established for #1230, surfaced through `olo_rt_scene_stats`. It partitions its own canonical instances into with-history and without-history, and the evidence test asserts that partition — a third number that is neither is a producer bug, not a property of a scene.

Skinned **virtual** meshes still ride the cluster-LOD DAG and are still counted under `Virtualized`; they now also appear as `unsupportedVariants` so a scene whose characters are all virtual reads as "0 canonical, N variants elsewhere" rather than as an empty census that looks like nothing was ever offered.

### 4. Raster submission and debug views resolve the same records, including concurrent command recording

`DrawAnimatedMesh`, `DrawAnimatedMeshParallel` and `MeshSubmitDesc` carry a link. It reaches the same choke point a rigid link reaches (`CommandDispatch::UploadModelInstance`), so a linked animated draw takes its transforms and material slot from the record instead of the per-entity cache — which was keyed on the entity alone and therefore gave every submesh after the first a zero velocity.

**This criterion was a trap, and I want it flagged explicitly.** Before this PR, `DrawAnimatedMeshParallel` and the `SubmitMeshesParallel` bracket had **no callers anywhere in the repository** — `Model::DrawParallel` is the only live user of the bracket and hard-codes `IsAnimated = false`. Threading a parameter through an entry point nobody calls and ticking the box would repeat #1226 exactly: its `HasBoneHistory()` guard went onto `Renderer3D::RenderAnimatedMeshes`, which also has no callers, and looked load-bearing through two review passes while the path that actually rendered stayed ungated (`skeletal-deformation-shared-output.md` §3).

So the parameter is threaded **and driven**. `AnimatedGPUSceneConcurrentRecordingTest` is the caller: it mints 24 distinct links on the main thread, records from the worker contexts with `minBatchSize = 1`, and compares the links the workers wrote into their packets against the ones the main thread minted. A branch that dropped the parameter submits the same number of packets and draws the same picture; only that comparison fails.

The thread-safety argument is structural, not a lock: links are **minted** on the main thread (`ExtractGPUSceneMesh` appends to an unsynchronised per-frame vector) and only **carried** across the boundary as a plain integer copied into the descriptor. The test pins the invariant that makes that safe — the link table does not grow during the parallel region — rather than chasing a race by repetition, which is TSan's job and not something this box can do.

## Verification

Hardware: Windows 11, RTX 4090, clang-cl Debug (`dev-cached`).

**Tests — full suite: 8552 tests / 1352 suites, 8550 passed, 2 failed, 7 skipped.**

Both failures are pre-existing on master and **not caused by this branch**: `ReSTIRPTDevice.ClosedCornellDiffuseAndGlossyIndirectMatchExactlyTruncatedReferenceSupport` and `ReSTIRPTDevice.DiffuseAndNearSpecularEnvironmentAgreeWithTheWholeReferenceIntegral`. I built base `7f818a8e7` detached and ran them there: they fail with **byte-identical** error values (0.078408916319963828 / 0.077918366921105986; 0.64058356137445349 / 0.54203124813806092 / 0.44348109450830953). An OBS Vulkan layer is injected on this box (`VK_LAYER_OBS_HOOK uses API version 1.3 … May cause issues`), which is a candidate but unconfirmed. Not investigated further — it is outside this issue, and no open issue names these tests.

Affected suites, re-run together: **52/52 passed** (`GPUScene.*`, `GPUSceneAnimated.*`, `SkeletalDeformationContract.*`, `AnimatedGPUSceneScene.*`, `AnimatedConcurrentRecordingScene.*`), including `GPUSceneLayoutTest`'s SPIRV-Cross reflection check of the new GLSL lanes against the C++ `offsetof` table, and `GPUSceneAntiDuplicationRatchetTest`.

`OloEditor`, `OloRuntime` and `OloServer` all link — they are not in the test build and have reached master broken before.

**Captures** — `OloEditor/assets/tests/visual/AnimatedGPUScene_*.png`, 13 images, all inspected:

- four angles (Front/Left/Right/Above): correct geometry and shadow, no transposition or double-shift;
- five frames of a bending bone (Bend0–4): the upper half leans progressively while the lower half stays put, so the pose is reaching the vertex stage;
- one after a declared discontinuity: still draws — a dropped history costs the surface its velocity, never its geometry;
- one per rendering path (Forward / Forward+ / Deferred).

**Live editor** — `fox.olo`, the real imported skinned character, driven over MCP:

- `olo_shader_errors`: `count: 0`, status ready. `OloEngine.log`: zero errors.
- `olo_rt_scene_stats` → `gpuScene.animated`: `canonicalEntities 1, canonicalInstances 1, surfacesWithHistory 1, surfacesWithoutHistory 0, unsupportedInstances 0, unsupportedVariants 0`; `notStagedByCategory.skinned: 0`; `notStagedTotal: 0`.
- `olo_skeletal_deformation_stats`: `skeletonsAdvanced 1, skeletonsWithHistory 1, boneMatricesAdvanced 24`, `historyResets.total 1` (all FirstUse). **The producer's `skeletonsWithHistory` and the census's `surfacesWithHistory` agree in a live scene**, which is the cross-check the unit test makes structurally.
- Screenshot: the fox renders correctly.

**Backends and paths.** `EveryRenderingPathConsumesTheSameAnimatedRecords` asserts a non-zero `CommandDispatch::GetGPUSceneConsumedDrawCount()` on Forward, Forward+ and Deferred — a path that quietly stopped carrying links would still draw, because an unresolved link falls back by design, so the counter is the only thing that catches it. The transform half of a link is consumed at the single-instance choke point both backends share; the material half is the Deferred G-buffer's read. Vulkan is covered by the existing `VulkanPassSuite` in the full-suite run above; I did not add a Vulkan-specific animated capture, and RT consumption is #1229 by the issue's own scope boundary.

**Cost and memory.** Fox scene at 1411×942: 5.03 ms frame, 2.73 ms CPU, 1.63 ms GPU, 198 fps; `ScenePass` 1.572 ms CPU / 0.793 ms GPU (DepthPrepass 0.045, Color 0.723); `ShadowPass` 0.048 ms GPU; total pass GPU 1.407 ms. No pass is attributable to this change — the deformation lanes are data, not work.

The honest cost is **memory**: `GPUSceneInstance` grew 16 bytes, so the instance table costs `16 × instance count` more — 1 KiB at the default capacity of 64, 320 KiB at 20k instances, and the same 12.5% on every upload of a dirty instance range. That is paid by rigid scenes too, since the lanes live on the shared record. The alternative — a sixth record kind — needs an SSBO binding, and the namespace has none free (`gpu-readback-stats-channel.md` §8).

## Known limitations, stated rather than left to be found

- **`CommandBucket` auto-batching drops draw links** when it collapses N draws into one instanced call, and #1031 batches skinned draws that share a pose — so a batched animated draw falls back even though its link resolved. Pre-existing for rigid draws and now reachable for skinned ones. It is **counted**, not silent (`GPUSceneFallbackDraws`), and the exit condition is the per-instance link lane in `InstanceData` already named in `GPUSceneLegacyAdapters.h`.
- **The concurrent animated path has no production caller.** This PR makes it correct and pins it with a test; it does not invent a caller for it.
- `Renderer3D::RenderAnimatedMeshes` / `RenderAnimatedMesh` remain callerless and were left alone.

## Docs

New guide `docs/agent-rules/animated-surface-records.md` — one line in the README's renderer index and one row in the failure-mode table, per CLAUDE.md. Its rule: a record carrying two histories derives them by two different rules, and the deformation one must come from the producer that owns the palettes.

## Summary by CodeRabbit

- **New Features**
  - Animated meshes now retain deformation history across frames for smoother rendering.
  - GPU scene diagnostics report animated-surface counts, history continuity, unsupported instances, and variants.
  - Animated rendering supports consistent tracking across parallel submissions and multiple submeshes.
  - Reset causes are recorded to improve deformation-history diagnostics.

- **Documentation**
  - Added guidance covering animated surface records, deformation continuity, and GPU scene tracking.

- **Tests**
  - Added coverage for animated rendering, record stability, concurrent submissions, layout validation, and deformation-history behavior.


drsnuggles8:
The two limitations named in the PR body now have scored follow-ups, so neither is left as prose in a merged description:

- **#1290** — `CommandBucket` auto-batching drops resolved GPU Scene draw links. Not caused by this PR; this PR makes it reachable for skinned draws too, because #1031 batches skinned draws that share a pose. The `InstanceData::GPUSceneRef` lane already exists per instance (#994) and four fields in the same upload loop already survive the N-into-1 collapse via `FrameDataBuffer` streams, so the shape is known — the blast radius is `GPUFrustumCuller`, which rewrites the stream and must preserve the lane. Giving the *explicitly*-instanced path links is deliberately out of its scope: that path passes `GPUSceneDrawLinkRequest::None` on purpose.
- **#1289** — the concurrent animated submission API has no production caller, and two more entry points in the same family have none at all. Filed as a decision issue (wire a caller vs. retire the API) with `confidence: 0.5`, because a Release crowd-scene measurement is what should pick the branch. Deliberately not resolved here: this PR makes the path correct and pins it with a test that is itself the caller, which is the honest state for a criterion that named concurrent recording, but "the test is the only caller" should be a deliberate state rather than an inherited one.

Neither was folded into this PR. Both are rigid-draw / submission-architecture work rather than animated-mesh work, and #1228's scope boundary is canonical records and raster consumption for animated surfaces.

drsnuggles8:
@coderabbitai review

drsnuggles8:
Reviewed all five findings against the code. **All five were valid**, and three were real defects introduced by this PR. Fixed in b2c3bfca5.

**1 — `GPUSceneDrawLink` not forwarded in the serial animated calls (`Renderer3DMeshSubmission.cpp`).** Confirmed and the most serious of the five. I forwarded the link on the serial `DrawMesh` branch and both parallel animated branches and missed these two. Which branch runs is decided by `numMeshes < minBatchSize * 2` alone, so the same scene would take the canonical record above the threshold and the legacy per-entity history below it — a path split on batch size, which is the exact failure mode `SubmitMeshSourceClassic`'s own comment warns about. Both calls now forward it, with a comment saying why the two arms must agree.

**2 — `m_CanonicalEntities` never incremented in `SubmitMeshSourceClassic`.** Confirmed. `GPUSceneAnimatedStats` documents "an entity contributes one here and one instance per submesh", and the virtual-mesh classic fallback violated it — canonical instances with zero entities. Latched with a `countedCanonicalEntity` flag across the per-submesh loop so the entity is counted exactly once. Your note that the per-submesh `Skinned` over-report is unreachable through current callers is correct, and I left that path as-is rather than adding a guard for a state no caller can produce.

**3 — variant census not gated on `VirtualGeometryEnabled`.** Confirmed, and it is the sharper of the two census bugs. `virtualPathOwnsMeshEntities` is gated on `Path == Deferred` only, so with virtual geometry switched off on Deferred the virtual loop takes its classic fallback (staging canonical records and counting the entity as a canonical instance) while the animated loop also ticked `m_UnsupportedVariants` — one entity counted twice under two contradictory headings, in the census whose entire purpose is to say which of the two happened. Gated on `Renderer3D::GetRendererSettings().VirtualGeometryEnabled`, with the reasoning written in.

**4 — capture write failures ignored.** Confirmed and taken. Adopted your suggestion. Worth noting this file had *already* been bitten by this class of bug during development: the captures initially landed in `OloEditor/OloEditor/assets/tests/visual/` because renderer init leaves the process cwd inside `OloEditor/`, so a relative path evaluated after the first frame doubles the prefix. That directory is created happily and every assertion passed. The path is anchored now, and with your change a write that fails is a failure rather than a silent absence.

**5 — no submission carried a prev≠current palette delta.** Confirmed, and this was the important one. Your analysis of the ordering is exactly right: `Capture` runs editor frames, each frame runs `AdvanceHistory` (prev := current) before rendering, and the fixture wrote the pose from outside the frame — so every submission carried `prev == current` and every draw emitted exactly zero bone motion. The surface moved *between* frames but never *within* one. The revision, continuity and luminance assertions all still passed, so the PR's claim of "captures with motion" was weaker than I had stated.

Fixed the way you indicated rather than by moving the call between frames. Edit mode previews animation through `AnimationSystem::Update`, which runs *after* `AdvanceHistory` and *before* submission — the production ordering — so a real playing clip is what produces the delta. `APlayingClipSubmitsANonZeroPaletteDeltaUnderAContinuousRevision` authors a clip on the primitive's `Upper` bone (keyed by name, since `AnimationSystem` samples by name and a wrong name would sample nothing and pass for the wrong reason), asserts a non-zero palette delta actually reaches submission, and then declares a mid-animation `Teleport` to check the record cancels continuity while the palette is genuinely moving.

**Verified by negative control**, since a test that passes is not a test that measures: with `m_IsPlaying = false` it fails on precisely the delta assertion —

```
no frame submitted a previous palette that differs from the current one, so every draw emitted
exactly zero bone motion and nothing here tested a velocity at all
```

— then passes again when re-enabled. The manual-bend test is kept, because it is still valid for the revision and continuity claims; both tests now state why they are separate.

Affected suites after the fixes: 41/41 (`GPUScene.*`, `GPUSceneAnimated.*`, `AnimatedGPUSceneScene.*`, `AnimatedConcurrentRecordingScene.*`), plus `SkeletalDeformationContract.*` green. Five new captures committed.

Issue #1288, opened 2026-09-15
ReSTIR PT: the metal / near-specular arms return ~1.5% of the reference integral


drsnuggles8:
Found while verifying #1241; unrelated to it, filed rather than folded in.

Two `ReSTIRPTDevice` tests fail reproducibly on this box (RTX 4090, Debug, `build-cached`):

- `ReSTIRPTDevice.ClosedCornellDiffuseAndGlossyIndirectMatchExactlyTruncatedReferenceSupport`
- `ReSTIRPTDevice.DiffuseAndNearSpecularEnvironmentAgreeWithTheWholeReferenceIntegral`

**Only the metal / glossy arms fail. The diffuse arms pass.**

Furnace test, `metal=1`:

```
reference = 0.650039, 0.550032, 0.450027
observed  = 0.009455, 0.008001, 0.006546
```

about **1.5%** of the expected value. `metal=0` in the same test passes
(`reference=0.661059,... observed=0.657516,...`).

Closed Cornell, `glossy=1`: `reference=0.11546 observed=0.037051` on the initial-proposals arm and
`0.037542` on temporal-plus-spatial-MIS, against a tolerance of `4*stderr + 5e-4 = 0.0037`.

## Attribution

Not from #1241's branch:

- reproduces in isolation, so it is not the known full-suite load flake;
- `ReSTIR_PT.glsl` includes `PBRCommon.glsl`, which that branch edits, so I restored master's
  `PBRCommon.glsl` and re-ran: the metal arm returned **byte-identical** values. The only shared
  shader is exonerated;
- no C++ file that branch touches is in the path tracer's path, and the test scene has no skin
  material in it.

## Why it may have sat unnoticed

These are L7 Vulkan ray-query tests. The self-hosted runners have no usable Vulkan RT device, so
this suite does not run in CI and a regression here is invisible to the PR gate.

## Where to start

The split between a passing diffuse arm and a ~1.5% metal arm points at the metal/near-specular
branch of the BRDF evaluation or its PDF/MIS weight in `ReSTIR_PT.glsl`, not at the reservoir
machinery - a broken reservoir would move both arms.

## Score
```olo-score
capability: 5
craft: 3
stability: 5
decay: 3
effort: 3
confidence: 0.8
learning: 5
fun: 5
kano: table-stakes
blocked_by: []
blocks: []
```
[...]

Issue #1272, opened 2026-09-14
Asset-imported MeshSources carry no bone influences, so a VirtualMeshComponent can never be skinned


drsnuggles8:
Split out of #1150, which made the renderer able to draw skinned virtual geometry and then could
not demonstrate it on a real asset — because **no `MeshSource` asset in the engine has ever carried
bone influences**.

## What is true on `master` (verified live, 2026-09-14)

The asset path for a mesh is

    MeshSourceSerializer::TryLoadData
      -> MeshImporterRegistry::Import
        -> AssimpMeshImporter
          -> Model(path)                      <- never extracts bones
          -> Model::CreateCombinedMeshSource  <- copies vertices, indices, submeshes;
                                                 not influences, bone info or skeleton

`Model` has no `mNumBones` handling at all. `AnimatedModel` does — it extracts influences, bone info
and the skeleton, and its own combine carries all three (`AnimatedModel.cpp:54-90`) — but nothing on
the **asset** path routes to it. So `MeshSource::HasBoneInfluences()` is false for every registered
mesh asset, whatever the source file contains.

Consequences, only the first of which is new:

- a `VirtualMeshComponent` (which addresses its mesh by `AssetHandle`) can never be skinned, so
  #1150's feature is unreachable from the editor;
- a `MeshComponent` can only be skinned when its `Ref<MeshSource>` came from an `AnimatedModel`
  (i.e. via `AnimationStateComponent`'s `SourceFilePath`), never from an asset handle.

## Reproduction

`OloEditor/SandboxProject/Assets/Scenes/VirtualGeometrySkinned.olo` (added by #1150). Deferred path,
press Play. Three CesiumMan characters:

- the CLASSIC one (left) walks — its `MeshComponent` holds the `AnimatedModel` source;
- both VIRTUAL ones stand in the bind pose — their `AssetHandle` resolves to a `MeshSource` with no
  influences.

`OloEngine.log` says so directly, once per mesh:

> virtual mesh 43f484b877c67484 is submitted with a bone palette but no part of its cooked cluster
> DAG carries a skinning payload — it renders in its REST POSE through the virtual path.

The renderer side is not in question: #1150's `VirtualSkinnedBounds` suite and
`VirtualGeometryVisualEvidence.ASkinnedVirtualMeshDeformsOnScreenFromEveryAngle` drive the same
pipeline from a procedurally rigged `MeshSource` and it deforms correctly.

## Shape of the fix

[...]

Issue #1247, opened 2026-09-13
Groom: physically grounded fibre scattering for dark and pale hair/fur


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1223

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Light the strand visibility samples with a fibre scattering model supporting directional reflection/transmission and controlled multiple-scattering approximation.

## Acceptance criteria

- [ ] Define tangent frame, width/roughness and pigment or absorption parameter meaning; material controls cook and round-trip.
- [ ] Compare measured/reference fibre lobes before choosing approximations, and provide matching evaluation/sampling/PDF where used by stochastic estimators.
- [ ] Dark, pale and coloured fibres respond plausibly under frontal, grazing and backlight with separated diagnostic contributions.
- [ ] Lighting terms conserve energy within declared approximations; scene lights and environment lighting use consistent material parameters.

## Scope boundaries

Local fibre response. Inter-fibre occlusion/density transport is owned by the following shadow slice.

## Blocked by

- #1246

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 8
craft: 13
stability: 3
decay: 1
effort: 5
confidence: 0.5
learning: 13
fun: 13
kano: performance
blocked_by: [1246]
blocks: [1223, 1248]
```
[...]

Issue #1246, opened 2026-09-13
Groom: measured strand visibility and stable subpixel composition


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1223

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Build the first production strand visibility path from a cooked groom. Start with a bounded comparison of viable coverage/composition approaches, then render a static coat with neutral lighting and inspectable visibility buffers.

## Acceptance criteria

- [ ] Compare near silhouette, dense overlap, subpixel coverage, depth/composition cost and memory before recording the selected approach.
- [ ] Curves render with width and depth-correct overlap against the body and ordinary scene geometry.
- [ ] Camera movement, MSAA/TAA/upscaling and varying resolution have measured coverage error and stable silhouettes.
- [ ] Capability fallbacks are explicit on OpenGL/Vulkan; diagnostic curve rendering is not advertised as a finished quality tier.

## Scope boundaries

Neutral-lit coat isolates visibility. Fibre optics, dense self-shadowing and simulation are later slices.

## Blocked by

- #1232
- #1239

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 8
craft: 8
stability: 3
decay: 1
effort: 5
confidence: 0.5
learning: 13
fun: 13
kano: performance
blocked_by: [1232]
blocks: [1223, 1247, 1249]
```
[...]

PR #1292, opened 2026-09-15
Groom: curve assets, deterministic cooking and static editor preview (#1232)

### LINKED ISSUE #1232: Groom: curve assets, deterministic cooking and static editor preview

## Parent

#1223

## Agent Brief

**Category:** enhancement

## What to build

Import groom curves into a dedicated asset with group/root/width attributes and deterministic cooked data. Provide a static debug preview so import correctness is visible before production hair shading.

## Acceptance criteria

- [ ] Support a documented curve input convention, including Alembic curves; do not treat existing polygon Alembic import as curve support.
- [ ] Root UVs, widths, curve offsets, groups, guide designation and bounds survive import/reimport/cook/save/load.
- [ ] Reject malformed curve data and unsupported attributes explicitly; preview shows roots, groups and curve direction.
- [ ] Reference animal and human grooms load with source provenance and meaningful diagnostics.

## Scope boundaries

Static debug preview is the acceptance target, not a photoreal coat. No in-engine full grooming DCC replacement.

## Blocked by

None - can start immediately.

## Score
```olo-score
capability: 8
craft: 5
stability: 5
decay: 3
effort: 3
confidence: 0.8
learning: 5
fun: 8
kano: performance
blocked_by: []
blocks: [1223, 1246]
```

drsnuggles8:
Closes #1232.

A groom becomes its own asset type: Alembic `ICurves` in, a deterministic `.ologroom` out, and a
static debug preview that makes an import checkable by eye before any hair shading exists.

### 1. A documented curve input convention, including Alembic curves — and polygon Alembic import is not treated as curve support

The convention is written down in the header comment of
[`AlembicGroomImporter.h`](../blob/feature/groom-curve-assets-1232/OloEngine/src/OloEngine/Asset/Interchange/Alembic/AlembicGroomImporter.h)
(what is read, at which scopes, and what is refused) and summarised as a table in the new guide
[`docs/agent-rules/groom-curve-import.md`](../blob/feature/groom-curve-assets-1232/docs/agent-rules/groom-curve-import.md).

The separation from polygon import is structural, not a comment. `AlembicMeshImporter` keeps
`IPolyMesh` / `ISubD`; the new `AlembicGroomImporter` reads `ICurves`; and an `.abc` is routed by
**content** (`ArchiveContainsCurves`), because one extension carries both schemas. Importing a groom
as a mesh loses per-strand identity, the root parameterisation and the guide flag — the three things
every later groom feature is defined in terms of.

Evidence — `GroomRejection.APolygonOnlyArchiveIsRejectedWithTheCurveSupportDistinction`: an archive
with no `ICurves` is refused with a diagnostic that says so in as many words, rather than falling
through to the mesh path.

### 2. Root UVs, widths, curve offsets, groups, guide designation and bounds survive import / reimport / cook / save / load

`GroomRoundTripTest` asserts all six as one unit at each stage of the chain, so no case can pass by
checking a subset. Positions and widths are compared **exactly** (`==`), not with an epsilon: the
pipeline stores IEEE-754 bit patterns end to end and does no arithmetic on them after import, so an
epsilon would hide exactly the bug this criterion exists to catch.

| Case | What it walks |
|---|---|
| `ImportReimportCookSaveLoadPreservesEverything` | a real 256-strand `.abc` → import → import again → cook → file → load |
| `CookSaveLoadPreservesEverySurvivingProperty` | the cook/save/load half with no Alembic involved |
| `ImportPreservesRootUVsAndPerVertexWidthsExactly` | hand-checkable: 2 strands, known UVs, known per-vertex widths |
| `SubGroupsBecomeNamedGroupsAndSurviveTheCook` | 120 strands, 3 interleaved sub-groups |
| `SourceTransformIsBakedIntoPointsAndWidths` | a 2x scale moves the tip to y=2 and makes the strand 2x thick |
| `BoundsIncludeStrandRadiusNotJustTheCentreline` | the AABB is widened by the largest radius |
| `CookSortsCurvesIntoContiguousGroupRanges` | every group is one range, and the ranges tile the curve array exactly once |

**Determinism** is treated as a contract. `GroomCookDeterminismTest` cooks the same source twice,
cooks two **separately imported** assets (independent allocations, independent hash maps), and cooks
the same curves in two **different source orders** — all must be byte-identical, and the failure
message names the first differing byte. The leaks that make a cook non-deterministic are each closed
deliberately and documented in `GroomCooker.h`: group ids come from first-appearance order rather
than hash-map order, the grouping sort is `std::stable_sort` (an unstable sort passes the
cook-twice case and fails the different-source-order one), every wire struct is explicitly padded
and `static_assert`ed, provenance carries no timestamp and no absolute path, and floats go to disk
as bit patterns rather than text.

The cook is a **canonicalisation** — a stable sort into contiguous group ranges — and never
resamples or remaps widths. That is what keeps the round-trip test a test of the pipeline rather
than of the cook's own rounding, and it leaves the "how many samples does a strand need?" decision
where it belongs, in #1246.

The cooked container is its own schema rather than an extension of `MeshBinaryFormat`. A groom has
no index buffer, no submesh and no material slot, and it does have an offset table, a root
parameterisation and a guide bit; sharing `.omesh`'s container would put two disjoint meanings on
one version line. (It also keeps this PR clear of #1272 / PR #1283, which owns those files.)

### 3. Malformed curve data and unsupported attributes are rejected explicitly; the preview shows roots, groups and curve direction

`GroomRejectionTest` (24 cases) asserts two things every time, and the second is the one that
matters: that the import failed, **and** that the diagnostic names the problem. The substring checks
are on the domain word ("periodic", "variable-order", "vertex counts", `groom_clumpid`), so
rewording a message does not break the suite while dropping the fact does.

Refused: periodic curves, variable-order curves, mixed bases in one archive, vertex counts that
disagree with the point array, a `widths` param with the wrong value count, non-finite or
out-of-range positions/widths/UVs, a negative width, an unregistered group, a declared group with no
curves, and **any unimplemented `groom_*` attribute, by name**. Accepted: a non-`groom_`
arbGeomParam (DCC bookkeeping — refusing those would make ordinary exports un-importable), a UV
outside the unit square (UDIM), and a zero width (a strand tapering to nothing). A missing `widths`
or `uvs` still imports, with the substitution **announced** rather than silent.

Corrupt cooked files are refused the same way: wrong magic, a future version (with "re-import" in
the message, because a `.ologroom` is a derived artifact), a payload that fails CRC32, a truncated
file, and a buffer too small to hold a header.

**The preview** draws strands, root crosses, one hue per group, and a root→tip brightness ramp. The
ramp is the only direction cue in the picture: a groom imported tip-first still looks like hair
without it.

Captures (committed as goldens, RMSE-compared on every run):

| | |
|---|---|
| `GroomPreview_Front.png` | the silhouette — strands growing inward instead of outward is obvious here and nowhere else |
| `GroomPreview_Side.png` | the droop profile, and whether roots sit **on** the surface |
| `GroomPreview_Top.png` | the two groups read as two hues across the hemisphere |
| `GroomPreview_RootCloseUp.png` | individual root crosses, a white guide strand, and the root-end darkening |

`GroomPreviewVisualEvidenceTest` also asserts four driver-independent contracts, each aimed at a way
the preview could be broken while still producing a plausible picture: the strands actually change
pixels versus a preview-off frame; roots and strands are independently visible; the direction ramp
darkens the root end (measured as a root-vs-tip **gap** against a ramp-off control, because the debug
line is emissive and a frame-wide mean just measures which strand won the depth test); and
colour-by-group produces more than one hue, with fewer when it is switched off.

### 4. Reference animal and human grooms load with source provenance and meaningful diagnostics

Two reference grooms are committed under `OloEditor/SandboxProject/Assets/Grooms/`, with a scene
(`Scenes/GroomReference.olo`) that places them:

| | curves | points | groups | guides | CPU | cooked |
|---|---|---|---|---|---|---|
| `reference-human-scalp` | 2 000 | 16 000 | 1 | 32 | 0.27 MiB | 0.18 MiB |
| `reference-animal-pelt` | 2 500 | 10 000 | 3 | 148 | 0.19 MiB | 0.12 MiB |

They are produced by `GroomReferenceAssets.ExportsTheEditorFixtures` — the same code path the
round-trip and determinism tests run — so the assets in the repo and the assets under test cannot
drift apart. Regenerate with `OLO_GROOM_EXPORT_DIR=<...>/Assets/Grooms OloEngine-Tests.exe
--gtest_filter=GroomReferenceAssets.ExportsTheEditorFixtures`.

Provenance is checked on load: source path (asserted **relative**, since an absolute path would make
the cooked bytes machine-specific), source format, a 64-bit FNV-1a content hash of the source bytes,
and the importer version. It must also survive the cook/load round trip. The editor's Groom
inspector shows all of it, plus the per-group curve counts with the same swatch colour the viewport
uses.

## Measurements (RTX 4090, Debug, `build-cached`)

At production-plausible scale, from `GroomReferenceAssets`:

| | curves | points | import | cook | CPU | cooked |
|---|---|---|---|---|---|---|
| human scalp | 20 000 | 160 000 | 66–105 ms | 148–216 ms | 2.73 MiB | 1.81 MiB |
| animal pelt | 30 000 | 120 000 | 112–213 ms | 129–181 ms | 2.26 MiB | 1.42 MiB |

The cook compresses (asserted, not just reported): an `.ologroom` larger than the resident arrays
would mean the payload is not being deflated at all.

## Two defects found by running it

Both were found because the preview was actually put on screen, and both are fixed here.

**`Renderer3D::DrawLine` builds a command packet and returns it — it does not queue it.** The
preview ran, logged thousands of submitted segments, and drew *nothing*, with no warning anywhere.
Every in-engine caller pairs it with `SubmitPacket` (`DrawWorldAxisHelper` is the readable example);
forgetting the pair is completely silent. Its `thickness` is likewise a **world-space** width in
5 mm units, not pixels, so a 9 cm scalp needs a thickness derived from its own bounds or 2 000
strands merge into a single blob. Both traps are now written down in the guide.

**The preview overflowed the frame's shared 65 536-entry transform buffer.** Two grooms in one scene
logged `FrameDataBuffer: Transform buffer overflow!` every tick. A strand costs one line per segment
**plus three** for its root cross, so a strand cap alone cannot bound it. `PlanGroomPreview` now
derives the stride from **both** a strand cap and a segment budget, whichever is tighter, always as a
stride rather than a truncation (a cap that drew the first N strands would show one group of a
three-group groom and look perfectly fine doing it). It reports which cap bound, and the inspector
says so — otherwise "I raised Max Strands and nothing changed" has no answer. Verified live: zero
overflows and zero errors in `OloEngine.log` afterwards.

`PlanGroomPreview` is split out as a pure function precisely so this selection rule is testable with
no GL context.

## Verification

* **Full suite:** 8 599 ran, **8 554 passed**, 2 failed, 43 skipped, exit 1.
  The two failures are `ReSTIRPTDevice.ClosedCornellDiffuseAndGlossyIndirectMatchExactlyTruncatedReferenceSupport`
  and `ReSTIRPTDevice.DiffuseAndNearSpecularEnvironmentAgreeWithTheWholeReferenceIntegral` — both are
  **open issue #1288**, which names these two tests exactly and attributes them away from any branch.
  They also fail in isolation here, so they are not the documented load flake either.
* **Groom suites:** 60/60 pass (`Groom*`), including the 5 GL visual-evidence cases against the
  committed goldens. The cross-binding guards pass too (`AssetExtensions*`, `BitwiseEqualLayout*`,
  `ComponentTupleCoverage*`, `SaveGame*` — 135/135).
* **OpenGL, live editor:** both reference grooms load and preview correctly (`GroomSerializer`
  reports `2500 curves, 10000 points, 3 groups, 148 guides; source 'Grooms/reference-animal-pelt.abc'
  [AlembicCurves]`), `olo_shader_errors` returns 0, and the log has no errors.
* **Vulkan, live editor** (`--rhi=vulkan`): the preview draws, in the right place, with the right
  topology, but the frame is wrong — see the caveat below. **Not claimed as parity.** Tracked as
  #1293.
* `pre-commit run --all-files` clean; `OloEditor` and `OloEngine-Tests` both build (the editor is not
  in the test build, so it was built explicitly).

## Vulkan caveat — filed as #1293, not fixed here

**Corrected since this PR was opened.** It originally listed *two* Vulkan defects. One of them was
wrong and is retracted: I claimed debug lines "render stippled on Vulkan", but the world-axis helper
looks **identical** on OpenGL and Vulkan when captured at the same camera — the original comparison
used two different camera angles. There is one defect, not two.

The remaining one is real: on Vulkan, enabling **any** groom preview drops one fixed 805 KB
`VulkanFrameArena` allocation **every frame**, so the frame renders with unbound buffers (the visible
symptom is washed-out strand colour in the Vulkan capture). The engine reports it loudly.

It is **not a volume problem**, which is the part that decided where it gets fixed. Sweeping the
strand cap from 10 to 2 000 (≈100 → ≈14 000 debug lines) leaves the request size, the arena cursor at
failure, and the per-frame drop count all unchanged; only disabling the preview entirely gives zero.
The arena is already ~96 % consumed *before* the request arrives, by consumers #1293 has yet to
identify. So there is no groom-side mitigation: this PR's segment budget changes the line count and
therefore changes nothing here — it fixes the equivalent **OpenGL** problem (the shared 65 536-entry
transform buffer), which is verified down to zero overflows.

Fixing it means answering "what fills 16 MiB on an ordinary editor scene?" inside the Vulkan RHI,
which is a different change with its own risk and no groom code in it. Full measurements and repro
are in #1293.

## A CI bug fixed on the way (own commit, 6f07925)

`ShaderCompileBudget.*` failed the routed TSan job on a data race whose **both** racing accesses are
inside `libvulkan_radeon.so` — RADV's barrier, on RADV's own shader-disk-cache threads
(`OloEngi:disk$1`, `disk$2`), during RADV's teardown:

```
Write of size 1 by thread T6:    #0 pthread_barrier_destroy
                                 #1 (libvulkan_radeon.so+0x2a3860)
Previous read of size 1 by T57:  #0 pthread_barrier_wait
                                 #1 (libvulkan_radeon.so+0x2a479c)
```

The engine appears in neither stack, only as the caller that created the null device
(`ShaderCompileBudgetTest.cpp:508`). `.github/sanitizer-suppressions/tsan.txt` already carries
`called_from_lib:libgallium` for exactly this family and for exactly this reason — the racing frames
*are* the pthread interceptors and carry no symbol of their own, so there is no `race:<symbol>` to
scope to — it simply named the GL driver and not the Vulkan one.

The gap opened on 2026-09-12 when the RADV null-device compile-budget guard landed: that test loads
`libvulkan_radeon` regardless of `--olo-gl-backend=none`, which falsified this file's own comment
that the routed TSan job "never loads a driver". That comment is corrected in the same commit, since
leaving it would send the next reader looking in the wrong place.

Not caused by this PR — master is green on the same shard, and nothing here touches a shader,
threading or Vulkan path (`git diff --name-only master...HEAD | grep -iE "shader|thread|vulkan"` is
empty). The likely trigger is that 66 added test cases reshuffle ctest sharding and so change what
runs concurrently in shard 1. The rule is scoped to the one library, so a genuine engine-side race
still reports normally — including one merely *guarded* by a driver barrier, whose racing accesses
would be in engine code.

## Also worth a follow-up

`VolumeVDBVisualEvidenceTest` (and several other evidence tests) rewrite their committed PNGs on an
ordinary, non-`--olo-golden-rebase` run, so a full-suite run dirties ~300 tracked files. I reverted
that churn here rather than committing it, but it makes "is my tree clean?" unanswerable after a
suite run, and it is an easy way to land unrelated binary churn in someone's commit. Not touched
in this PR; say the word and I'll file it.

## Scope

Static debug preview only, as the issue draws the boundary. No hair shading, no strand expansion, no
in-engine grooming DCC — the preview exists so that import correctness is visible *before* #1246 and
#1247 have anything to shade.

## Summary by CodeRabbit

* **New Features**
  * Added groom and hair/fur asset support, including Alembic curve imports and native `.ologroom` assets.
  * Added Content Browser support for importing Alembic files as groom assets.
  * Added scene support for attaching groom assets and viewing strands, roots, guides, direction ramps, and group colors.
  * Added groom inspection tools showing asset statistics, provenance, bounds, memory usage, and preview limits.
  * Groom preview settings now persist in saved scenes and projects.

* **Documentation**
  * Added guidance for groom importing, validation, canonicalization, and preview behavior.

* **Bug Fixes**
  * Improved diagnostics and rejection of malformed or unsupported groom data.


drsnuggles8:
@coderabbitai review

drsnuggles8:
All 16 addressed in dfbf3e6 — no pushback on any of them. Thank you; three were real and one was the feature not working.

**The importer had no production caller** (comment on `AssetImporter.cpp`). Correct, and the most important finding here. `AlembicGroomImporter` and `ArchiveContainsCurves` were referenced only by tests and by comments, so a groom `.abc` could not be imported in the editor at all — and the PR body claimed routing happened "by content", naming a predicate nothing called.

Wired as the content browser's **"Import as Groom"** action, following the documented `.vdb` → `.olovol` precedent beside it rather than hooking `EditorAssetManager::ImportAsset`: deciding by content means reading the file, and that entry point is metadata-only by design (the watcher calls it on files still being flushed). The sequence itself is now `AlembicGroomImporter::ImportAndCookToSidecar` in the engine, specifically so it is testable — the root cause was that nothing tested *reachability*, because every test called `Import()` directly. Three new cases cover it: loadable sidecar output, byte-identical to the direct route, and a polygon archive refused by name without leaving a sidecar behind.

**Resize before validation** (`GroomSerializer.cpp`). Correct, including the detail that the compressed path does not help. Replaced the six hand-rolled blocks with one `ReadArraySection` helper that validates, then sizes, then reads, so the wrong order is no longer expressible. A new test patches a cooked file's `Info` section to claim `MaxPointCount - 1` points, recompresses and fixes the CRC so the checksum cannot be what rejects it, and asserts the section check is.

**Group coverage** (`GroomPreview.cpp`). Correct — and my own subsampling test passed only because its fixture used equal group sizes, which is exactly the case where a global stride happens to hit everything. The stride is now phased per group (the cook's contiguous ranges make that possible); an un-canonicalised groom with interleaved ids falls back to a global stride and reports it, since phasing on interleaved ids would select everything. New test uses group sizes `{500, 5, 500}`.

Two notes where I did something other than what was suggested:

- **Assert the width-substitution diagnostic** — `Result::Diagnostic` is empty on success, so there was nothing to assert; the substitution was only ever a log line. Since "meaningful diagnostics" is an acceptance criterion, announced substitutions now reach the caller as `Result::Warnings`, and the test asserts on that. The editor logs them too.
- **`==` on `glm`** — fixed in `GroomPreviewTest`, `GroomReferenceAssetsTest` and the hue sector in the visual-evidence test. The byte-exact comparison in `GroomRoundTripTest` stays: exactness is that test's entire contract (an epsilon would hide a stage that quietly rescales, which is the bug AC 2 exists to catch), and the file header says so.

Also fixed: `MaxSegments` now enforced against each curve's exact cost at submission rather than only through the average-derived stride (with a new test using a layout where the average understates the selected curves' real cost); `tc.GetTransform()` → `GetWorldTransform` so a parented groom no longer draws at the wrong transform; bounded Alembic traversal depth; the builder's incomplete reset; `is_sorted` before building the permutation; the `RecordProperty` `int` cast; the missing `<cctype>`; the wrapped code span; and the `called_from_lib` overclaim — it suppresses *every* interceptor call originating in that library, and the comment now states that trade and names the narrowing escape hatch.

The selection rule moved into `SelectGroomPreviewCurves`, a pure function, so tests assert the real selection instead of re-deriving it — two existing tests were reimplementing the rule and would have kept passing while it drifted.

143 tests pass (`Groom*` plus the cross-binding guards).

One gap I am **not** claiming as verified: the "Import as Groom" menu item renders and dispatches. The engine half is unit-tested three ways, the action bit is unique, and the handler is a thin call mirroring the adjacent `.vdb` one — but I could not get injected input onto the content browser's context menu (dpiScale 1.5 plus a downscaled PrintWindow capture made the coordinate mapping unreliable), so that last six lines of ImGui wiring is argued, not demonstrated.

Issue #822, opened 2026-08-17
Build: CMake 4.3/4.4 native build instrumentation + per-file-set compile job pools


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

We cap every build at `-j6` / `OLO_LINK_JOBS=2` and, per #759, know from a *hand-rolled*
instrumentation run (the CMake Instrumentation API opted in manually via its pre-4.3
experimental gate UUID, recipe in
[docs/agent-rules/build-trees-and-windows-asan.md §5](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/build-trees-and-windows-asan.md#5-instrumenting-a-build-with-the-cmake-instrumentation-api-issue-759))
that compilation, not linking, is the serialization point, and that peak memory is set by a
handful of heavy TUs (`LuaScriptGlue.cpp` ~150s, `McpFieldRegistry.cpp` ~52s×2), not by `-j` width.

That measurement was a one-off requiring a manual UUID-gated query file and is not repeatable by
a future agent without re-deriving the gotcha. And even with the measurement in hand, we have no
lever to act on it: today it's a blunt choice between "throttle every TU equally via `-j`" or
"don't." There's no way to cap *just* the memory-heavy TUs while letting the rest of the tree
compile at full width.

**Describe the solution you'd like**

Adopt two features that shipped in CMake **4.3** and **4.4** (this machine was on a 4.2 floor;
upgraded to 4.4.2 as part of clearing this issue's blocker):

1. **`cmake_instrumentation()` (4.3)** — the same Instrumentation API #759 used, but through its
   now-stable native command instead of the experimental UUID-gated query file. Wire it behind
   `CMAKE_VERSION VERSION_GREATER_EQUAL 4.3` and an opt-in cache variable (default `OFF` — not
   every developer build should pay the snippet-volume cost), mirroring the validation rigor of
   the existing `OLO_LINK_JOBS` link pool at
   [CMakeLists.txt:219-239](https://github.com/drsnuggles8/OloEngineBase/blob/master/CMakeLists.txt#L219-L239).
2. **File-set `JOB_POOL_COMPILE` (4.4)** — a `SOURCES`-type file set can carry `JOB_POOL_COMPILE`,
   which overrides both target- and source-level settings, Ninja-only. Migrate the heavy TUs
   identified from a fresh instrumented trace into a file set bound to a new job pool
[...]

Issue #759, opened 2026-08-08
Build: instrument the build (CMake Instrumentation API), then revisit the -j6 parallelism cap with data


drsnuggles8:
## Problem

We cap every build at 6 jobs (`--parallel 6` / `CMAKE_BUILD_PARALLEL_LEVEL=6`, plus `OLO_LINK_JOBS=2` for the Ninja link pool) because the dev box is 16c/31GB, also hosts three CI runners for another repo, and has been OOM-killed once by a full-width build.

That cap is a **memory decision made without memory data.** We don't know:

- whether the link pool or compilation is the actual serialisation point,
- what peak RSS at `-j6` really is, and therefore whether there's headroom,
- which targets dominate wall-clock, and
- whether targets are blocked waiting on dependency edges rather than on CPU.

## Proposal

Bill Hoffman's *CMake's Recent Evolution* (C++Now 2026, Kitware) covers the **Instrumentation API**, which answers exactly these. It emits one JSON snippet per configure / generate / compile / link / install / test step carrying `timeStart`, `duration`, `role`, `language`, `outputSizes`, and — the interesting part here — `beforeCPULoadAverage` / `afterCPULoadAverage` and `beforeHostMemoryUsed` / `afterHostMemoryUsed`. Snippets collate into an index and render as a Google Trace Event file.

Work:

1. **Enable instrumentation** for one clean build of `build/` (msvc) and read the trace.
2. **Answer the two questions above with numbers** — is the link pool the long pole, and is there memory headroom at `-j6`? Record the answer in this issue so the cap stops being folklore. If the data says the cap can move, move it; if it says the cap is right, that's equally valuable and this issue closes as "measured, no change".
3. **Try `OPTIMIZE_DEPENDENCIES ON`.** The talk flags it as the fix for the "compilation blocked by dependent targets" shape, and it drops unnecessary dependency edges from *static and object libraries*. `OloEngine` is a static lib with four dependents (`OloEditor`, `OloRuntime`, `OloServer`, `OloEngine-Tests`), which is precisely the configuration it targets. Re-trace and compare.
4. Optionally check `INSTALL_PARALLEL` — likely irrelevant since we don't install, noted only so it isn't re-discovered later.

## Acceptance

- A committed trace-derived summary (in this issue, not a tracked file) giving: total configure/generate/build time, the top 5 targets by compile+link time, peak host memory at `-j6`, and whether the link pool is the constraint.
[...]

PR #1308, opened 2026-09-17
build: measure per-TU and per-link peak RSS, and publish the ranking (#1305)

### LINKED ISSUE #1305: Build: per-TU peak RSS is unmeasured — `--parallel 2` and the 14 GiB cap rest on five numbers that disagree

Every Linux build on the `olo-ci` box runs at `--parallel 2` on a 16-thread host, and each runner
unit carries a 14 GiB `memory.max`. Both numbers exist because a single `clang++` TU is believed to
need ~9 GB. **Nothing measures that.** Five numbers are in play and no two agree:

| source | per-TU figure it assumes |
|---|---|
| `asan.yml` comments | ~3 GB/TU ("4 + 4 compiles is ~24 GB peak") |
| measured on the box, 2026-09-07 | **9.0 GB** for one `clang++` TU; one hit **13.9 GB** RSS |
| same build, link step | **8.7 GB** for `ld.lld` |
| `actions-runner-*.service` `MemoryMax` | 14 GiB |
| `user-1004.slice` `MemoryMax` | 19 GiB, justified in its drop-in as "~2 x 9 = 18 G" |

**This is not a request to instrument the build — that is done.** #759 and #822 landed it, it is
still wired (`OLO_BUILD_INSTRUMENTATION=ON`, native CMake 4.3+, recipe in
[`docs/agent-rules/build-trees-and-windows-asan.md` §5](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/build-trees-and-windows-asan.md#5-instrumenting-a-build-and-a-per-file-set-compile-job-pool-issues-759-822)),
and it produces per-TU compile *time* and a host-memory sample. The gap is narrower and is called
out in our own doc: the field it records is
`args.dynamicSystemInformation.afterHostMemoryUsed`, which is **system-wide host memory in KiB, not
the command's own RSS**, sampled at that command's completion. So the existing traces cannot
attribute memory to a TU at all.

#759's conclusion — *"peak memory is set by a handful of heavy TUs, not by `-j` width"* — is almost
certainly right, but the handful was identified by **compile time** (`LuaScriptGlue.cpp` ~150 s,
`McpFieldRegistry.cpp` ~52 s x2), which is a proxy for memory, not a measurement of it. Both files
are ~100-190 lines, so whatever they cost is coming from header fan-out, template instantiation or
codegen — and time and memory do not have to rank the same way.

What the gap costs:

- **`--parallel` cannot be raised on evidence.** 2 on a 16-thread box is the largest single term in
  job service time. #1284 had to shed UBSan and TSan to hosted runners because the box sits at
  rho 1.69 — buying capacity by removing work rather than by making the work cheaper. The cheaper
  lever was never priced.
- **`OLO_HEAVY_COMPILE_JOBS` / the `olo_heavy` pool from #822 is aimed at exactly this** — throttle
  the expensive TUs separately instead of throttling every TU equally via `-j`. Choosing its file-set
  membership needs per-TU RSS; today it is populated from compile time.
- **The cgroup caps are derived from the unmeasured number.** If 9.0 GB is wrong in either
  direction, the 14 GiB unit cap and the 19 GiB slice cap are both wrong with it.
- **It drifts silently.** ~3 GB to 9.0 GB already happened, and nothing noticed until a cgroup
  started killing jobs.

Recent context, stated precisely so it is not over-claimed: on **2026-09-17** three `systemd-oomd`
kills destroyed CI jobs on this box (the visible symptom was *"The self-hosted runner lost
communication with the server"* on `Steam (stub SDK)`). The root cause was **page-cache reclaim
against the slice's `MemoryHigh`, not compiler RSS**, so this issue would not have prevented those
failures. It is relevant because the remediation deliberately left `MemoryHigh=18G` in place —
nobody can say what the builds actually need, so the safe move was to change nothing.

Measure first, act on the ranking, refactor last.

1. **Per-TU peak RSS.** clang's `-fproc-stat-report=<file>` appends one record per invocation with
   wall time and **peak RSS of that process**, at essentially zero overhead. Wire it behind the
   existing `OLO_BUILD_INSTRUMENTATION` option so it rides the switch that is already there.
   Clang-only, which is fine — the self-hosted Linux builds are clang, and that is the host whose
   caps we are trying to justify.
2. **Link steps too.** `CMAKE_CXX_LINKER_LAUNCHER` is already wired for `LinkSemaphore.cmake`, so
   there is a hook. `ld.lld` at 8.7 GB is the other half of the ceiling and `OLO_LINK_JOBS=2`
   rests on it.
3. **Publish the ranking as a CI artifact** — top ~20 TUs by peak RSS, plus the link steps. This is
   the part that stops the number drifting back; a one-off measurement decays the way #759's did.
4. **Price the cheap config levers against the ranking before touching any code.** The Linux
   sanitizer jobs are `CMAKE_BUILD_TYPE=Debug` + ASan with full `-g` and no split DWARF, which is a
   plausible explanation for both the 9 GB TUs and the 8.7 GB link:

   | lever | why it might matter here | risk |
   |---|---|---|
   | `-gsplit-dwarf` + `--gdb-index` | debug info stays in `.dwo`; the linker stops loading it | verify ccache's `.dwo` handling — do not assume |
   | `-g1` on sanitizer CI only | keeps line tables + function names, which is all ASan needs to symbolize; drops variable/type DWARF | CI only; not for local debugging |
   | ThinLTO instead of full LTO | `INTERPROCEDURAL_OPTIMIZATION` is on for Release/Dist | does not affect the Debug sanitizer jobs |

5. **Re-derive the caps and `--parallel` from the measurement** — including whether the `olo_heavy`
   file set is holding the right TUs.
6. **Only then refactor** whatever the ranking actually names (`extern template`, header fan-out,
   splitting a god-TU). Deliberately out of scope for this issue: scope creep here is how a build
   task becomes a month.

- *Extend the CMake Instrumentation API to record per-command RSS.* Correct place for it long-term,
  but it is an upstream change and blocks this behind someone else's release. `-fproc-stat-report`
  gets the same number today.
- *Wrap compiles in `/usr/bin/time -v`.* Works, but it has to compose with the ccache launcher
  chain (`CMAKE_CXX_COMPILER_LAUNCHER`), and a wrapper that silently breaks caching would cost far
  more than it measures. A compiler flag has no such interaction.
- *Raise the cgroup caps instead.* Treats the symptom, and the box has livelocked twice (2026-09-06,
  2026-09-07). The caps should follow the measurement, not lead it.
- *Buy a third runner.* Rejected in #1219 on measurement — a hardware purchase against a box whose
  cold sanitizer builds already reach the 14 GiB ceiling.
- *Just refactor `LuaScriptGlue.cpp` and `McpFieldRegistry.cpp`.* They were named by compile time.
  Without per-TU RSS there is no way to tell whether a refactor moved the memory number, which is
  the definition of an unfalsifiable optimization.

**Additional context**

- Prior art, both **completed**: #759 (instrumentation + the `-j6` question), #822 (native CMake
  4.3/4.4 instrumentation + per-file-set compile job pools). Read both before starting; this issue
  is the one field they could not record.
- Capacity context: #1084 (umbrella, per-PR critical path), #1219 (box over capacity, closed),
  #1284 (shed UBSan/TSan to hosted).
- Existing knobs this would inform: `OLO_HEAVY_COMPILE_JOBS`, `OLO_LINK_JOBS`,
  `OLO_LINK_SEMAPHORE_SLOTS`, and the `--parallel` sites capped to 2 by #1158/#1159.
- **Adjacent but a different problem, do not merge them:** `OloEngine-Tests` has been hitting the
  same 14 GiB unit cap *at runtime* — `gtest_discover_tests` gives one process per case, so that is
  a single test case using 7-14.5 GB. That is a runtime leak question, not a compile-memory one, and
  a build-width fix cannot touch it.

## Score
```olo-score
capability: 5
craft: 1
stability: 3
decay: 3
effort: 3
confidence: 0.8
learning: 3
fun: 2
kano: performance
blocked_by: []
blocks: []
```

Scoring notes: **capability 5** — no engine capability, but it is the measurement several workflows'
throughput is gated on: `--parallel`, `OLO_HEAVY_COMPILE_JOBS`, `OLO_LINK_JOBS` and both cgroup caps
are all currently set from one unverified figure, and every self-hosted job pays for it.
**craft 1** — invisible in a frame, no portfolio value. **stability 3** — mis-set caps produce real
job-killing failures, but they bite occasionally rather than corrupting anything. **decay 3** — mild
compounding: per-TU memory climbs quietly with template and codegen accretion, and the ~3 GB to
9.0 GB drift already happened unnoticed. **effort 3, confidence 0.8** — steps 1-3 are a small
self-contained change on top of harnesses that already exist; the 0.8 is on how much the config
levers actually buy, which is exactly what the measurement settles. Refactoring is explicitly
deferred so this stays a 3. **performance** — it makes existing work cheaper rather than adding
capability, and the box is the constraint.

### LINKED ISSUE #1306: Build: OLO_BUILD_INSTRUMENTATION=ON cannot build on Windows — ctest --instrument execs vendored glad's shell-builtin custom command
**Describe the bug**

`cmake --preset dev-cached -DOLO_BUILD_INSTRUMENTATION=ON` (or `clangcl`, or any Windows
Ninja tree) configures cleanly and then **fails to build**, at the first custom command:

```
[0/1879] glad-generate
Batch file failed at line 3 with errorcode 1
FAILED: [code=1] .../OloEngine/vendor/glad-build/include/KHR/khrplatform.h
                 .../glad-build/include/glad/gl.h .../glad-build/src/gl.c
                 .../glad-build/args.txt
ninja: build stopped: subcommand failed.
```

There is no diagnostic from the failing command itself — only the batch wrapper's line
number. The recipe in
[`docs/agent-rules/build-trees-and-windows-asan.md` §5a](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/build-trees-and-windows-asan.md)
is therefore currently unusable on Windows.

**Cause**

`cmake_instrumentation()` (wired in the root `CMakeLists.txt` by #822) wraps *every*
custom command in

```
ctest --instrument --command-type custom ... -- <argv>
```

and `ctest --instrument` **executes argv directly** rather than handing it to a shell.
Vendored glad's own generator rule
(`OloEngine/vendor/clang/glad-src/cmake/GladConfig.cmake`, a FetchContent download) is
built from four `COMMAND echo …` lines:

```cmake
add_custom_command(
    OUTPUT ${GLAD_FILES} ${GLAD_ARGS_PATH}
    COMMAND echo Cleaning ${GLAD_DIR}
    ...
    COMMAND echo Generating with args ${GLAD_ARGS}
    ...
    COMMAND echo ${GLAD_ARGS} > ${GLAD_ARGS_PATH}
    ...
)
```

Under `cmd.exe` `echo` is a **shell builtin with no executable**, and the last line also
needs shell **redirection**. Neither survives being exec'd.

**Reproduced in isolation** (CMake 4.4.2), which is what identifies the mechanism rather
than blaming glad or the network:

| shell | command | result |
|---|---|---|
| Git bash (POSIX) | `ctest --instrument … -- echo Cleaning` | prints `Cleaning`, **exit 0** — `echo` is a real binary at `/usr/bin/echo` |
| `cmd.exe` | same command | **exit 1, no output at all** |

That asymmetry also explains why this was never seen before: Linux CI never enables the
option, and #759/#822's numbers were taken on trees where the glad outputs were already
present and up to date, so the rule did not re-run. It re-runs as soon as the option
changes, because the option changes the command line.

**Impact**

`OLO_BUILD_INSTRUMENTATION=ON` cannot build this project on Windows at all. That is the
switch the documented per-build-profiling recipe uses, and it is the only tree where
`OLO_LINK_JOBS` and `olo_heavy` exist, so it is also the correct tree for any
link-concurrency or heavy-TU question.

**Worked around, not fixed, by #1305**

#1305 needed per-TU peak RSS on Windows and could not use the umbrella, so
`cmake/ProcStatReport.cmake` added `OLO_PROC_STAT_REPORT`, which is a plain compiler flag
with no launcher and is unaffected. `-DOLO_PROC_STAT_REPORT=ON` works on Windows today.
**The CMake Instrumentation API half — per-command timing, `postCMakeBuild` traces, the
whole of #759/#822's tooling — remains broken there.**

**Options, none of them obviously right**

1. **Patch the vendored glad rule** at FetchContent time (the repo already has a patch
   mechanism for fsr2: `cmake/fsr2-apply-patches.cmake`). `COMMAND echo X` →
   `COMMAND ${CMAKE_COMMAND} -E echo X` fixes three of the four lines, but
   `COMMAND echo ${GLAD_ARGS} > ${GLAD_ARGS_PATH}` needs restructuring — `cmake -E echo`
   cannot redirect either. Probably `cmake -E copy`/a generated file, or a tiny
   `-P` script.
2. **Generate glad's sources once and commit them**, dropping the custom command. glad is
   already fetched with `REPRODUCIBLE` (#824) so the output is deterministic; this trades
   a vendored generator for a vendored artifact.
3. **Ask upstream** (`Dav1dde/glad`) to make its rule launcher-safe. Correct long-term,
   blocks on someone else's release.
4. **Document it as a Windows limitation** and point at `OLO_PROC_STAT_REPORT` for the
   memory question. Cheapest; leaves the timing half of #759/#822 Windows-broken.

Option 1 is the smallest real fix. Worth confirming first whether any *other* vendored
custom command has the same shape — the check is a scan for `COMMAND echo` and for `>`
inside `add_custom_command` across `OloEngine/vendor/`.

**Additional context**

- Prior art: #759 and #822 (both closed) wired the instrumentation; #1305 hit this.
- `docs/agent-rules/build-memory-per-tu.md` records the reproduction and the workaround.

## Score
```olo-score
capability: 3
craft: 1
stability: 2
decay: 2
effort: 2
confidence: 0.7
learning: 2
fun: 1
kano: table-stakes
blocked_by: []
blocks: []
```

Scoring notes: **capability 3** — no engine capability, but it restores a documented
build-profiling workflow that is currently dead on the only OS that runs the editor, and
the one tree where both memory job pools exist. **craft 1** — invisible. **stability 2** —
it breaks a build loudly and only under an opt-in flag, so nothing ships wrong because of
it. **decay 2** — mild: the longer it sits, the more likely a second vendored custom
command acquires the same shape unnoticed. **effort 2, confidence 0.7** — the patch itself
is small; the 0.7 is on the `> args.txt` line, which needs a restructure rather than a
substitution, and on whether other vendored rules share the shape.
**table-stakes** — a documented switch should build.

### LINKED ISSUE #1307: Build: the olo_heavy file set is mispopulated in both directions — re-derive it from #1305's ranking

#1305 built the per-TU peak-RSS measurement the `olo_heavy` file set was always supposed to
be chosen from, and the first full ranking says the set is **mispopulated in both
directions**. Not badly, and not randomly — it tracks an older ranking, which is exactly
the drift #1305 exists to stop.

This is the follow-up #1305's step 6 deliberately deferred ("only then refactor whatever
the ranking actually names ... scope creep here is how a build task becomes a month").
#1305 measures and publishes; this one acts on it.

**The evidence**

`scripts/analyze_proc_stat.py`, 1833 compiles at 99.0% coverage, clang-cl 23.1.0, Debug,
PCH and unity off, `OloEditor` + `OloEngine-Tests`. Distribution: median 0.38 GiB,
p95 1.08 GiB, max 5.38 GiB; 17 TUs over 2 GiB, 5 over 3 GiB, 2 over 5 GiB.

In the pool but ranking low:

| TU | rank | peak RSS |
|---|---:|---:|
| `Functional/SaveGame/AbilityComponentRoundTripTest.cpp` | #278 | 0.96 GiB |
| `Rendering/RenderGraphTest.cpp` | #180 | 1.03 GiB |
| `OloEngine/Scene/Prefab.cpp` | #25 | 1.45 GiB |
| `OloEngine/Scripting/Lua/LuaScriptGlue_EngineApi.cpp` | #24 | 1.48 GiB |

Outside the pool but ranking high:

| TU | rank | peak RSS |
|---|---:|---:|
| `OloEditor: Panels/SceneHierarchyPanel.cpp` | **#5** | 3.08 GiB |
| `OloEngineContent: OloEngine/Scene/Scene.cpp` | **#6** | 2.72 GiB |
| `OloEngineContent: OloEngine/Scripting/Lua/LuaScriptGlue.cpp` | **#13** | 2.22 GiB |
| `OloEngineContent: OloEngine/Scene/SceneSerializer.cpp` | #15 | 2.07 GiB |
| `OloEngineContent: VisualScript/ComponentFieldRegistry.cpp` | #16 | 2.06 GiB |
| `OloEditor: EditorLayer.cpp` | #17 | 2.05 GiB |
| `OloEngineContent: OloEngine/Scripting/C#/ScriptGlue.cpp` | #19 | 1.98 GiB |
| `OloEngineContent: OloEngine/SaveGame/SaveGameSerializer.cpp` | #20 | 1.91 GiB |

Two of these are rules rather than data points:

- **`LuaScriptGlue.cpp` is 13th of 1833.** It was deliberately excluded when the glue was
  split, and `OloEngine/src/CMakeLists.txt` says why: "the registrations moved out of it
  when the glue was split (issue #822), and the dispatcher left behind is an ordinary small
  TU." At 2.22 GiB it is heavier than six of the nine `LuaScriptGlue_*` parts that *are*
  pooled. Whatever is left in the dispatcher is not small.
- **`Prefab.cpp` drifted, and its comment did not.** The same file carries
  `# 6,400 MB peak compiler RSS in 22 s`; it now measures **1,488 MB in 7.4 s**. The wall
  time fell with the memory, so the TU genuinely got cheaper. It is still worth pooling on
  the "cheap insurance" argument, but the justification beside it is four times off.

**What the pool gets right**, so this is a correction and not a rewrite:
`McpFieldRegistry.cpp` is #1 **and** #2 (5.38 / 5.37 GiB — compiled independently into
`OloEditor` and `OloEngine-Tests`), and eight of the nine `LuaScriptGlue_*` parts land in
the top 18. #822's split did what it set out to do.

1. **Take the numbers from the Linux `Debug + ASan` artifact, not from the table above.**
   The ranking above is Windows clang-cl Debug, non-sanitizer — the wrong configuration to
   set a Linux cgroup cap from. `.github/workflows/build-memory.yml` publishes
   `build-memory-debug-asan` weekly; use that, and confirm the *identities* agree before
   moving membership. If the two hosts disagree about which TUs are heavy, that is a
   finding in itself and this issue should stop and report it.
2. **Re-populate the file set from the ranking**, via the existing
   `olo_bind_heavy_compile_pool()` call sites (`OloEngine/src/CMakeLists.txt`,
   `OloEditor/src/CMakeLists.txt`, `OloEngine/tests/CMakeLists.txt` ×2). Pick the
   threshold explicitly and write it down — "every TU over N GiB" is a rule a future
   reader can re-apply; a hand-picked list is what decayed.
3. **Refresh or delete the committed per-TU figures** in those files' comments. A number
   in a comment with no date, compiler version or PCH state cannot be checked, which is how
   `Prefab.cpp`'s ended up 4x off. Prefer citing the artifact over restating a figure.
4. **Consider whether the pool should exist for the entries ranked #180 and #278 at all.**
   A depth-2 pool that includes a 0.96 GiB TU is throttling something that does not need
   it, which costs wall-clock for no memory benefit.

Explicitly **not** in scope here either: splitting or refactoring any of the named TUs
(`SceneHierarchyPanel.cpp`, `Scene.cpp`, `McpFieldRegistry.cpp`). That is a third issue,
and it should be justified by the same artifact.

- *Leave it.* Defensible — #822's measured verdict was that depth 2 is ~7% faster and ~4-5%
  less peak memory than depth 6, and the pool is framed as cheap insurance rather than a
  headline win. But the insurance is currently pointed at the wrong TUs, and on the
  Makefiles generator every Linux job uses the pool does not exist at all
  (build-trees-and-windows-asan.md §5e), so the mispopulation only costs on Ninja.
- *Widen the pool to "everything over 2 GiB" (17 TUs).* Simple rule, but a depth-2 pool
  over 17 members serialises much more of the build; needs the same A/B #822 ran.

**Additional context**

- Prior art: #759 (instrumentation, ranked by time), #822 (the pool, ranked by an early
  RSS survey), #1305 (the measurement and the weekly artifact).
- `docs/agent-rules/build-memory-per-tu.md` has the recipe, the four traps and this verdict.
- Related, do not merge: #1306 (`OLO_BUILD_INSTRUMENTATION=ON` cannot build on Windows).

## Score
```olo-score
capability: 3
craft: 1
stability: 2
decay: 3
effort: 2
confidence: 0.6
learning: 2
fun: 1
kano: performance
blocked_by: []
blocks: []
```

Scoring notes: **capability 3** — no engine capability; it points an existing memory
throttle at the TUs that actually need it, on the tree where the throttle exists.
**craft 1** — invisible. **stability 2** — a mispopulated pool makes an OOM marginally more
likely on a constrained host, but nothing is corrupted and the Ninja-only scope limits the
blast radius. **decay 3** — this is the second time the set has gone stale, and it will
drift again between measurements; part of the work is choosing a re-appliable threshold
rather than another hand-picked list. **effort 2, confidence 0.6** — the edit is four
one-line call sites, but the 0.6 is real: the membership must come from the Linux ASan
artifact rather than the Windows ranking above, and widening a depth-2 pool has a
wall-clock cost that needs the A/B #822 ran. **performance** — makes existing work cheaper.

### LINKED ISSUE #1313: Build: one ld.lld link is 11.17 GiB — the Linux memory ceiling is the link, and nothing bounds it there

#1305's measurement landed, and the headline is not the one anyone expected: **the single
largest memory consumer in a Linux build is one `ld.lld` link, not any translation unit —
and nothing bounds it there.**

Measured by `.github/workflows/build-memory.yml` on `olo-ci-1` (PR #1308), 1761 compiles per
cell at 99.9% coverage, compiler cache off, PCH and unity off:

| | plain `Debug` | `Debug + ASan` |
|---|---:|---:|
| median compile | 0.31 GiB | 0.32 GiB |
| p95 compile | 0.84 GiB | 0.85 GiB |
| heaviest compile | 4.65 GiB | 8.09 GiB |
| **heaviest `ld.lld` link (`OloEngine-Tests`)** | **10.03 GiB** | **11.17 GiB** |
| link ÷ heaviest compile | **2.16x** | **1.38x** |

**11.17 GiB is 80% of the `actions-runner-*.service` `MemoryMax=14G`.** The figure it
replaces was 8.7 GB, so the real number is 38% higher — and it is 10.03 GiB with no
sanitizer instrumentation at all, so this is not an ASan artifact.

**Why nothing stops it**

`OLO_LINK_JOBS` (the `olo_link` pool) and `olo_heavy` are **Ninja job pools**, and the root
`CMakeLists.txt` guards both with `if(CMAKE_GENERATOR MATCHES "Ninja")`. Every Linux job in
this repo configures with no `-G` and therefore gets **Unix Makefiles**, where both pools
silently do not exist — see
[`build-trees-and-windows-asan.md` §5e](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/agent-rules/build-trees-and-windows-asan.md).
`cmake/LinkSemaphore.cmake` does not cover it either, and this is confirmed in the run's own
log rather than inferred — the box has no PowerShell, so the wrapper is never wired and the
module fails open by design:

```
Link semaphore: neither 'pwsh' nor 'powershell' found on PATH — linking
WITHOUT cross-tree throttling.
```

So under `make -j2` nothing prevents that link running concurrently with a compile:

- link + heaviest compile = **19.26 GiB** — past the 14 GiB unit cap, at the 19 GiB
  `user-1004.slice` cap
- link + two p95 compiles = **12.9 GiB** — inside 14 GiB, but with 1.1 GiB to spare

In practice the `OloEngine-Tests` link lands at the very end of the build when little else
is left, which is almost certainly why this has not been failing constantly. That is luck of
scheduling, not a guarantee.

**Why this is worth its own issue**

It reframes two long-standing conclusions without contradicting either:

- **#759 concluded `OLO_LINK_JOBS=2` is not the serialisation point, compilation is.** That
  was a **wall-clock** finding and is still correct — links are off the critical path. For
  **memory** the opposite holds, and the two were never distinguished.
- **The cgroup caps were sized against per-TU compile memory.** At 8.09 GiB the "~9.0 GB per
  `clang++` TU" folklore turns out roughly right, so the caps are not wrong *for compiles* —
  they are simply not sized against the thing that actually peaks. This is also a better
  candidate explanation for historical OOM kills than compile memory ever was, though note
  the 2026-09-17 kills were separately diagnosed as page-cache reclaim against `MemoryHigh`
  and are **not** evidence for this.

1. **Bound link concurrency under Makefiles.** Options, cheapest first:
   - `CMAKE_JOB_POOLS` is Ninja-only, so the pool approach cannot be ported. The Makefile
     generator has no equivalent.
   - A **`CMAKE_CXX_LINKER_LAUNCHER` wrapper that takes a real semaphore** — this is what
     `LinkSemaphore.cmake` already is, and making it work on Linux (a `flock`-based shell
     wrapper instead of `pwsh`) would cover both the intra-tree and cross-slot cases. The two
     `olo-ci` slots are in ONE 19 GiB slice, so two concurrent links across slots is
     22.3 GiB — already over, and equally unguarded today.
   - Or accept it and **size the caps for it**: a 14 GiB unit cap has 2.8 GiB of headroom
     over one link, which is the number to defend rather than `--parallel`.
2. **Ask whether the link needs 11 GiB at all.** `OloEngine-Tests` is a single enormous
   executable. `-gsplit-dwarf` + `--gdb-index` is the obvious lever (debug info stops being
   loaded by the linker) and #1305 verified ccache caches and restores `.dwo` correctly.
   **Unverified before adopting:** this repo maps `DW_AT_comp_dir` to `/olo` via
   `-ffile-prefix-map`, and that is what a sanitizer symbolizer would have to resolve the
   `.dwo` through — confirm a stack still symbolises first.
3. **Do not raise `--parallel` as a substitute.** The compile side has room (non-sanitizer
   Debug fits 4 lanes), but the link is what is tight and more lanes only increases the
   chance something is running beside it.

- *Lower the caps' pressure by splitting `OloEngine-Tests`.* Real, and much larger work; the
  test binary's size is also a runtime problem already noted in #1305 (`gtest_discover_tests`
  gives one process per case, at 7-14.5 GB each). Different problem, do not merge them.
- *Leave it.* Defensible while the link stays at the end of the build — but it is unguarded,
  and the margin is 2.8 GiB on a box that has livelocked twice.

**Additional context**

- Measurement, recipe and traps: `docs/agent-rules/build-memory-per-tu.md` (PR #1308).
- Artifacts: `build-memory-debug` / `build-memory-debug-asan`, published weekly.
- Related: #1305 (the measurement), #1307 (`olo_heavy` membership — compile side only),
  #1084 (CI critical path), #796 (the Makefiles/`--parallel` generator trap).

## Score
```olo-score
capability: 4
craft: 1
stability: 3
decay: 2
effort: 3
confidence: 0.6
learning: 3
fun: 1
kano: performance
blocked_by: []
blocks: []
```

Scoring notes: **capability 4** — no engine capability, but it is the memory ceiling every
self-hosted Linux job runs against, and it is currently unguarded on the generator CI
actually uses; both cgroup caps are sized against the wrong quantity. **craft 1** —
invisible. **stability 3** — an unbounded 11 GiB step under a 14 GiB cap produces real
job-killing failures, and two concurrent links across the slice's two slots exceed it
outright; it bites occasionally rather than corrupting anything. **decay 2** — the link grows
with the test binary, so the margin shrinks quietly, but slowly. **effort 3, confidence
0.6** — a `flock` link wrapper is small, but making it compose with the existing pwsh
semaphore without a fail-closed deadlock needs care, and the `-gsplit-dwarf` arm depends on
the symbolizer question, which is genuinely open. **performance** — makes existing work
survivable rather than adding capability.

### LINKED ISSUE #1314: Build cost: PCH is disabled on clang for an MSVC reason, and 47 editor TUs compile twice — the two measured levers

#1305 built the instrument; this is what it says about where build and link cost actually
live, and which code changes would move it. Everything below is measured from the
`build-memory-debug-asan` / `build-memory-debug` artifacts (Linux, `olo-ci-1`, clang 23.1.0,
1761 compiles at 99.9% coverage, cache off, PCH and unity off) — not estimated.

### The cost structure

**Compile: 2.01 h of CPU-wall across 1761 TUs, and it is FLAT.**

| top N TUs by compile time | share of all compile time |
|---|---|
| top 1 | 1.3% |
| top 10 | 5.8% |
| top 20 | 8.8% |
| top 100 | 19.0% |

There is no hotspot. The mean TU is 4.1 s and the work is spread across ~1800 of them, which
means **per-TU surgery cannot move the build's wall-clock** — the lever has to be something
that applies to every TU at once, or something structural.

**Where the TUs are:**

| target | TUs | compile wall |
|---|---:|---:|
| `OloEngine-Tests` | 944 | **1.24 h (62%)** |
| `OloEngine` | 338 | 0.30 h |
| `OloEngineRenderer` | 322 | 0.25 h |
| `OloEngineContent` | 97 | 0.21 h |

Inside the test target: 897 genuine test TUs (mean **4.6 s**, vs **3.4 s** for an engine TU)
plus 47 `OloEditor` sources recompiled wholesale. `Rendering/` alone is 407 TUs and 29 min —
45% of the test TUs.

**Link: one step, and it is the memory ceiling.** `ld.lld` on `OloEngine-Tests` peaks at
**11.17 GiB** (ASan) / **10.03 GiB** (plain Debug), against a heaviest *compile* of 8.09 /
4.65 GiB. See #1313.

Four changes, ranked by measured leverage. The first two are the ones worth doing.

### 1. Re-enable PCH on the clang path — the only lever that touches all 1761 TUs

`cmake/CompilerCache.cmake` force-disables precompiled headers whenever the compiler cache
is on:

```cmake
set(OLO_ENABLE_PCH OFF CACHE BOOL "Disabled: MSVC PCH (/Fp) is non-cacheable under sccache/ccache" FORCE)
```

The justification in that reason string is **MSVC-specific**, but the setting is global — so
every Linux CI build compiles ~1800 TUs with no PCH, each re-parsing the engine's header
closure. That is where a flat 2 h of compile time comes from.

Meanwhile this repo has already established the opposite for clang.
`docs/agent-rules/compiler-cache-uncacheable-compiles.md` says plainly: ccache caches a
PCH-using compile given `CCACHE_SLOPPINESS=pch_defines,time_macros`, and **"clang needs
nothing beyond what CMake emits"** — and `.github/actions/setup-linux-build` *already sets
exactly that sloppiness on both self-hosted setups*, with a comment saying "every engine
target uses one." The CI action is configured for a PCH the CMake then turns off.

One of those two is wrong, and it is worth finding out which.

The second argument in `CompilerCache.cmake` is general rather than MSVC-specific and has to
be answered on evidence, not dismissed: *"PCH and the cache optimize opposite cases — PCH
speeds a cold compile, the cache makes a warm compile a near-instant hit."* The reason to
think it may not bind on clang: the PCH's constituent headers are inputs to the cache key
either way, so enabling PCH should not add invalidation — it should be additive. **Measure
rather than assume**; #1305's workflow is exactly the instrument, and the A/B is one
`workflow_dispatch` per arm plus a ccache-hit-rate comparison using `CCACHE_STATSLOG` (not
`ccache -s`, which is polluted by the sibling slot).

Note `olo_enable_pch()` is applied `PUBLIC` on the engine parts, so `OloEngine-Tests`
inherits `INTERFACE_PRECOMPILE_HEADERS` and the 944-TU majority benefits too.

### 2. Stop compiling 47 `OloEditor` sources twice

`OloEngine-Tests` does not link `OloEditor`; it recompiles its `.cpp` files as its own
sources. Measured on a build of both targets: **47 duplicated TUs, 231 s of redundant
compile time (3.8%)**, and the duplication is worst exactly where it hurts —
`McpFieldRegistry.cpp` is compiled twice and is the #1 memory TU in every cell (8.09 GiB
under ASan), `AutomationComponentRegistry.cpp` twice at #2.

Fix: factor the editor's testable code into an `OloEditorLib` static library. `OloEditor`
becomes a thin `main` that links it; `OloEngine-Tests` links it instead of recompiling.
Removes the redundant compiles *and* one 8 GiB peak from the build's memory profile. This is
ordinary CMake surgery, and it is the change with the best effort-to-benefit ratio here.

### 3. Attack the 11.17 GiB link (tracked in #1313)

In expected-value order:

- **`-gsplit-dwarf` + `--gdb-index`.** The evidence that the link is debug-info bound rather
  than instrumentation bound: ASan adds **+74%** to the heaviest compile but only **+11%** to
  the link. Whatever dominates that 10-11 GiB is present without sanitizers. #1305 verified
  ccache caches and restores `.dwo` correctly; the open question is whether a sanitizer stack
  still symbolises through this repo's `-ffile-prefix-map=…=/olo`.
- **`ld.lld --threads=N`.** lld parallelises across all cores by default and each thread
  carries working set; on a 16-thread box capping it is a one-line experiment with a
  potentially large effect on peak.
- **Split `OloEngine-Tests` into per-area binaries.** `Rendering/` is 45% of the test TUs on
  its own, and CI already shards test *execution* — per-area executables would align the
  build with that sharding and cut the single largest link into pieces. Largest change here,
  and it multiplies link count, so price it against the previous two first.

### 4. Do NOT expect wall-clock from splitting heavy TUs

#1307 re-points the `olo_heavy` pool, and that is worth doing **for memory**. It will not
make the build faster: the top 20 TUs are 8.8% of compile time, so even halving all of them
buys ~4%. Stating it here so the two are not conflated when #1307 is scheduled.

- *Unity/jumbo builds.* Also force-disabled alongside PCH, and here the reasoning genuinely
  is cache-general (a 16-TU jumbo object's key is busted by any of its 16 inputs). Lower
  priority than PCH and a worse fit for a warm-cache CI.
- *Raise `--parallel`.* Addressed in #1305: the compile side has room on non-sanitizer Debug,
  but the link is what is tight and more lanes only increases what runs beside it.
- *Trim header fan-out directly.* Almost certainly the real root cause of a flat 2 h across
  1800 TUs, but "reduce includes" is unbounded work without data. The right next instrument
  is **`-ftime-trace`** (clang emits a per-TU JSON of where frontend time goes), which would
  name the headers to attack — the natural complement to #1305's memory census and a much
  smaller change than acting blind.

**Additional context**

- Measurement and traps: `docs/agent-rules/build-memory-per-tu.md`, artifacts
  `build-memory-*` (weekly, `.github/workflows/build-memory.yml`).
- Related: #1305 (the instrument), #1313 (the link ceiling), #1307 (`olo_heavy` membership),
  #1084 (CI critical path).
- `docs/agent-rules/compiler-cache-uncacheable-compiles.md` is the prior art for item 1 and
  should be updated with whatever the A/B finds.

## Score
```olo-score
capability: 4
craft: 1
stability: 2
decay: 2
effort: 3
confidence: 0.5
learning: 3
fun: 2
kano: performance
blocked_by: []
blocks: []
```

Scoring notes: **capability 4** — no engine capability, but compile time is the dominant term
in every self-hosted job's service time, and items 1 and 2 are the only measured levers that
apply broadly rather than to a handful of TUs. **craft 1** — invisible in a frame.
**stability 2** — removing a duplicated 8 GiB compile peak helps the memory picture slightly;
nothing else here changes failure behaviour. **decay 2** — header fan-out and the test-TU
count both grow quietly, so the cost compounds mildly. **effort 3, confidence 0.5** — item 2
is bounded CMake surgery, but item 1's payoff is genuinely unknown until the A/B runs (the
warm-cache argument may well survive contact with the data), and that uncertainty is most of
the 0.5. Split item 1 out if it turns out to be a measurement task plus a one-line change.
**performance** — makes existing work cheaper; the box is the constraint.

drsnuggles8:
Closes #1305. Closes #1306. Closes #1307. Closes #1313. Closes #1314.

`--parallel 2` on a 16-thread box, a 14 GiB runner-unit cap and a 19 GiB slice cap all
rested on a belief that one `clang++` TU needs ~9 GB, and five sources disagreed about that
by 3x. #759/#822 wired the CMake Instrumentation API, but its per-command memory field is
`afterHostMemoryUsed` — **system-wide** host memory, not the command's own RSS — so the
existing traces cannot attribute memory to a TU at all. This adds the field they could not
record, and the published artifact that stops it decaying again.

## What landed

| | |
|---|---|
| `cmake/ProcStatReport.cmake` | clang's `-fproc-stat-report` on every compile, and on every link where the compiler drives it. Rides `OLO_BUILD_INSTRUMENTATION`. |
| `scripts/analyze_proc_stat.py` | Ranks compiles and links, states coverage and provenance, derives the largest `--parallel N` whose worst case fits a cap. |
| `.github/workflows/build-memory.yml` | Weekly + on-demand artifact for `Debug` and `Debug + ASan`, compiler cache off for a full census, coverage-gated. |
| `docs/agent-rules/build-memory-per-tu.md` | The recipe, four traps, the measured results, the levers priced. Indexed in the README and the failure-mode table. |

No linker launcher was needed — `clang++ … -o app` runs `ld.lld` as a driver subprocess, so
the same flag on the link line reports the linker's own peak RSS. `LinkSemaphore.cmake`'s
launcher chain is untouched.

## Acceptance criteria (#1305's numbered steps)

- [x] **1. Per-TU peak RSS behind `OLO_BUILD_INSTRUMENTATION`.** Clang-only, as intended.
- [x] **2. Link steps too.** Via the same flag rather than `CMAKE_CXX_LINKER_LAUNCHER`. Linux
      only — see the clang-cl finding below, which is stated rather than skipped.
- [x] **3. Published as a CI artifact.** Top 20 by peak RSS plus the link steps, in the job
      summary and as `build-memory-{debug,debug-asan}`, retained 90 days for diffing.
- [x] **4. Cheap config levers priced before touching code.** Table below.
- [x] **5. Caps and `--parallel` re-derived, and the `olo_heavy` membership checked.** Below.
      **No cap or `--parallel` value is changed in this PR** — that is a live runner
      configuration change and it is the user's call; the recommendation and the numbers are
      the deliverable.
- [x] **6. Stopped there.** The refactor the ranking justifies is #1307, not this PR.

## Four things measurement changed, that assumption would have got wrong

**1. The records path must stay relative, or every other job loses its compiler cache.**
ccache hashes the flag's value and `base_dir` cannot rewrite it — base_dir relativises
arguments ccache recognises as paths to *existing* files, and this one names a file that
does not exist yet at hash time. Two source-identical trees under one `base_dir`, second
expected to hit:

| command line | hit rate |
|---|---|
| no flag at all (baseline) | 1/2 — **50%** |
| `-fproc-stat-report=<absolute path>` | 0/2 — **0%** |
| `-fproc-stat-report=<relative path>` | 1/2 — **50%**, identical to baseline |

This matters beyond local worktrees: the two `olo-ci` slots share one `CCACHE_DIR` and have
different `GITHUB_WORKSPACE` paths. A `FATAL_ERROR` and a CI grep guard it, because the
failure mode is a cache that stays *alive* and never warm.

**2. The Unix Makefiles generator writes one records file per subdirectory.** A relative
path resolves against the *build tool's* working directory, and the generators differ —
confirmed by generating both and reading the emitted rule, not by assumption:

- Ninja runs every command from the top build dir → one `build/olo-proc-stat.csv`
- Unix Makefiles emits `cd <build>/<subdir> && clang++ … -c …` → one file **per subdirectory**

Every Linux job here configures with no `-G`. A Ninja-only check would have shipped a Linux
measurement that ranked a fraction of the build — and the fraction it omits is the engine
and test subdirectories, i.e. the heavy ones. The analyser takes the build *tree* and
gathers them all, re-rooting outputs so two targets cannot collide.

**3. `OLO_BUILD_INSTRUMENTATION=ON` cannot build this project on Windows at all.**
Pre-existing, filed as **#1306**. `cmake_instrumentation()` wraps every custom command in
`ctest --instrument -- <argv>`, which execs argv; vendored glad's generator rule needs a
shell for both `echo` and `>`. Dies at `glad-generate` with `Batch file failed at line 3
with errorcode 1` and no diagnostic. Reproduced in isolation: the same wrapped `echo`
succeeds under a POSIX shell and fails under `cmd.exe`, which is why Linux CI never saw it.

Fixing that means restructuring a vendored third-party custom command, so it stays in
#1306. What landed here is the split that makes the measurement reachable anyway:
`OLO_PROC_STAT_REPORT` is the memory half alone — a compiler flag, no launcher, no CMake
floor, no generator restriction — defaulting to the umbrella, so the "one switch" contract
holds. It is deliberately *not* a cache variable: that would write the umbrella's first
value into the cache and keep reading it after the umbrella went back OFF, the "set forever
after the first configure" trap `CompilerCache.cmake` documents.

**4. The flag must not go on a clang-cl link line.** `CMAKE_CXX_LINK_EXECUTABLE` only drives
the link through the compiler for a GNU-frontend clang. Under clang-cl CMake invokes
`lld-link` **directly**, so there is no driver to take the option and `/clang:…` is read as
an input file — every link failed with `could not open
'/clang:-fproc-stat-report=olo-proc-stat.csv'`. Compiles only there, and the configure line
says so: "no link records on Windows" would otherwise read as a measurement that found
linking free.

## The cheap config levers, priced

| lever | verdict |
|---|---|
| ThinLTO instead of full LTO | **Cannot help the sanitizer jobs.** `CommonProperties.cmake` sets `INTERPROCEDURAL_OPTIMIZATION_RELEASE`/`_DIST` only, and every sanitizer job is `CMAKE_BUILD_TYPE=Debug`. There is no LTO there to convert. |
| `-gsplit-dwarf` + `--gdb-index` | **ccache handles it** — the thing #1305 said to verify rather than assume. Measured on ccache 4.13.6: a cold compile produced `.o` + `.dwo`; with both deleted, a cache **hit** restored both at identical sizes. The `.o` names its companion as a bare `t.dwo` and takes the directory from `DW_AT_comp_dir`, which this repo already rewrites to `/olo`, so objects stay path-independent and cross-slot sharing is unaffected. **Unverified risk before adopting:** that same `/olo` mapping is what ASan's symbolizer must resolve the `.dwo` through. Confirm a sanitizer stack still symbolises first. |
| `-g1` on sanitizer CI only | Not yet priced against the right numbers. It should be measured on the specific TUs the ranking names rather than on the tree average, and from the Linux ASan artifact this PR's workflow produces — not from the Windows run below. |

## Step 5: the measurement — the LINK is the ceiling, not any compile

**Both Linux cells passed on `olo-ci-1`**, 1761 compiles each at **99.9% coverage**, cache
off, PCH and unity off. This is the configuration the caps were derived from, so these are
the numbers to use. Artifacts: `build-memory-debug`, `build-memory-debug-asan`.

| | plain `Debug` | `Debug + ASan` |
|---|---:|---:|
| median compile | 0.31 GiB | 0.32 GiB |
| p95 compile | 0.84 GiB | 0.85 GiB |
| compiles over 2 GiB | 10 | 12 |
| **heaviest compile** | 4.65 GiB | **8.09 GiB** |
| **heaviest `ld.lld` link** | **10.03 GiB** | **11.17 GiB** |
| link ÷ heaviest compile | **2.16x** | **1.38x** |

**One `ld.lld` link of `OloEngine-Tests` takes 11.17 GiB — 80% of the runner unit's 14 GiB
`memory.max`, and more than any translation unit in the build.** The belief it replaces was
8.7 GB, so it is **38% higher**, and it is 10.03 GiB even with no sanitizer. Meanwhile the
heaviest compile is 8.09 GiB: the "~9.0 GB for one `clang++` TU" folklore was close, and the
"one TU hit 13.9 GB RSS" claim does **not** reproduce.

This completes #759 rather than contradicting it. #759's result was about **wall-clock** —
links are off the critical path, still true. For **memory**, the link is the largest single
consumer in both configurations.

**And nothing bounds it on Linux.** `OLO_LINK_JOBS` and `olo_heavy` are Ninja job pools;
every Linux job gets Unix Makefiles (§5e), so `make -j2` is free to run that 11.17 GiB link
alongside a compile — **19.26 GiB** with the heaviest one, past the 14 GiB unit cap and at
the 19 GiB slice cap. In practice the tests link lands at the end when little else remains,
which is why this has not been failing constantly; it is unguarded, not safe by
construction. That is a plausible contributor to the historical OOM kills that per-TU
compile memory never explained.

### Recommendations — applied to nothing

Changing a live runner's configuration is out of #1305's scope and is yours to make.

- **Sanitizer jobs: keep `--parallel 2`.** Three lanes is a 15.24 GiB worst case, past the
  cap. The pin was right and now has evidence behind it.
- **Non-sanitizer Linux Debug fits 4 lanes** on compile grounds, but the same link bounds
  the benefit — raise it only if the queue needs it.
- **The lever actually worth having is a link bound that works under Makefiles.** 2.8 GiB of
  headroom over a single link is what is tight, not `--parallel`.

### `olo_heavy` — two hosts, same verdict

The Linux+ASan ranking independently reproduces the Windows census's verdict, same names and
same direction. It holds `AbilityComponentRoundTripTest.cpp` (#281), `RenderGraphTest.cpp`
(#34), `LuaScriptGlue_EngineApi.cpp` (#21), `Prefab.cpp` (#18) and `McpToolsRender.cpp`
(#17), while `Scene.cpp` (**#3**), `ComponentFieldRegistry.cpp` (**#5**),
`SceneSerializer.cpp` (#7), `LuaScriptGlue.cpp` (#9) and `SaveGameSerializer.cpp` (#11) sit
outside it. `LuaScriptGlue.cpp` was excluded as "an ordinary small TU" and ranks **9th**.
What the pool gets right: `McpFieldRegistry.cpp` is #1 in every cell and the
`LuaScriptGlue_*` parts cluster in the top 15 — #822's split worked. Re-populating it is
**#1307**, which now has two hosts agreeing on the membership.

### One correction to the issue's premise, and one to my own first reading

#1305 says the heavy set "is populated by compile *time*, which is a proxy, not a
measurement". That is no longer accurate — `OloEngine/src/CMakeLists.txt` and
`OloEditor/src/CMakeLists.txt` already carry per-TU **peak RSS** figures from an #822
follow-up.

Those figures are **up to 8.7x too high, and were wrong the day they were written** — not
decayed. The cause is a recipe bug: the survey ran
`cmake --build <dir> --target X -- '<obj>'`, which passes *both* the target and the object
to ninja and therefore builds the whole target, then credited the largest of a dozen
concurrent unrelated compiles to whichever TU was named. An isolated re-measurement on
2026-09-08 (per-PID `PeakWorkingSet64`, exactly one `clang-cl` asserted) had already caught
it. I first read the gap as drift; it is not, and this census is a third, independent
method that settles it:

| TU | committed | isolated (2026-09-08) | this census | vs isolated |
|---|---:|---:|---:|---:|
| `Scene.cpp` | 2,498 MB | 2,499 MB | 2,785 MB | 1.11x |
| `ComponentRoundTripTest.cpp` | 12,389 MB | 1,415 MB | 1,599 MB | 1.13x |
| `McpToolsRender.cpp` | 7,838 MB | 1,369 MB | 1,680 MB | 1.23x |
| `Prefab.cpp` | 6,400 MB | 1,282 MB | 1,488 MB | 1.16x |
| `RenderGraphTest.cpp` | 8,619 MB | 949 MB | 1,053 MB | 1.11x |

Different mechanism entirely — OS-sampled peak working set of an isolated compile, versus
clang's own `-fproc-stat-report` during a 6-wide parallel build — agreeing within
**1.11–1.23x**. `Scene.cpp` is the control: it is the one row that reproduced exactly under
the buggy recipe, and it lands at the same 1.11x here, so the residual is the parallel-build
environment rather than a systematic error in either method. Use this census for *ranking*
and the isolated recipe for a single TU's *absolute* figure; treat the numbers in those
CMakeLists comments as void until re-measured.

What none of this could be established *from* is the committed numbers themselves — only by
re-running them. That is why the report now carries its own provenance (timestamp, compiler
id/version, build type, generator, PCH, unity, cache, each sanitizer, LTO, both pool depths)
read from `CMakeCache.txt`: a figure that does not say how it was produced cannot be checked.

## Verification

| axis | cells | how |
|---|---|---|
| instrumentation switch | `ON` / `OFF` | Diffed the generated Ninja rules. Strip *only* the flag token from the ON tree and **23,748 of 23,750 lines are identical**; the 2 that differ are `OLO_BUILD_GIT_HASH`/`_DIRTY`/`_DESCRIBE`/`_TIMESTAMP` defines, which moved because I committed between the two configures. The switch's entire effect on the build is the flag itself, so OFF cannot cost anyone anything. |
| step kind | compile / link | 1849 compile occurrences in the generated rules, 0 on any `LINK_FLAGS` line under clang-cl (correct — see finding 4). Link records are asserted on the Linux run via `--require-link-records`. |
| compiler cache | miss / hit | Controlled 2-file A/B isolating the variable (table above), **plus a real-tree check on the 1833-TU build**: delete the objects for the 9 heaviest TUs and rebuild with the flag present — `9 direct_cache_hit`, `9 local_storage_hit`, **zero misses**. `direct_cache_hit` also confirms the flag does not break the depfile handling from #858. Measured with `CCACHE_STATSLOG` after a first attempt using before/after `ccache -s` proved unusable: a sibling worktree shares the same `CCACHE_DIR`, so the global counters moved by 2378 calls for a build that ran 28 steps. |
| host | Windows clang-cl / Linux clang | Windows artefact-backed above; **Linux is live-only and runs on this PR** (the paths filter matches the three files this PR adds). |
| toolchain | clang / MSVC | MSVC is not a cell — `-fproc-stat-report` is clang-only. The `msvc` preset must still configure cleanly with the option ON, and does: checked against a real Visual Studio 18 2026 tree, which warns, configures at exit 0, and puts the flag in zero `.vcxproj` files. |

A late catch worth naming, because it shows why the outcome assertion replaced a textual
one: the workflow originally greped `link.txt` for the flag and **would have failed the job
on a correct build**. Generating a real Unix Makefiles tree showed the generator inlining
the executable's link command into `build.make` instead — 2 `flags.make` files carrying the
flag, zero `link.txt`, and the flag plainly present on the link line inside `build.make`.
`--require-link-records` asserts the measurement instead, which cannot pass while the thing
it guards is broken.

Nothing here touches runtime behaviour, so the `{backend} × {path}` grid does not apply and
MSAA/upscale/resolution/serialization/scripting/physics/assets are all unreached.

## Cost note

This PR triggers `build-memory.yml`: two serial cold builds (`Debug`, `Debug + ASan`),
bounded at 180 minutes each, on a box at rho 1.69. That is the only way to verify the Linux
link half and the Makefiles multi-file path, and the trigger is narrow — only the three
files that make up the measurement itself — so it will not fire on unrelated CMake changes.

## Follow-ups filed

Nothing from this work is left open.

## Acting on the measurement: #1307 and #1314 are fixed here too

Folded in at the maintainer's request rather than deferred.

**#1307 — `olo_heavy` re-derived**, with a stated rule so it can be re-applied instead of
accumulating: **every TU ≥ 1.5 GiB peak RSS**. Not a round number — the 17th TU is 1.50 GiB
and the 18th is 1.37, against a tree p95 of 0.85. Added five TUs ≥2 GiB that were outside it
(`Scene.cpp` #3, `ComponentFieldRegistry.cpp` #5, `SceneSerializer.cpp` #7,
`LuaScriptGlue.cpp` #9, `SaveGameSerializer.cpp` #11); dropped three below it (`Prefab.cpp`
#18, `LuaScriptGlue_EngineApi.cpp` #21, `RenderGraphTest.cpp` #34). Depth stays 2 — two
concurrent is 11.94 GiB, and the pooled set totals 587 s so a 2-wide floor is 294 s against a
-j6 ideal of 1206 s, i.e. it cannot become the critical path.

**Correction to an earlier claim in this PR:** I reported the pool as holding TUs ranked #180
and #281. That was a suffix-matching bug — `endswith("ComponentRoundTripTest.cpp")` also
matches `AbilityComponentRoundTripTest.cpp`, which is **not** in the pool and never was.
Re-derived with exact matching the pool held ranks #1–#34: mostly right, missing five heavies
and carrying three light ones. The fix is real but smaller than I first described.

**#1313 — the link is bounded on Linux now, and it was not before.** The measurement's
headline was that the ceiling is not a compile: one `ld.lld` link peaks at **11.17 GiB**
(10.03 without sanitizers) against a heaviest compile of 8.09 / 4.65. Three things that
should have bounded it did not — `OLO_LINK_JOBS` and `olo_heavy` are Ninja pools and every
Linux job gets Unix Makefiles, and `LinkSemaphore.cmake` shelled out to `pwsh`, which the
box does not have (confirmed in the runner's own log, not inferred). So `make -j2` was free
to run that link beside a compile: 19.26 GiB with the heaviest, past the 14 GiB unit cap and
at the 19 GiB slice cap.

`scripts/link-semaphore.py` is the POSIX half of that throttle — N permits via `flock`, where
the kernel releases the lock when the holder dies, so a killed linker cannot leak a permit
(the Windows half needs N named mutexes to get the same property). It fails open throughout.

It also fixes a knob that did nothing: `OLO_LINK_SEMAPHORE_SLOTS` is a cache variable but
both wrappers read it from the *environment*, and nothing put it there — so the documented
value only changed a status message while the built-in default of 2 was the real setting on
**both** platforms. It now travels via `cmake -E env`.

**Split DWARF, measured rather than assumed.** #1313 recorded "confirm a sanitizer stack
still symbolises" as an unverified risk. It is verified now, with a control on each variable
(llvm-symbolizer, clang 23.1.0, cross-compiled to x86_64-linux):

| arm | inline frames |
|---|---|
| `-gsplit-dwarf` alone, `.dwo` present | resolve |
| `-ffile-prefix-map=/olo` alone | resolve |
| **both together** | **lost** |

It is the *combination* that breaks, and this repo is in it — the `/olo` map exists so the two
runner slots can share a ccache, and split DWARF resolves its side-car through
`DW_AT_comp_dir`. Function names and file:line survive in every arm, and the sanitizer jobs
are Debug builds with little inlining to lose. `OLO_SPLIT_DWARF` therefore ships **default
OFF**, and `--gdb-index` is deliberately *not* paired with it: lld builds that index by
reading the debug info the flag exists to skip, so it works against the memory goal.

**#1306 — the instrumentation API can build on Windows again.** It could not, at all: the
API executes every custom command instead of shelling out, and vendored glad's generator
rule needed a shell for both `echo` and `>`, so the build died at `glad-generate` with
`Batch file failed at line 3 with errorcode 1` and no diagnostic. glad is a FetchContent
download, so the fix is a patch on top of the pin using the mechanism this repo already has
for FSR2 — now shared (`cmake/apply-vendor-patches.cmake`) rather than duplicated.

Verified in three steps that isolate each claim: the exact batch file that failed exits 0
and writes all four outputs; deleting `glad-build` and rebuilding through ninja regenerates
it with the wrapper in place; and the API **actually emits a trace on Windows now** — 20
events, roles `{custom: 15, compile: 2, link: 2, cmakeBuild: 1}`, where those 15 custom
commands are the ones that used to kill the build. `REPRODUCIBLE` (#824) still holds.

`OLO_PROC_STAT_REPORT` stays, but for a better reason than before: it is the narrower
switch. A compiler flag costs nothing per build step; the API spends a process on every
compile, link and custom command.

**#1314 — the two measured levers.**

*PCH was disabled on clang for an MSVC reason.* `CompilerCache.cmake` force-disabled it
unconditionally while its own reason string said "MSVC PCH (/Fp) is non-cacheable" — yet
`compiler-cache-uncacheable-compiles.md` already records that clang PCH compiles cache fine
with the right sloppiness, and `setup-linux-build` **already exports exactly that sloppiness**.
The CI action was configured for a PCH the CMake turned off, so ~1800 Linux TUs re-parsed the
header closure every cold build. The guard now tracks its reason, and **checks** the sloppiness
rather than assuming it. Honest caveat: the payoff is a prediction, not a result — the
"PCH and cache optimise opposite cases" argument is general, and this is one conditional to
flip back if a warm-cache regression appears.

*41 editor sources compiled twice.* `OloEngine-Tests` cannot link an executable, so it
recompiled them — 231 s, 3.8% of compile time, and it paid for the tree's #1 memory TU twice.
`OloEditorCore` is a static library holding exactly the shared set (Automation/ 18, MCP/ 21,
2 ContentBrowser helpers); the editor keeps its 34 ImGui panels. The seam is the one the tests
already drew, so it is discovered rather than invented, and nothing there self-registers —
every command goes through an explicit `Register*Commands()`, so the static-library
symbol-dropping trap does not apply. Checked before moving anything.

Verified: `McpFieldRegistry.cpp` went from 3 on-disk objects to **one live build edge**, total
objects 1849 → 1809, and **1118 Automation/MCP/ContentBrowser tests: 1116 passed, 0 failed**,
2 skipped for environment.

### A note on this workflow's cost, for whoever merges

GitHub evaluates a `pull_request` `paths` filter against the **whole PR diff**, not the last
push — so because this PR touches the three measurement files, *every* push to it re-runs
both cold builds. That is correct behaviour for a PR that changes the measurement, but it
means the trigger is more expensive than it looks while such a PR is in flight. If that
proves annoying in practice, the cheap fix is to run only the `debug-asan` cell on
`pull_request` and keep both on the weekly schedule. I have deliberately stopped pushing to
this branch for that reason.

## Summary by CodeRabbit

- **New Features**
  - Added build-memory measurement and reporting, including per-translation-unit peak usage, coverage, rankings, and recommended parallelism.
  - Added optional split DWARF support to reduce linker memory usage.
  - Added cross-platform link concurrency throttling.
  - Added a shared editor core library to streamline editor and test builds.

- **Bug Fixes**
  - Improved instrumentation compatibility for Windows builds.
  - Compiler caching now preserves precompiled headers when supported.

- **Documentation**
  - Added guidance on build-memory measurement, cache behavior, parallelism, and memory limits.


drsnuggles8:
All ten findings were valid and all ten are fixed in `db563564a`. Three were real defects
rather than polish, so recording what each actually broke:

**PCH outputs classified as links** — the sharpest catch. `.pch`/`.gch`/`.pcm` have no object
suffix and fell through to the `link` fallback, which would have let `--require-link-records`
(the CI gate added earlier in this PR) pass on a build where **no linker ever ran**.
`OLO_ENABLE_PCH` defaults ON whenever the compiler cache is off, so that is the default local
configuration. Regression-tested with a 3-compile + 1-`.gch` fixture: link count 0, gate
correctly fails.

**Flag validated on C++ but applied to C** — correct, and it was also *under*-instrumenting.
Each language is now gated on its own compiler id with `COMPILE_LANGUAGE`/`LINK_LANGUAGE`
generator expressions. On the real tree the configure line changed from `CXX compiles` to
`C/CXX compiles`, so the C TUs (vendored glad, lua, bc7enc) were never being measured.

**Derivation ran past the compile count** — it repeated the last total and could claim 16 safe
lanes from 3 measured TUs. Capped at `min(max_lanes, len(compiles))`.

The `OLO_PROC_STAT_FILE` path-component case is closed at the source rather than half-supported:
CMake now rejects any value containing a separator, since a basename is the entire contract the
analyser's tree-walk can honour. The `--name` suggestion was applied as written.

On the workflow path filter: `CMakeLists.txt` added as suggested — a wiring regression there
would make this workflow vacuous rather than red, which is the failure mode worth guarding. To
keep that affordable on a box at rho 1.69, a pull request now runs only the `debug-asan` cell
via a matrix `if`; both cells still run weekly. Net effect is a wider trigger for half the
per-PR cost.

Two doc points taken verbatim ("a hit produces no compiler RSS record", `CCACHE_STATSLOG` is
build-scoped and appends) and the spot-check count now actually lists five.

Issue #1301, opened 2026-09-16
Testing: host the Vulkan suite under a software driver, as cross-vendor.yml already does for GL


drsnuggles8:
## Problem

`VulkanPassSuite` contains 58 tests that run the render graph on a real Vulkan device and assert on read-back pixels. **They execute in no automated job.** Hardware Vulkan is unavailable everywhere CI runs: hosted runners have no GPU, and the self-hosted AMD box cannot host the suite (RADV has no descriptor heap). So the tests only ever run when a developer on a Vulkan-capable machine chooses to run them.

The cost is late detection rather than no detection — but "late" here has meant months. `VulkanPassSuite.FoliageInstancePullDrawsThreeTintedCards` held a stale brightness assertion from #1234 until someone ran the suite by hand during unrelated work.

## The opportunity

The repository already solves the equivalent problem for OpenGL. `.github/workflows/cross-vendor.yml` downloads Mesa3D, drops the software `opengl32.dll` next to the test binary, and runs the whole suite under **llvmpipe** on a hosted Windows runner — roughly 25 minutes, on a nightly schedule plus manual dispatch, with goldens scoped via `--olo-golden-vendor=llvmpipe`.

**The Mesa archive that job already downloads also ships lavapipe**, Mesa's software Vulkan ICD. So hosting the Vulkan suite may be as small as copying two more files out of an archive already in the pipeline and pointing `VK_ICD_FILENAMES` at the ICD manifest.

## Known risk, to settle before committing to the approach

The engine requires Vulkan 1.4 (`OloEngine/src/Platform/Vulkan/VulkanCapabilities.h:51`, `kMinApiVersion = VK_API_VERSION_1_4`). The pinned Mesa build is 24.3.4, whose lavapipe advertises Vulkan 1.3; 1.4 support landed in Mesa 25.x. So this needs at minimum a pin bump, and then lavapipe still has to satisfy whatever else the engine demands at instance and device creation.

**This is unverified.** It is a prototype question, not an analysis question — stand the job up against a current Mesa and see what `vkCreateInstance` / device selection says. A negative result is a perfectly good outcome for this issue, provided it is recorded with the evidence so nobody re-derives it later.

## Acceptance criteria

- [ ] Either `VulkanPassSuite` executes in an automated job, or the issue is closed with concrete evidence of why a software Vulkan driver cannot host it (which call fails, on which Mesa version, with what message).
[...]

Issue #1294, opened 2026-09-15
CI: the L7 Vulkan ray-query suites never run, so a red ReSTIR PT arm reaches master green


drsnuggles8:
**Is your feature request related to a problem? Please describe.**

The L7 Vulkan ray-query suites — `ReSTIRPTDevice`, and every other test gated on
`ProbeVulkanDeviceTestGate` + `IsRayQueryEnabled()` — never run on any runner. The self-hosted AMD
box has no usable Vulkan RT path (RADV has no descriptor heap), the hosted runners have no GPU, and
the tests `GTEST_SKIP` cleanly when the device is absent. So the PR gate is green whether these
tests pass or not, and the only thing that ever runs them is a human on a local RTX box.

#1288 is what that costs. The G-Buffer flags-lane layout moved in 327cc38fa (#1231): the PBR closure
model went from `model * 2` to `model * 64` to make room for a material kind and a skin-profile
slot. `ReSTIRPTDeviceTest.cpp` hand-encodes that lane for its synthetic G-Buffer and still wrote the
literal `2.0`, which had meant *ClosureV2, lit* and now means *Legacy*. The primary receiver was
shaded with the Legacy closure while the CPU oracle and every secondary vertex used ClosureV2, and
Legacy's `distributionGGX` clamps its denominator where its sampling density does not — so the
near-specular metal arm returned **1.45%** of the reference integral. It shipped to master and sat
there for a day. Two arms of two tests were red and no gate noticed.

The failure is loud, deterministic and byte-identical across runs. It is exactly the class of defect
CI exists to catch, and the only reason it escaped is that nothing runs the suite.

**Describe the solution you'd like**

A runner that can execute the L7 ray-query suites, and a gate that fails when they do not run at
all rather than when they fail. Roughly, in order of how much they buy:

1. **Make "skipped" visible.** Today a skip and a pass are the same colour. A job that asserts a
   minimum executed-case count for the RT-gated suites — or reports the skip reason in the summary —
   turns "this never ran" into something readable without any new hardware.
2. **A runner with a usable Vulkan RT device.** `radv-lacks-descriptor-heap` rules out the existing
   AMD box. This is a hardware or hosted-GPU-runner question and is where the real cost is.
3. **Failing that, a scheduled (not per-PR) run on the developer box**, so drift surfaces within a
   day rather than whenever someone next runs the suite by hand.

**Describe alternatives you've considered**

[...]

Issue #1234, opened 2026-09-13
Foliage: two-sided leaf transmission and mapped surface shading across render paths


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1224

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Replace basic forward foliage lighting and hardcoded deferred surface attributes with a dedicated vegetation material demonstrated on a backlit clump and shrub.

## Acceptance criteria

- [ ] Leaf normals, roughness and thickness/transmission maps are authorable and survive import/cook/save/load.
- [ ] Both sides respond correctly to lights, shadows and environment lighting; leaf transmission is not an unshadowed ambient constant.
- [ ] Forward/forward+/deferred preserve the same material meaning and normal orientation; share evaluation where appropriate.
- [ ] Mapped detail and alpha coverage remain stable under motion and distance; compare front/side/backlight and inspect diffuse/specular/transmission.

## Scope boundaries

Reuse the material-kind plumbing from skin, not skin diffusion. Leaves need a distinct thin-surface model.

## Blocked by

- #1233
- #1231

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 8
craft: 13
stability: 3
decay: 1
effort: 3
confidence: 0.8
learning: 8
fun: 13
kano: performance
blocked_by: [1233]
blocks: [1224, 1237, 1255]
```
[...]

PR #1309, opened 2026-09-17
test(vulkan): pass-level Vulkan coverage as the convention, and a run that says whether it got any #1300

### LINKED ISSUE #1300: Testing: Vulkan backend coverage is manual, per-feature, and evaporates after the PR
## Problem

The engine ships two backends behind an RHI, but only one of them is covered by tests that anyone runs.

OpenGL has scene-level automated coverage: the `RendererAttachedTest` family loads a real `.olo` scene, runs frames through the actual pipeline, reads pixels back and compares them. Vulkan has `VulkanPassSuite` — 58 pass-level tests on a real device — which nothing routinely executes, and no scene-level equivalent at all.

The practical consequence is that **every rendering feature re-does the same manual Vulkan ritual**: launch the editor with `--rhi vulkan`, switch forward / forward+ / deferred, look at the frame, scan `OloEngine.log` for errors and VUIDs, screenshot for the PR. It is slow, it is not repeatable, it depends on the author remembering to do it, and it guards nothing once the PR merges.

## Why this is worth fixing

Two concrete cases, both from recent work:

- **A Vulkan-only silent failure that no CPU test could see.** In #1246 the strand pass rendered nothing on Vulkan: 0 errors, 0 validation warnings, the draw recorded, the buffer contents correct — a missing `abs()` on `projection[1][1]`, which is negative under Vulkan's clip convention. Finding it took a two-run live shader bisect. A test asserting *"the coat covers a nonzero pixel count, and an occluder reduces it"* fails on that in seconds.
- **A stale assertion nobody saw for months.** `VulkanPassSuite.FoliageInstancePullDrawsThreeTintedCards` kept asserting a brightness that stopped being correct when #1234 replaced the foliage flat ambient with IBL. It was caught only because someone ran the suite by hand while working on something else.

The second case is the important one: the coverage *existed* and still reported late, because there is no habit or mechanism that runs it.

## What to decide and build

This is a testing-configuration question, not a prescription. Options to weigh:

1. **Make the scene-level fixture family backend-selectable**, so a visual-evidence test can run under either RHI rather than booting GL unconditionally. Highest value, largest job — `RendererAttachedTest` is wired to GL end to end.
2. **Establish a convention** that a rendering feature adds a `VulkanPassSuite` case for its pass, the way it already adds an L1 contract test. Cheaper, and the suite's existing 58 tests show the pattern works.
3. **Some mix**: pass-level Vulkan coverage as the default expectation, scene-level for features where composition against other geometry is the point.

Whatever is chosen, one thing needs settling regardless: **a skipped suite currently looks like a passing one.** `VulkanPassSuite` skips cleanly with no device, no Vulkan loader, or a driver below the required version. That is correct behaviour, but it means "green" carries no information unless the run says which of the two happened.

## Acceptance criteria

- [ ] A rendering feature can be verified on Vulkan by running a test, not by driving the editor by hand.
- [ ] Backend-parity verification is repeatable by a later contributor who did not write the feature.
- [ ] A run that skipped for want of a device is distinguishable from a run that executed and passed, in the output a developer actually reads.
- [ ] The chosen convention is written down in `docs/agent-rules/testing-architecture.md`, including which kind of Vulkan coverage a new rendering feature is expected to add.

## Scope boundaries

This is about making Vulkan coverage *possible and habitual on a developer machine*. Getting it to run automatically is a separate concern — see #1301. The two are independent: this one pays off on a dev box whether or not CI can ever run Vulkan, and should not be blocked on it.

## Score

```olo-score
capability: 8
craft: 1
stability: 5
decay: 5
effort: 3
confidence: 0.8
learning: 3
fun: 3
kano: table-stakes
blocked_by: []
blocks: [1301]
```

drsnuggles8:
## The decision, first

#1300 asked for a choice between three shapes and did not prescribe one. **The answer is option 3, the mix, and the line falls where §9 of `testing-architecture.md` already put it** — this PR writes down the half that was missing.

- **Option 1 (backend-selectable scene fixtures) is refused**, on §9's numbers: 118 test files inherit `RendererAttachedTest`, and its Vulkan instantiation would `SKIP` for nearly all of them because `Scene::OnUpdateRuntime` is unreachable once the backend is displaced. Parameterising does not remove that cost; it multiplies it by 118 and hides it behind a green instantiation that ran nothing.
- **Option 2 (pass-level convention) is where the bugs actually are.** #1106 ×2, #994's `.length()` and #1246's missing `abs()` are all single-pass, single-shader faults reachable by driving one pass with production inputs.
- **So: pass-level `VulkanPassSuite` tenancy is the expectation for a new rendering feature; scene-level composition stays a GL artefact plus a live Vulkan capture in the PR body.**

The reasoning, the cost of a new tenant, and what one owes (cross the production seam, anchor analytically, pin the seam's per-backend answers outright) are in **`docs/agent-rules/testing-architecture.md` §10**, linked from both README indexes.

## What was actually holding this up

A skipped Vulkan suite reports exactly like a passing one. `VulkanPassSuite` skips cleanly with no device, no loader, or a driver below the ADR 0010 contract — correct behaviour, and why the binary is green on a hosted runner — but gtest reports that identically to a run that exercised the backend and found it correct. The skips scroll past hundreds of lines above a `[  PASSED  ]` nobody reads them against. That is how `FoliageInstancePullDrawsThreeTintedCards` kept asserting a brightness that #1234 invalidated, for months.

**One gate, and it reports.** `OLO_VULKAN_DEVICE_OR_SKIP()` (`VulkanTestSupport.h`) replaces eleven hand-rolled `ProbeVulkanDeviceTestGate` + `GTEST_SKIP` pairs *and* the inline probe ladder `VulkanPassSuite` still carried — the ladder that made it the last Vulkan suite not honouring `--olo-gl-backend=none` (#1107). It marks the running test as device-gated, so the tally cannot drift from the call sites.

**A verdict below gtest's own summary**, as the last thing on screen:

```
=============================================================================
Vulkan backend coverage: EXERCISED - 57/57 device-gated tests ran.
  device: NVIDIA GeForce RTX 4090
=============================================================================
```
```
=============================================================================
Vulkan backend coverage: NOT EXERCISED - none of the 2 device-gated tests ran.
  reason: No GPU / GL 4.5+ context available in this environment (GL backend
  pinned to 'none' by --olo-gl-backend=none; the Vulkan gate honours it too).
  This run says NOTHING about the Vulkan backend. Green here is green for
  OpenGL only. Pass --olo-require-vulkan to make it a failure instead.
  See docs/agent-rules/testing-architecture.md section 10.
=============================================================================
```

A third verdict, `PARTIAL`, covers "a device is present and some tests still skipped" and names the first skip past the gate — telling that developer "no device" would send them hunting for a driver they already have. Nothing prints when no device-gated test was selected, so an ordinary `--gtest_filter` run stays quiet.

**`--olo-require-vulkan`**, the twin of `--olo-require-gpu`: a gate refusal becomes a FAILURE, a run where nothing executed exits 3 (a real test failure outranks it), and the flag is rejected at parse time together with `--olo-gl-backend=none`.

## Proving it catches a real defect

`VulkanPassSuite.GroomStrandCoatCoversPixelsUnderTheVulkanClipConvention` drives the real `GroomRenderPass` through the real render graph, with the camera UBO uploaded through `RHI::AdjustProjectionForBackend`, and asserts the coat covers a nonzero pixel count **and** that an opaque depth prefill takes those pixels away. The groom pass had no Vulkan coverage at all before this.

With #1246's defect reintroduced locally (the `abs()` removed from `oloGroomPixelsPerUnitAtUnitW`, **not committed**):

| | with the defect | clean |
|---|---|---|
| the new Vulkan tenant | **FAILS in 4 s** — `covered NO pixels while the pass reported 1792 segments drawn` | passes |
| all 4 `GroomStrandVisualEvidenceTest` GL tests | **all pass** | all pass |

That second row is the point of the whole issue: the defect is invisible to the entire OpenGL scene-level coverage, on every rendering path, and finding it originally took a two-run live shader bisect.

Per §9 the tenant also asserts the seam's answers outright (`rasterProjection[1][1] < 0`, `WindowSpaceFrontFaceSign() == -1`) before looking at a single pixel — otherwise a regression in `AdjustProjectionForBackend` would make this test green for the wrong reason.

## Acceptance criteria

- [x] **A rendering feature can be verified on Vulkan by running a test** — the groom tenant, 0.5 s, no editor.
- [x] **Repeatable by a later contributor** — `--gtest_filter=VulkanPassSuite.* --olo-require-vulkan`; the convention and the cost of a tenant are in §10.
- [x] **A skipped run is distinguishable from a passing one, in the output a developer reads** — the banner, plus a non-zero exit under `--olo-require-vulkan`.
- [x] **The convention is written down in `testing-architecture.md`** — §10, including which kind of Vulkan coverage a new rendering feature is expected to add.

## Verification

| axis | cell | result |
|---|---|---|
| backend | OpenGL | full suite: 8845 tests, 8798 passed (see below) |
| backend | Vulkan (RTX 4090) | 221 tests across 24 device suites: 220 passed, 1 skipped (opt-in `--olo-vegetation-experiment`); banner `PARTIAL - 121 of 122`, naming that skip and its reason |
| device availability | present | `EXERCISED - 57/57`, device named |
| device availability | absent (`--olo-gl-backend=none`) | `NOT EXERCISED`, gate reason carried, exit 0 |
| device availability | absent (no ICD, `VK_DRIVER_FILES` to nowhere) | `NOT EXERCISED - vkCreateInstance failed (no Vulkan 1.4-capable ICD available)` |
| `--olo-require-vulkan` | gate refuses | test FAILS with the flag named, process exits 3 |
| `--olo-require-vulkan` | + `--olo-gl-backend=none` | rejected at parse time, exit 2 |
| existing-suite regression | full `OloEngine-Tests` | 8798 passed / 1 failed / rest skipped, 29 min |

**The one full-suite failure is not this change — now measured, not argued.** `FoliageWindEvidenceTest.EveryRasterPathShowsWindAndMotion` failed in the full run and passes in isolation. Investigated and filed as **#1315**, with a 30-second two-suite reproducer:

```powershell
OloEngine-Tests.exe --gtest_filter="TerrainCameraRelativeVisualEvidenceTest.*:FoliageWindEvidenceTest.EveryRasterPathShowsWindAndMotion"
```

It is a deterministic cross-test state leak, not a flake, and the symptom is far worse than the SSIM number suggested: the foliage is **entirely absent** (0% subject coverage) while the terrain, lighting and splat pattern are unchanged. Two independent triggers, both tessellated-terrain evidence tests.

It is not reachable from this diff, and that is now provenance rather than inference: the `FoliageWind_*` goldens were last regenerated by `6b7c97132` (PR #1303), an **ancestor of this branch's base** `3b2e23aeb`, and the reproducing filter contains no test this PR touches. #1315 records what was ruled out by measurement (wind-phase accumulation, render origin, renderer-configuration leaks, the terrain GPU-LOD global) so the next person does not re-run those.

The third skip reason the gate can produce (driver below the required version) could not be forced on this box. It is one `RecordGateRefusal(reason)` call carrying whatever text the gate produced — the same code path as the two reasons that *were* forced live — and all four reason texts are covered device-free by `VulkanCoverageReport.EveryGateRefusalReasonSurvivesIntoTheBanner`.

**Render path selection:** the mechanism is path-agnostic because a `VulkanPassSuite` tenant hand-drives its pass rather than going through `RenderPipeline`; the groom tenant therefore selects none of Forward / Forward+ / Deferred, which matches production — `GroomRenderPass` is registered on every path and composes through one forward-style pass on all three (see its header). Path selection on Vulkan remains an editor-level question.

**CI does not run any of this, and that is a fact about the hardware.** The AMD self-hosted runner cannot run Vulkan at all (RADV has no descriptor heap) and the hosted runners have no GPU. Vulkan execution is a developer-box capability here; making CI able to run it is #1301 and is deliberately out of scope. The banner is what keeps that honest in the meantime — a CI log that prints `NOT EXERCISED` is telling the truth about itself.

**Overlap with #1294** is recorded rather than duplicated: `VulkanCoverage::FormatBanner` is a pure function over a `Tally`, tested device-free, and #1294's "make skipped visible" for the L7 ray-query gap should extend it rather than grow a second reporter. Said so in §10 too.

## Review guide

**Where I'd look hardest**

1. `OloEngine/tests/Rendering/VulkanCoverageReport.cpp:60-80` — the listener's executed/not-exercised classification. It is the one piece that is *not* covered by the device-free tests (they exercise `FormatBanner`, which is pure), and it already had one bug: a gate refusal turned into a `FAIL()` by `--olo-require-vulkan` is not `Skipped()`, so the first version counted it as EXERCISED. Now keyed on a per-test admission flag. If that flag is ever set without a matching `OnTestStart` reset, the tally lies in the reassuring direction.
2. `OloEngine/tests/Rendering/VulkanPassSuiteTest.cpp:494` — replacing the fixture's inline probe ladder with the shared gate changes its volk handling (`volkFinalize()` + re-init on the way out, which the ladder did not do) for all 57 tenants. All 57 pass, but this is the change with the widest blast radius in the diff.
3. `OloEngine/tests/Rendering/VulkanPassSuiteTest.cpp:7860` — the groom tenant's coat is at **editor scale, not anatomical scale** (`rootDiameter` 0.05 rather than a human hair's 70 µm), for the reason `GroomStrandVisualEvidenceTest` gives for its own groom: at real scale the OpaqueRibbon tier's compensating alpha sits under the 0.5 cutoff and the coat is legitimately empty. The defect being pinned is a sign, not a magnitude, so this does not weaken it — but it is a judgement call and the comment says so.

**What I verified, and how** — the defect A/B above (`abs()` removed: new tenant fails in 4 s, all 4 GL groom evidence tests pass); all three banner verdicts live, two of them with a genuinely refused gate; `--olo-require-vulkan` in both directions plus the parse-time contradiction; 221 Vulkan device tests; a full 8845-test suite run. Evidence PNGs reverted before staging.

**Least confident about** — the groom tenant's editor-scale coat (point 3 above). It is the one place I chose a number to make a test produce signal, and although the defect it pins is a sign rather than a magnitude, a reviewer who knows the strand pass should sanity-check that widening the ribbons did not also widen what the test would accept. (The full-suite foliage failure that used to sit here is resolved as a question: it is #1315, pre-existing, with provenance.)

**Deliberately not tested** — the driver-below-required-version gate reason (cannot be forced on an RTX 4090; same code path as the two that were); CI execution of any Vulkan test (impossible on this repo's runners — #1301); and the listener's classification logic device-free, because driving gtest's listener callbacks from inside a gtest test is worse than the coverage it would buy.

Closes #1300

## Summary by CodeRabbit

- **New Features**
  - Added end-of-run Vulkan coverage reporting with **EXERCISED**, **PARTIAL**, or **NOT EXERCISED** results.
  - Added `--olo-require-vulkan` to fail test runs that do not execute Vulkan tests.
  - Added Vulkan pass-level coverage for groom strand rendering and depth occlusion.

- **Bug Fixes**
  - Vulkan test skips now consistently report device availability and skip reasons.

- **Documentation**
  - Documented Vulkan coverage requirements, reporting behavior, and testing guidance.


drsnuggles8:
@coderabbitai review

PR #1320, opened 2026-09-19
ci(vulkan): host the Vulkan suites on Mesa lavapipe, nightly (#1301)

### LINKED ISSUE #1301: Testing: host the Vulkan suite under a software driver, as cross-vendor.yml already does for GL

drsnuggles8:
`VulkanPassSuite`'s 58 tests — and the 105 device-gated Vulkan tests around them — executed in **no automated job**. Hosted runners have no GPU; the self-hosted AMD box's RADV has no descriptor heap. They ran only when a developer chose to run them, which is how `FoliageInstancePullDrawsThreeTintedCards` kept a stale assertion from #1234 for months.

This adds `.github/workflows/vulkan-software.yml`: the Vulkan twin of `cross-vendor.yml`, hosting those suites on Mesa's software Vulkan ICD (lavapipe), nightly plus dispatch.

## The feasibility question, settled by measurement

The issue named Vulkan 1.4 as the known blocker. It is not the binding one — **ADR 0010's extension contract is**:

| Mesa | `apiVersion` | `VK_EXT_descriptor_heap` | `VK_KHR_shader_untyped_pointers` | `VK_KHR_device_address_commands` | contract |
|---|---|---|---|---|---|
| 24.3.4 (the GL job's pin) | 1.3.296 | no | no | no | no |
| 26.1.8 | **1.4.354** | **no** | **no** | **no** | **no** |
| 26.2.0 | 1.4.354 | yes | yes | yes | **yes** |

26.1.8 is the interesting row: it clears the 1.4 bar everyone checks and still fails on all three. So `MESA_VERSION: '26.2.0'` is a floor, not a preference, and the workflow says so.

Probed with SDK 1.4.357.0, `VK_DRIVER_FILES` naming one ICD so nothing else could answer.

## Acceptance criteria

- [x] **`VulkanPassSuite` executes in an automated job.** 206 cases, **104/104 device-gated `EXERCISED`**, exit 0 under lavapipe 26.2.0.
- [x] **Wall clock measured and stated.** Debug, developer box, whole Vulkan filter: **118 s under lavapipe vs 190 s on an RTX 4090**. Software Vulkan is the *faster* arm — these suites are 128×128 targets and one-group dispatches, so the run is launch-bound, not fill-bound. The ~25 min llvmpipe GL figure is not the right comparison, and budgeting against it would be wrong by an order of magnitude. The 90 min cap is for a hang, not an expectation.
- [x] **"Ran and passed" is distinguishable from "skipped — no usable device".** `--olo-require-vulkan` (#1300) does it, and it is what makes this job worth having rather than merely harmless. Demonstrated against **Mesa 24.3.4's lavapipe — a real driver below the contract, not a broken path**:

  ```text
  WITH --olo-require-vulkan     exit 1   [  FAILED  ] 1   Vulkan backend coverage: NOT EXERCISED
  WITHOUT it                    exit 0   [  SKIPPED ] 1   Vulkan backend coverage: NOT EXERCISED
  ```

  Same driver, same filter, same banner. Only the exit code differs — and the exit code is the only part CI reads.
- [x] **Golden scoping decided as the GL job does it.** `--olo-golden-vendor=lavapipe`, plus `--olo-perf-machine=ci-lavapipe-windows` so perf rows can never clobber a hardware baseline. Worth recording: the FXAA golden compare **passed against the NVIDIA-baselined PNG** under lavapipe, so no `golden/lavapipe/` set is needed yet. One caveat for whoever needs it later — `VulkanPassSuiteTest.cpp`'s golden load is a hardcoded path that does *not* route through `--olo-golden-vendor`. It costs nothing today and I left it alone rather than trade away the cross-vendor equality evidence it currently provides.
- [x] **Written down.** `docs/agent-rules/vulkan-software-driver-ci.md` (8.2 KB), plus the README subsystem-index line and failure-mode row, and `testing-architecture.md` §10 no longer says CI cannot run this.

## The job found a real engine bug before it ever ran in CI

Three `VulkanPassSuite` tenants failed their zero-validation-errors assertion on lavapipe and nowhere else:

> `vkCmdBindPipeline(): The pipeline being bound is the same as the currently bound pipeline and between the calls, a dynamic state was set which is static in this pipeline.`

`VulkanRendererAPI::SetViewport` emitted the plain `vkCmdSetViewport`, while every pipeline `VulkanPipelineBuilder` creates declares `VIEWPORT_WITH_COUNT` — as the draw front-end's own comment already said, *"the plain setters cannot satisfy it"*. It set a state that is **static** in every pipeline it could affect, so it changed no draw: inert at driver conformance ≥ 1.3.8.0 (hence invisible on NVIDIA), a validation error below it.

Fixed in its own commit (d269ae6), per CLAUDE.md's bug-on-the-way rule. `RecordedViewport` is unchanged; the call is simply gone. NVIDIA was **207/207 before and after**.

## One test is excluded, and it is a driver defect

`VulkanDrawPath.FramebufferBlitPreservesUnrequestedDepthStencilAspects` asserts that a single-aspect copy or resolve onto a combined depth/stencil image leaves the other aspect alone. lavapipe overwrites it — all 8 depth-only permutations lose the destination's stencil, the stencil-only ones symmetrically lose its depth, on all 256 pixels, **with no validation message at all**. Legal usage, wrong result. The same binary passes all 24 permutations on NVIDIA.

The exclusion is a named `-Test` in the filter with the evidence in a comment above it. It costs no coverage that existed before this PR — the test still runs on developer hardware, which is where it has always run. Re-test on the next Mesa bump.

## Scope

Nightly + `workflow_dispatch`, per the issue's scope boundary — **not** per-PR gating. The `pull_request` trigger is anti-rot only, filtered to this workflow file and `VulkanCapabilities.*`: the two changes that can silently turn this into a job that verifies nothing. Same trade `vulkan-off.yml` and `steam-stub.yml` already make.

**#1294 is not absorbed.** The filter deliberately excludes the L7 ray-query suites; that is a separately scored issue. This PR makes it a one-line change — widen the filter — once someone confirms lavapipe serves ray queries. lavapipe does advertise ray query and ray-tracing pipelines in the bring-up log, so it looks likely, but I did not test it.

## Verification

Everything here is artefact-backed; this PR changes no rendered pixel, so the renderer grid does not apply. All runs are Debug, on the developer box, `--gtest_filter=Vulkan*:VirtualShadowMapVulkanShaders*`:

| Arm | Result |
|---|---|
| NVIDIA RTX 4090, before the fix | 207/207 pass, 105/105 `EXERCISED` |
| NVIDIA RTX 4090, after the fix | 207/207 pass, 105/105 `EXERCISED` |
| lavapipe 26.2.0, before the fix | 204/207 — 3 validation-count failures |
| lavapipe 26.2.0, after the fix | 206/207 — only the driver-defect test |
| lavapipe 26.2.0, final job filter | **206/206 pass, 104/104 `EXERCISED`, exit 0** |
| Mesa 24.3.4 lavapipe + `--olo-require-vulkan` | exit 1, `NOT EXERCISED` |

**Not verified, and cannot be before merge:** the job running on a hosted runner. `workflow_dispatch` 404s unless the workflow exists on the default branch, so the first real dispatch is only possible after merge. The `pull_request` trigger fires on this PR because it touches the workflow file, which is the closest thing to a real CI run available beforehand — I have not faked a green for it. Two things are runner-specific and unproven locally: that the hosted image's `vulkan-1.dll` honours `VK_DRIVER_FILES`/`VK_ICD_FILENAMES` (both are set for exactly this reason), and that llvmpipe GL serves the shared test process's context, where this box used hardware GL. If either bites, it bites loudly — `--olo-require-vulkan` cannot report green through it.

Closes #1301

## Summary by CodeRabbit

- **New Features**
  - Added nightly software-based Vulkan validation using Mesa’s lavapipe driver.
  - Vulkan test runs now require a supported device and publish test results and image-difference artifacts when failures occur.
  - Viewport state is applied during rendering to improve command sequencing.

- **Documentation**
  - Documented supported software-driver requirements, verification rules, limitations, and runtime considerations.
  - Clarified the distinction between pull-request Vulkan reporting and nightly device-backed coverage.
  - Added the software-driver validation guide to testing documentation.


drsnuggles8:
@coderabbitai review

drsnuggles8:
@coderabbitai review

drsnuggles8:
## It runs green on a hosted runner — and the first attempt found why it wouldn't

Updating the "not verified, and cannot be before merge" caveat in the PR body: the `pull_request` anti-rot trigger fired this job on its own PR, so it **is** verified now, on `windows-2025`.

**Run [35442179862](https://github.com/drsnuggles8/OloEngineBase/actions/runs/35442179862):**

```text
apiVersion    = 1.4.354
driverVersion = 26.2.0
deviceName    = llvmpipe (LLVM 22.1.8, 256 bits)

[==========] 206 tests from 22 test suites ran. (100736 ms total)
[  PASSED  ] 206 tests.

Vulkan backend coverage: EXERCISED - 104/104 device-gated tests ran.
  device: llvmpipe (LLVM 22.1.8, 256 bits)
```

### The first attempt failed, and the diagnostic step earned its place

Attempt one got through the build and the Mesa download, then stopped at the device probe:

```text
ERROR: [Loader Message] windows_read_data_files_in_registry: Registry lookup failed to get
       ICD manifest files.  Possibly missing Vulkan driver?
ERROR: vkCreateInstance failed with ERROR_INCOMPATIBLE_DRIVER
```

The manifest was correct and both `VK_DRIVER_FILES` and `VK_ICD_FILENAMES` were set — the step printed the ICD's own `api_version: 1.4.354` immediately before failing. **The loader ignored them.** It reads both through its *secure-getenv* path, which returns nothing for a process running with elevated privileges, and GitHub's Windows runners run steps as an administrator. It fell back to registry enumeration, found no driver on a GPU-less runner, and refused.

This is exactly the runner-specific unknown the PR body flagged as unprovable locally, and it genuinely cannot reproduce on a developer box, where the shell is not elevated and the variables work as documented. Fixed in 815aeda4b by registering the manifest under `HKLM\SOFTWARE\Khronos\Vulkan\Drivers` — the loader's own Windows discovery mechanism, which works elevated or not. The env vars stay as the fallback; they are what keeps a local repro honest.

Two things worth noting about how it failed:

- The diagnostic step named the cause in **one line before any test ran**, instead of burying it under 206 identical skips. That is what it was added for.
- It could not have reported green. `--olo-require-vulkan` was the second gate behind it.

### Wall clock, measured on the runner — and a correction

| Step | Time |
|---|---|
| Configure CMake (warm vcpkg cache) | 3 m 31 s |
| **Build tests** | **2 h 09 m** |
| Download Mesa3D | 5 s |
| Report the device | 1 s |
| **Run the Vulkan suites under lavapipe** | **1 m 42 s** |
| **Job total** | **2 h 16 m** |

The PR body compared against "the ~25 minutes the llvmpipe GL job takes", because that is what `cross-vendor.yml`'s header and the issue both say. **That figure is wrong.** Measured on its 2026-09-19 nightly ([run 35419666930](https://github.com/drsnuggles8/OloEngineBase/actions/runs/35419666930)): **2 h 45 m total — build 2 h 37 m, `Run tests under llvmpipe` 2 m 24 s.** Off by more than 6x, and inverted: that job is not test-dominated, it is ~95% cold build.

So the right comparison is not 1m42s against 25 minutes. It is that **both software jobs are almost entirely build**, and their software test runs are minutes:

| | build | software test run | total |
|---|---|---|---|
| llvmpipe (GL, `cross-vendor.yml`) | 2 h 37 m | 2 m 24 s | 2 h 45 m |
| lavapipe (Vulkan, this job) | 2 h 09 m | **1 m 42 s** | 2 h 16 m |

The 90-minute cap on the test step has ~50x headroom, which is the right shape for a hang guard. I have corrected this job's inherited comment to state the measured figures.

**Follow-up, not taken here:** `cross-vendor.yml`'s no-sccache rationale rests on the inverted premise ("caching the minority build slice is a marginal win"). The build *is* the entire cost for both jobs, so switching them to Ninja + sccache is worth a real look. The mechanical blocker it names is still real — the default Visual Studio generator ignores `CMAKE_<LANG>_COMPILER_LAUNCHER` — so this is a separate change, not a rider on standing this job up. I have not edited that file.

### On the earlier Vulkan OFF red

It failed twice on `ad2284cea`, then **passed on `815aeda4b`** — where the only difference is 57 lines of YAML and Markdown, with `VulkanRendererAPI.cpp` byte-identical between them. Same runner (`olo-ci-1`), and the passing run rebuilt the same `OloEngine.dir/cmake_pch.hxx.pch` and `OloEnginePCH.cpp.o`, so it exercised the identical path rather than skipping it.

A transient in the ccache/PCH layer, then, not this branch's code. The include chains are the tell: the failing runs show a relative `../../../` chain that never passes through `cmake_pch.hxx`, while the passing run — and the sibling PR that passed alongside the first failure — show the absolute chain *through* it. Different compile mode, same source. Recording it here because "only this branch in 25 runs" looked like attribution and was not.

PR #1322, opened 2026-09-19
feat(groom): dense-coat self-shadowing and light transmission (#1248)

### LINKED ISSUE #1248: Groom: dense-coat self-shadowing and light transmission

## Parent

#1223

## Agent Brief

**Category:** enhancement

## What to build

Give a coat convincing internal depth with a density/deep-shadow representation and explicit body-to-hair/hair-to-body shadow integration.

## Acceptance criteria

- [ ] Prototype density-volume/deep-shadow options against a dense static coat and select using quality, cost and memory evidence.
- [ ] Undercoat, guard-hair silhouettes and light transmission remain distinct under front/side/backlight, including pale coats.
- [ ] Hair shadows the body and scene and is shadowed by them; representation resolution and update policy are inspectable.
- [ ] Animated light/camera movement and shadow LOD changes do not cause uncontrolled flicker or stale density.

## Scope boundaries

Raster-compatible coat shadows first. Hybrid RT proxies are a separate slice; avoid double-counting shadow/transmission terms.

## Blocked by

- #1247

## Score
```olo-score
capability: 8
craft: 13
stability: 3
decay: 1
effort: 5
confidence: 0.5
learning: 13
fun: 13
kano: performance
blocked_by: [1247]
blocks: [1223, 1250, 1251, 1255]
```

drsnuggles8:
Gives a dense coat internal depth: a density representation, selected on measured evidence, marched
per light in the strand shader so the hair underneath is occluded by the hair on top.

## The bake-off, and what it selected

Acceptance criterion 1 asks for a bounded comparison **before** an approach is recorded as selected.
`docs/analysis/groom-coat-self-shadowing-1248.md` is that comparison; every number in it is an
assertion in `GroomCoatShadowPropertyTests.cpp`, so the argument fails loudly rather than ageing in
a merged PR body.

The measured quantity is `tau(x, L)` — the **expected number of fibre crossings** between a point
and the light — from which transmittance is `exp(-kappa * tau)`. It is purely geometric and sees no
colour, which is what keeps it from double-counting #1247's per-fibre absorption.

Best of breed, mean absolute transmittance error against an exact ray-cylinder ground truth:

| case | `None` | Isotropic 64³ | **Anisotropic 64³** | Deep map 128²×8 |
|---|---|---|---|---|
| pelt, side light | 0.8317 | 0.0573 | **0.0297** | 0.0515 |
| pelt, top light | 0.8516 | 0.0411 | **0.0236** | 0.0495 |
| pelt, oblique light | 0.8418 | 0.0526 | **0.0263** | 0.0542 |
| scalp, side light | 0.8937 | 0.0207 | **0.0130** | 0.0275 |
| scalp, oblique light | 0.8783 | 0.0181 | **0.0134** | 0.0218 |

**Selected: an anisotropic density volume, 64 voxels on the longest axis, marched at three voxels
per step.** Most accurate in every case, and — decisively — **light-independent**, so an animated
light costs exactly zero rebuilds. That is criterion 4 answered structurally rather than by tuning.

**Rejected: the deep opacity map**, with its numbers kept. Cheaper per query (3 fetches against ~10
taps, 320 KiB against the volume) but **per light**: it must be rebuilt whenever the light or the
coat moves, which is the exact failure criterion 4 names, and it is ~2× less accurate at its own
optimum.

### Two findings that changed the decision rather than decorating it

**Finer is worse, for both representations.** They optimise at roughly *two strand spacings* and
alias below it, because a cell holding one strand or none stores a sample rather than an average:

| deep map | mean \|e\| | | volume | mean \|e\| |
|---|---|---|---|---|
| 64² | 0.0522 | | 16³ | 0.0741 |
| **128²** | **0.0515** | | 32³ | 0.0480 |
| 256² | 0.0944 | | **64³** | **0.0297** |
| 512² | 0.1709 | | 128³ | 0.0369 |
| 1024² | 0.2508 | | | |

The first run of this comparison scored the deep map at 512² — four times past its own optimum —
and therefore rated it a third as good as it actually is. **The resolution sweep is what made the
comparison fair**, and it is the reason the selection is evidence rather than a preference.

**A coarser march is free.** Ten taps beat thirty, because the volume's error is a systematic bias
rather than quadrature. Three voxels per step is at or within noise of the minimum on both
reference coats; past four it degrades sharply on a dense coat.

## What is on screen

The coat term multiplies each light's incoming radiance; the BCSDF is untouched. Live, on
`GroomCoatShadowCoat.olo`, the pale coat goes from a featureless blown-out mass to one with a lit
rim and a dark interior, with guard hairs standing out against a darker undercoat.

**The root-to-tip ramp #1246 shipped is bypassed while coat shadowing is active.** That ramp existed
because "a strand is darker near the root because it is deeper in the coat" — precisely what the
volume now measures instead of assuming — so keeping both would darken the roots twice. A visible
consequence, pinned rather than hidden: **a coat too sparse to actually self-shadow comes out
brighter**, because the fake darkening it replaced was the stronger of the two. Measured at +375 053
summed luma on 1500 strands against −513 403 on 6000.

## Verification matrix

Every cell, both kinds. Artefact-backed rows cite a committed file; live-only rows cite what was
measured and what the log said.

| Cell | Kind | Evidence |
|---|---|---|
| OpenGL × Forward | artefact | `GroomCoatShadow_GL_Forward.png` + `GroomCoatShadowOff_GL_Forward.png`; 22 318 px differ, max delta 106/255, luma −513 403 (darkens) |
| OpenGL × Forward+ | artefact | `GroomCoatShadow_GL_ForwardPlus.png` + control; identical deltas to Forward — one directional light, and the Deferred control differs, so the path knob is taking effect |
| OpenGL × Deferred | artefact | `GroomCoatShadow_GL_Deferred.png` + control; 22 318 px, max 106/255 |
| Vulkan × Forward | **live-only** | live A/B on `GroomCoatShadowCoat.olo`: 298 572 px differ, max 249/255, luma −23 553 532. `[RHI] Backend: Vulkan (source: --rhi flag)` in `OloEngine.log`; **0 `[error]`, 0 VUID** across the session |
| Vulkan × Forward+ | **live-only** | same capture set, same numbers |
| Vulkan × Deferred | **live-only** | 307 769 px differ, max 251/255, luma −24 235 863. The path-set ack timed out, so `renderpath` was **read back** as `deferred` rather than assumed |
| MSAA 1 vs 4 | artefact | `GroomCoatShadow_GL_Deferred_{NoMsaa,Msaa4}.png`; the pass still reports 1 shadowed groom, 0 fallbacks |
| Non-native resolution | artefact | `GroomCoatShadow_GL_Deferred_{Native,Scaled}.png`; resolution in force 64 → 64, i.e. the bake does not move with the render target |
| Serialization: scene YAML | artefact | generated by OloHeaderTool; `GroomCoatShadowCoat.olo` authors all eight fields |
| Serialization: save-game | test | `GroomCoatShadowSaveLoad`, 4 cases incl. a corrupt-save sanitisation |
| Prior on-disk version | n/a | **not a cell**: the component is new, so a save written before it existed does not contain its key and the load never reaches the function |

GL live A/B for comparison with the Vulkan rows: Forward 296 736 px / −23 431 258, Deferred
306 444 px / −24 118 109 — Vulkan tracks GL to within 0.7 %.

Tests: **337/337 pass** across the groom suites and the binding-layout guards, including #1246's and #1247's evidence, which this change's
shader edits could have broken. 55 new cases across the bake-off, the selection seam, the pixels and the
save-game cell.

## What this does NOT do

**Neither direction of coat-to-scene shadow integration**, which is the other half of criterion 3.
The volume contains the groom's own strands and nothing else, so hair does not cast onto the body or
the scene, and the body does not cast onto the hair.

**Tracked as #1323**, scored. That is a separate slice rather than an oversight, for a structural
reason worth stating: the strand
geometry is built inside `GroomRenderPass::Execute`, which runs **after** `ShadowRenderPass`. Making
a groom a shadow caster therefore means lifting #1246's strand-geometry cache somewhere both passes
can reach — a refactor of that cache, not a new caster list. And once grooms *are* casters, a strand
reading the cascade map would be shadowed by its own coat twice; the intended fix is to offset the
receiver to the coat's light-exit point, which the march already computes.

**A bound, deforming groom is refused rather than approximated.** The bake reads rest-pose curves,
so on an animating body the strands move and the volume does not. It reports `GroomIsDeformed` and
renders unshadowed instead of carrying a bind-pose shadow around.

**No GPU parity test for the CPU/GLSL twin.** #1246 and #1247 each ship one; this twin deserves the
same and does not have it yet. The header says so rather than implying a guard that is not there.

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Renderer/Passes/GroomRenderPass.cpp` — `AcquireCoatVolume`. The bake,
   the shadow LOD, the hysteresis and the resident budget all live here, and six of the nine
   findings from my own `/code-review` pass were in this function. The march step in particular
   *was* derived as `extent / resolution`, which is the cubic voxel size on the longest axis only —
   so on a flat coat it shipped a one-voxel march instead of the measured three-voxel one. It now
   carries `volume.VoxelSize()`.
2. `OloEditor/assets/shaders/GroomStrand.glsl` — the ramp gate. It is `lit && coat`, not `coat`,
   because the coat term only runs on the lit path; gating on the coat mode alone removed an unlit
   groom's only depth cue and put nothing in its place.
3. `ShaderBindingLayout.h` — `TEX_GROOM_COAT_VOLUME = 75` was the last unclaimed sampler index. It
   moved `TEX_SHADER_GRAPH_0` 77 → 78 and `HEAP_IMAGE_SLOT_BASE` 78 → 79, and that base is mirrored
   in GLSL **by hand** in two places. Both moved in the same commit and
   `BindlessShaderPipeline.HeapImageBaseMatchesTheBindingLayout` passes.

**Re-verified after the fixes.** The self-review changed the shader and the march, so the live
numbers above are from the **fixed** build, not the one I first measured — re-running was the point,
since a PR citing numbers from a binary that no longer exists is the failure mode.

**What I verified, and how** — the bake-off numbers are re-derived by
`GroomCoatShadowPropertyTests` on every run, not quoted; the pixels by
`GroomCoatShadowVisualEvidenceTest` across three paths plus MSAA and a non-native resolution; the
Vulkan cells by a live editor A/B with the backend line and the log confirmed; criterion 4 by a
**rebuild counter** that must read zero across eight light directions and six camera poses, which is
the check that would have failed had a per-light representation been selected.

**Least confident about** — the shadow-LOD pixel-size metric. It now takes the transform scale into
account, but it is derived from the cull view and a bounds radius and I have only exercised it at
unit scale and one framing; a groom parented under a scaled hierarchy is the case I have not seen.
Also the `RGBA32F` volume: `RGBA16F` would halve the memory and is what the packing wants, but
`Texture3D`'s RGBA16F declares 8 bytes a texel while uploading `GL_FLOAT`, so `SetData` rejects the
buffer. I chose to record that rather than fix a shared RHI path from this branch.

**Deliberately not tested** — the deep opacity map has no GPU implementation at all: it exists in
the CPU model as the rejected candidate, and `SelectGroomCoatShadow` now **refuses** it with
`ModeNotImplemented` so it cannot be selected into a frame. (An earlier revision of this branch did
not refuse it, and silently substituted an isotropic density volume while reporting the deep map as
delivered — see the fix commit and the comment below.) Also no Linux/CI-GPU coverage of the visual
cells, which is the existing situation for every groom evidence test.

Closes #1248


drsnuggles8:
Two corrections to this PR's own claims, from a question about the "deliberately not tested"
section.

**1. The deep opacity map was not merely untested on the GPU — it was silently substituted.**

Requesting `DeepOpacityMap` passed every guard in the selection seam: `wantsVolume` is false so the
3D-texture check did not fire, and `HasDirectionalLight` is a constant so the per-light check did not
either. The pass then baked a **density volume**, because the bake does not branch on the mode,
handed the shader mode 3, and the shader's only test (`mode != ANISOTROPIC`) marched that volume
isotropically.

So the coat rendered, `GroomCoatShadowStats` reported the deep map as **delivered**, and the picture
was a different representation than the one asked for. A rejected candidate that still reports
success is worse than a missing one, because nothing looks wrong — and it is precisely the pair of
failures (a silent fallback, and a counter that lies) this seam exists to prevent.

The previous test asserted the *wrong* behaviour and passed, which is how it survived my own review:
it checked that a deep-map request was delivered, and was happy that it was.

Fixed in b4c5ceb8f: `GroomCoatShadowModeIsImplemented` names the difference between what the
bake-off measures (four modes) and what a shader renders (three); the seam refuses the rest with
`ModeNotImplemented`; the pass refuses before baking so no memory is spent on a draw that will
render unshadowed. The editor's mode label now says "REJECTED - renders unshadowed" rather than
"not wired". 57/57 coat tests pass.

So the honest version of that line is **"not implemented on the GPU, and now refused loudly"** —
not "implemented but untested".

**2. The missing shadow integration now has an issue: #1323.**

Both directions of coat-to-scene shadowing (hair→body/scene and body/scene→hair), scored
`capability 8, craft 8, effort 5, confidence 0.5`, with the structural reason it was not part of
this PR written down: the strand geometry is built inside `GroomRenderPass::Execute`, which runs
*after* `ShadowRenderPass`, so it is a refactor of #1246's geometry cache rather than a new caster
list. It also carries the double-count fix that becomes necessary the moment grooms are casters, and
notes that the same refactor is what would let a deforming groom's bake follow its body.

drsnuggles8:
All thirteen findings addressed in 3689aa7f3 — every one was valid, and no dismissals.

The four that could actually bite:

| # | Finding | Why it mattered |
|---|---|---|
| 7 | float `!=` on the width-scale key | `CLAUDE.md` forbids it outright. Now `Math::BitwiseEqual`, which also keeps the exact invalidation semantics an epsilon would have blurred. |
| 9 | coat bytes not subtracted on eviction | `m_CacheBytes` stayed permanently inflated by every evicted coat, after which the pass evicts healthy entries forever and reports itself over budget with nothing left to free. One `ReleaseCoatVolume` helper now owns all three steps at all five release sites. |
| 11 | unclamped `m_Resolution` | `MakeGroomCoatLodPolicy` is the one boundary a direct MCP or native write crosses. 1024 reached `BuildDensityVolume`, which accepts it — about 1.07e9 voxels, roughly **17 GB**. |
| 10 | resident set not bounded | `residentCoatVolumes` counted only the current frame's draws, so a groom that stopped being visible kept its volume uncounted and the next newcomer was still granted a slot. Residency is now counted across the whole cache, and a newcomer **reclaims the least-recently-used volume not in use this frame** rather than being refused while stale ones sit there. |

Finding 5 was subtler than it looked: `CoatSegmentBounds` took the min/max over a NaN endpoint and returned false, so **one** corrupt segment failed the entire bake — the opposite of the drop-the-curve promise `BuildCoatSegments` makes. The test only had a lone good segment, so it could not see it; it now uses a mixed set and asserts 1 binned / 2 rejected.

Finding 13 was the most important of the test findings and it was right: the MSAA and scaled-resolution cells asserted "the coat rendered", which is true even if the shader never samples the coat-shadow term. Both are now A/Bs against their own shadow-disabled controls:

    [groom-coat] MSAA 1: 22318 px differ, luma delta -513403, shadowed=1 fallback=0
    [groom-coat] MSAA 4: 22319 px differ, luma delta -513432, shadowed=1 fallback=0
    [groom-coat] MSAA effect ratio 4x/1x: 1.000
    [groom-coat] scaled effect: 11021 px differ, luma delta -265267 (native -513403)

Findings 1–4 were all genuinely wrong claims in the docs, and #2's arithmetic correction is the one I would most have wanted caught: the deep map's memory crossover is about **twelve** lights, not three — I had computed it against the CPU model's tighter packing rather than against the RGBA32F volume that actually ships. The map is rejected on rebuild cost and accuracy, not on memory, and the analysis now says so.

Finding 8 was fixed independently in b4c5ceb before this pass, prompted by the same question from the repo owner.

435/435 tests pass across the groom suites, the binding-layout guards and the component round-trips.

Issue #1361, opened 2026-09-19
Skin: the Burley fit runs narrow and the support radius truncates transport above a diffuse albedo of ~0.7


drsnuggles8:
Found by the independent skin diffusion reference added in #1255. Not fixed there: that slice is
read-only on production material code by design.

## What is wrong

Two related things, both measured against a semi-infinite searchlight Monte Carlo walk — the
transport Christensen and Burley's fit was fitted to.

**1. The fit runs narrow at high diffuse albedo, and the error grows into the tail.**

| Authored `ScatterColor` | Quantile agreement, transport vs fit |
|---|---|
| 0.35 – 0.70 | within **10%** at the 25/50/75/90% quantiles |
| 0.85 | fit is ~13% narrow at the median, ~23% at q75, ~**35% at q90** |
| 0.95 | ~24% / ~44% / ~83% |

The growth across quantiles is the diagnostic: a scale error would give the same ratio everywhere, so
this is the fit tracking the near field and losing the tail.

**2. The support radius truncates real transport at the same albedos.**

`SkinDiffusionSupportRadiusMM` places the outermost tap at the radius holding 99.5% of the *profile's*
energy. Where the profile is accurate that is also 99.5% of the transport. Where it is not:

| Authored `ScatterColor` | Fraction of the walk's exiting energy inside the support |
|---|---|
| 0.35 – 0.55 | > 99% |
| 0.85 | **97%** |
| 0.95 | **93%** |

## Why it matters more than the numbers suggest

The default `SkinProfileParameters::ScatterColor` is `(0.85, 0.55, 0.45)`. Green and blue sit inside
the accurate band; **red does not** — and red is the channel that carries the visible bleed past a
shadow terminator, which is the thing subsurface scattering exists for. So the one channel a reviewer
would look at is the one running narrow and losing 3% of its energy to the support.

## What it does *not* affect, which bounds the work

The pass's **edge response** — a shadow terminator crossing skin — barely moves: the largest
disagreement against transport is 0.040 of a unit step at an authored 0.55 and 0.045 at 0.85, against
a discretisation floor of 0.052 that `SkinDiffusion.cpp` already documents for the 17-tap kernel. An
edge response is cumulative and dominated by where the bulk of the energy sits, and the fit's error is
in the tail. So this is **not** a terminator bug; it is a tail/haloing question, and it should be
judged on a bright small feature against dark skin rather than on a terminator.

[...]

Issue #1360, opened 2026-09-19
Groom coat transmittance is a Jensen lower bound: exp(-kappa E[N]) over-darkens a disordered coat


drsnuggles8:
Found by the independent coat reference added in #1255. Not fixed there: that slice is read-only on
production material code by design, and the fix belongs with someone who can re-measure the coat's
appearance.

## What is wrong

`GroomCoatShadow::CoatTransmittance` returns `exp(-kappa * tau)` where `tau = E[N]`, the **expected**
fibre crossing count over the ray bundle. The transmittance a fragment footprint actually receives is
the mean of the per-ray transmittances:

```
T_true = E[exp(-kappa * N)]        what the footprint gets
T_ship = exp(-kappa * E[N])        what is returned
```

By Jensen's inequality `T_true >= T_ship`, with equality only when `N` has no variance. **The shipped
form is a lower bound on transmittance: it over-darkens**, and by an amount that grows with the
disorder of the coat rather than with any authored parameter.

## Evidence

`OloEngine/tests/Rendering/PathTracing/NonlocalTransportReferenceTest.cpp`, suite
`CoatTransportReference`:

* `TheProductionFormIsALowerBoundAndTheGapIsMeasured` — the inequality holds at every `kappa`
  sampled, and the largest gap over the sweep exceeds **0.02 in transmittance** on a Poisson fibre
  medium.
* `TheGapClosesOnAnOrderedCoat` — the control. A dense touching lattice, at matched optical depth
  1.2, has dispersion `Var/E < 0.1` against the Poisson medium's `1.0`, and a relative transmittance
  gap below **0.005** against more than four times that. So the gap is a property of the medium's
  disorder, not an artefact of the estimator.
* `ThePoissonMediumWalkMatchesItsClosedForm` — the reference itself is validated against an analytic
  ground truth with no sampling error.

## The fix is closed form and costs the same

For a Poisson medium the crossing count is Poisson distributed, so the exact mean transmittance is
that distribution's probability generating function at `exp(-kappa)`:

```
E[exp(-kappa N)] = exp(-mu * (1 - exp(-kappa)))
```

which is one extra `exp` on a value that is already being exponentiated, evaluated per pixel at most.
`MaterialReference::PoissonMediumTransmittance` is that expression and is already under test.

A real coat is not exactly Poisson — it is somewhere between the lattice and the Poisson medium — so
the honest options are (a) ship the Poisson form, which is correct for a disordered coat and slightly
[...]

PR #1362, opened 2026-09-19
test(rendering): independent skin, leaf and fibre material references and AOV comparisons (#1255)

### LINKED ISSUE #1255: Renderer validation: independent skin, leaf and fibre material references and AOV comparisons

## Parent

#1225

## Agent Brief

**Category:** enhancement

## What to build

Extend material validation beyond generic PBR so skin diffusion, leaf transmission and fibre response are checked against suitable independent references.

## Acceptance criteria

- [ ] Add analytic/numerical or offline reference cases with declared units, coordinate conventions and tolerances for each specialised model.
- [ ] Test energy/reciprocity/sampling consistency where applicable; explicitly distinguish local BSDF tests from nonlocal skin diffusion and coat transport.
- [ ] Produce aligned reference/production AOV comparisons across representative parameter ranges and light directions.
- [ ] If the existing CPU/GPU reference tracer cannot represent a model, report that limitation and add a suitable reference instead of treating generic PBR as ground truth.

## Scope boundaries

Each feature still owns its initial correctness tests; this slice consolidates independent references and comparisons.

## Blocked by

- #1241
- #1248
- #1234

## Score
```olo-score
capability: 8
craft: 8
stability: 5
decay: 1
effort: 3
confidence: 0.8
learning: 8
fun: 8
kano: performance
blocked_by: [1241, 1248, 1234]
blocks: [1225, 1259]
```

drsnuggles8:
Closes #1255.

## What this is

The engine already has a reference apparatus — `PathTracing/ReferenceBRDF.h` and the `Reference*Test`
files beside it — and it is a faithful CPU mirror of `PBRCommon.glsl`'s Cook-Torrance BRDF. It is the
right oracle for a metal-rough surface and the wrong one for skin, leaf and fibre: none of the three
*is* a Cook-Torrance surface, so "compared against the reference tracer" would mean "compared against
a model that cannot represent it". That is the failure criterion 4 names by name.

So this adds an independent oracle per model, each computed by a route that shares no code — and
where possible no derivation — with the model it judges:

| Model | Reference | Shares with production |
|---|---|---|
| Fibre | dielectric-cylinder cross-section walk: explicit 3D entry geometry, full vector Snell, a term-by-term path sum, **plus** a stochastic walk that never forms the recurrence's products at all | nothing but the convention |
| Leaf | plane-parallel translucent slab by Monte Carlo random walk | nothing — they share no parameter |
| Skin | semi-infinite searchlight random walk — the transport Christensen-Burley was *fitted to* | the medium's definition |
| Coat | stochastic fibre medium, by Monte Carlo **and** in closed form for a Poisson medium | nothing; the intersection test is test-local |

**No production code is touched.** This slice reads the three material models and adds references, one
test-only probe shader and a decision record — which is also what the handover's read-only constraint
required, with two sibling worktrees editing adjacent groom and skin code.

## Acceptance criteria

- [x] **Analytic/numerical or offline reference cases with declared units, coordinate conventions and
  tolerances for each specialised model.** All four are declared per entry point in
  `MaterialReference.h` and tabulated in
  [`docs/analysis/material-reference-validation-1255.md`](docs/analysis/material-reference-validation-1255.md).
  Every tolerance states where its number comes from — an f32 mantissa, N standard errors at a named
  sample count, or a quadrature rule's measured error — rather than being taste.
- [x] **Energy / reciprocity / sampling consistency where applicable; local BSDF tests explicitly
  distinguished from nonlocal skin diffusion and coat transport.** The file split *is* the
  distinction, and §0 of the doc tabulates which questions have a referent for which model. Asking a
  diffusion profile for reciprocity would be asserting a coincidence, and that is said rather than
  quietly skipped.
- [x] **Aligned reference/production AOV comparisons across representative parameter ranges and light
  directions.** Linear RGBA32F probes with no camera, no exposure and no tonemap, so the alignment is
  a property of the setup rather than a claim about it — deliberately chosen over a beauty A/B, since
  a ratio of tone-mapped sums reads backwards here and evidence PNG shading depends on test order.
- [x] **Report the limitation rather than treating generic PBR as ground truth.** Four reported below.

## What the references measured

Confirming the models is most of the result — the shipped `GroomFibreCommon.glsl` agrees with the
cylinder walk over 64 light directions at 4e-3, and three production formulas (the incidence cosine
`cos θ_o cos γ_o`, Bravais' modified index, the closed-form geometric tail) come out as *consequences*
of the reference rather than inputs to it. Beyond that:

**1. Coat transmittance is a Jensen lower bound.** `CoatTransmittance` returns `exp(-κ·E[N])`; the
transmittance a fragment footprint actually receives is `E[exp(-κ·N)]`. Jensen puts the second at or
above the first, so the shipped form **over-darkens a disordered coat**. Measured gap above 0.02 in
transmittance over a κ sweep, and near zero on an ordered control at matched optical depth. Filed as
#1360 — the fix is a closed form of the same cost.

**2. The Burley fit runs narrow at high albedo, on the channel that matters.** Within 10% of transport
up to a diffuse albedo of 0.70; at 0.85 it is ~13% narrow at the median and ~35% at the 90th
percentile, and the support radius holds 97% of the walk rather than 99.5%. The default
`ScatterColor` is `(0.85, 0.55, 0.45)` — so green and blue are inside the accurate band and **red**,
the channel carrying the visible bleed past a terminator, is not. Filed as #1361.

**3. The leaf lobe is a gain, not a transmittance.** Its directional-hemispherical transmittance
exceeds 1 inside the authored ranges (`wrap = 1`, `thickness = 1` suffices), and it is non-reciprocal
by over 25%. Both are legitimate consequences of a phenomenological term that #1234 chose deliberately
for a lamina whose physical thickness is never authored — but they mean **no absolute comparison
against transport can ever be made for this model**, only shape and ordering, and that had to be
written down rather than papered over.

**4. A documentation nit, not filed.** `GroomFibreScattering.h` describes `σ_a` as "in units of the
fibre DIAMETER"; the expression it ships, `exp(-σ_a · 2cos γ_t / cos θ_t)`, is per **radius** (a
head-on chord on a unit-radius cylinder is 2). Nothing behaves wrongly — the comment is one word out.
Left alone because that file is off-limits to this slice while `feature/fur-coat-authoring-1251` is in
flight; a one-word fix for whoever touches it next.

## Three mistakes the tests caught in the references themselves

Worth recording, because each was a plausible-looking green arm away from shipping:

- **The reciprocal quantity is the BTDF, not the binned energy.** The first slab reciprocity test
  compared binned energies and read a 0.61 "reciprocity failure". The measured ratio was exactly
  `mu_o/mu_i` (0.6163 against a predicted 0.613) — which is the missing exit cosine, and the exact
  match is how it was identified rather than mistaken for a broken solver. With the cosine,
  reciprocity holds to 0.55%.
- **`ScatterColor` is a diffuse *surface* albedo, not a single-scattering albedo.** Driving the walk
  with it directly compares two different media and reports Burley as twice too wide. Cross-checking
  the walk against Chandrasekhar's `A = 1 − √(1−a)·H(1)` (agreement within 1%) is what showed the
  error was in the test. The inversion is now a named function and an asserted bridge.
- **"Regular" is a property of a medium *and* a ray direction.** The first ordered-coat control laid
  its fibres along `+z` and fired along `+x`, so a ray crossed either a whole row of forty or none of
  it: variance **thirty times** the mean, more disordered than the Poisson medium it was controlling
  for.

A fourth, in my own claims rather than the code: the leaf wrap term does **not** make the lobe
isotropic at `wrap = 1` (measured spread 0.83). What is exactly true, and is now what is asserted, is
that it adds a *constant over the view hemisphere* — an authored floor. Transport has no such slider.

## Verification

| Axis | Cell | Result |
|---|---|---|
| Reference evaluation | CPU reference | 31 L1 cases, all pass |
| Reference evaluation | GPU parity | 3 shaderpipe cases, all pass |
| Production AOV | OpenGL | fibre ambient over 64 light directions; leaf over 4096 view/light directions × 2 profiles × 2 controls |
| Production AOV | Vulkan | **not reachable** — the headless fixtures need a GL 4.6 context and skip without one. Both probes are fullscreen draws with no camera and no path selection, so forward / forward+ / deferred do not enter either |
| Upscale / non-native resolution | cell for the screen-space models | passes — world footprint invariant, exactly |
| Field of view | cell | passes — exactly |
| MSAA | **not a cell** | no AOV is read before a resolve; both probes render single-sample with no resolve in the path |
| Serialization, scripting, physics | **not reachable** | no persisted format added; the one new asset is a test probe shader under `assets/shaders/tests/`, not cooked |

```
OloEngine-Tests.exe --gtest_filter='SkinDiffusionReference.*:CoatTransportReference.*:LeafTransmissionReference.*:FibreScatteringReference.*:MaterialReferenceAovParity.*'
[==========] 34 tests from 5 test suites ran. (21429 ms total)
[  PASSED  ] 34 tests.
```

Adjacent suites, for regressions from the CMake and probe additions:

```
OloEngine-Tests.exe --gtest_filter='GroomFibre*:Foliage*:Skin*:GroomCoat*'
[==========] 391 tests from 70 test suites ran. (303539 ms total)
[  PASSED  ] 391 tests.
```

**No visual verification, and that is the honest answer rather than a gap.** This slice changes no
renderer or shader code, so there is nothing on screen that can have moved. The one new shader is a
test probe, and its compilation is covered by the shaderpipe test (and by `glslc`, both stages clean).

**Runtime budget**, since this adds four test files to the ctest shards: 22 s for all 34 cases in a
Debug build, dominated by the searchlight walks at high albedo. Every estimator is seeded, so a
failure reproduces exactly.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive reference-based validation for fibre scattering, leaf transmission, skin diffusion, and coat transport.
  * Added GPU shading parity checks to verify rendered material outputs against independent CPU calculations.
  * Added deterministic checks for energy conservation, reciprocity, sampling, attenuation, and parameter behavior.

* **Documentation**
  * Documented reference methodologies, tolerances, test coverage, limitations, and identified material-model behavior.


drsnuggles8:
Addressed all three review findings in e80f5df29.

**1. Slab energy closure was a tautology — fixed, and this was the one that mattered.** `Absorbed` was defined as `1 - R - T`, so `TheSlabWalkConservesEnergy` asserted `R + T + A == 1` to 1e-12 whatever the walk did. The assertion could not fail; my own comment beside it said "the identity is arithmetic" without drawing the conclusion. The absorbed share is now accumulated from the weight each scattering event removes, making the three independent measurements.

Agreed with the review on the roulette too: a killed path is sampling termination, not absorption, and its weight is already accounted for by survivors being raised to the floor — booking it would double-count. That makes the closure hold only in expectation, so the tolerance is now statistical: **1e-3 absolute against a measured worst deviation of 9.4e-5** across the tau/albedo grid, which is two orders below the percent-scale leak it exists to catch.

**2. `Pcg32` determinism — added.** Fair: every tolerance in this directory rests on the generator producing the same bytes everywhere, and nothing pinned it. Added same-seed reproducibility over 1000 draws, a different-seed divergence check so that cannot pass on a constant, a pinned eight-draw sequence for seed `0x1255`, and a bound on the unit-float conversion — every walk takes `-log(1 - u)`, so a `u` that could reach 1 returns an infinity rather than a number.

Worth noting how the pinned values were obtained, since a sequence generated from the implementation it is meant to guard is circular: they came from an independent Python reimplementation of PCG32's state transition and output permutation, written from the algorithm rather than from this code. The C++ agreeing with it on the first run is the cross-check.

**3. Memoise the searchlight walk — taken.** Correct count: six authored albedos reached 24 times, 18 of them repeats. Cache is keyed on the authored value and the seed parameter is dropped entirely rather than conditioned on, which removes the dangling-reference hazard in returning a reference from the non-cached path. **22.2 s -> 9.3 s** for the new suites in Debug, skin alone 9.8 s -> 2.4 s — worth having given four new files reshuffle the ctest shards.

36 cases pass. No production code touched.

Issue #1250, opened 2026-09-13
Groom: guide simulation and body collision for moving long fur and hair


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1223

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Simulate a limited guide set and interpolate the rendered coat, with stable rest shape and body collision on a moving long-coated animal and human hair asset.

## Acceptance criteria

- [ ] Guides preserve length/rest shape within declared tolerances across variable frame rate, pause/resume and teleport.
- [ ] Collision handles representative body shapes and prevents conspicuous body penetration during the benchmark motions.
- [ ] Simulation, interpolation and density/shadow updates consume a coherent frame state and generate correct motion history.
- [ ] Per-group simulation budgets, reset controls and debug views are authored/cooked; record cost as guide/render-hair counts change.

## Scope boundaries

First minimise and measure a solver/collision prototype. Full individual-hair self-collision is not assumed necessary for the real-time target.

## Blocked by

- #1249
- #1248

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 8
craft: 13
stability: 3
decay: 1
effort: 5
confidence: 0.5
learning: 13
fun: 13
kano: performance
blocked_by: [1249, 1248]
blocks: [1223, 1252]
```
[...]

PR #1363, opened 2026-09-19
feat(groom): coat groups, per-role budgets and texture-driven coat variation (#1251)

### LINKED ISSUE #1251: Fur: undercoat, guard hairs and regional coat authoring on two animals

## Parent

#1223

## Agent Brief

**Category:** enhancement

## What to build

Expose coat groups and texture-driven length/density/clump/colour variation, then author short- and long-coated animal references with distinct muzzle, ears, body and tail regions.

## Acceptance criteria

- [ ] Undercoat and guard hair groups can be adjusted independently while preserving density and silhouette during cooking.
- [ ] Regional maps and root UV colour patterns remain attached under body deformation; deterministic variation avoids visible repetition.
- [ ] Editor preview exposes group visibility, density, widths and clumps with bounded runtime parameter edits.
- [ ] Capture both animals under controlled front/side/backlight; include whisker/long-hair groups where present and avoid a uniformly fuzzy coat.

## Scope boundaries

Dry coat authoring on supplied assets. Wetness/mud and interactive grooming remain later extensions.

## Blocked by

- #1248
- #1249
- #1239

## Score
```olo-score
capability: 5
craft: 13
stability: 3
decay: 1
effort: 3
confidence: 0.8
learning: 8
fun: 13
kano: performance
blocked_by: [1248, 1249]
blocks: [1223, 1252]
```

drsnuggles8:
Closes #1251.

Coat groups become a first-class authored concept, and the strand budget is spent **per role** so a
budget cut costs the animal's interior before its outline.

A cooked groom already carried *groups*; it did not carry what a group **is**. Every knob was
therefore global, and turning the density down thinned the sparse guard hairs that draw the
silhouette by the same factor as the dense undercoat — the "uniformly fuzzy coat" criterion 4
forbids, stated as a mechanism rather than a judgement.

---

### 1. Undercoat and guard hair groups adjusted independently, density and silhouette preserved during cooking

`GroomCoatRole` (Unassigned / Undercoat / GuardHair / Whisker / LongHair) plus the per-group
density, length, width, clump and tint live in the **asset** and are cooked into the `.ologroom` as
section 9. `GroomCoatComponent` gives the two roles the criterion names their own bounded runtime
multipliers.

The budget is solved per role by bisection for the `k` with `sum_r count_r * min(1, k * w_r) ==
budget`, weighted Whisker 64, GuardHair 8, LongHair 4, rest 1. Measured on the long-coated reference
at 12000 → 1500 strands, the fraction of the animal's **outline** that survives:

| budget allocation | outline retained |
|---|---|
| single global stride (the build before this PR) | **0.293** |
| per-role weighted | **0.511** |

Same groom, same geometry, same budget; only the allocation differs. The pair is
`FurCoatBudgetThin_GL_Forward.png` vs `FurCoatBudgetThinGlobalStride_GL_Forward.png` — the global
stride leaves a visibly bald animal with the body edge showing through and two whiskers left.

Counter-side, on the unit fixture at a 200-strand budget: guard retained > undercoat retained, and
whiskers retained **exactly 1.0** (`GroomCoatAuthoring.TheBudgetSpendsItselfOnTheSilhouetteLast`).

*Preserved during cooking*: `GroomCooker::CookToBytes` copies field by field, so the coat table had
to be named there explicitly — the trap is that the loose asset keeps the roles and the cooked one
drops them, and the difference is a coat that is only slightly too uniform.
`TheCoatTableSurvivesTheCook` cooks, reloads and asserts exact equality of the whole description.

### 2. Regional maps and root-UV colour patterns stay attached under deformation; variation avoids repetition

`EvaluateGroomCoatStrand` takes a curve index, a root UV and a group id — and nothing else. No pose,
no world position, no frame counter, no camera. The coat's shape changes are applied in **rest
space** (`ApplyGroomCoatShape`) and the binding's root transform is applied to the result, so a
regional map cannot slide off a moving animal. `ApplyGroomCoatShape` returns the root **bit for bit**
at `t == 0` for both the length scale and the clump pull.

Two root-UV maps: a region map (R length, G density, B clump) and a colour map, read back to the CPU
once and sampled **per strand** — two of the region map's channels decide whether a strand exists
and how long it is, which no fragment shader can answer.

*No repetition*: the variation is a hash of the curve index, so it has no period. `TheVariationHasNoPeriod`
checks every lag from 1 to 256 over 4096 samples for exact repeats and asserts the four salts are
uncorrelated. `TheDensityDrawIsSpatiallyUncorrelated` checks the density draw does not stripe (longest
run < 25 over 600 curves at 0.5 density) — a stride would decimate in cooked order, which within a
group is spatial order.

### 3. Editor preview exposes group visibility, density, widths and clumps with bounded runtime edits

New **Groom Coat** inspector section: a read-only table of the asset's groups (name, role, curve
count, authored parameters, with each group's preview hue beside it), per-role visibility checkboxes,
the undercoat/guard sliders, the two map slots, the variation controls, and a live "resulting build"
readout of available/built/stride per role. Every slider is a **multiplier** with a hard range.

Verified live: setting `RoleVisibilityMask` through MCP on Vulkan removed that entity's coat
entirely and left its two siblings untouched, rebuilding the geometry through the existing cache key.

### 4. Both animals under controlled front/side/backlight, whisker and long-hair groups, not uniformly fuzzy

Two committed reference animals (`reference-shortcoat-animal`, `reference-longcoat-animal`) with
muzzle, ears, body and tail regions in **disjoint root-UV bands** — without which a regional map is
indistinguishable from a global multiplier — plus whiskers on both and a mane and tail plume on the
long coat. Nine groups each, every one with an authored role
(`TheReferenceAnimalsCarryRealCoatRoles`).

Six captures, two animals × front/side/backlight, each with its control. The authored/control pair is
the clearest evidence: the control is a smooth combed mass with a hard outline; the authored coat has
longer varied guard hairs breaking the silhouette, visible tufting and per-strand shade.

Pixels changed by enabling the authoring: ShortCoat 33.6k / 31.6k / 28.0k, LongCoat 130.9k / 123.5k /
96.1k (front/side/back).

---

### Primary — {OpenGL, Vulkan} × {Forward, Forward+, Deferred}

| backend | path | evidence | result |
|---|---|---|---|
| OpenGL | Forward | artefact — `FurCoatGroups_GL_Forward.png` + control | **123 499 px** differ |
| OpenGL | Forward+ | artefact — `FurCoatGroups_GL_ForwardPlus.png` + control | **123 499 px** differ |
| OpenGL | Deferred | artefact — `FurCoatGroups_GL_Deferred.png` + control | **123 506 px** differ |
| Vulkan | Forward | live — editor session, matches the GL frame | pass, no log errors |
| Vulkan | Forward+ | live | pass, no log errors |
| Vulkan | Deferred | live — matches the GL Deferred frame | pass, no log errors |

Every Vulkan cell is live-only by construction: the headless fixtures need a real GL 4.6 context and
skip without one. The Vulkan arm is the one that matters most here — it **pulls** the vertex as a
flat float array, so the tint lane at index 11 is a stride contract, not an attribute binding.

### Assets — {loose, cooked}

Both cells. `TheCoatTableSurvivesTheCook` cooks a groom with non-default parameters on every group and
asserts exact equality after reload. `TheCookIsStillDeterministicWithACoatTable` cooks twice and
compares the buffers byte for byte (section 9's explicit pads are zeroed on the way out).

### Serialization

| cell | status |
|---|---|
| scene YAML | **a cell, and it passes.** The coat lives on a component, so scene YAML is generated by OloHeaderTool (all-trivial layout); `GroomCoatAuthoring.olo` round-trips through it in the live editor. |
| asset pack | **a cell.** The groom asset is what is extended: section 9, `CurrentVersion` 1 → 2. |
| save-game | **a cell.** `Serialize` overload + `RegisterAll` hand-written (not generated), with every float re-clamped on load because `OLO_SERIALIZE` does not reach the archive. |
| a prior on-disk version | **a cell.** A `.ologroom` is a derived artifact, so `MinSupportedVersion` moves with `CurrentVersion`: a v1 file is **rejected by version**, loudly, and re-imported. `AFileFromBeforeTheCoatTableIsRejectedByVersion` forges one and asserts the refusal names the version. The two committed fixtures were regenerated by their own generator test. |

### Conditional renderer axes

| cell | result |
|---|---|
| MSAA on/off | **run** (Deferred; `MSAASampleCount` lives on `DeferredSettings`). 1× 123 506 px, 4× 123 505 px, ratio 1.000. Each sample count is its own A/B. |
| Upscale / non-native resolution | **run.** Coat footprint 0.1340 of the frame at 1280×720, 0.1128 at 960×540 — the fraction, not the pixel count, because that is what a coat going bald or solid would move. |
| Cross-entity variation | **run.** Three entities in the live scene share two groom assets at different seeds and densities and render as three different animals. |

---

## Two measurements that were wrong first

Both produced a plausible number, and both are written up in
[docs/agent-rules/coat-authoring-and-per-role-budgets.md](docs/agent-rules/coat-authoring-and-per-role-budgets.md) §5.

**Absolute variance falls when a multiplicative tint is applied.** Shade jitter 0.0 → 0.6 *lowered*
the coat's luminance variance on a frame where 27 000 pixels visibly changed: a tint cannot push an
albedo above the authored one, so the upper half of a symmetric jitter clamps and the coat darkens,
and absolute variance scales with the square of the mean. The assertion now uses the **relative
spread** (σ/μ), which rises 0.735 → 0.808 at maximum jitter.

**Interior pixel counts are saturated; outline counts are not.** The first silhouette test compared
the outline's retained fraction against the whole coat's under one budget cut. It fails, and not
because the weighting is broken: thousands of undercoat strands land on the same pixels, so the
interior barely moves while the sparse outline responds almost linearly. The control arm had to be
*the other budget allocation*, not the full coat — which is what produced the 0.293 vs 0.511 above.

The "is the coat uniform" statistic in the six capture cells is **printed, not asserted**, because
the authored and control coats occupy different pixel sets and the comparison selects for that.

## Self-review pass before CodeRabbit (second commit)

Four findings, three real:

* **The per-role stride rounded to nearest**, which is the wrong direction for a budget — 1/0.4 is
  2.5, kept half a role where the solver had decided on four tenths, and overshot `MaxStrands` by
  about a third. Ceil also reproduces the pre-#1251 arithmetic exactly on a groom with no roles,
  which round did not. The ceil needs a relative tolerance: the bisection converges on the largest
  `k` it *proved* affordable, so 500 strands into a budget of 100 returned 0.2 minus ~1e-12 and a
  bare ceil made that a stride of six.
* **The root-UV map cache grew without bound.** Keyed by asset handle, so entity destruction is not
  its eviction point; repointing a coat's map left 16 MB of pixels resident for the Scene's lifetime.
  Now rebuilt from the handles the frame asked for, with the failure latch preserved.
* **The whisker evidence measured a confounded quantity** — hiding a role frees its budget, so the
  rest of the coat got denser and the A/B mixed two effects. It now measures the whiskers against an
  all-hidden coat, and the claim is an equality: **2280 px at a budget of 400 and 2280 px at 12000**.
* An unused lambda capture.

## Fixed on the way (separate concerns, same commit)

* `GroomDeformation.TheSelectionStopsOnTheSameCurveTheBuildStopsOn` pinned a premise that this
  change improves: the old selection left the stride at 1 below one affordable curve, so a tiny
  budget selected the whole groom and built one strand of it. The per-role solver has no such floor.
  The invariant under test is unchanged; the premise moved, and the test says why.
* `GroomCookDeterminism.EveryEnumeratedSectionIsWrittenExactlyOnce` pins the section count
  deliberately — 9 → 10, with a note that bumping it is the moment to check the encoder, the
  decoder's fixed order, the payload arithmetic and both version constants.

## Not done, deliberately

Wetness, mud and interactive grooming are the issue's stated later extensions. The fibre BCSDF
(#1247, under independent validation in #1255) is **not touched**: the per-strand tint multiplies the
shaded radiance at the end of `GroomStrand.glsl` rather than modulating `sigma_a`, which is an
approximation named at the point it is applied. Guide-driven interpolation is #1250.

## Tests

299 of 299 pass across `*Groom*` and `*FurCoat*` (37 suites, 180 s). Pre-commit clean.


drsnuggles8:
@coderabbitai review

PR #1366, opened 2026-09-19
measure(skin): the support fraction and the Burley fit both stay (#1361)

### LINKED ISSUE #1361: Skin: the Burley fit runs narrow and the support radius truncates transport above a diffuse albedo of ~0.7

drsnuggles8:
#1255 measured two errors in the skin screen-space diffusion pass. #1361 asked which to repair,
with a documented prior that widening the support fraction was "probably the whole fix".

**Measured answer: neither, and the prior was wrong in an interesting way.** There is a third
approximation in the pass that neither lever touches, and it is the dominant one.

No behaviour change ships here. Every added line in `SkinDiffusion.h` is a comment — verified — so
the kernel is byte-identical. What ships is the measurement that decided it, the decision written
down where the constant lives, and two reference tests that assert the new relationship.

## The measurement

All at the default `SkinProfileParameters::ScatterColor`'s red channel, authored **0.85** — the one
channel of the shipped default that sits outside the fit's accurate band. Against the same Monte
Carlo searchlight walk #1255 introduced. The image is a **bright small feature against dark skin**,
which is what the issue asked for: an edge response is cumulative and a tail error barely reaches
it, which is why #1255's terminator bound could not settle this.

The halo metric is *the fraction of the feature's energy landing at the wrong radius*, read at
2/3/5/8/12/18/25 mm and taken as the worst.

### Criterion 1 — the support radius: leave it at 0.995

| | |
|---|---|
| support the fit asks for | **15.02 mm**, holding **97.09%** of the transport |
| support to cover 99.5% of the *transport* | **27.05 mm** — a **1.80×** widening |
| centre tap's share of the profile | 0.143 → **0.217** (a sixth of the blur becomes no blur) |
| centre tap's span | 0.23 mm → **0.43 mm** |
| halo error | 0.0558 → **0.0598** |

A cost with no benefit. The reason is the load-bearing one: **a support radius carries no energy —
the tap weights do**, and they come from the same narrow fit. Moving the outer taps further out
moves a near-zero weight further out, and because the tap count does not grow with them, every
offset scales together and the kernel gets uniformly coarser.

### Criterion 2 — the fit: do not revisit it

The control is the decisive measurement: a separable pass carrying the **exact transport profile**
with **unlimited taps** — no Burley fit, no 0.995 truncation, no 17-tap discretisation. Everything
the issue proposed repairing, repaired perfectly and for free.

| kernel | halo error |
|---|---|
| shipped (Burley fit, 17 taps, 0.995 support) | **0.0558** |
| support widened 1.80× | 0.0598 |
| **exact transport profile, unlimited taps** | **0.0806** (converged control: 0.0782) |

Repairing the profile moves the image **further** from transport. The pass is **separable**, and at
a small bright feature that projection over-spreads the core along the axes while the narrow fit and
the truncated support pull energy back in. The two errors partly cancel — which is a coincidence,
not a design, and is explicitly *not* asserted as a virtue. What is asserted is the ordering that
makes both levers pointless.

The best correction available anyway is tiny: sweeping a scale on the profile's `d` against this
metric, the optimum is **1.06×** at 0.85, worth 0.006, and **1.00×** at every albedo at or below
0.70. Nothing here justifies a format change.

### Criterion 3 — the tests

`SkinDiffusionReference` gains the two measurements above, and the two existing tests now carry the
resolved claim instead of leaving a number hanging as an implied defect. The reference stays; the
claims it makes changed, as the issue asked.

Two quality notes on the new tests:

- The widened-support arm needs weights **rebuilt** at the new support, not just rescaled offsets —
  raising the fraction gives each tap a wider strip of the same profile. Since
  `kSkinDiffusionSupportFraction` is a compile-time constant, that needed a mirror of the production
  tap binning; the mirror is **asserted against the real tap table** at the natural support before
  it is used, so the test cannot end up measuring its own transcription.
- Both decisions `GTEST_LOG_(INFO)` their numbers. A decision to change nothing leaves no trace in
  the diff, so the evidence has to be reachable without re-deriving it.

Metric cost is negligible: both sides are closed-form circle–circle overlaps (the separable PSF is a
weighted point set, the walk is weighted annuli), so there is no grid, no FFT and no resolution to
tune. The suite runs in 2.4 s.

## Verification

| | |
|---|---|
| `SkinDiffusionReference` | 10/10 pass (8 existing + 2 new) |
| `CoatTransportReference` (co-tenant of the shared file) | 4/4 pass |
| `*Skin*` across the suite | **220 pass**, Vulkan device-gated coverage exercised 1/1 |
| `pre-commit run --all-files` | clean, no reformatting |

**The renderer axes (GL/Vulkan × Forward/Forward+/Deferred, resolution, upscale, MSAA, prior
on-disk version) are all answered by one fact rather than by captures: the production diff is
comment-only.** Every added line in the only production file is a comment — so the kernel, the
shaders, the serialized format and every pixel are unchanged. There is no before/after image to
take, because there is no before/after. `kSkinDiffusionSupportFraction` keeps its value, no
`SkinEvaluationModel` was added, and the cumulative transport version lists are untouched.

## Also

- `docs/guides/skin-diffusion.md` — one entry under *Explicit limitations*, stating both errors and
  why they stand.
- `docs/agent-rules/notes-renderer.md` + one README failure-mode row — the reusable half: *build the
  control that removes the error and check the image improves, before tuning the constant that
  causes it.* An approximation's two named error terms were both real, both measured, and neither
  was dominant.

Closes #1361

## Summary by CodeRabbit

- **New Features**
  - Added an isotropic gather skin-diffusion mode for improved single-pass subsurface scattering.
  - Isotropic gather profiles now support skin transmission, oral, ocular, and layered-specular effects.

- **Documentation**
  - Clarified limitations at high diffuse albedo values, including transport coverage and projection errors.
  - Added guidance on blur support and profile refinement trade-offs.

- **Tests**
  - Added visual and reference coverage for gather diffusion, bright details, halo energy, and renderer paths.


drsnuggles8:
## Live verification in the editor

The PR body argues the renderer axes are answered by the diff being comment-only, which is sound for
*"did I break anything"* but says nothing about whether the shipped look is **acceptable**. So the
head went up: `ReferenceHead.olo`, GL **Deferred**, `olo_shader_errors` clean, `ScatterColor`
`(0.85, 0.55, 0.45)` at the shipped default.

Two changes to the fixture to make the effect legible, both ephemeral (MCP session state, nothing
saved): Suzanne scaled to **human head size** (~0.16 m, since the fixture's subject is ~8x human
scale and a 1.55 mm profile covers 8x less of it than it should), and the studio rig replaced by a
single close point light — **a bright small feature against dark skin**, which is the image #1361
asked to be judged on.

### The pass does what it says

Skin diffusion OFF vs ON at an identical pose:

| | |
|---|---|
| pixels differing > 2/255 | 35,371 |
| max difference | 49/255 |
| mean signed change where it differs | **R +1.84, G +0.30, B +0.06** |

Red spreads, blue stays put — the colour signature the feature exists for, on a real frame rather
than in a tap table. The difference concentrates at the brow ridge, the eye sockets and the
silhouette, which is where the irradiance gradient is.

### And the live frame confirms the decision this PR makes

The claim in `kSkinDiffusionSupportFraction`'s comment is that **a support radius carries no
energy**. That is visible directly. Measuring the ON−OFF difference in rings around the strongest
difference pixel, at the measured 3.41 px/mm of this capture:

| ring | mean \|difference\| |
|---|---|
| 0–1 mm | 18.3/255 |
| 1–2 mm | 13.9/255 |
| 2–4 mm | 11.1/255 |
| 4–8 mm | 7.3/255 |
| 8–15 mm | 3.4/255 |
| **15–23 mm** (the outer third of the support) | **2.6/255** |
| 23–40 mm (*beyond* the support — the noise floor) | 2.0/255 |

Red's Burley `d` is 1.55 mm; its 99.5% support is **23.3 mm**. By 15–23 mm the blur's contribution is
2.6/255 against a 2.0/255 floor measured outside the support entirely — the outermost third of the
support is already doing nothing you can see. Widening it to 27 mm extends a region that is at the
noise floor, which is the same conclusion the CPU measurement reached (0.004 of the feature's energy)
arrived at from the other end.

### What this does not show

I did not build an editor with a raised `kSkinDiffusionSupportFraction` to A/B it live. A
comment-only header change already forces a wide rebuild here, and the predicted difference is
0.004 of the feature's energy — well under 1/255, i.e. below this capture's own noise floor. The
falloff table above is the indirect form of that experiment and it is the one that was affordable.
Flagging it rather than implying the live A/B was run.

drsnuggles8:
## Scope change: this PR now also fixes #1368

The PR above argued #1361's two levers are individually worthless and should be left alone. That
conclusion stands and its tests stay. What follows is the pair that *does* work, shipped as
`SkinEvaluationModel::IsotropicGather = 6`.

### The coupling, which is why it is one version

At the default `ScatterColor`'s red channel (authored 0.85), halo of a bright small feature:

| | halo error |
|---|---|
| version 1 — separable projection + narrow Burley fit | 0.0558 |
| the disc alone, on the **uncorrected** profile | 0.0607 — *worse* |
| version 6's correction alone, through the **separable** pass | 0.1179 — *worse* |
| **version 6 — both** | **0.0186** |

Either lever alone is a regression. Shipping them as two versions would have shipped a version
nobody should author against, so they are one.

### Across the band

| authored | v1 | v6 | gain |
|---|---|---|---|
| 0.35 | 0.0224 | 0.0137 | 1.6× |
| 0.45 | 0.0374 | 0.0192 | 2.0× |
| 0.55 | 0.0481 | 0.0226 | 2.1× |
| 0.70 | 0.0419 | 0.0171 | 2.5× |
| **0.85** | **0.0558** | **0.0186** | **3.0×** |
| 0.95 | 0.1032 | 0.0327 | 3.2× |

**It is cheaper than what it replaces** — 25 taps in one pass against 17 along each of two. The
angle comes from the tap index (a golden-angle Vogel spiral) instead of the table, so a 2D kernel
ships through the same one-float-per-tap upload and no UBO is resized.

**The profile** is the same two exponentials with Burley's fixed 0.25 and 3 opened up, plus an
albedo-dependent scaling correction. Burley is exactly (0.25, 3, 1), so version 1 is the special
case, not a second code path. The correction is monotone in albedo deliberately: each channel
carries its own, so a kink would sit between red and green and read as a colour shift.

**No shipped content moves.** ADR 0024's rule: a version-1 head keeps the pixels it was signed off
with. Version 6 is *not* neutral on arrival, unlike versions 3–5 — there is no field to author, so
moving a profile there changes its look immediately, by design.

### Two corrections to what #1368 itself claims

- The **6.3×** headline assumed a *perfect* profile. With a realistic refit it is **3.0×** at the
  default. I will amend the issue.
- "Leaves 0.35–0.70 unchanged" is **wrong** as a criterion — the new mixture weight and rate ratio
  change the profile's shape everywhere. The measurement is better than the criterion asked for:
  every albedo in the band *improves*, 1.6–2.5× there. `TheGatherImprovesEveryAlbedoAndNeverDamages
  TheAccurateBand` asserts the sweep rather than the original wording.

### Verification

| | |
|---|---|
| `SkinDiffusionReference` | 13 tests — the coupling four-way, the band sweep, disc well-formedness |
| `SkinDiffusionEvidenceTest` | GL **Forward, Forward+, Deferred**, artifact-backed |
| whole skin surface | **229 pass**, Vulkan device-gated coverage exercised |
| `pre-commit` | clean |

The evidence test's fourth claim is the one a CPU test cannot make: a version-6 frame must **differ**
from a version-1 frame. Every reference number would be unchanged if the shader never took its
gather branch and quietly blurred along one axis — that assertion is what fails if
`u_SkinDiffusionSlots[slot].z` stops arriving, or if the vertical pass stops being an identity over
the disc.

**Live, GL and Vulkan**, human-scale head under a single close light: renders smooth, zero shader
errors on both, GL and Vulkan agree to within 0.5% on mean and standard deviation. The radial
luminance profile is monotone with no sawtooth — **no ring aliasing** from the disc, which is
exactly what the CPU metric is structurally blind to (a rotation maps every tap to the same radius,
so a radial measure cannot see angular sampling at all; the test says so in as many words).

The cumulative version lists this needed: **9 CPU sites and 8 GLSL ones**.

### Things I got wrong and caught

- `kSkinGatherTapCount = 32` did not fit the 25-entry tap table, behind a `static_assert` I had
  written to pass vacuously. 25 measures the same and is cheaper.
- The shader's tap count was pass-global; with a gather slot and a separable slot on screen together
  one of them would have looped the wrong bound. It is per-slot now.
- The disc well-formedness test asserted red reaches further than blue on a profile authored
  `vec3(albedo)` — all three channels the same medium, so it passed as an equality until it was run.
- The evidence test's pixel comparison used a floating-point threshold on 8-bit captures, which
  would have counted quantisation as a difference.

drsnuggles8:
@coderabbitai review

Issue #1323, opened 2026-09-19
Groom: route hair into the scene's shadow techniques, both directions


drsnuggles8:
## Parent

#1223

## What to build

Route grooms into the engine's shadow techniques in **both** directions, which is the half of
#1248's acceptance criterion 3 that PR #1322 did not land:

- **hair → body and scene.** A groom is not a shadow caster in any technique today, so a coat
  throws no shadow onto the body it grows on, onto the ground, or onto anything else.
- **body and scene → hair.** `GroomStrand.glsl` reads no cascade or atlas lookup, so a character
  standing in shade has a coat lit as if it were in the open.

#1248 shipped the coat's **internal** self-shadowing — a density volume containing the groom's own
strands and nothing else. That term is finished and is not what this issue is about.

## Why it was not part of #1248

Not an oversight, and not a scope trim for its own sake — a structural obstacle found while
implementing:

**The strand geometry is built inside `GroomRenderPass::Execute`, which runs *after*
`ShadowRenderPass`.** `ShadowRenderPass` has five caster families plus the GPU-driven virtual
geometry route, and a sixth is normally a POD list plus a depth shader — but the buffers a groom
caster would draw do not exist yet when the shadow pass runs. So this is a refactor of #1246's
per-(asset, settings) strand-geometry cache into something both passes can reach, not a new caster
list. That is a different shape of change from #1248's, and it lands in *shared* shadow code.

## Acceptance criteria

- [ ] Grooms are a caster family in **both** directional techniques — CSM cascades and the Virtual
      Shadow Map. `virtual-geometry-into-a-second-shadow-technique.md` is explicit that a family
      reaches a technique only if somebody wired it there, and that nothing detects the gap.
- [ ] Strands are widened to at least one shadow texel when rasterised from the light, for the same
      sub-pixel reason `groom-strand-visibility.md` rule 2 gives for the main pass. A 70 µm hair is
      far below a cascade texel; rasterised honestly it casts nothing.
- [ ] `GroomStrand.glsl` samples the scene shadow term, and **does not double-count the coat's own
      occlusion**. Once grooms are casters, a strand reading the cascade map is shadowed by its own
      coat twice — once in the map and once in the density volume. The intended fix is to offset the
      receiver to the coat's **light-exit point**, which `oloGroomCoatOpticalDepth` already computes
      as `tExit`; a strand has no surface normal, so the usual normal-offset bias does not apply.
[...]

PR #1369, opened 2026-09-19
fix(groom): coat transmittance is the mean of the transmittances, not the transmittance of the mean (#1360)

### LINKED ISSUE #1360: Groom coat transmittance is a Jensen lower bound: exp(-kappa E[N]) over-darkens a disordered coat

drsnuggles8:
`GroomCoatShadow::CoatTransmittance` returned `exp(-kappa * tau)` where `tau` is `E[N]`, the
**expected** fibre crossing count over a fragment's footprint. What the footprint receives is
`E[exp(-kappa N)]`, and Jensen's inequality puts the second at or above the first — so the shipped
form was a strict **lower bound on transmittance: it over-darkened**, by an amount set by the coat's
disorder rather than by anything anybody authored.

Modelling `N` as Poisson makes the exact mean that distribution's probability generating function at
`exp(-kappa)`:

```
E[exp(-kappa N)] = exp(-tau * (1 - exp(-kappa)))
```

one extra `exp` on a value already being exponentiated.

## The (a)-vs-(b) decision, as a number

The issue offered **(a)** the Poisson closed form, or **(b)** carrying the measured crossing
**variance** alongside the mean — which needs a second channel in the density volume *and* in the
deep opacity map. It asked for the choice to be made on measured evidence, and
`CoatTransportReference.TheGapClosesOnAnOrderedCoat` now measures both arms at matched optical
depth 1.2:

| arm | walk | **(a)** shipped | alternative |
|---|---|---|---|
| ordered lattice — *the arm (a) is wrong about* | 0.3012 | **0.0054** | (b) second cumulant: 0.0000 |
| disordered Poisson medium | 0.3304 | **0.0011** | retired form: 0.0292 |

**(a) costs 0.0054 on the arm it is wrong about to remove 0.0292 from the arm it is right about — a
5.4:1 trade, so (b)'s format change is not bought.** `TheGapClosesOnAnOrderedCoat` asserts that
inequality directly, so if a future coat representation makes it false, that case is what will say
so. On the disordered sweep the shipped form tracks the walk to **≤ 0.0019** where the retired one
was out by up to **0.0379**.

The residual on an ordered coat, stated because the issue asked for it: **0.0054 absolute, 1.8%
relative**. It shrinks as `kappa` falls, which is the direction a finer coat moves.

## The authored `kappa` moved 1.0 → 4.0

The same number now attenuates less — at `kappa = 1` a crossing costs `1 - 1/e = 0.63` of an e-fold
rather than a whole one — and the 6000-strand evidence coat went from **−510 380** summed luma to
**+22 955**, i.e. it stopped reading as shadowed at all. So the default and the sample scene were
re-judged against a measured sweep:

| kappa | sparse (1500) | dense (6000) |
|---|---|---|
| 1.0 | +495 375 | **+22 955** |
| 2.0 | +418 935 | −319 523 |
| **4.0** | **+380 857** | **−487 404** |
| 6.0 | +375 833 | −509 840 |
| 16.0 | +375 053 | −513 403 |

4.0 is the knee: it buys **94.9%** of all the darkening there is, 6.0 buys 99.3%, and 16.0 adds
0.7% more — a slider with nothing left in it. The saturation is the model's, not the renderer's:
transmittance **floors at `exp(-tau)`**, because a coat of opaque fibres still passes light wherever
the footprint crossed nothing, which for a Poisson `N` is exactly `P(N = 0)`.

At the re-authored `kappa` the committed evidence captures land **within 2/255 per pixel** of their
previous appearance (+25 999 luma, **zero** pixels over the 8/255 threshold). The transport is
corrected without the coat's look moving.

**The sparse arm is permanently brighter** (+380 857 against +338 020) and cannot be tuned back — it
floors at +375 053, because a sparse coat is the most disordered relative to its own mean and gains
the most from the correction. That is the fix landing, not a tuning failure.

**A scene authored before this keeps its own value and renders brighter.** That is the correction
reaching it; a dense coat authored at 1.0 wants re-judging.

## The parity test this twin's own header said it was missing

`GroomCoatShadowCommon.glsl` stated in its header that there was no GPU parity test for this pair.
A formula changed twice by hand is exactly the one that lands on one side only, and that failure is
**silent** — a coat shaded with the old form still looks like a coat, every CPU-side number in the
suite keeps agreeing with itself because they all come from the C++ arm, and the visual-evidence A/B
compares the coat against itself with the feature off, so it cannot see the two arms drifting the
same way.

`GroomCoatTransmittanceParityTest` compiles the real shader and compares it against the C++ twin
over the whole authored `kappa` range plus the non-finite case, with a companion test asserting the
grid actually **reaches** the region where the two candidate formulas disagree — without which the
comparison could be two formulas that happen to agree everywhere it looks. The march
(`oloGroomCoatOpticalDepth` vs `SampleDensityVolume`) is still unguarded, and the header now says
which half is covered rather than implying both.

## A bug found on the way — separate commit

`006763de0`: `oloGroomCoatTransmittance` guarded with `!(opticalDepth > 0.0)`, which catches NaN but
**not `+infinity`** — the one non-finite value that passes `> 0.0`. It fell through to `exp(-inf)`
and returned a **black** coat, breaking rule 10 of `groom-coat-self-shadowing.md` in the direction
the rule exists to prevent, and disagreeing with the C++ twin on exactly the input that matters
most.

## Verification matrix

Complete — 6 cells from {GL, Vulkan} × {Forward, Forward+, Deferred}, plus the conditional axes.

| backend | path | kind | evidence | result |
|---|---|---|---|---|
| GL | Forward | artefact | `GroomCoatShadow_GL_Forward.png` | 21 297 px, max 105/255, luma −487 404 |
| GL | Forward+ | artefact | `GroomCoatShadow_GL_ForwardPlus.png` | 21 297 px, max 105/255, luma −487 404 |
| GL | Deferred | artefact | `GroomCoatShadow_GL_Deferred.png` | 21 297 px, max 105/255, luma −487 404 |
| GL | all three | live | master-vs-branch A/B, 3 camera angles | 116–174 px, max 41–74/255, luma +10 903…+54 212 |
| Vulkan | Forward | live | A/B + log: **0 errors, 0 VUIDs** | 186 px, max 49/255, luma +56 624 |
| Vulkan | Forward+ | live | A/B + log: **0 errors, 0 VUIDs** | 186 px, max 49/255, luma +56 624 |
| Vulkan | Deferred | live | A/B + log: **0 errors, 0 VUIDs** | 201 px, max 52/255, luma +73 374 |

Conditional axes, each answered:

| axis | cell | evidence |
|---|---|---|
| CPU/GPU parity | **run** | `GroomCoatTransmittanceParity.*`, 4096 grid points + the infinity case |
| MSAA | **run**, one confirmation | 1×: 21 297 px / −487 404; 4×: 21 298 px / −487 433; ratio 1.000 |
| Resolution | **run**, one confirmation | native −487 404, scaled −254 398, resolution in force 64/64 |
| Ordered vs disordered coat | **run** | `TheGapClosesOnAnOrderedCoat`, both arms kept |
| Prior on-disk version | **not reachable** | (a) changes no serialized data; the `kappa` *value* moved but its type, range and wire position did not, and the save-game band has no version gate to advance |

Both Vulkan arms were captured after clearing `%LOCALAPPDATA%\OloEngine\ShaderCache\*\GroomStrand*`
with the recompile confirmed in the log, since the cache key excludes `#include`s. Vulkan Forward
and Forward+ are byte-identical: this scene has one directional light, so Forward+'s tile culling
has nothing to cull. The `deferred` capture does differ and the setting reported
`changed: true` / `previousValue: forward`, so the switch took.

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Scene/Components.h` — the `m_Kappa` default moving 1.0 → 4.0 is the
   part with real consequences for anyone's existing content. The formula change is arithmetic;
   this is a judgement, and it is the one I would most want a second opinion on.
2. `OloEditor/assets/shaders/include/GroomCoatShadowCommon.glsl:268` vs
   `OloEngine/src/OloEngine/Groom/GroomCoatShadow.cpp:689` — the two must stay one expression. The
   C++ arm uses `expm1`, GLSL spells it `1 - exp(-kappa)`; that difference is deliberate and is
   below the f32 epsilon of the `exp()` it feeds, but it is the kind of thing worth a second look.
3. `GroomCoatTransmittanceParityTest.cpp` — a new GPU test. If the grid or the tolerance is wrong,
   it passes while proving nothing, which is why the discrimination companion exists.

**What I verified, and how** — `CoatTransportReference.*` (4), `GroomCoatShadow*` +
`GroomCoatTransmittance*` + save-load (66 total) all green; the twelve matrix cells above; a
measured `kappa` sweep for the re-authoring rather than a guess; before/after pixel diffs of the
committed evidence captures and of live GL and Vulkan frames. The check that would have failed if
the formula were wrong on one side is `GroomCoatTransmittanceParity.TheCompiledShaderMatchesTheCppTwin`;
the one that would have failed if `kappa` were left at 1.0 is
`ADenserCoatIsDarkenedMoreThanASparseOne`, and it did, which is how the re-authoring got measured.

**Least confident about** — the choice of 4.0 rather than 6.0. 6.0 reproduces the old dense-coat
darkening almost exactly (−509 840 against −510 380); I picked 4.0 because past ~6 the slider is
dead, and because reproducing the old pixels is not actually the goal when the old pixels came from
an over-darkening formula. That reasoning is defensible but it is a taste call sitting on top of a
measurement, and someone who looks at the coats may disagree.

**Deliberately not tested** — the march half of the CPU/GPU twin still has no parity probe; it needs
a 3D volume uploaded and read back identically on both sides, which is a larger harness than this
slice and is recorded in the guide rather than silently skipped. `docs/agent-rules/groom-coat-self-shadowing.md`
is now 12.1 KB against the repo's ~10 KB guideline — it was already 10.6 KB before this, and
splitting it would collide with `feature/groom-guide-simulation-1250`, which is live in the same
directory.

## Note for #1323

That issue's double-count fix offsets the receiver to the coat's light-exit point using `tExit`,
which `oloGroomCoatOpticalDepth` computes. **This change does not touch the march or `tExit`** — it
is confined to `oloGroomCoatTransmittance`, the function that consumes `tau` afterwards. So #1323
starts no harder than it would have. The one thing that changed for it: the transmittance a strand
reads is now brighter for the same `tau`, so a double-count there is slightly less visible than it
was, not more.

Closes #1360

## Summary by CodeRabbit

- **New Features**
  - Updated groom coat self-shadowing to use a more physically accurate transmittance model, reducing excessive darkening and preserving light through gaps between fibres.
  - New default fibre opacity value improves coat shadow appearance for newly created components.

- **Bug Fixes**
  - Existing scenes using authored opacity values now render with the corrected transmittance behavior.
  - Clarified the “Fibre opacity” setting tooltip to explain light transmission per fibre crossing.

- **Documentation**
  - Updated groom self-shadowing guidance and reference material to reflect the revised model.


drsnuggles8:
## 🤖 Self-review @ `2ed700f84bca82f7a025efa0d212dfcb907af005`
Reviewed the PR diff at **high** effort before opening, per the task loop's Phase 3.

- **Findings:** 6 · **Fixed:** 5
  - two `restore the default` fallbacks still handed out the retired 1.0 (`MakeGroomCoatKappa`, the save-game sanitiser) — both now read a default-constructed component
  - `GroomCoatShadow.h`'s file header still stated `exp(-kappa * tau)` as shipping
  - three sites in `MaterialReference.h` still described the retired form as the production answer
  - the parity test's infinity arm compared against a literal `1.0f`, leaving the input that file calls most important as the one input the twins were not compared on
  - `kKappa`'s comment in the property tests carried the retired aggregate semantics
- **Dismissed:** 1 — the comparison suite's absolute thresholds were flagged as possibly needing re-derivation against a brighter truth. All nine `GroomCoatShadowComparison` cases run green post-change, so they are left as authored rather than loosened.

One finding was not a defect: `GroomStrandRequest::CoatKappa` is a third `1.0f` and deliberately stays one — a default-constructed *request* should not claim a coat four times as opaque as the component it never read. Documented rather than left looking stale.

drsnuggles8:
## CI status: 29/30 green, one job blocked on a runner outage

**`ASan + LSan (Linux/Clang)` is not a failure of this change.** Attempt 1 died mid-compile and the re-run (attempt 2) has now been queued 90 minutes with no runner.

Evidence it is infrastructure:

- Attempt 1's log ends mid-compilation of `Networking/Replication/SnapshotInterpolator.cpp` (an EnTT template instantiation, untouched by this PR) with **no error line and no `##[error]`**.
- Its step conclusions show `Build tests` stuck at `in_progress` with `conclusion: null` and every later step `pending` — a runner drop, not a compile failure. The `sccache` restore steps were skipped, so it was a cold build.
- All three self-hosted runners currently read `offline`:

  ```
  olo-ci-1: offline busy=false
  olo-ci-2: offline busy=false
  olo-gpu-amd: offline busy=false
  ```

That is the documented signature of the shared `user@1004.service` slice going down and taking every Linux sanitizer arm with it. Recovery is `systemctl restart user@1004.service` on the box, which needs root.

**No code change was made in response**, per the repo's rule that an infra flake earns a re-run and nothing else. The other two Linux sanitizers — **TSan and UBSan, including all 4+4 test shards and both shard-coverage gates** — went green on this same commit before the runners dropped, so the Linux sanitizer tier is not unexercised.

Everything else is green: Windows `build` + all 3 test shards + shard coverage, ASan (Windows) build + all 4 shards + coverage, pre-commit, shader toolchain floor, Steam stub SDK, and the Steam-seam gates.

Issue #1243, opened 2026-09-13
Skin: layered specular, pore filtering and expression-driven detail


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1222

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Give a close-up animated head controlled skin surface lobes, mapped roughness/pore detail and expression-driven wrinkle/detail blending.

## Acceptance criteria

- [ ] Authorable specular/roughness and detail controls render consistent energy and do not erase the underlying diffusion.
- [ ] Pore/normal filtering avoids sparkling under camera motion and distance changes.
- [ ] Expression-driven detail blends deterministically with existing morph/animation state and emits valid history changes.
- [ ] Demonstrate several skin tones under soft, hard and grazing light with baseline/detail-only AOV comparisons.

## Scope boundaries

Begin with a measured lobe/detail comparison. This is material response and animation linkage, not a facial-rig authoring tool.

## Blocked by

- #1241
- #1227

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 5
craft: 13
stability: 3
decay: 1
effort: 5
confidence: 0.5
learning: 8
fun: 13
kano: performance
blocked_by: [1241]
blocks: [1222, 1245, 1256]
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

Issue #1242, opened 2026-09-13
Skin: thickness-aware transmission for backlit ears and thin regions


drsnuggles8:
> *This was generated by AI during triage.*

## Parent

#1222

## Agent Brief

**Category:** enhancement

**Type:** AFK (dependency-gated)

## What to build

Add authored thickness and a shadow-aware thin-region transmission term to skin, demonstrated by a backlit animated ear/head.

## Acceptance criteria

- [ ] Thickness source, units, import/cook and editor controls are explicit and survive round-trip.
- [ ] Front lighting, backlighting and occlusion vary the contribution correctly without turning the head uniformly emissive.
- [ ] Transmission and diffusion do not double-count the same energy; missing thickness has a documented conservative fallback.
- [ ] Animation, shadow mode changes and different skin tones retain plausible thin-region colour and stable history.

## Scope boundaries

Keep the work within the described feature; adjacent roadmap children own the remaining capabilities.

## Blocked by

- #1241

## Delivery and validation contract

- This is an AFK implementation brief after its listed blockers land. Recheck the relevant code at implementation time; baseline inspection was master `e86a630d6` on 2026-09-13, not a claim based on historical issue checkboxes.
- Deliver the narrow feature through editor/asset/runtime integration where applicable, including persistence/cooking, finite input validation, resource lifetime and diagnostics. Do not leave an isolated shader or unused data structure as completion.
- Pin important math/state contracts with meaningful tests, capture and inspect multiple views plus motion, and run the real editor with shader/render validation. Exercise OpenGL/Vulkan and forward/forward+/deferred where the feature applies; Vulkan RT has explicit raster fallback. Report exact test totals/exits and material skips.
- Record image quality, per-pass cost and memory on named hardware. Initial benchmark budgets are measurements to establish, not permission to reduce the AAA ambition. If a confidence-0.5 experiment rejects an approach, report the evidence and rescope before production implementation.

## Score
```olo-score
capability: 5
craft: 8
stability: 3
decay: 1
effort: 3
confidence: 0.8
learning: 5
fun: 8
kano: performance
blocked_by: [1241]
blocks: [1222, 1245]
```
<sub>Rated per [issue-scoring](https://github.com/drsnuggles8/OloEngineBase/blob/master/docs/process/issue-scoring.md) · score = confidence × (capability + craft + stability + decay) / effort, derived by the picker.</sub>

drsnuggles8:
[...]

PR #1377, opened 2026-09-20
fix(skin): close #1368 — the version-6 list the inspector never got, and the tripwire that never fired

### LINKED ISSUE #1368: Skin: the separable projection and the Burley fit are coupled — fixing either alone is worthless, fixing both is 6.3x
Follow-up to #1361, which measured the skin diffusion pass's three approximations and decided to
leave two of them alone. This is the third one, and the finding that makes it worth doing.

## The two errors are COUPLED, and that is why #1361 found nothing worth fixing

#1255 measured two errors in the pass and #1361 asked which to repair. The answer was "neither",
because repairing either **alone** is no better than shipping as-is. Extending that measurement to a
2D gather shows why — and shows that the pair together is worth a great deal.

All figures are the worst disagreement with a Monte Carlo searchlight walk over a bright small
feature's halo, read at 2/3/5/8/12/18/25 mm, as a fraction of the feature's own energy. The same
metric `SkinDiffusionReference` uses.

| authored `ScatterColor` | shipped (separable + Burley fit) | 2D gather + fit | separable + transport profile | **2D gather + transport profile** |
|---|---|---|---|---|
| 0.55 | 0.0469 | 0.0106 | 0.0420 | **0.0049** |
| **0.85** — the default's red channel | **0.0559** | 0.0593 | 0.0779 | **0.0088** |
| 0.95 | 0.1043 | 0.1011 | 0.0807 | **0.0211** |

Read the 0.85 row, which is the one that ships. Going 2D alone: **0.0593, worse than doing nothing.**
Fixing the profile alone: **0.0779, worse still.** Doing both: **0.0088 — a 6.3x improvement.**

The mechanism is that the two errors have opposite signs. The separable projection over-spreads a
point response along the axes; the narrow fit and the truncated support pull energy back in. Remove
one and the other is exposed. That is also exactly why #1361's two candidate levers each measured as
worthless, and it is not a fact either #1255 or #1361 had.

Note the 0.55 row: where the fit is already accurate, going 2D alone is worth **4.4x** on its own.
The coupling is specifically a high-albedo problem, which is to say it is a *skin* problem.

## What NOT to build, measured

**A second separable pass on the 45° diagonal buys nothing.** This is the obvious cheap approximation
to isotropy and it cannot work here: rotating a separable kernel maps every tap pair to the **same
radius**, so its radial response is identical and averaging the two changes the radial CDF not at
all. It would reduce the star-shaped *angular* artifact, which is a different defect and not the one
measured above. Worth writing down before someone spends a sprint on it.

## Scope

Two halves, and the issue is only worth doing as both — that is the whole finding:

1. **An isotropic gather** in place of the two separable passes. A Burley-importance-sampled disc is
   the standard form. The tap budget question is real and open: the shipped pass is 2N fetches for
   N=17, and a gather is not.
2. **A profile that holds at high albedo.** The Burley searchlight fit runs 35% narrow at the q90
   radius at an authored 0.85 (#1255). A scale correction on `d` recovers most of it — the optimum
   is 1.06x at 0.85 and 1.14x at 0.95 against this metric, and **1.00x at every albedo at or below
   0.70**, so any correction must be albedo-dependent or it will damage the band that is currently
   accurate.

Half 2 changes what existing profiles render, so it needs the `SkinEvaluationModel` treatment — and
those version lists are cumulative across 5+ explicit CPU lists plus the GLSL ones, per
`docs/agent-rules/component-serializer-codegen.md` and the warnings in `SkinDiffusion.cpp`.

## Acceptance criteria

- [ ] Decide the gather form and **state its tap budget against the current 2x17 fetches**, measured,
      not estimated. A 6.3x accuracy win that costs 3x the frame time is a different trade and the
      issue should say which one it made.
- [ ] `SkinDiffusionReference` gains the coupled measurement: both halves, and the two single-lever
      arms that must stay worse. `TheSeparableProjectionOutweighsBothErrorsAtASmallBrightFeature`
      is written to FAIL when the projection stops being dominant — it is the tripwire for this
      work, and closing this issue means deleting or inverting it.
- [ ] Judged on a bright small feature against dark skin, per #1361, not on a terminator.
- [ ] The albedo-dependent correction leaves 0.35–0.70 unchanged, asserted.
- [ ] If half 2 lands, a new `SkinEvaluationModel` with every cumulative version list edited.

## Evidence

- #1361 / PR #1366 — the three-way measurement, the live editor confirmation, and the decision to
  leave the single levers alone.
- `SkinDiffusionReference` in `OloEngine/tests/Rendering/PathTracing/NonlocalTransportReferenceTest.cpp`.
- `docs/agent-rules/notes-renderer.md` → *Measure which error term dominates before repairing either
  of the others*.

## Score
```olo-score
capability: 5
craft: 8
stability: 1
decay: 1
effort: 5
confidence: 0.7
learning: 8
fun: 7
kano: performance
blocked_by: []
blocks: []
```

drsnuggles8:
## What this is, and what it is not

**#1368's feature already shipped.** Transport version 6 — the isotropic gather on a refitted
profile — landed on master in `926f6bff0` + `b081e8cc0`, carried in on PR #1366. The HANDOVER for
this task said "nothing done yet"; that was wrong, and rebuilding the feature would have been
manufactured work.

So this PR audits #1368's acceptance criteria against what actually shipped and closes the three
gaps it found. One of them is a live bug.

## 1. A missed cumulative version list — the bug

#1368 updated **eight** of the nine CPU version lists and all eight GLSL ones. The ninth is the
editor's Skin Profile inspector, and it has two blocks. A version-6 head therefore told its author
two false things:

- **"Not diffused — this profile is authored at transport version 0"**, while the head diffused on
  screen with the new gather;
- **nothing at all about transmission** — that block does not say the wrong thing, it *disappears*,
  so a version-6 ear that is not glowing offered no readout to explain why.

Silent by construction: `OloEditor` is not in the test build, so no evidence test can reach either
block. That is also why this is the **third** time this readout has gone stale one version behind
the renderer (#1242 and #1243 hit the same class of miss on the kernel builder's copy).

**The fix is not a tenth copy.** The two lists are now one named predicate each —
`SkinEvaluatesScreenSpaceDiffusion` and `SkinEvaluatesThicknessTransmission` — and the nine call
sites ask them. The lists stay *explicit* rather than becoming `>= ScreenSpaceDiffusion`: a version
that does not diffuse must not inherit diffusion silently, which is the property the copies were
paying for. There is now one of each to forget instead of nine.

## 2. #1361's tripwire was never retired — and never fired

Acceptance criterion 2 says closing #1368 means deleting or inverting
`TheSeparableProjectionOutweighsBothErrorsAtASmallBrightFeature`. It was neither, and the reason is
the interesting part: **its own comment promised it would fail** when a gather shipped, and it
stayed green, because it built its kernel from a version-1 profile and went on measuring version 1
after version 6 existed. *A guard that names a fixed enum version cannot notice a new one.*

Inverted rather than deleted, because its control — the walk's own profile at full support through a
separable pass — is the only thing that can separate "our profile was wrong" from "our *projection*
was wrong", which is the whole of #1368. Measured at authored 0.85:

| arm | halo error |
|---|---|
| version 1 | 0.0558 |
| a **perfect** profile through the separable pass | 0.0822 |
| **version 6** | **0.0186** |

That ordering is the strongest form of the claim. "Version 6 beats version 1" is measured next door
at every albedo, but version 6 changed two things and that comparison cannot say which one paid.
This one can: the control has an exact profile and a separable projection, version 6 has an
approximate profile and no projection, and version 6 wins by 4.4x.

## 3. A dangling doc pointer

`skin-diffusion.md` said "**version 6 fixes both** — see below" and there was no below. There is
now: the band table, the coupling at 0.85, the tap budget (25 fetches in one pass against 17 along
each of two — a cost **cut**), and the 45° diagonal that was measured and rejected.
`notes-renderer.md`'s "measure which error term dominates" section ended at #1361's "repair
neither"; it now carries the sequel and this PR's two lessons.

## Verification matrix

Complete — every cell the task's axes produce. The engine change is a behaviour-preserving
refactor, so the artefact-backed cells are asserted as **byte-identical**, which is a stronger
result than "passes".

| axis | cell | kind | evidence | result |
|---|---|---|---|---|
| Backend x path | GL Forward | artefact | `SkinDiffusion_GL_Forward{,_Gather}.png` | rewritten by the run, **0 bytes changed** |
| Backend x path | GL Forward+ | artefact | `SkinDiffusion_GL_ForwardPlus{,_Gather}.png` | rewritten, **0 bytes changed** |
| Backend x path | GL Deferred | artefact | `SkinDiffusion_GL_Deferred{,_Gather,_Separable}.png` | rewritten, **0 bytes changed** |
| Backend x path | Vulkan, all paths | live | editor `-Rhi vulkan`; log `[RHI] Backend: Vulkan (source: --rhi flag)` | v6 head renders; **0 `[error]`, 0 VUID**; FPS 59 |
| Lighting path | Forward / Forward+ reachability | artefact | the three GL rows above | the pass runs on all three; not deferred-only |
| MSAA | on / off | — | **Not a cell.** This PR adds no shader, no GLSL and no render-pass code; the diff is a CPU predicate, an ImGui readout, tests and docs. Nothing reads a resolved or per-sample target. |
| Upscale | on / off | — | **Not a cell**, same reason — no screen-space radius is computed or changed here. |
| Resolution | >=2 resolutions | — | **Not a cell**, same reason. |
| Prior on-disk version | v1..v5 still load and render | artefact | 60 CPU tests + the 7 `SkinDiffusionScene` captures, all byte-identical | unchanged |
| Numeric | reference table at 0.55 / 0.85 / 0.95 | artefact | `SkinDiffusionReference`, 73 tests | pass; numbers above |
| Albedo band | 0.35–0.70 not damaged | artefact | `TheGatherImprovesEveryAlbedoAndNeverDamagesTheAccurateBand` | pass (1.6x–3.2x, every albedo improves) |
| Editor inspector | v6 readout | live | screenshot, GL, `ReferenceHead.oloskin` temporarily at v6 | `Diffusion reach (mm, derived): 38.51` and the transmission block both render |

The inspector's **before** state is not a screenshot: it is the deleted `else` branch, visible in
the diff, which printed *"Not diffused — this profile is authored at transport version 0."* I did
not rebuild the old editor to photograph a branch whose text is in the diff.

## Review guide

**Where I'd look hardest**

1. `OloEngine/src/OloEngine/Renderer/SkinDiffusion.h` + `SkinTransmission.h` — the two predicates
   are now the single source of truth for nine call sites. If either list is wrong, it is wrong
   *everywhere at once*. I checked them set-for-set against what they replaced; a second pair of
   eyes on the two sets is the highest-value thing in this PR.
2. `OloEngine/tests/SkinTransmissionTest.cpp` — the energy-bound tripwire. It took **three**
   attempts to make it capable of failing (see the third commit message), and its correctness now
   depends on a thickness chosen for margin (0.5 mm, transmittance 0.62). That is a number I
   derived; if it is wrong, the guard silently reverts to decoration.
3. `NonlocalTransportReferenceTest.cpp` — the inverted tripwire's 0.01 slack against a stochastic
   walk. The measured margin is 0.064, so it is not close, but the walk is reseeded per machine.

**What I verified, and how** — 73 `SkinDiffusionReference` + 60 skin CPU tests + 7
`SkinDiffusionScene` evidence captures + 11 parity/evidence tests, all green. The refactor is
proven pixel-neutral by the evidence PNGs coming back **byte-identical** after being rewritten.
Both new classification guards were **falsified on purpose** — each predicate was broken and each
count failed ("5 of 7", "4 of 7") — and, importantly, that falsification is what showed the loop
arms cannot fail, which the third commit then corrected in the comments. The bound tripwire was
falsified by raising `kMaxSkinTransmissionStrength` to 4 and watching it fail at 1.2309 vs 0.5.

**Least confident about** — whether inverting the tripwire was the right call against deleting it.
The criterion allows either. I kept it because its control is expensive machinery that uniquely
separates the two error terms, but it is now a test that asserts a historical fact about version 1
alongside a live one about version 6, and a reviewer could reasonably say the historical half
belongs in the guide rather than in an assertion.

**Deliberately not tested** — MSAA, upscale and non-native resolution, for the reason in the matrix
(no shader or pass code in the diff). I also did not re-measure version 6's *frame time* against
version 1: acceptance criterion 1 asks for a measured tap budget, and the shipped answer is a fetch
count (25 in one pass vs 34 in two) rather than a timing. Two sibling worktrees were building
throughout this session and GPU timings on this box swing up to 4x under that load, so a timing I
took today would have been worse evidence than the fetch count, not better.

**Note on tooling** — the `sonarqube` MCP server failed to connect this session
(`CONNECTION_CLOSED`), so `sonarqube-review-alignment.md`'s pre-review check could not be run.

Closes #1368

## Summary by CodeRabbit

* **New Features**
  * Added corrected version 6 isotropic-gather skin diffusion for improved approximation accuracy and reduced rendering cost.
  * Skin materials now consistently support diffusion and thin-region transmission across supported transport versions.

* **Bug Fixes**
  * Updated skin-profile diagnostics to accurately identify diffusion and transmission capabilities.

* **Documentation**
  * Added guidance and measurements for version 6 skin diffusion, including usage details and known trade-offs.


drsnuggles8:
## 🤖 Self-review @ `23e63178dac8db1a27e5c18c183982f0ad6c51ea`
Reviewed the PR diff at **high** effort (`/code-review high`), before the PR existed.

- **Findings:** 5 · **Fixed:** all 5, in `23e63178d`.
  1. `SkinTransmissionTest` energy-bound assertion was **vacuous** (`term <= bound` with albedo 1.0 is true by construction) — replaced with a dead-branch tripwire, then falsified twice more before it could actually fail. Details in the commit message.
  2. + 3. Both `EveryTransportVersionIsClassified…` **loop arms are circular** — the engine asks the same predicate, so they cannot fail. Comments corrected to name the arm with teeth (the count), with the falsification numbers.
  4. MCP diffusing-profile note still steered authors to `EvaluationModel 1` while this PR calls version 6 the corrected transport.
  5. Superseded comment paragraphs in both editor files, one still telling the next author to grow a list that no longer exists there.
- **Dismissed:** none.

Worth flagging for a human reviewer: finding 1 is the kind of defect this PR is *about* — a guard that reads as a guard and cannot fail — and I shipped one while fixing three others. It took three attempts and an on-purpose sabotage to make it real.

PR #1378, opened 2026-09-20
feat(groom): distance LOD with cooked card levels and coverage compensation (#1252)

### LINKED ISSUE #1252: Groom: strand/card/mesh LOD with stable coverage and bounded work

## Parent

#1223

## Agent Brief

**Category:** enhancement

## What to build

Cook and select distance-appropriate groom representations with independent visibility, simulation and shadow budgets, demonstrated on several animals at mixed distances.

## Acceptance criteria

- [ ] LOD assets and selection thresholds are authored/cooked; preserve silhouette, apparent density, colour and highlight response across transitions.
- [ ] Strand count/width compensation and alternate representations are measured in motion; hysteresis prevents thrashing.
- [ ] Simulation/interpolation/shadow work scales down independently without visible snapping or missing coats.
- [ ] Capability fallbacks and residency requirements are explicit; report GPU cost and memory by animal/group/representation.

## Scope boundaries

Compare representations first; shell techniques may be evaluated for suitable short coats, but do not replace the close-up strand target.

## Blocked by

- #1250
- #1251

## Score
```olo-score
capability: 8
craft: 13
stability: 3
decay: 1
effort: 5
confidence: 0.5
learning: 8
fun: 13
kano: performance
blocked_by: [1250, 1251]
blocks: [1223, 1253, 1256, 1257, 1258]
```

drsnuggles8:
Closes #1252.

A groom now picks its representation and its three budgets from apparent size, and keeps its
apparent density while doing it.

The issue carries `confidence: 0.5`, which `issue-scoring.md` reads as *spike first*. The spike ran
first, and it **changed the design three times**. Every number below is an assertion in
`GroomLodComparisonTest`, and they are written up in
[`docs/analysis/groom-representation-lod-1252.md`](../blob/feature/groom-representation-lod-1252/docs/analysis/groom-representation-lod-1252.md).

---

## What the measurement decided

**1. The width compensation is `1/k`, linear — not `1/sqrt(k)`.** A strand is a *band* whose covered
area is length × width, so a coat's area goes as `count × width`. `FoliageLod` uses `1/sqrt(k)`
because a foliage instance is a *sprite*; borrowing that here under-compensates a sixteenth-density
coat by 4×. Measured on `human-scalp`, geometry area ratio against the full coat:

| Thinning | 256 px | 128 px | 64 px | 32 px |
|---|---|---|---|---|
| ⅛, **uncompensated** | 0.207 | 0.210 | 0.216 | 0.218 |
| ⅛, **compensated** | 0.996 | 1.003 | 0.996 | 1.053 |

**2. A mean-centreline card — the textbook construction — lost on every coat at every distance.**
Averaging curves that diverge produces a shorter, straighter curve, so it loses exactly the spread
that gives a tuft its silhouette. On the short coat it lost **45 % of the covered area outright**
(0.548 against a kept member's 0.963). The shipped construction keeps the cluster's most central
*real* strand and scales its widths to the cluster total; `GroomCardAggregation::MeanCentreline`
survives as the measured-and-rejected alternative, the way `AlphaToCoverage` does for #1246.

**3. Cards only earn their cook ABOVE the compensation cap.** The runtime stride can thin by `k` and
widen by `1/k` for free, and at a modest reduction it ties the card within noise. But the runtime
widening is capped at 8× (a strand widened sixty times is a flat band, not a fibre), and a card's
width is *baked*, so the cap does not apply to it:

| Coat @ 64 px, coarse cell | Reduction | Card area | Stride area | Card mean&#124;e&#124; | Stride mean&#124;e&#124; |
|---|---|---|---|---|---|
| `human-scalp` | 58.8× | **1.007** | 0.234 | **0.0740** | 0.2096 |
| `short-coat` | 25.0× | **0.964** | 0.409 | **0.2059** | 0.2545 |
| `long-coat` | 16.7× | **1.006** | 0.618 | **0.0889** | 0.1879 |

**4. The shell ("mesh") tier is refused, on measurement.** A shell claims coverage 1 inside the
silhouette, so it is only honest where the coat has saturated. Across three coats and six distances
down to 4 px the highest *solid* fraction anywhere was **0.064**, and mean coverage never exceeded
0.50 — these coats get *less* solid with distance. `GroomLodFallbackReason::MeshTierNotSelected`
reports it by name, and `TheShellTierIsMeasuredBeforeItIsRefused` **fails** if a future coat ever
saturates, which is the correct signal to revisit. The scope boundary asked for shell techniques to
be evaluated and not to replace the close-up strand target: they were, none is suitable, and the
strand target is untouched.

**0. The first cook measured the fixture, not the representation.** `GroomStrandFixture` records an
unwrapped `phi / 2π` as a root U, reaching **11 459**; `GroomCoatClumpCell` *clamps* a UV to ±16, so
every strand past the clamp landed in one cell. The first card cook produced 64 cards for 20 000
strands carrying **16 %** of the coat's coverage, and it looked like a working level.
`GroomLodBuilder` now refuses such a groom by name.

---

## What shipped

- **`Groom/GroomLod.h`** — the ladder as a value: pure, constexpr, backend-neutral, from apparent
  size plus what the cook produced to the tier and the first reason it is not the selected one.
  Three independent budget curves **and three independent stability counters** (criterion 3 — one
  shared counter recouples them through the hysteresis).
- **Two anti-thrash mechanisms with two different bounds.** The threshold slide gives a dead band
  `2h × threshold` wide, so a camera travelling inside it causes at most one transition *ever*; the
  frame hold means a camera alternating every frame never coarsens at all. `FoliageLod` only has the
  first, because a plant carries no per-instance state — a groom is an entity and its LOD state
  lives in `Scene`, keyed by UUID, because a pass runs once per *camera* and a hysteresis must
  advance once per *frame*.
- **Cooked card levels**, section 10 of the `.ologroom` (format **v3**), carrying the map back to
  the base curve each card stands in for — without it a coat at card range stops following its body
  and stops moving.
- **One shading path.** A card is a cooked *curve*, so the fibre BCSDF (#1247), the coat shadow
  volume (#1248), the per-strand tint (#1251) and the guide simulation (#1250) all keep working
  across a transition. Criterion 1's colour/highlight preservation is true by construction.
- `GroomLodComponent` with a full inspector (authored knobs + a live readout of tier, strands and
  memory *by representation*, and the dominant fallback reason), scene YAML, save-game and MCP
  field registry.

**⚠️ Format v3 moves `MinSupportedVersion` with `CurrentVersion`, so every `.ologroom` on disk must
be re-imported.** The failure is a readable version error, pinned by a test. The four committed
reference grooms were regenerated in this PR.

---

## Verification

| Axis | Cell | Result |
|---|---|---|
| Backend | OpenGL | ✅ `GroomLodVisualEvidenceTest`, 10 committed PNGs |
| Backend | **Vulkan** | ✅ live editor, `"backend": "vulkan"`, both tiers built, 0 shader errors, 0 VUIDs |
| Lighting path | Forward / Forward+ / Deferred | ✅ all three assert the same ratios (1.202 coverage at 1250 vs 12000 strands) |
| MSAA | 4 / 1 | ✅ live; G-Buffer confirmed recreated `x4MSAA` in the log |
| Upscale | quality / off | ⚠️ setting reports `changed: true`; **no corroborating log line** — the known `olo_renderer_settings_set` caveat. Coat renders correctly either way |
| Resolution | 256², 1280×720, 1411×942 live | ✅ thresholds key on render height via `Renderer3D::GetLODViewParams()` |
| **Motion** | camera across every threshold, both directions | ✅ 40 alternating frames → **0** representation changes; slow sweep out → **1**, back → **1** |
| **Prior on-disk version** | v2 `.ologroom` | ✅ refused by version with a readable reason; reference assets regenerated |
| Vulkan RT | raster fallback | untouched — the strand pass is forward-style on every path |

**Tests:** 373 `Groom*` pass, 1167 pass across `SaveGame|Scene|Serializ|BitwiseEqual|Component`.
Exact totals: `[==========] 373 tests from 48 test suites ran. [ PASSED ] 373 tests.`

**Evidence PNGs** (`OloEditor/assets/tests/visual/GroomLod*`) — the compensation A/B is the one to
look at: identical geometry (1500 strands both arms), 13 580 coat pixels uncompensated against
16 208 compensated, and the uncompensated frame is visibly stringy where the compensated one is a
full coat.

**Live editor**, `GroomStrandCoat.olo` on Vulkan + Deferred: one groom on **2000 strands**, the
other on **400 cards**, from the real cooked v3 asset, in one frame.

## GPU cost and memory, by animal / group / representation

RTX 4090, Debug, 1411×942. Per-representation cost is reported live by the renderer panel
(`GroomLodStats::StrandsByRepresentation` / `BytesByRepresentation`) and per-asset by the inspector.

| Coat | Strands | Strand tier | Cards | Card tier | Reduction |
|---|---|---|---|---|---|
| `reference-human-scalp` | 20 000 | 37.4 MiB | 400 | 0.75 MiB | 50× |
| `reference-animal-pelt` | 30 000 | 24.0 MiB | 1 200 | 1.60 MiB | 25× |
| `reference-shortcoat-animal` | 5 649 | — | 815 | — | 6.9× |
| `reference-longcoat-animal` | 6 399 | — | 835 | — | 7.7× |

Cooked size grew 1.85 MiB → 1.85 MiB (scalp) / 1.48 MiB (pelt) including section 10; cook time
141 ms / 124 ms.

## Notes for adjacent work

- **#1323** (route hair into the scene's shadow techniques) refactors #1246's per-(asset, settings)
  strand cache out of `GroomRenderPass::Execute`. That cache key now also carries the
  **representation** — a card level and the base groom are different curve sets that would otherwise
  hash the same. Worth knowing before that refactor starts; nothing here implements #1323's routing.
- A **textured** hair card (an alpha/tangent atlas baked from the strands it replaces) would
  preserve a tuft's spread through the texture rather than through width. It needs an atlas bake and
  a second shading path, and it fixes the coat's look at bake time — out of scope here, and worth
  its own issue if the card tier's silhouette error ever becomes the limiting factor.
- `GroomCoatClumpCell`'s ±16 UV clamp is a real limitation for UDIM charts that #1251's clumping has
  **silently**; this PR only makes the card cook refuse it loudly.

## Summary by CodeRabbit

* **New Features**
  * Added groom representation LOD, transitioning hair and fur between strand and cooked card representations based on screen size.
  * Added independent visibility, simulation, and shadow budgets with hysteresis and width compensation.
  * Added Groom LOD settings, live renderer statistics, and fallback details to the Scene Hierarchy inspector.
  * Alembic imports can generate cooked card LOD data.
  * Groom LOD settings are preserved in save games.

* **Documentation**
  * Added design guidance and measured analysis for groom representation LOD.

* **Compatibility**
  * Updated the groom asset format; older assets must be re-imported.


drsnuggles8:
All ten addressed in 8c2bb75d8. Two were genuine correctness bugs and two were real gaps in the guards' own coverage — thank you, the first one in particular.

**The hold counting the ideal tier (`GroomLod.cpp`) was the serious one.** Confirmed reachable and confirmed to deadlock: the shell tier ships in no build, so a coat straddling the *mesh* edge has an `ideal` that alternates Mesh/Card while `target` is Card on both frames. I added `StraddlingTheMeshEdgeStillHandsOverToCards` and verified it fails on the old ordering — 20 held frames out of 40, still held at frame 39, i.e. the hand-over never happens at all. Fixed by advancing the counter on `target`, after the fallback walk.

One refinement to the suggested test shape: `HeldByHysteresis` on the first few frames is *correct* (the coat starts on Strand and the hold genuinely has not elapsed), so the assertion is on **transience** — the last held frame must be within `HoldFrames` — rather than on its absence. My first draft asserted absence and failed for the right reason, which was a useful correction.

**Byte double-count (`GroomRenderPass.cpp`)** — agreed, and the reasoning about `CachedBytes` is the right frame: a resident figure that can exceed the cache's own total is not a resident figure. I used a `BytesCountedTick` on the entry rather than a per-frame `unordered_set`, which avoids the allocation and matches the `CoatBuiltTick` idiom already in that struct.

**Group-id bound (`GroomAsset.cpp`)** — agreed, and the `GroupDesc`-answers-identity detail is exactly why it matters: the failure is silent, not a crash. I put the check inside `GroomLodLevel::Validate` (which now takes the group count) rather than in the asset's loop, so the invariant lives with the type that owns it and the builder gets it too.

**Decoder-side rejection (`GroomLodRoundTripTest.cpp`)** — you were right, and my comment claiming a forged payload was not buildable "without reimplementing the writer" was simply wrong; decompress/patch/recompress/re-CRC is four calls. The new case does that, plus a **control** that puts the original value back through the same forging path — otherwise it would pass for a decoder that rejected every re-compressed file, which is a different bug wearing this one's clothes.

**The rung-rule comment (#8)** was off by one in the *header* as well, not just the test: "step k is asked for below `FullPixelSize / 2^k`" describes step k+1. At a 512 px full size, 100 px is step 3 (it clears the 64 px rung), not step 2. Both are corrected; `GroomLodBudget`'s sweep is what pins the real rule.

The remaining four (rule-number in the failure table, section 9 vs 10, `<cmath>`, the double `FindLodLevel`) applied as suggested.

375 `Groom*` tests pass; pre-commit clean.

drsnuggles8:
@coderabbitai review

drsnuggles8:
Both addressed in e41c441c7.

## The red sanitizer jobs were mine, and one cause

ASan, TSan and UBSan all failed on the same line: `GroomLodComparison.*` hit ctest's **600 s per-entry timeout**. (The `LLVMgold.so` noise in the ASan log is a non-fatal LTO support probe, not the failure.)

I calibrated the instrumentation factor rather than guessing at it: #1246's analogous suite `GroomCoverageComparison.*` takes **50 s locally and 496 s under TSan** — ~10×. My combined suite was ~118 s locally, so ~1200 s instrumented. It was never going to fit.

Three changes, none of which alter what is measured:

- **One gtest suite per coat.** ctest registers one entry per suite and times each out separately, so three entries share the budget and run in parallel. The coats were only ever in one case because they were written in one loop.
- **The 256 px row is gone.** Cost scales with the coat's pixel *footprint*, so that row alone was ~¾ of the measurement — and it is the one size at which none of the deciding claims apply, since they are gated to `pixelSize < CardPixelSize`. The shell case still sweeps 128 px → 4 px.
- **Two redundant recomputations removed.** `strand-full` re-projected and re-rasterised the geometry whose coverage had just been computed *as the reference it is compared against*; and the matched stride was measured once per card *arm* rather than once per card *count* (the two aggregations at one cell produce the same count).

Worst entry: **21 s locally, ~210 s instrumented**, against a 600 s limit.

I did not add the suite to `OLO_TSAN_LINUX_CTEST_EXCLUDE` and friends. Those lists hold genuinely sanitizer-incompatible tests (network, worker-restart), not slow ones, and putting a correct-but-slow suite there would be hiding the cost rather than paying it — quite apart from `asan.yml` belonging to another branch in this batch.

**Worth flagging separately:** `GroomFibreEnergyTest.*` (572 s) and `GroomCoverageComparison.*` (496 s) are already sitting against that 600 s ceiling under TSan. Neither is mine and both are green today, but the next test file added anywhere reshuffles the shards and either could tip. Not fixing them in this PR — just noting where the cliff is.

## While in there: an assertion that was a coincidence

The card beats the matched stride on silhouette error by a factor nearly everywhere (2.8× on the scalp at 64 px) — but by **1.3 %** on the short coat at 128 px, a margin a different float summation order on another platform could flip either way. That was a flake waiting to happen, on CI, at a size the finding doesn't depend on.

It is now two claims: *never materially worse* (5 % tolerance) across the whole card band, and a *strict win* at and below 64 px where the margin is a factor. That the top of the band is a tie is itself the finding that the tier is distance-gated.

## The `MaxCards` finding — correct, and the reasoning was the useful part

`MaxCards` is the one card setting that **sizes an allocation** (the build reserves `cardCount × PointsPerCard` points) and it was the one I left unvalidated while validating the other three by name. Now bounded like them, with the point total checked against the format cap *before* the reserve.

You were also right that `GroomLodLevel::Validate` lacked the point-count bound its base-groom counterpart has. Worth being precise about which path that guards: the **decoder** already bounds a file-supplied `PointCount` before sizing anything, so this is the **authored** path — a level reaching `AttachLodLevels` from a builder or a tool, which has no file behind it to blame. The two paths disagreeing about what a valid level is was the real defect.

Covered by `TheCardCapItselfIsBoundedBecauseItSizesAnAllocation`.

378 `Groom*` tests pass (was 375); pre-commit clean. The analysis doc and the rules guide were updated to match the sweep that now runs — numbers quoted from the dropped 256 px row were replaced with the measured 128 px ones rather than left to rot.

